PR #110, opened 2020-01-02
Update trust-dns-* dependencies

nevi-me:
This has the side effect of replacing tokio=0.1 with 0.3, and reducing the number of dependencies being built.

I'm trying to reduce the number of dependencies on my projects as I'm moving them to async_await syntax. I noticed that `trust-dns-resolver` is still using tokio=0.1


saghm:
The failures are due to issues with our CI, not due to the driver, so I'll merge this now. Thanks for the contribution!

nevi-me:
Hey @saghm, I see that the RUST project on JIRA is now visible to the public, but I can't create a new JIRA.

I wanted to create a bug for `FindOneOptions`, it seems like the `sort` option is not included, but it's there for `FindOneAnd*`.

I've been using `find()` with a limit of 1 instead of `find_one()` for now, but this omission might confuse some people.

saghm:
Hello! You're correct, we do need to add the sort field to FindOptions. I'm actually having trouble filing a ticket for it now myself; I think that the settings may have gotten messed up when the JIRA was made public this morning. Once we get that sorted out, I'll comment here with the ticket number. Thanks!

nevi-me:
> Hello! You're correct, we do need to add the sort field to FindOptions. I'm actually having trouble filing a ticket for it now myself; I think that the settings may have gotten messed up when the JIRA was made public this morning. Once we get that sorted out, I'll comment here with the ticket number. Thanks!

I think it's just `FindOneOptions` as `FindOptions` has it already

saghm:
Oops, sorry for the typo; I meant to say `FindOneOptions`

saghm:
@nevi-me any chance you could try filing a JIRA ticket for this now? We just changed the permissions, and we want to make sure users are able to file tickets now

nevi-me:
Here it is: https://jira.mongodb.org/browse/RUST-266

saghm:
Awesome, thanks a bunch!

PR #115, opened 2020-01-14
Default on ReplaceOptions.hint is missing

andywswan:
So it'll throw a not very helpful error at you:

> error[E0599]: no method named `build` found for type `mongodb::coll::options::TypedBuilder_BuilderFor_ReplaceOptions<(), (std::option::Option<bool>,), (), (), ()>` in the current scope


saghm:
Good catch, and thanks for the pull request!

PR #117, opened 2020-01-24
RUST-276 Properly construct findAndModify without options

saghm:


src/operation/find_and_modify/mod.rs
```diff
@@ -34,7 +34,8 @@ impl FindAndModify {
         query: Document,
         options: Option<FindOneAndDeleteOptions>,
     ) -> Self {
-        let options = options.map(FindAndModifyOptions::from_find_one_and_delete_options);
+        let options =
+            FindAndModifyOptions::from_find_one_and_delete_options(options.unwrap_or_default());
```
patrickfreed: As an added guard rail, what do you think about introducing an enum for the update/delete that is non-optional in `FindAndModifyOptions`? Since one of the two must be specified, it would be nice to cement that at the type level.

Also these constructors should probably be rewritten as `From` implementations, though that's more of a cosmetic issue.

saghm: We could do that, but it would make the serialization a bit more tricky, since we'd be making one field map to one of two different names based on the contents.

saghm: Updated

PR #119, opened 2020-01-28
RUST-242 Properly handle network errors during operations

saghm:


src/sdam/state/mod.rs
```diff
@@ -175,6 +175,54 @@ impl Topology {
     }
 }
 
+pub(crate) fn handle_pre_handshake_error(
```
patrickfreed: The logic all looks good, but all the methods in this file seem as if they would fit better on `Topology` itself rather than as free floating functions. Doing that might a bit of restructuring of the type though, and that's likely out of scope for this ticket. Perhaps some follow up work could be done to internalize the state updates of `Topology`?

saghm: The reason these are all top-level functions is that the `update_topology` function is top-level, which in turn is top-level because we don't want to have to hold the lock on the topology the entire time that the method executes. We could probably refactor this by renaming `Topology` to `TopologyInner` and then making a `Topology` type that contains an `Arc<RwLock<TopologyInner>>` and then delegates as needed, but I agree this is out of scope for this ticket. My instinct is to just save this work for when we restructure SDAM a bit for async so that we can take into account the new design constraints rather than potentially having to do this work different ways twice, but I don't feel that strongly against doing the work earlier if we have some spare cycles.

patrickfreed: Sounds good. I filed a generic [ticket](https://jira.mongodb.org/browse/RUST-279) for this work to track it, and it can just be closed out either as part of async or if it happens to get picked up earlier.

PR #121, opened 2020-02-04
RUST-258 Replace connection in monitoring threads when errors occur

saghm:


src/sdam/monitor.rs
```diff
@@ -28,20 +29,33 @@ lazy_static! {
...
+        let make_connection = || {
+            Connection::new(
+                0,
+                address.clone(),
+                0,
+                options.connect_timeout,
+                options.tls_options(),
+                options.cmap_event_handler.clone(),
+            )
+            .unwrap()
```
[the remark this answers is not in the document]
patrickfreed: The only place in which `monitor_server` is called is in `add_new_server`, which returns a `Result`. So we can just make `monitor_server` return a `Result<()>` and not unwrap this connection.

`is_master` uses `make_connection`, but it too returns a `Result`.

src/sdam/monitor.rs
```diff
@@ -114,10 +130,28 @@ fn check_server(
...
+        .unwrap_or(false)
+    {
+        std::mem::replace(conn, make_connection());
+    }
+
+    result
 }
 
-fn is_master(conn: &mut Connection) -> Result<IsMasterReply> {
+fn is_master_inner(conn: &mut Connection) -> Result<IsMasterReply> {
```
patrickfreed: this function is only called above and it's short enough that I think it makes sense to just include it inline in `is_master`.

src/sdam/monitor.rs
```diff
@@ -74,6 +88,7 @@ fn monitor_server_check(
     conn: &mut Connection,
     mut server_type: ServerType,
     server: &Weak<Server>,
+    make_connection: impl Fn() -> Connection,
```
patrickfreed: a docstring explaining why this lambda is necessary might be useful.

src/sdam/monitor.rs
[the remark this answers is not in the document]
saghm: This is inside a call to `std::thread::spawn`, though; `add_new_server` doesn't (and shouldn't) wait for this thread to finish before returning. Barring some sort of mechanism to pass errors back to the main thread, the options are to either panic and crash the monitoring thread or continually try to create a new connection until the thread is shut down by other means (i.e. the server being removed from the cluster or the client being dropped).

src/sdam/monitor.rs
saghm: We need to make a new connection to replace the old one, which requires knowing the address, the connect timeout, and the TLS options. Abstracting them into a lambda that could be passed along rather than passing along three separate arguments that were used for only one thing felt cleaner.

src/sdam/monitor.rs
saghm: I separated it so that we could use the `?` operator in the body. The idea is that we want to short-circuit doing the rest of this method in any of the numerous places errors could occur, but we don't want to exit from `is_master` itself because we want to replace the connection if an error occurred.

src/sdam/monitor.rs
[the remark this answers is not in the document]
patrickfreed: We could just establish the connection outside of the thread spawn and then move it into it. If it fails, don't bother spawning the thread and just return early.

src/sdam/monitor.rs
patrickfreed: I meant leave that comment in the code itself. I understand why it's necessary, but future readers of the codebase may not, or we may forget the exact reasoning in the future.

src/sdam/monitor.rs
patrickfreed: Makes sense. Now that I think about it though, we should just call `make_connection` in the error handling portion of `check_server`, rather than inside of `is_master`. Connection reestablishment isn't really a concern for an isMaster helper anyways. This would remove the need for an `is_master_inner`.

Also, the spec says that we should only be doing some of this logic in response to connection errors and timeout errors, but we seem to just be doing it for all errors.

src/sdam/monitor.rs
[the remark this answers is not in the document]
saghm: What do we do when we need to replace the connection in response to an error? That's what this PR is supposed to be handling; the thread will have already been spawned by that point, so we only have the options I mentioned above.

src/sdam/monitor.rs
saghm: We call `is_master` twice from `check_server`, which is why it's abstracted into a function. I think it's cleaner not to duplicate the code.

src/sdam/monitor.rs
saghm: Okay, will do.

src/sdam/monitor.rs
saghm: > Also, the spec says that we should only be doing some of this logic in response to connection errors and timeout errors, but we seem to just be doing it for all errors.

The code checks to see if the error is a network error (i.e. an I/O error). If it's not, the connection isn't replaced.

src/sdam/monitor.rs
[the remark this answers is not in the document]
patrickfreed: Assuming `make_connection` returns a `Result`, we can just treat the error cases as network errors and follow the regular SDAM logic specified in the [psuedocode](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#network-error-during-server-check).

src/sdam/monitor.rs
patrickfreed: I was referring to the error checking in `check_server`. Another benefit of moving the connection re-establishment there is that the error handling of network errors is consolidated, as it is in the SDAM spec psuedocode.

src/sdam/monitor.rs
```diff
@@ -28,20 +29,33 @@ lazy_static! {
...
-) {
+pub(super) fn monitor_server(address: StreamAddress, server: Weak<Server>, options: ClientOptions) {
     std::thread::spawn(move || {
         let mut server_type = ServerType::Unknown;
-        let heartbeat_frequency = heartbeat_frequency.unwrap_or(DEFAULT_HEARTBEAT_FREQUENCY);
+        let heartbeat_frequency = options
+            .heartbeat_freq
+            .unwrap_or(DEFAULT_HEARTBEAT_FREQUENCY);
+
+        let make_connection = || {
```
mbroadst: I took a slightly different approach in node, and I wonder what your thoughts are as they may apply to Rust. 

Instead of passing a function around here, I just made the "connection" a member of the `Monitor` (in this case it might be a member of the `Server` since there is no `Monitor` type). Then I have a method called `checkServer` which either reuses the existing connection (known good), or if the existing connection is `null` then it makes a new connection. The added benefit is that, since the initial handshake of connection establishment _is_ itself a heartbeat, you can still call the whole process `checkServer`.

saghm: I took a stab at rewriting this a bit to use a `Monitor` struct; let me know what you think of it

src/sdam/monitor.rs
```diff
@@ -56,84 +86,109 @@ pub(super) fn monitor_server(
...
-    })
+        Ok(IsMasterReply {
+            command_response,
+            round_trip_time: Some(start_time.to(end_time).to_std().unwrap()),
+        })
+    }
+
+    fn make_new_connection(&self) -> Result<Connection> {
+        Connection::new(
+            0,
```
patrickfreed: The id and generation being hard coded to 0 seems like an issue. We should probably get these from the pool.

src/sdam/monitor.rs
```diff
@@ -56,84 +86,109 @@ pub(super) fn monitor_server(
...
+    fn is_master(&mut self) -> Result<IsMasterReply> {
+        let result = self.is_master_inner();
+
+        if result
+            .as_ref()
+            .err()
+            .map(|e| e.kind.is_network_error())
+            .unwrap_or(false)
+        {
+            self.connection = self.make_new_connection()?;
```
patrickfreed: To consolidate the existing threads, I'll just start a new one here.
This connection re-establishment logic makes a lot more sense to me if it is included with the rest of the error handling done in `check_server`, and seems out of place here inside of the `is_master` helper itself.
e.g. I'm thinking our `check_server` could look very similar to the pseudocode presented in the SDAM spec:
```
def checkServer(description):
    try:
        call ismaster
        return new ServerDescription
    except NetworkError as e0:
        clear connection pool for the server

        if description.type is Unknown or PossiblePrimary:
            # Failed on first try to reach this server, give up.
            return new ServerDescription with type=Unknown, error=e0
        else:
            # We've been connected to this server in the past, retry once.
            try:
                reconnect and call ismaster
                return new ServerDescription
            except NetworkError as e1:
                return new ServerDescription with type=Unknown, error=e1
```

patrickfreed: Because connection establishment could fail, I think it would make sense to have `self.connection` be an `Option`. Then when we encounter a network error, we null out the connection and only assign back to it if we successfully re-establish. Otherwise, we may attempt to use a broken/old connection next time we check the server.

src/sdam/monitor.rs
saghm: Per the SDAM and CMAP specs, [monitoring connections cannot be from the regular connection pool on the server](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#a-monitor-should-not-use-the-client-s-regular-connection-pool), and [connection pools must start their auto-incrementing ID at 1](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection). Given that no CMAP events are emitted from the SDAM monitoring connections, the connection ID doesn't really matter, but if it did, picking 0 guarantees that it will never overlap with any of the IDs from the pool for the same server (unless we have more than `usize::MAX` connections, but that would cause overlap regardless of how we picked the ID here).

src/sdam/monitor.rs
saghm: That seems reasonable to me; I'll update the code to do that.

saghm: It's somewhat tough to model the code like that given that Rust has no concept of nested try/catch; the closest analog to try/catch we have is `?`, which always returns from the function on an error. We'd have to break it up into multiple functions, which is what I've done here.

patrickfreed: That makes sense; I see why it's necessary as is. I like the name change to `perform_is_master`, only remaining suggestion I have wrt this is to update `check_server` to remove the `return` from the success branch, and move the retry attempt into the error branch. I think that better communicates the actual control flow, and it more closely mirrors the spec as well.

src/sdam/monitor.rs
saghm: Good idea; done

src/sdam/monitor.rs
saghm: Ah, yeah, that does make it much more similar; done

src/sdam/monitor.rs
mbroadst: nice, LGTM!

PR #122, opened 2020-02-05
add Sort option to FindOneOptions

ttdonovan:
Adds the sort option to `FindOneOptions`. Unclear if a specific test is needed.


patrickfreed:
Thanks for your contribution!

PR #123, opened 2020-02-11
RUST-257 Properly fallback when unable to connect to first address

saghm:


src/cmap/conn/stream.rs
```diff
@@ -38,27 +38,53 @@ pub(super) enum Stream {
...
+
+    socket_addrs.sort_by_key(|addr| if addr.is_ipv4() { 0 } else { 1 });
+
+    let mut connect_error = match try_connect(&socket_addrs[0], timeout) {
+        Ok(stream) => return Ok(stream),
+        Err(err) => err,
+    };
+
+    for address in &socket_addrs[1..] {
+        connect_error = match try_connect(&socket_addrs[0], timeout) {
```
patrickfreed: should be connecting to `address`. Surprised clippy didn't catch it as `address` isn't being used.

src/cmap/conn/stream.rs
```diff
@@ -38,27 +38,53 @@ pub(super) enum Stream {
...
+    }
+
+    socket_addrs.sort_by_key(|addr| if addr.is_ipv4() { 0 } else { 1 });
+
+    let mut connect_error = match try_connect(&socket_addrs[0], timeout) {
+        Ok(stream) => return Ok(stream),
+        Err(err) => err,
+    };
+
+    for address in &socket_addrs[1..] {
```
patrickfreed: will this panic if `socket_addrs` only has one element?

src/error.rs
```diff
@@ -124,6 +124,9 @@ pub enum ErrorKind {
     #[error(display = "{}", _0)]
     Io(#[error(source)] std::io::Error),
 
+    #[error(display = "No DNS results for domain {}", _0)]
+    NoDnsResults(StreamAddress),
```
patrickfreed: Could we lump the `DnsName` error in with this one?

src/cmap/conn/stream.rs
```diff
@@ -38,27 +38,53 @@ pub(super) enum Stream {
...
+fn connect_stream(address: &StreamAddress, connect_timeout: Option<Duration>) -> Result<TcpStream> {
+    let timeout = connect_timeout.unwrap_or(DEFAULT_CONNECT_TIMEOUT);
+
+    let mut socket_addrs: Vec<_> = address.to_socket_addrs()?.collect();
+
+    if socket_addrs.is_empty() {
+        return Err(ErrorKind::NoDnsResults(address.clone()).into());
+    }
+
+    socket_addrs.sort_by_key(|addr| if addr.is_ipv4() { 0 } else { 1 });
```
patrickfreed: why was this methodology chosen over of happy eyeballs? It might be worth recording the rationale in `connect_stream`'s docstring.

saghm: We discussed this in a triage meeting a few weeks ago (which I think you were present for, but I could be wrong). Matt seemed unconvinced that the overheard of happy eyeballs (spawning additional threads) and the complexity of implementation were worth the potential gains from not checking them in sequence. He also brought up that C is the only driver that uses happy eyeballs currently, if I recall correctly.

src/cmap/conn/stream.rs
saghm: Nope; it would just yield an empty slice, meaning no iterations are done. See [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bf54f063b1349ed8139e8f65850f6596) for a playground showing this.

src/cmap/conn/stream.rs
saghm: Good catch! I'm also surprised that there was no warning for this.

src/error.rs
saghm: That would be tricky; `DnsName` wraps an external error (from `webpki`), so we'd have to turn it into an enum itself, which doesn't seem much better than just putting another variant here. Note also that `DnsName` [signifies that the domain name a user passed us isn't syntactically valid](https://docs.rs/webpki/0.21.2/webpki/struct.InvalidDNSNameError.html), which is a bit different from a DNS query returning no results.

src/cmap/conn/stream.rs
patrickfreed: Sounds good--could we leave a docstring or comment indicating that?

src/error.rs
patrickfreed: Perhaps we should rename the `DnsName` case to better indicate that now that we have a similarly named error case, e.g. `InvalidDnsName`. This would also match up with `InvalidHostname` nicely.

saghm: I'm not sure I understand why users would confuse `DnsName` with `NoDnsResults`. `NoDnsResults` seems more similar to `DnsResolve` than `DnsName` to me. That said, I'm not opposed to the name `InvalidDnsName`, but that would be a breaking change, so we probably want to be pretty certain that we want to change error names before doing it. In particular, if there's another name for `NoDnsResults` that you think would improve things instead, that would be a preferable solution since it's not an existing error type yet.

src/cmap/conn/stream.rs
saghm: This line is existing code from a previous commit; the only change to it is that it was moved to a different function. Is there a reason why it sticks out to you now when it didn't before?

patrickfreed: We put it in before as a temporary measure until we fully implemented happy eyeballs. Now that we've decided not to implement happy eyeballs and commit to the temporary measure, it makes sense to me for us to record why.

src/error.rs
patrickfreed: The driver is currently in alpha; breaking changes should not be a concern whatsoever. Also, I have no issue with `NoDnsResults` since it is named very specifically. `DnsName` is too generic in my opinion and should be changed to match `InvalidHostname`.

mbroadst: I'm deferring to both of you on this one, but I agree that we should feel comfortable making breaking changes until 1.0 comes out. If we feel like `InvalidDnsName` is going to help a user better understand what's going on, then we should make that change.

src/cmap/conn/stream.rs
mbroadst: I think we discussed this one offline, and decided to record that "other approaches were entertained, but ultimately we decided to do what other drivers are doing"

src/error.rs
saghm: How about we file another ticket to change it to `InvalidDnsName` and then schedule it? It's not really related to the work this ticket is doing at all, so I think it makes sense to follow up with a separate commit right after rather than bundle it in here.

src/cmap/conn/stream.rs
```diff
@@ -38,27 +38,55 @@ pub(super) enum Stream {
...
+    // After considering various approaches, we decided to do what other drivers do, namely try each
+    // of the addresses in sequence with a preference for IPv4.
+    socket_addrs.sort_by_key(|addr| if addr.is_ipv4() { 0 } else { 1 });
+
+    let mut connect_error = match try_connect(&socket_addrs[0], timeout) {
+        Ok(stream) => return Ok(stream),
+        Err(err) => err,
+    };
+
+    for address in &socket_addrs[1..] {
```
patrickfreed: could we combine the above connect into this loop? We could have `connect_error` start as `None` or something, then do an `unwrap_or_else` with `NoDnsResults`.

src/error.rs
patrickfreed: that's fine with me.

saghm: Filed https://jira.mongodb.org/browse/RUST-295

PR #130, opened 2020-02-28
minor: add static runtime variable

saghm:


src/lib.rs
```diff
@@ -74,11 +74,14 @@
 pub mod options;
 
 mod bson_util;
+#[cfg(any(feature = "tokio-runtime", feature = "async-std-runtime"))]
```
saghm: I added annotations to each module that will need to use `RUNTIME` either directly or transitively so that users who don't have tokio xor async-std enabled will only get the custom compiler error message we give below.

saghm: Actually, I missed a bunch; literally the only module we have that isn't in this set is `collation`.

saghm:
It seems to still work fine for me

PR #131, opened 2020-03-02
 Added "Bson" in the scope of " Finding documents in a collection" in README

UmerMIB:
It is necessary as it is using afterwards


patrickfreed:
Thank you for the contribution!

PR #133, opened 2020-03-03
RUST-298 Implement async SDAM

saghm:


src/sdam/message_manager.rs
```diff
@@ -0,0 +1,82 @@
...
+            topology_check_listener,
+            topology_change_notifier: Arc::new(topology_change_notifier),
+            topology_change_listener,
+        }
+    }
+
+    /// Gets a receiver which will receive messages whenever a topology check has been requested by
+    /// a server selection operation. This should be called by each SDAM background task when it
+    /// starts up.
+    pub(super) fn subscribe_to_topology_check_requests(&self) -> Receiver<()> {
```
patrickfreed: Do we need this method? I imagine the SDAM background tasks could just use `wait_for_topology_check_request`.

src/sdam/state/mod.rs
```diff
@@ -2,161 +2,302 @@ pub(super) mod server;
...
             servers: Default::default(),
-            condvar,
-            options,
         }));
 
+        let topology = Topology { common, state };
+
         {
-            let mut topology_lock = topology.write().unwrap();
+            let mut state = topology.state.write().await;
```
patrickfreed: Now that this is happening on the separate state struct, we can do this initialization before putting it behind the lock.

src/sdam/state/mod.rs
```diff
@@ -2,161 +2,302 @@ pub(super) mod server;
...
+        // properly inform users of errors that occur here.
+        let _ = state_clone.update(server_description, self.clone());
+
+        // Now that we have the proper state in the copy, acquire a lock on the proper topology and
+        // move the info over.
+        let mut state_lock = self.state.write().await;
+        state_lock.description = state_clone.description;
+        state_lock.servers = state_clone.servers;
+
+        self.common.message_manager.notify_topology_changed();
```
patrickfreed: According to the SDAM spec, we're only supposed to publish topology description changed events if the topology description actually changed. I figure that also applies to server selection, since if the topology description didn't actually change, then selection will fail again. Given that, we should probably check for equality of the server/topology descriptions here before notifying that the topology changed. Also, there isn't really a point to returning a bool from this method if it will always be true.

src/sdam/state/mod.rs
```diff
@@ -2,161 +2,302 @@ pub(super) mod server;
...
+            self.mark_server_as_unknown(error.clone(), server.address.clone())
+                .await;
+
+            self.common.message_manager.request_topology_check();
+
+            let wire_version = conn
+                .stream_description()
+                .map(|sd| sd.max_wire_version)
+                .ok()
+                .and_then(std::convert::identity)
```
patrickfreed: I think we can use `flatten` for this now.

src/sdam/message_manager.rs
saghm: Nope, we can delete it.

src/sdam/state/mod.rs
saghm: Ah, good point!

src/sdam/state/mod.rs
saghm: I agree that this should be updated, but when I realized that we hadn't done this already, I figured it would make sense to do in a follow-up commit rather than bundling the new equality implementation on `TopologyDescription` as part of this commit. Would you be okay with me adding a todo and filing a ticket for this to work on as soon as this is merged?

src/sdam/state/mod.rs
saghm: Sounds good; in that case, we'll also need to update the minimum version to 1.40 before we release, so we could also add the `nonexhaustive` attribute to our enums.

src/sdam/state/mod.rs
saghm: Actually, we can't do this easily, since we need to have the already created `Topology` to pass to `add_new_server` to be able to start the monitoring threads. Moving the lock to a separate struct doesn't really change anything.

patrickfreed: Oh right, forgot about that.

src/sdam/state/mod.rs
patrickfreed: Yep, sounds good to me.

saghm: Filed RUST-301

PR #135, opened 2020-03-04
RUST-301 Check topology updates for changes

saghm:


saghm:
@patrickfreed is this all set now?

PR #138, opened 2020-03-05
RUST-214 Implement async CMAP

patrickfreed:
[RUST-214](https://jira.mongodb.org/browse/RUST-214)
This PR updates the existing connection pool to be async.


src/cmap/background.rs
```diff
@@ -1,71 +1,78 @@
...
-    ensure_min_connections_in_pool(&pool);
+    pool.remove_perished_connections().await;
+    pool.ensure_min_connections().await;
 }
 
-/// Iterate over the connections and remove any that are stale or idle.
-fn remove_perished_connections_from_pool(pool: &ConnectionPool) {
-    let mut connections = pool.inner.connections.write().unwrap();
-    let mut i = 0;
+impl ConnectionPool {
```
patrickfreed: these are all pretty much the same thing, but I moved them to be methods on `ConnectionPool` itself.

src/cmap/conn/mod.rs
```diff
@@ -73,19 +75,21 @@ impl Connection {
...
             id,
             generation,
             pool: None,
-            stream_description: None,
             ready_and_available_time: None,
-            stream: Stream::connect(address.clone(), connect_timeout, tls_options)?,
+            stream: Stream::connect(
+                address.clone(),
+                options.as_mut().and_then(|options| options.connect_timeout.take()),
+                options.as_mut().and_then(|options| options.tls_options.take())
```
patrickfreed: `StreamOptions` could implement `From<ConnectionOptions>` instead of this manual extraction.

src/cmap/mod.rs
```diff
@@ -57,38 +54,139 @@ impl From<Arc<ConnectionPoolInner>> for ConnectionPool {
...
-/// The internal state of a connection pool.
-#[derive(Derivative)]
-#[derivative(Debug)]
-pub(crate) struct ConnectionPoolInner {
-    /// The address the pool's connections will connect to.
-    address: StreamAddress,
-
+/// A struct used to manage the creation, closing, and storage of connections for a `ConnectionPool`.
+#[derive(Debug)]
+struct ConnectionManager {
```
patrickfreed: This type encapsulates all the connection related state that needs to be accessed from behind the lock. 
- As discussed offline, it `total_connection_count` should only be accessed if a write lock is held on the `Vec<Connections>`, so it makes sense to group them. 
- The same is generally true for `next_connection_id`. 
- Based on just looking at where we used `generation`, I saw that it was _only_ being used when the lock was held, so there was no cost to keeping it here with the rest of the state. The only exception is that `clear` would previously update the generation without getting the write lock on the connections, but the argument could be made that it should.

src/cmap/mod.rs
```diff
@@ -105,34 +203,21 @@ pub(crate) struct ConnectionPoolInner {
...
-    /// is enforced by threads entering the wait queue when they first try to check out a
-    /// connection and then blocking until they are at the front of the queue.
+    /// The queue that operations wait in to check out a connection.
+    ///
+    /// A thread will only reach the front of the queue if a connection is available to be checked out
+    /// or if one could be created without going over `max_pool_size`.
     wait_queue: WaitQueue,
 }
 
 impl ConnectionPool {
```
patrickfreed: A lot of the changes here are simply moving the implementations to inner. Instead of having different methods implemented on the different structs, I decided it was clearer to just implement everything on the inner type, and simply call through to the methods from `ConnectionPool`. This nicely splits the "internal API" of the pool from the internal implementation of the pool itself.

src/cmap/mod.rs
```diff
@@ -57,38 +54,139 @@ impl From<Arc<ConnectionPoolInner>> for ConnectionPool {
...
+
+/// The internal state of a connection pool.
+#[derive(Derivative)]
+#[derivative(Debug)]
+pub(crate) struct ConnectionPoolInner {
+    /// The address the pool's connections will connect to.
+    address: StreamAddress,
+
+    /// The structure used to manage connection creation, closing, storage, and generation.
+    connection_manager: Arc<Mutex<ConnectionManager>>,
```
patrickfreed: this manager will only ever have a second reference to it when we move it to the task in `Drop` since we can't `await` on the mutex in a non-async function. It's unfortunate that we have two layers of `Arc` here, but it doesn't really seem like that big of a deal to me.

src/cmap/test/mod.rs
```diff
@@ -128,129 +134,143 @@ impl Executor {
...
+            self.wait_queue_timeout.matches(&expected.wait_queue_timeout)
     }
 }
 
-fn assert_events_match(actual: &Event, expected: &Event) {
-    match (actual, expected) {
-        (Event::ConnectionPoolCreated(actual), Event::ConnectionPoolCreated(ref expected)) => {
-            if let Some(ref expected_options) = expected.options {
-                assert_eq!(actual.options.as_ref(), Some(expected_options));
+impl Matchable for Event {
```
patrickfreed: since these tests use the "MATCH" function, I figured it was useful to use our test runner's facilities for that. This also helped catch a bug in the event emission.

src/cmap/mod.rs
```diff
@@ -388,37 +295,101 @@ impl ConnectionPoolInner {
...
+        Ok(conn)
+    }
+
+    /// Waits for the thread to reach the front of the wait queue, then attempts to check out a
+    /// connection. If none are available in the pool, one is created and checked out instead.
+    async fn acquire_or_create_connection(&self) -> Result<Connection> {
+        // Handle that will wake up the front of the queue when dropped.
+        // Before returning a valid connection, this handle must be disarmed to prevent the front
+        // from waking up early.
+        let mut wait_queue_handle = self.wait_queue.wait_until_at_front().await?;
```
patrickfreed: this handle allows us to use the `?` operator, which is nice. It does need to be disarmed when successfully returning though.

src/cmap/wait_queue.rs
```diff
@@ -1,177 +1,85 @@
...
+        })
     }
 
-    /// Notifies the wait queue that the next thread in the queue can be woken up.
-    pub(super) fn notify_ready(&self) {
-        self.inner.lock().unwrap().notify_ready();
+    /// Signal that the front of the queue (if there is one) is ready to wake up.
+    pub(super) fn wake_front(&self) {
+        if self.semaphore.permits() >= self.max_handles {
+            panic!("greater than {} connections checked back into pool with address {}",
```
patrickfreed: So far, we've generally been panicking when preconditions aren't met (e.g. `unwrap` on locks that could be poisoned), so I continued that here. We _could_ silently handle this here if we wanted to, though.

One thing that I'm noticing now that seems kind of troubling is that the semaphore [appears to be implemented using a parking lot mutex](https://docs.rs/futures-intrusive/0.3.0/futures_intrusive/sync/type.Semaphore.html), so that [`release()`](https://docs.rs/futures-intrusive/0.3.0/src/futures_intrusive/sync/semaphore.rs.html#524-526) and [`permits()`](https://docs.rs/futures-intrusive/0.3.0/src/futures_intrusive/sync/semaphore.rs.html#529-531) constitute blocking calls (note how this method isn't `async`). Should this be something we're worried about?

src/cmap/conn/mod.rs
```diff
@@ -228,26 +212,130 @@ impl Connection {
...
+    fn close(&mut self, reason: ConnectionClosedReason) {
+        self.detatch_from_pool();
+        if let Some(ref handler) = self.handler {
+            handler.handle_connection_closed_event(self.closed_event(reason));
+        }
+    }
+
+    /// Removes the weak reference to the parent pool from this connection.
+    /// If this connection is dropped, it will not emit any events.
+    fn detatch_from_pool(&mut self) {
```
saghm: I'm not super fond of this method name, as it seems weird to be calling something called `detach_from_pool` from `mark_checked_in`. I'm thinking a name that describes it at a lower-level might be better, e.g. `remove_pool_reference`. That being said, I'm not convinced this really needs to be a separate method; calling `self.pool.take()` in the two places internally doesn't seem like a big deal to me.

src/cmap/conn/mod.rs
```diff
@@ -228,26 +212,130 @@ impl Connection {
...
+/// Struct encapsulating the state of a connection that has been dropped.
+///
+/// Because `Drop::drop` cannot be async, we package the internal state of a dropped connection into this
+/// struct, and move it into an async task that will attempt to check the reconstructed connection back into the pool.
+///
+/// If the async runtime has been dropped, that task will not execute, and this state will be dropped. From there,
+/// we simply emit a connection closed event and do not attempt to reconstruct the `Connection`.
+#[derive(Derivative)]
+#[derivative(Debug)]
+struct DroppedConnectionState {
```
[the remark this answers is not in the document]
saghm: `DropedConnectionState` looks good to me, but I'm curious about how the runtime was getting dropped when we store it in a static variable. Was this happening due to the tests finishing before the drop tasks completed?

src/cmap/conn/mod.rs
```diff
@@ -228,26 +212,130 @@ impl Connection {
...
+                connection_id: self.id,
+                reason: ConnectionClosedReason::PoolClosed,
+            });
+        }
+    }
+}
+
+impl From<DroppedConnectionState> for Connection {
+    fn from(mut state: DroppedConnectionState) -> Self {
+        Self {
```
saghm: This is pretty similar to `Connection::take`; is there a way to share the code between them?

src/cmap/mod.rs
saghm: I like this refactor a lot! The separation makes the pool a lot cleaner.

Just to make sure I understand, the Establisher is in here so that it can be more easily accessed by the background thread?

src/cmap/background.rs
```diff
@@ -1,71 +1,78 @@
...
-            if pool.inner.total_connection_count.load(Ordering::SeqCst) < min_pool_size {
-                match pool.create_connection(false) {
-                    Ok(connection) => connections.push(connection),
-                    e @ Err(_) => {
-                        // Since we had to clear the pool, we return early from this function and
-                        // put the background thread back to sleep. Next time it wakes up, the
-                        // stale connections will be closed, and the thread can try to create new
-                        // ones after that.
-                        return;
+    /// Add connections until the min pool size it met. We explicitly release the lock at the end of
```
saghm: Kind of weird that git doesn't realize that this is really similar to the section on the left just because of an extra closing brace

src/cmap/mod.rs
```diff
@@ -57,38 +54,139 @@ impl From<Arc<ConnectionPoolInner>> for ConnectionPool {
...
-    credential: Option<Credential>,
+    /// The options used to create connections.
+    connection_options: Option<ConnectionOptions>,
 
     /// Contains the logic for "establishing" a connection. This includes handshaking and
     /// authenticating a connection when it's first created.
     establisher: ConnectionEstablisher,
+}
+
+impl ConnectionManager {
```
saghm: I assume the implementations here are basically straight copies of those from before, just made async?

src/cmap/mod.rs
saghm: We have two layers of `Arc` in other places in the driver, so I think it's fine.

src/cmap/test/integration.rs
```diff
@@ -26,7 +27,7 @@ async fn acquire_connection_and_send_command() {
     let pool_options = ConnectionPoolOptions::from_client_options(&client_options);
 
     let pool = ConnectionPool::new(client_options.hosts[0].clone(), Some(pool_options));
-    let mut connection = pool.check_out().unwrap();
+    let mut connection = RUNTIME.block_on(pool.check_out()).unwrap();
```
saghm: Since this is an async function (with the annotations to start the runtime), you can actually `await` here.

src/cmap/test/mod.rs
```diff
@@ -267,7 +287,7 @@ async fn cmap_spec_tests() {
             let _guard = LOCK.run_concurrently();
 
             let executor = Executor::new(test_file);
-            executor.execute_test();
+            RUNTIME.block_on(executor.execute_test());
```
saghm: Ditto above about being able to `await` here

src/cmap/wait_queue.rs
```diff
@@ -1,177 +1,85 @@
...
 
-    /// The timeout signifying how long a thread should wait in the queue before returning an
-    /// error. This will be the `wait_queue_timeout` for a given connection pool.
-    timeout: Option<Duration>,
+    /// A fair counting semaphore whose count corresponds to the number of connections available.
+    semaphore: Semaphore,
 
-    /// The address that the connection pool's connections will connect to. This is needed to
-    /// return a WaitQueueTimeoutError when the timeout has elapsed.
+    /// The address of the server whose pool this `WaitQueue` belongs to.
```
saghm: Is there a reason that the additional information in the comments for `address` and `timeout` was removed?

src/cmap/wait_queue.rs
```diff
@@ -1,177 +1,85 @@
...
 
 impl WaitQueue {
-    /// Creates a new `WaitQueue`.
-    pub(super) fn new(address: StreamAddress, timeout: Option<Duration>) -> Self {
-        let inner = WaitQueueInner {
-            queue: Default::default(),
-            address,
-            timeout,
-        };
+    /// Creat a new `WaitQueue`.
```
saghm: "Creates"

src/cmap/wait_queue.rs
```diff
@@ -1,177 +1,85 @@
...
+    pub(super) fn new(address: StreamAddress, max_handles: u32, timeout: Option<Duration>) -> Self {
+        let max_handles = max_handles.try_into().unwrap_or(usize::max_value());
```
saghm: If the number of handles doesn't fit into a `usize`, it might make more sense to just assume there's no max pool size; unless the driver is getting run on a 16-bit processor, I doubt that a max pool size of `usize::max_value()` would ever hit its limit.

src/cmap/wait_queue.rs
saghm: I don't think that they are actually blocking; it looks like they write their own wrapper around the parking lot primitives using things like `try_lock` rather than `lock`; note that in the body of the code you linked to, `self.state` refers to a `PollState`, not a `Mutex`. From glancing over the code [here](https://github.com/Matthias247/futures-intrusive/blob/master/src/sync/mutex.rs), it doesn't seem like they're actually calling `lock` on a mutex directly anywhere.

src/cmap/wait_queue.rs
```diff
@@ -1,177 +1,85 @@
...
-            condvar.notify_one();
-        }
+/// A handle to a `WaitQueue` that will wake up the front of the queue when dropped.
+/// To disable this behavior, call `WaitQueueHandle::disarm`.
+pub(super) struct WaitQueueHandle<'a> {
+    semaphore_releaser: SemaphoreReleaser<'a>
+}
+
+impl<'a> WaitQueueHandle<'a> {
+    pub(super) fn disarm(&mut self) {
```
saghm: Is there a reason we can't just call `disarm` from a `Drop` implementation on `WaitQueueHandle`? It's an `async` function, so I don't think there should be any issue.

src/runtime/join_handle.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// Wrapper around `tokio::task:JoinHandle`.
+    #[cfg(feature = "async-std-runtime")]
+    AsyncStd(async_std::task::JoinHandle<T>),
+}
+
+impl<T> Future for AsyncJoinHandle<T> {
+    // tokio wraps the Output of its JoinHandle in a Result, while async-std does not.
+    // In order to standardize on this without us panicking, we too wrap the output in a Result.
+    type Output = Result<T>;
```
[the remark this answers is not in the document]
saghm: This seems fine to me, given that we're only using this in tests right now.

src/runtime/mod.rs
```diff
@@ -63,4 +72,16 @@ impl AsyncRuntime {
     pub(crate) async fn connect_stream(self, options: StreamOptions) -> Result<AsyncStream> {
         AsyncStream::connect(options).await
     }
+
+    /// Await on a future for a maximum amount of time before returning an error.
+    pub(crate) async fn await_with_timeout<F>(self, future: F, timeout: Duration) -> Result<F::Output>
```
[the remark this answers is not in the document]
saghm: I'm fine with putting generic async helpers on the runtime type even if they don't need runtime-specific functionality.

src/cmap/wait_queue.rs
patrickfreed: Nope, I think I got rid of them and then tried to rewrite them from memory or something. Reverted them back. I also added a docstring for `max_handles`.

src/cmap/wait_queue.rs
patrickfreed: fixed. I dropped the "s" since the other docstrings were imperative too.

src/cmap/wait_queue.rs
patrickfreed: as discussed offline, to ensure fairness we want to keep using the semaphore, so we'll treat `usize::max_value` as essentially no limit. As part of this, I added handling 0 as an indication of no limit to the pool options parsing.

src/cmap/wait_queue.rs
patrickfreed: Ok cool.

src/cmap/wait_queue.rs
patrickfreed: We don't want the handle to be disarmed in drop as that would defeat its purpose. This handle is useful for scenarios where you get to the front of the queue but fail to create / acquire a connection and return early via `?`. That way, instead of waking the front on every early return, we only have to disarm before we return successfully, and the queue will otherwise be woken when the handle drops.

src/cmap/conn/mod.rs
saghm: Ah, good point

src/cmap/test/mod.rs
saghm: Ah, okay. Unfortunately async closures are not stable yet.

PR #140, opened 2020-03-09
RUST-308 Use separate client per test

saghm:


src/test/client.rs
```diff
@@ -76,10 +77,10 @@ async fn server_selection_timeout_message() {
         max_staleness: None,
     };
 
-    let mut options = CLIENT.options.clone();
+    let mut options = CLIENT_OPTIONS.clone();
     options.server_selection_timeout = Some(Duration::from_millis(500));
 
-    let client = Client::with_options(options).unwrap();
+    let client = TestClient::new();
```
patrickfreed: this line can probably just be reverted so it can take in the modified options set up in the previous line. below the reference to `client.options` would then have to be updated too, and `CLIENT_OPTIONS` would probably work fine.

PR #141, opened 2020-03-10
RUST-303 Implement async wire protocol

saghm:


src/cmap/options.rs
```diff
@@ -98,3 +98,37 @@ impl ConnectionPoolOptions {
             .build()
     }
 }
+
+/// Options used for constructing a `Connection`.
```
saghm: I moved `ConnectionOptions` here so that all of the CMAP options structs (`StreamOptions`, `ConnectionOptions`, and `ConnectionPoolOptions` are in the same place). I feel like this will make it easier to remember what import path to use when trying to import one of them, but let me know if you disagree and we can come out with another solution.

src/bson_util/async_decode.rs
```diff
@@ -0,0 +1,214 @@
+use bson::{oid, Array, Bson, DecoderError, DecoderResult, Document};
```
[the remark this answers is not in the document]
patrickfreed: It seems the decimal128 parts of it are missing. I'm assuming that's due to some weirdness with the feature flag. My concern with that is what if a decimal128 is included in a wire response? Will we not be able to read/write it?

As a side note, the decimal128 stuff appears to use `unsafe`, which is probably something we will want to change when we flesh it out.

src/bson_util/async_decode.rs
```diff
@@ -0,0 +1,214 @@
...
+                let pat = read_cstring(reader).await?;
+                let opt = read_cstring(reader).await?;
+                Ok(Bson::RegExp(pat, opt))
+            }
+            Some(JavaScriptCode) => read_string(reader, utf8_lossy)
+                .await
+                .map(Bson::JavaScriptCode),
+            Some(JavaScriptCodeWithScope) => {
+                // disregard the length:
+                //     using Read::take causes infinite type recursion
```
patrickfreed: what is this comment referring to?

src/bson_util/async_decode.rs
```diff
@@ -0,0 +1,214 @@
...
+                };
+
+                match Utc.timestamp_opt(sec, (msec as u32) * 1_000_000) {
+                    LocalResult::None => Err(DecoderError::InvalidTimestamp(time)),
+                    LocalResult::Ambiguous(..) => Err(DecoderError::AmbiguousTimestamp(time)),
+                    LocalResult::Single(t) => Ok(Bson::UtcDatetime(t)),
+                }
+            }
+            Some(Symbol) => read_string(reader, utf8_lossy).await.map(Bson::Symbol),
+            Some(Undefined) | Some(DbPointer) | Some(MaxKey) | Some(MinKey) | None => {
```
patrickfreed: being unable to read these seems like a problem that we will want to fix before 1.0. As with decimal128, we need to be able to at least read/round trip all the BSON types.

src/bson_util/async_encode.rs
```diff
@@ -0,0 +1,141 @@
...
+                write_i64(
+                    writer,
+                    (v.timestamp() * 1000) + (v.nanosecond() / 1_000_000) as i64,
+                )
+                .await
+            }
+            Bson::Null => Ok(()),
+            Bson::Symbol(ref v) => write_string(writer, &v).await,
+            #[cfg(feature = "decimal128")]
+            Bson::Decimal128(ref v) => write_f128(writer, v.clone()).await,
```
patrickfreed: so the decimal128 stuff is here, but is this feature flag now a feature of the driver? or is it checking the bson flag?

src/bson_util/async_decode.rs
saghm: It was a pre-existing comment from the BSON implementation. I think they're referring to the fact that using `take` to limit the number of bytes that can be read similar to how it's used elsewhere in the file. I've removed all the comments like this with the most recent commit.

src/bson_util/async_decode.rs
saghm: I don't think reading them is necessarily required for 1.0, as adding in the support later isn't a breaking change, but given how easy it would be to implement each of these four, I'm not opposed to doing it now. Would you prefer that I do it as part of this ticket or a follow-up?

src/bson_util/async_encode.rs
saghm: Good catch! I don't think that the `cfg(feature)` attribute requires the feature to exist in order to check for it, so I believe this is looking for a feature that the driver doesn't define, so it the line will just never be compiled in. I've removed it for now.

src/bson_util/async_decode.rs
patrickfreed: A follow up would be better I think.

Also, from [Bernie's comments on the roadmap](https://docs.google.com/document/d/1A48UMVSpd-l3QTQs4Alh7g5KZZ2XX9zOdqzRViOHSh4/edit?disco=AAAADiol9nc), we were actually supposed to implement [RUST-217](https://jira.mongodb.org/browse/RUST-217) for the alpha so that users could understand why decimal128s (and other types, I guess) couldn't be parsed while we hadn't implemented them. The implication with that seemed to be that once 1.0 came out, we would support decoding all those types. Either way, that's not work we need to do in this PR, but it is something to keep in mind as we get closer to 1.0.

src/bson_util/async_decode.rs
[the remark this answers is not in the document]
patrickfreed: See my new comment below w.r.t. decimal128. I was talking more about for 1.0 than in this PR specifically.

re: getting this in the BSON library: Ideally we would want maintain a single imlementation of this stuff, so I guess it would have to be the async one. For sync, I guess `bson` spins up its own runtime and blocks on it similar to how trust-dns handles their sync work? also cc @mbroadst, since the BSON stuff seems to be something we hadn't considered in the design phase.

src/bson_util/async_decode.rs
patrickfreed: I think that commit hasn't been pushed yet.

saghm: Oops; now it has been pushed

src/bson_util/async_decode.rs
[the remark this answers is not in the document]
saghm: I'm a bit on the fence about whether we want the sync BSON to wrap the async BSON like we do for the driver. The driver is already a pretty heavyweight dependency, so I don't think requiring sync users to pull in a async runtime is a huge deal, but BSON is much lighter weight, so I'd be a bit more hesitant to require the inclusion of an async runtime potentially just read and write a few bytes. As I mentioned in my earlier comment, right now I'm leaning more towards providing async functionality as either a feature flag or a separate crate. I'm not as worried about having to maintain two implementations for the BSON crate because the implementations would be a lot smaller (both due to the fact that the BSON crate is smaller than the driver and that most of the BSON crate is orthogonal to I/O).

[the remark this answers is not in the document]
patrickfreed: In Swift we decided that pulling in the runtime in the BSON library wouldn't be a huge deal, since a large majority of consumers of the BSON library will be consumers of the driver which already pulls it in. I'm unsure if that's also the case in the Rust world, but it is something to consider. Regardless, this seems like something that we should sketch out at least informally, even if the solution ends up being to just duplicate all the code under a feature flag.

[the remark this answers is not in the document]
patrickfreed: Given that we're not actually doing any blocking I/O here, could we use [`futures::io::AllowStdIo`](https://docs.rs/futures/0.3.4/futures/io/struct.AllowStdIo.html) to reduce the duplication?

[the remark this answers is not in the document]
saghm: I don't think we can assume that there won't be able blocking I/O done; `decode_document` and `encode_document` are generic over the type of reader and writer they use, and they're public functions, so theoretically they could be used with something like a stream (which is in fact how we use them in the driver).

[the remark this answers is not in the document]
univerz: > which is in fact how we use them in the driver

and that makes the driver really slow without buffering. i have strong susspicion that async machinery would make it even slower.

src/bson_util/async_decode.rs
patrickfreed: Filed [RUST-311](https://jira.mongodb.org/browse/RUST-311) for the follow on work.

src/bson_util/async_decode.rs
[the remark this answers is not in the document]
patrickfreed: As per offline discussion, we now buffer the entire document into a vec, and then subsequently use the existing sync decode implementation from `bson`.

PR #143, opened 2020-03-10
minor: fix rustfmt checks

saghm:


src/cmap/background.rs
```diff
@@ -40,10 +37,12 @@ impl ConnectionPool {
 
         let mut i = 0;
         while i < connection_manager.checked_in_connections.len() {
-            if connection_manager.checked_in_connections[i].is_stale(connection_manager.generation) {
+            if connection_manager.checked_in_connections[i].is_stale(connection_manager.generation)
+            {
```
patrickfreed: this hanging brace is kind of weird. could we configure rustfmt to break on one of the dot-operators instead?

.evergreen/check-rustfmt.sh
```diff
@@ -1,6 +1,6 @@
-#!/bin/sh
+#!/bin/bash
 
 set -o errexit
 
 . ~/.cargo/env
-cargo +nightly fmt -- --check
+rustfmt +nightly --unstable-features --check src/**/*.rs
```
patrickfreed: do we need to keep using the nightly version if we're using `unstable-features`?

saghm: `--unstable-features` is a nightly-only flag

src/cmap/background.rs
saghm: Unfortunately I think we're already set to do that, and it's failing to; see the default of the option [here](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#control_brace_style)

PR #144, opened 2020-03-10
RUST-309 Ensure runtime exists before executing background tasks

saghm:


src/runtime/mod.rs
```diff
@@ -29,20 +29,35 @@ pub(crate) enum AsyncRuntime {
 
 impl AsyncRuntime {
     /// Spawn a task in the background to run a future.
```
patrickfreed: We should update this docstring to reflect why it's returning an optional / how it's different from `execute`.

saghm: Updated

PR #146, opened 2020-03-12
minor: fix new clippy lint failures with rust 1.42.0

saghm:


src/client/auth/scram.rs
```diff
@@ -299,7 +299,7 @@ fn h_i<M: Mac + Sync>(str: &str, salt: &[u8], iterations: usize, output_size: us
 
 /// Parses a string slice of the form "<expected_key>=<body>" into "<body>", if possible.
 fn parse_kvp(str: &str, expected_key: char) -> Result<String> {
-    if str.chars().nth(0) != Some(expected_key) || str.chars().nth(1) != Some('=') {
+    if !str.starts_with(expected_key) || str.chars().nth(1) != Some('=') {
```
saghm: I'm constantly amazed at how many specific lints like this exist. It even handles the fact that it's negated!

src/bson_util.rs
```diff
@@ -27,7 +27,7 @@ pub(crate) fn to_bson_array(docs: &[Document]) -> Bson {
 
 #[cfg(test)]
 pub(crate) fn sort_document(document: &mut Document) {
-    let temp = std::mem::replace(document, Default::default());
+    let temp = std::mem::take(document);
```
saghm: I did not know that this existed, but I love it

src/client/options/mod.rs
```diff
@@ -839,7 +839,7 @@ impl ClientOptionsParser {
             }
         }
 
-        let db_str = db.as_ref().map(String::as_str);
+        let db_str = db.as_deref();
```
saghm: Also didn't know this existed

PR #147, opened 2020-03-13
minor: remove dev dependency on assert_matches

patrickfreed:
This PR removes the dependency on the third party `assert_matches` crate in favor of using the newly introduced `matches!` in Rust 1.42

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,10 @@ This repository contains the officially supported MongoDB Rust driver, a client 
 
 ## Installation
 ### Requirements
- - Rust 1.42+
+| Driver Version | Required Rust Version |
+|:--------------:|:---------------------:|
+| master         | 1.42+                 |
+| 0.9.x          | 1.39+                 |
 - MongoDB 3.6+
 
 ### Importing
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ This repository contains the officially supported MongoDB Rust driver, a client 
 
 ## Installation
 ### Requirements
- - Rust 1.39+
+ - Rust 1.42+
 - MongoDB 3.6+
 
 ### Importing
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ This repository contains the officially supported MongoDB Rust driver, a client
 
 ## Installation
 ### Requirements
- - Rust 1.39+
+ - Rust 1.42+
```
patrickfreed: I wasn't sure what our policy on updating this was. I did it now since the driver won't compile without 1.42. I'm not sure that's the right move though, since the latest released driver doesn't need 1.42, and the master branch README will be what people probably look at. Maybe this should only be updated during the release process (I think that's what we do in Swift). What do you think?

saghm: Maybe the best solution is to have a table with the minimum Rust versions for each release? Since we've already branched off of 0.9.x, we could list it per minor version

patrickfreed: That sounds like a good idea, added the chart.

nevi-me: Comment from an observer: the bump to 1.39 was a necessity due to async/await, but what will the policy likely be for ergonomics-related and smaller features? There is for example the non_exhaustive JIRA, which would bump to 1.40, and then this `matches!`. I'd understand if there isn't an answer yet :)

saghm: While we're still pre-1.0, we're planning on taking advantage of newer features whenever we do a minor version bump (e.g. 0.9.x to 0.10.0). Once we release 1.0, the API will stabilize, so we'll continue supporting whatever the latest stable version that was available at the time of release indefinitely for the 1.0 branch. We consider bumping the minimum Rust version supported to be a breaking change, but we aren't going to restrict ourselves from making breaking changes until we hit 1.0. We do plan on getting to 1.0 fairly quickly though; after we finish the async work, there are only a few more driver features we want to add, along with some work on the BSON API.

PR #149, opened 2020-03-16
RUST-300 Implement async DNS resolver

patrickfreed:
[RUST-300](https://jira.mongodb.org/browse/RUST-300)

This PR implements an async DNS resolver for the driver. 

Originally we intended to introduce a crate that provides `trust-dns` support for `async-std` as part of this work and then conditionally depend upon it. I think a more proper path would be just to introduce that functionality straight into `trust-dns` itself, similar to how it already has [builtin support for tokio](https://docs.rs/trust-dns-resolver/0.19.3/trust_dns_resolver/#using-the-tokioasync-resolver).

As far as the driver's concerned, we don't actually need that builtin support to work since we have a runtime abstraction that we can just pass to `trust-dns` instead. With the goal of advancing the driver's functionality, I decided to just implement it straight in the driver and relegate any third party library contributions to our own time if we so wish. 

As a side note: our entire `runtime` module would probably be useful as its own crate.


Cargo.toml
```diff
@@ -17,11 +17,10 @@ async-std-runtime = ["async-std", "async-std/attributes"]
 
 [dependencies]
```
patrickfreed: One change made as part of this PR is to migrate away from `tokio::io` based dependencies in favor of the `futures::io` ones. While the tokio ones are fine for our needs, it appears that the community (including `async-std` itself) is adopting the `futures::io` ones more often for runtime-agnostic contexts. (e.g. `async-tls`, `trust-dns`). 

The `AsyncWrite` / `AsyncRead` traits are nearly identical between `futures` and `tokio` (`poll_shutdown` vs `poll_close` being the main difference), so it required very little changes on our side though. `AsyncWriteExt` and `AsyncReadExt` are more fleshed out in tokio though, so I had to supplement them in the driver.

Cargo.toml
```diff
@@ -54,7 +52,7 @@ version = "1.5.0"
 optional = true
 
 [dependencies.rustls]
-version = "0.17.0"
+version = "0.16.0"
```
patrickfreed: unfortunately, this is required for compatibility with `async-tls`. They have an [issue](https://github.com/async-rs/async-tls/issues/20) tracking the update to fix this.

src/bson_util/async_encoding.rs
```diff
@@ -1,27 +1,32 @@
-use bson::{DecoderResult, Document, EncoderResult};
-use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
-use tokio_byteorder::{AsyncReadBytesExt, AsyncWriteBytesExt, LittleEndian};
+use bson::Document;
+use futures::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
+
+use crate::{
+    error::Result,
+    runtime::{AsyncLittleEndianRead, AsyncLittleEndianWrite},
```
patrickfreed: `futures::io::Async{Read|Write}Ext` don't have the helpers for each integer type that tokio does, so I implemented them in the driver. 

This also helps guarantee that we use little endian bye order for all our reading and writing, which I think we were not doing in `cmap/conn/wire/message.rs` due to tokios helpers for these all reading/writing in big endian. The places where we were explicitly specifying little endian were of course already okay. I think the reason it didn't show up in any of our testing is that we were only using big endian for the flag bits and the checksum, neither of which we actually use regularly (at all?).

src/cmap/wait_queue.rs
```diff
@@ -55,12 +55,11 @@ impl WaitQueue {
         let future = self.semaphore.acquire(1);
 
         let releaser = if let Some(timeout) = self.timeout {
-            RUNTIME
-                .await_with_timeout(Box::pin(future), timeout)
-                .await
-                .map_err(|_| ErrorKind::WaitQueueTimeoutError {
+            RUNTIME.timeout(timeout, future).await.map_err(|_| {
```
patrickfreed: now that we use tokio/async-std's builtin timeouts, we don't need the `Box` anymore, which is nice.

src/runtime/mod.rs
```diff
@@ -75,20 +81,23 @@ impl AsyncRuntime {
     }
 
     /// Await on a future for a maximum amount of time before returning an error.
-    pub(crate) async fn await_with_timeout<F>(
+    pub(crate) async fn timeout<F: Future>(
```
patrickfreed: this was required for the resolver, and I figured it was a better replacement for the one I wrote.

src/runtime/resolver.rs
```diff
@@ -0,0 +1,160 @@
...
+    error::{Error, Result},
+    runtime::AsyncRuntime,
+    RUNTIME,
+};
+
+// Some shorter type names for the generic constraints of the underlying trust-dns
+// `AsyncResolver`.
+type AsyncConnectionProvider = GenericConnectionProvider<AsyncRuntimeProvider>;
+type TrustDnsResolver =
+    trust_dns_resolver::AsyncResolver<GenericConnection, AsyncConnectionProvider>;
```
patrickfreed: trust-dns already has built in support for tokio, but since I had to implement all of this anyways for async-std, it was easier to not conditionally use the built in support and to just have a single type that relies on our agnostic runtime.

src/runtime/resolver.rs
```diff
@@ -0,0 +1,160 @@
...
+    async fn connect(addr: SocketAddr) -> std::io::Result<Self::Transport> {
+        AsyncTcpStream::connect_socket_addr(&addr, None)
+            .await
+            .map_err(Error::into_io_error)
+    }
+}
+
+/// A runtime-agnostic UdpSocket necessary for trustdns' resolver.
+/// This does not need to be used elsewhere in the driver.
+enum AsyncUdpSocket {
```
patrickfreed: this is defined here instead of its own module because I can't imagine we'd need UDP anywhere else in the driver. Let me know if you think it should still have its own private module or something, or if this is fine.

src/runtime/stream.rs
```diff
@@ -127,12 +133,11 @@ impl AsyncStream {
         // If there are TLS options, wrap the inner stream with rustls.
         match options.tls_options {
             Some(cfg) => {
-                let name = DNSNameRef::try_from_ascii_str(&options.address.hostname)?;
```
patrickfreed: `connector.connect` [does this for us now, and it automatically converts the error to an io error](https://docs.rs/async-tls/0.6.0/src/async_tls/connector.rs.html#109). Because of that, I removed the `InvalidDnsName` error type as it is no longer being returned anywhere.

Cargo.toml
patrickfreed: as discussed offline, we'll just keep using tokio-rustls.

src/bson_util/async_encoding.rs
saghm: Do we still want top keep these even though we're continuing to use tokio? The benefit you pointed out still applies, so I'm not opposed to keeping them

src/bson_util/async_encoding.rs
```diff
@@ -1,27 +1,32 @@
...
+};
 
 pub(crate) async fn decode_document<R: AsyncRead + Unpin + Send>(
     reader: &mut R,
-) -> DecoderResult<Document> {
-    let length = AsyncReadBytesExt::read_i32::<LittleEndian>(reader).await?;
```
saghm: It is nice to not need to disambiguate between with `read_i32` we mean

src/runtime/async_read_ext.rs
```diff
@@ -0,0 +1,30 @@
+use async_trait::async_trait;
+use futures::io::AsyncRead;
+
+use crate::error::Result;
+
+/// Trait providing helpers that read various integer types in little-endian order.
+#[async_trait]
+pub(crate) trait AsyncLittleEndianRead: Unpin + futures::io::AsyncReadExt {
```
saghm: One thing I've seen done before for traits that should only be implemented generically (like done below) is to make a private subtrait that this trait requires, e.g.

```rust
pub(crate) trait AsyncLittleEndianRead: Sealed + Unpin + futures::io::AsyncReadExt { ... }

trait Sealed {}

impl<R: AsyncRead + ?Sized + Unpin> Sealed for R {} 
```

This would prevent any outside implementations of `AsyncLittleEndianRead`. It's probably overkill given that this isn't a public trait, but I figured it was worth mentioning the pattern anyways.

src/runtime/resolver.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    pub async fn txt_lookup<N: IntoName>(&self, query: N) -> Result<TxtLookup> {
+        let lookup = self.resolver.txt_lookup(query).await?;
+        Ok(lookup)
+    }
+}
+
+/// The provider type used by trustdns to operate using the proper async runtime.
+#[derive(Copy, Clone)]
+struct AsyncRuntimeProvider {}
```
saghm: Interestingly, in Rust there's a slight difference between a struct with zero fields and a singleton struct; in this case, I think it's more idiomatic to use the latter (i.e. `struct AsyncRuntimeProvider;` rather than `struct AsyncRuntimeProvider {}`).

src/test/atlas_connectivity.rs
```diff
@@ -25,17 +23,14 @@ fn run_test(uri_env_var: &str) {
...
 async fn atlas_repl_set() {
     run_test("MONGO_ATLAS_FREE_TIER_REPL_URI");
 }
 
-// TODO RUST-300: re-enable these tests once async SRV resolution is implemented.
-// #[cfg_attr(feature = "tokio-runtime", tokio::test(core_threads = 2))]
-// #[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[cfg_attr(feature = "tokio-runtime", tokio::test(core_threads = 2))]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
 #[allow(dead_code)]
```
saghm: We can remove the dead_code annotation now too, right?

src/bson_util/async_encoding.rs
saghm: I think we could still use tokio_byteorder if we wanted, given that we implement the tokio traits. It's not huge deal to implement our own version though, so I'm fine with leaving it.

src/client/options/mod.rs
```diff
@@ -466,8 +467,8 @@ impl TlsOptions {
                 }
             };
 
-            // TODO: Get rid of unwrap
-            config.set_single_client_cert(certs, key.into_iter().next().unwrap())?;
+            // TODO: Get rid of unwrap.
```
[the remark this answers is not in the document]
saghm: Interesting! Apologies for both misreading the diff and not realizing that I had broken my own rule with the original comment.

saghm:
Looks like the changes I made to the evergreen project caused this not to get run, but I think it should work fine for new pull requests. For now, can you do a manual patch to verify this? You can merge it assuming it passes.

patrickfreed:
Patch running [here](https://evergreen.mongodb.com/version/5e726b582a60ed6b3b8dd0a7).

saghm:
Oh, looks like it's pulling in the newer BSON library, which is causes clippy failures. I think you should be able to cherry-pick the commit the commit from the 0.9.x branch to fix this. To keep it as a separate commit in the history, you can squash before cherry-picking and then use the "rebase and merge" option for merging this in, which will take all of the commits in the PR and put them on top of master as-is.

patrickfreed:
New [patch](https://evergreen.mongodb.com/version/5e72839b3e8e86373861356b) with cherry-picked bson commit and a [skip for the failing test](https://github.com/patrickfreed/mongo-rust-driver/blob/RUST-300/async-dns/src/test/spec/initial_dns_seedlist_discovery.rs#L35). Since we don't have a ticket for it yet, I just linked the DRIVERS ticket in the TODO.

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -28,16 +28,19 @@ struct ParsedOptions {
...
-// TODO RUST-300: re-enable these tests once async SRV resolution is implemented.
-// #[cfg_attr(feature = "tokio-runtime", tokio::test(core_threads = 2))]
-// #[cfg_attr(feature = "async-std-runtime", async_std::test)]
-#[allow(dead_code)]
+#[cfg_attr(feature = "tokio-runtime", tokio::test(core_threads = 2))]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
 async fn run() {
     let run_test = |mut test_file: TestFile| {
+        // TODO DRIVERS-796: unskip this test
+        if test_file.uri == "mongodb+srv://test5.test.build.10gen.cc/?authSource=otherDB" {
```
saghm: Does the test have a name we can use to skip instead of hard-coding the URI?

patrickfreed: Unfortunately not. The seedlist discovery tests only include a "comment" string for tests that are supposed to error. We could skip based on test file name, but that would require changes to the spec test runner function.

saghm: I see. This is better than skipping based on the file name.

PR #151, opened 2020-03-19
RUST-300 Update StreamAddress resolution to be async

patrickfreed:
This is a follow up PR for [RUST-300](https://jira.mongodb.org/browse/RUST-300)

This previous PR (#149) only updated the SRV/TXT DNS resolution to be async. This PR updates the regular DNS resolution to also be async by replacing the usage of `std::net::ToSocketAddrs` with the async-std/tokio equivalents.

### DIFF
--- a/src/runtime/mod.rs
+++ b/src/runtime/mod.rs
@@ -4,7 +4,7 @@ mod join_handle;
 mod resolver;
 mod stream;
 
-use std::{future::Future, time::Duration};
+use std::{future::Future, net::SocketAddr, time::Duration};
 
 pub(crate) use self::{
     async_read_ext::AsyncLittleEndianRead,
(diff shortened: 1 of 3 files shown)


src/runtime/mod.rs
```diff
@@ -100,4 +103,24 @@ impl AsyncRuntime {
                 .map_err(|_| ErrorKind::Io(std::io::ErrorKind::TimedOut.into()).into())
         }
     }
+
+    pub(crate) async fn resolve_address(
+        self,
+        address: &StreamAddress,
+    ) -> Result<impl Iterator<Item = SocketAddr>> {
```
patrickfreed: tokio's [version of the `ToSocketAddrs` trait](https://docs.rs/tokio/0.2.13/tokio/net/trait.ToSocketAddrs.html) is opaque, so you need to use `lookup_host` instead invoking trait methods directly. async-std's is basically just the same thing.

patrickfreed:
evergreen patch [here](https://evergreen.mongodb.com/version/5e73ef0e2a60ed78fb78ea2f)

PR #152, opened 2020-03-20
RUST-324 Add async cursor API

saghm:


src/operation/get_more/mod.rs
```diff
@@ -15,7 +15,7 @@ use crate::{
     Namespace,
 };
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
saghm: This is needed so that we can send owned GetMore operations to `execute_operation` as described above.

src/cursor.rs
```diff
@@ -90,32 +121,92 @@ impl Drop for Cursor {
...
 
-        let _ = self.client.database(&namespace.db).run_command(
-            doc! {
-                "killCursors": &namespace.coll,
-                "cursors": [self.get_more.cursor_id()]
-            },
-            None,
-        );
+        RUNTIME.execute(async move {
+            let _: Result<_> = client.database(&namespace.db).run_command(
```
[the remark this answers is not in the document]
patrickfreed: Yeah that sounds like a good idea.

src/cursor.rs
```diff
@@ -21,48 +37,63 @@ use crate::{error::Result, operation::GetMore, options::StreamAddress, Client, N
...
 /// ```
 #[derive(Debug)]
 pub struct Cursor {
     client: Client,
-    buffer: VecDeque<Document>,
     get_more: GetMore,
     exhausted: bool,
+
+    /// This will be None when the cursor is exhausted.
+    state: Option<PollState>,
```
patrickfreed: Rather than storing this as an `Option<PollState>`, I think it would be more straightforward if we just introduced a third state called `Exhausted`. This would remove the need to document what `None` means here and below. Given that `Exhausted` wouldn't necessarily be a state of polling, it may make sense to rename it to `CursorState` or something.

src/cursor.rs
```diff
@@ -90,32 +121,92 @@ impl Drop for Cursor {
...
+                // If the getMore is finished and successful, then we pop off the first document
+                // from the batch, set the poll state to Buffer, record whether the cursor is
+                // exhausted, and return the popped document.
+                Poll::Ready(Ok(get_more_result)) => {
+                    let mut buffer: VecDeque<_> = get_more_result.batch.into_iter().collect();
+                    let next_doc = buffer.pop_front();
+
+                    self.state = Some(PollState::Buffer(buffer));
                     self.exhausted = get_more_result.exhausted;
+                    Poll::Ready(next_doc.map(Ok))
```
patrickfreed: So this looks like we're doing the "tryNext" behavior here. After lots of discussion in Swift we ended up defaulting to the resending-getMore-on-empty-batch behavior for `next` while also offering a `tryNext` method if users wanted it. Do we plan on doing something similar here?

saghm: Right now, we don't have tailable cursors, so there wouldn't be any user-facing difference regardless of the behavior we implement, correct? If so, we might as well wait until we implement tailable cursors to decide.

src/cursor.rs
saghm: I've updated it as you suggested. Since the enum is private to the cursor module, I just named it `State`. I also needed to move things around in the `poll_next` implementation a bit; because we're not using an option anymore, we can't match on `self.state.take()`, so I had to switch all of the captures to take references and then resolve all of the borrow checker errors that resulted.

src/cursor.rs
patrickfreed: We currently support tailable cursors via setting the cursor type on `FindOptions`. You can just keep calling `next` on the resulting cursor even after it returns an empty batch, since it'll never return `exhausted: true` from a getmore.

saghm: Ah, true. Users generally won't be calling `poll_next` directly due to ergonomic reasons; it's mostly used under the hood in the `StreamExt` trait, where `next` is defined using it. If we implement `poll_next` to eagerly resend the getMore, I don't think there's any way we could define a higher-level alternative to `StreamExt::next` that doesn't eagerly send it. Since I wasn't involved in any of the Swift conversations, I think it would be helpful for me to ask you both about what the intended use cases are of `next` versus `tryNext` in Swift (although even if we do the same thing, we'll need to name them differently, since `StreamExt::try_next` is already a thing that means something different). Specifically, I'm not sure if an eagerly-getMore-sending `next` is intended to be used with the higher-level constructs that `StreamExt` provides (e.g. `map` and `collect`) or if it's just intended to be called directly. If the latter, I think it would make more sense to do the reverse of what Swift does and have `poll_next` not eagerly resend getMore, but if the former, we might need to do something more complex like having two separate cursors types (one eager and one lazy) and provide methods to convert them back and forth.

src/cursor/impatient.rs
```diff
@@ -3,81 +3,17 @@ use std::{
...
-/// # let coll = client.database("foo").collection("bar");
-/// # let cursor = coll.find(Some(doc! { "x": 1 }), None)?;
-/// #
-/// let results: Vec<Result<Document>> = cursor.collect().await;
-/// # Ok(())
-/// # }
-/// ```
 #[derive(Debug)]
-pub struct Cursor {
+pub struct ImpatientCursor {
```
saghm: Final choice of name pending tomorrow's discussion

src/cursor/mod.rs
```diff
@@ -21,101 +30,73 @@ use crate::{error::Result, operation::GetMore, options::StreamAddress, Client, N
...
     type Item = Result<Document>;
 
-    fn next(&mut self) -> Option<Self::Item> {
-        if self.buffer.is_empty() && !self.exhausted {
-            match RUNTIME.block_on(self.client.execute_operation(&self.get_more, None)) {
-                Ok(get_more_result) => {
-                    self.buffer.extend(get_more_result.batch);
-                    self.exhausted = get_more_result.exhausted;
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
+        if self.tailable {
```
patrickfreed: I don't think we need a special flag for tailable. I think we can instead just check if the underlying cursor is still alive. As it is right now, once a tailable cursor dies / is exhausted, this will loop forever.

As a side note, we should probably return an error if a user tries to iterate a dead cursor. We have to be careful not to return an error if the cursor dies after we start repeatedly polling it though.

saghm: I've made the suggested changes except for the part about returning an error when the cursor is exhausted. Is this something we really want to do? I don't think I've seen any other iterators in Rust do that.

patrickfreed: I just mentioned it since it's something we do in Swift. Note that I'm only referring to returning errors when beginning a new iteration the cursor after it has indicated it is done, not just whenever it becomes exhausted a la Python.The rationale for doing so is that a user iterating a dead cursor is probably indicative of a programmer error:
- In the "patient" case, it could only happen if they're not using a `while let` loop or if they start iterating after their `while let` loop ended, both likely errors.
- In the "non-patient" case, it means they're not distinguishing between a cursor being exhausted and returning `None` (like we were doing here), which is also usually an error since it can lead to infinite loops.

All that being said, I'm completely fine with not including it here and discussing it later.

src/cursor.rs
patrickfreed: To summarize offline conversation, we've elected to go with a two cursor strategy:
- Default `Cursor` implements the keep-resending-getMore-until-nonempty-batch behavior
- `Cursor` has a method on it that takes ownership and returns a new wrapping cursor (temporarily has type name `NonPatientCursor`) which does not resend getMore and instead returns empty batches.

src/cursor/mod.rs
patrickfreed: I've filed [RUST-337](https://jira.mongodb.org/browse/RUST-337) to discuss it.

PR #153, opened 2020-03-21
RUST-322 Implement async Client / Database API

patrickfreed:
[RUST-322](https://jira.mongodb.org/browse/RUST-322)

This PR updates the client / database API to be async.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -19,7 +19,6 @@ use crate::{
     operation::ListDatabases,
     options::{ClientOptions, DatabaseOptions, ReadPreference, SelectionCriteria},
     sdam::{Server, Topology},
-    RUNTIME,
 };
 
 const DEFAULT_SERVER_SELECTION_TIMEOUT: Duration = Duration::from_secs(30);
@@ -34,8 +33,8 @@ const DEFAULT_SERVER_SELECTION_TIMEOUT: Duration = Duration::from_secs(30);
 /// ```rust
 /// # use mongodb::{Client, error::Result};
 /// #
-/// # fn start_workers() -> Result<()> {
-/// let client = Client::with_uri_str("mongodb://example.com")?;
+/// # async fn start_workers() -> Result<()> {
+/// let client = Client::with_uri_str("mongodb://example.com").await?;
 ///
 /// for i in 0..5 {
 ///     let client_ref = client.clone();
@@ -71,8 +70,8 @@ impl Client {
     ///
     /// See the documentation on
     /// [`ClientOptions::parse`](options/struct.ClientOptions.html#method.parse) for more details.
-    pub fn with_uri_str(uri: &str) -> Result<Self> {
-        let options = ClientOptions::parse(uri)?;
+    pub async fn with_uri_str(uri: &str) -> Result<Self> {
+        let options = ClientOptions::parse(uri).await?;
 
         Client::with_options(options)
     }
@@ -82,7 +81,7 @@ impl Client {
         options.validate()?;
 
         let inner = Arc::new(ClientInner {
-            topology: RUNTIME.block_on(Topology::new(options.clone()))?,
+            topology: Topology::new(options.clone())?,
             options,
         });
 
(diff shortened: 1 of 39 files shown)


src/client/mod.rs
```diff
@@ -82,7 +81,7 @@ impl Client {
         options.validate()?;
 
         let inner = Arc::new(ClientInner {
-            topology: RUNTIME.block_on(Topology::new(options.clone()))?,
+            topology: Topology::new(options.clone())?,
```
patrickfreed: `Topology::new` didn't need to be async, since the only blocking behavior it had was waiting on a lock that it created itself. I refactored some stuff to make acquiring that lock not necessary, and this subsequently allowed me to make `Client::with_options` not async.

src/client/options/test.rs
```diff
@@ -239,7 +240,9 @@ fn run_test(test_file: TestFile) {
                 });
 
                 if !is_unsupported_host_type {
-                    let options = ClientOptions::parse(&test_case.uri).unwrap();
+                    let options = RUNTIME
```
patrickfreed: spec test functions can't be async until `run_spec_tests` is updated. I figured doing so would be out of scope for this work, so for now I just block on any async methods.

src/coll/mod.rs
```diff
@@ -47,8 +48,8 @@ const MAX_INSERT_DOCS_BYTES: usize = 16 * 1000 * 1000;
 /// # use bson::{bson, doc};
 /// # use mongodb::{Client, error::Result};
 /// #
-/// # fn start_workers() -> Result<()> {
-/// # let client = Client::with_uri_str("mongodb://example.com")?;
+/// # async fn start_workers() -> Result<()> {
```
patrickfreed: these examples show using the various objects across threads. Now that we're async, these should probably be using tasks, but I wasn't sure how to represent that in a runtime agnostic way.

src/sdam/state/mod.rs
```diff
@@ -65,19 +65,20 @@ impl Topology {
...
-        let state = Arc::new(RwLock::new(TopologyState {
+        let mut topology_state = TopologyState {
             description,
             servers: Default::default(),
-        }));
+        };
 
-        let topology = Topology { state, common };
+        for address in hosts.iter() {
+            topology_state.add_new_server(address.clone(), options.clone());
```
patrickfreed: I moved the weak reference to the topology off of `Server` and onto `Monitor` to allow us to create the servers before the topology itself was created. That way, we can start the monitors without having to grab the lock here.

src/sdam/state/mod.rs
```diff
@@ -214,21 +215,26 @@ impl Topology {
         // doesn't check the fields of an Unknown server, and we only return Unknown server
         // descriptions when errors occur. Once we implement SDAM monitoring, we can
         // properly inform users of errors that occur here.
-        let _ = state_clone.update(server_description, self.clone());
+        let _ = state_clone.update(server_description, self.common.options.clone());
 
-        if old_description == state_clone.description {
-            return false;
-        }
+        match old_description.diff(&state_clone.description) {
```
patrickfreed: because `add_new_server` isn't starting the monitors anymore, we need to know which servers were added by the latest update to the topology. This diffing might be a little over engineered, but I figured it might also be useful for SDAM monitoring in the future.

src/client/options/test.rs
saghm: That was one of the things I ran into when I first tried to convert the entire API in one go.

src/coll/mod.rs
saghm: That's a good question. For now, let's leave it, but can we file a ticket so we can revisit before releasing?

src/client/mod.rs
saghm: Good work on this! I'm honestly not sure this will ever need to be called in a non-async context, but the code is definitely cleaner regardless.

src/coll/mod.rs
patrickfreed: filed [RUST-329](https://jira.mongodb.org/browse/RUST-329)

PR #154, opened 2020-03-23
Rust 275

anthonyjchriste:
This PR includes two changes:

* The first is an update to the SCRAM conversation as defined by: https://github.com/mongodb/specifications/commit/d02b627adbc77bb52ca26465a95ef1e37bb5960a#diff-4ae4e931060899b3147a7255caadd990 which introduces a new top level "options" field to the "saslStart" command.
* The second is removing the done check in validate which allows this client to authenticate with AWS DocumentDB.

It should be noted that updating the SCRAM conversation causes a new error to be returned:

> OrderedDocument({"saslStart": I32(1), "mechanism": String("SCRAM-SHA-1"), "payload": BinData(0, <redacted, is this sensitive?>), "options": Document(OrderedDocument({"skipEmptyExchange": Boolean(true)}))})


saghm:
If you look at the section starting [here](https://github.com/mongodb/mongo-rust-driver/pull/154/files#diff-909a7ab3b25df6b1da57330056201c10L142-L149) (you'll need to click the "expand" button at the top left of the file a few times to see it; Github unfortunately won't let me comment on lines of code that weren't changed, and it doesn't seem to be smart enough to expand automatically to the linked section), the driver is doing a final round-trip after the regular SCRAM conversation has finished. This is the part that the new server versions are going to get rid of (and that DocumentDB doesn't perform). Right now, it'll still happening regardless of the server's previous response, which presumably will cause an error due to the server not expecting another message in the conversation. The fix would be to return early from that function right after the call to `validate` if the server response contains `done: true`.

anthonyjchriste:
Thank you for the information. 

The error I'm getting actually occurs during the parsing of server_first (https://github.com/anthonyjchriste/mongo-rust-driver/blob/RUST-275/src/client/auth/scram.rs#L102) which calls validate_command_success (https://github.com/anthonyjchriste/mongo-rust-driver/blob/RUST-275/src/client/auth/scram.rs#L309) which eventually fails.

The reason validate_command_success fails is because the server_first_response (https://github.com/anthonyjchriste/mongo-rust-driver/blob/RUST-275/src/client/auth/scram.rs#L101) indicates the following error when include the new top level "options" field in the "saslStart" command. 

This is what server_first_response looks like on my end:

```
[/home/opq/IdeaProjects/mongo-rust-driver/src/client/auth/scram.rs:103] &server_first_response = CommandResponse {
    source: StreamAddress {
        hostname: "localhost",
        port: Some(
            27017,
        ),
    },
    raw_response: OrderedDocument({"ok": FloatingPoint(0.0), "errmsg": String(" : { saslStart: 1, mechanism: \'SCRAM-SHA-1\', payload: \'...\' }"), "code": I32(37)}),
}
```
Is it possible that DocumentDB is validating all messages to fit a certain schema? It's also entirely possible that I am misunderstanding the SCRAM conversation and changes to it.

As always, I appreciate the patience.

saghm:
Hmm, I'm not getting that error message when I run against MongoDB 4.2. I also have never seen an error message that just shows the document and doesn't give any other info. Error code 37 also doesn't seem to be referring to a MongoDB thing; although the server doesn't currently offer a stable error-code API, at least as of the 4.2 branch, error code 37 appears to refer to [FileRenameFailed](https://github.com/mongodb/mongo/blob/r4.2.0/src/mongo/base/error_codes.err#L40).

anthonyjchriste:
Fascinating. I'll install a local version of MongoDB and configure it to be as similar as possible to DocumentDB. It would make more sense for me to test against both.

andrewleverette:
Has there been any movement on this pull request? I find myself in a similar situation to Anthony where I need to connect to DocumentDB, which uses MongoDB v3.6, via tls. The documentation seems straight forward but it doesn't connect. I get the following error `AuthenticationError { message: "SCRAM failure: handshake terminated early" }`. Any guidance on how to connect to DocumentDB would be great. For more info, here is the connection string that i'm using: `mongodb://username:password@host:27017/?tls=true&tlsCAFile=ca_file.pem&replicaSet=rs0&readPreference=secondaryPreferred&retryWrites=false`. This connection string is the one recommended by DocumentDB except it uses ssl options instead of tls.

anthonyjchriste:
@andrewleverette The MongoDB team has recently added support for the shorter scram conversation. I believe this support is in the latest master branch. I've been able to successfully connect to DocumentDB as of this past week from the master branch.

saghm:
For additional context, DocumentDB does not actually use MongoDB 3.6, but attempts to emulate the same API. MongoDB drivers do not officially support DocumentDB and are not tested with it. That said, we did implement shortening the SCRAM conversation for use with MongoDB [in this commit](https://github.com/mongodb/mongo-rust-driver/commit/57794c0d6ed424d55e5f13539d00410e0b205a9e), so it's possible that the issue you were having might be resolved now.

andrewleverette:
@anthonyjchriste and @saghm Thank you both for you comments. I was able to track those changes down in the master branch. So for now, I can just point Cargo.toml file to the master branch for this dependency and it works exactly like I expected. Thanks, looking forward to a new release with these changes.

patrickfreed:
Closing this PR now that #239 is merged and seems to have resolved the issue in RUST-275. Nonetheless, thank you for your help in investigating and working on this @anthonyjchriste !

PR #155, opened 2020-03-25
RUST-331 Replace futures_timer with runtime-specific delays

saghm:


src/sdam/message_manager.rs
```diff
@@ -65,7 +66,7 @@ impl TopologyMessageManager {
 }
 
 async fn wait_for_notification(receiver: &mut Receiver<()>, timeout: Duration) -> bool {
-    let timeout = Delay::new(timeout);
+    let timeout = RUNTIME.delay_for(timeout);
     let message_received = Box::pin(receiver.recv());
 
     match futures::future::select(timeout, message_received).await {
```
patrickfreed: I think if we replace this `select` with `RUNTIME.timeout`, we can remove the need for `BoxFuture` on `delay_for` and just make it a regular async method.

saghm: Updated

patrickfreed: And now `delay_for` can just be `async`, right?

saghm: Ah, yep. We may need to wrap the return value in a box depending on how we use it, but no need to do that in this case.

PR #158, opened 2020-04-02
RUST-323 Implement async Collection API

patrickfreed:
[RUST-323](https://jira.mongodb.org/browse/RUST-323)

This PR implements the async `Collection` API. It also updates most of the test suite to be async.

### DIFF
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -28,7 +28,6 @@ use crate::{
     Client,
     Cursor,
     Database,
-    RUNTIME,
 };
 
 /// Maximum size in bytes of an insert batch.
@@ -196,7 +195,7 @@ impl Collection {
     ///
     /// Note that using [`Collection::estimated_document_count`](#method.estimated_document_count)
     /// is recommended instead of this method is most cases.
-    pub fn count_documents(
+    pub async fn count_documents(
         &self,
         filter: impl Into<Option<Document>>,
         options: impl Into<Option<CountOptions>>,
@@ -234,7 +233,7 @@ impl Collection {
                 .build()
         });
 
-        let result = match RUNTIME.block_on(RUNTIME.block_on(self.aggregate(pipeline, aggregate_options))?.next()) {
+        let result = match self.aggregate(pipeline, aggregate_options).await?.next().await {
             Some(doc) => doc?,
             None => return Ok(0),
         };
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -555,29 +555,6 @@ impl Collection {
         let update = Update::new(self.namespace(), query, update.into(), false, options);
         self.client().execute_operation(&update, None).await
     }
-
-    /// Creates the indexes specified by `models`.
-    pub async fn create_indexes(
-        &self,
-        models: impl IntoIterator<Item = IndexModel>,
-    ) -> Result<Vec<String>> {
-        unimplemented!()
-    }
-
-    /// Drops the index specified by `name`.
-    pub async fn drop_index(&self, name: &str) -> Result<Document> {
-        unimplemented!()
-    }
-
-    /// Drops the index with the given `keys`.
-    pub async fn drop_index_with_keys(&self, keys: Document) -> Result<Document> {
-        unimplemented!()
-    }
-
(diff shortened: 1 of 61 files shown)


src/test/spec/command_monitoring/operation.rs
```diff
@@ -267,18 +281,39 @@ pub(super) struct UpdateOne {
...
+        &self,
+        operation: AnyTestOperation,
+        database_name: &str,
+        collection_name: &str,
+    ) -> Vec<CommandEvent> {
+        let _: Result<_> = operation
+            .execute(self.database(database_name).collection(collection_name))
+            .await;
+        self.command_events
+            .write()
```
patrickfreed, line 313: All these events still use regular sync locks. I experimented with updating the event handlers to fire off background tasks for populating the event lists, but it was too nondeterministic for testing purposes. Because we're using a different client for each test nowadays, I think the risk for deadlock is minimal, so I decided to just leave all the tests' sync event handling code alone until we decide something else. Filed [RUST-340](https://jira.mongodb.org/browse/RUST-340) to discuss options there.

src/coll/mod.rs
```diff
@@ -546,29 +553,29 @@ impl Collection {
...
-    pub fn create_indexes(
+    pub async fn create_indexes(
         &self,
         models: impl IntoIterator<Item = IndexModel>,
     ) -> Result<Vec<String>> {
         unimplemented!()
     }
 
     /// Drops the index specified by `name`.
-    pub fn drop_index(&self, name: &str) -> Result<Document> {
```
saghm, line 561: We said we wanted to get rid of these since they aren't implemented, right? It might make sense to do it now, since we're already making a bunch of other breaking changes here.

src/test/spec/mod.rs
```diff
@@ -8,16 +8,18 @@ mod read_write_concern;
...
 
 use bson::Bson;
 use serde::Deserialize;
 use serde_json::Value;
 
-pub(crate) fn run_spec_test<'a, T, F>(spec: &[&str], run_test_file: F)
+pub(crate) async fn run_spec_test<'a, T, F, G>(spec: &[&str], run_test_file: F)
 where
-    F: Fn(T),
+    F: Fn(T) -> G,
```
saghm, line 21: Out of curiosity, would it let you do `F: Fn(T) -> impl Future<Output = ()>`? I'm not sure how impl trait plays in generic bounds, but if ti works, it will save a generic parameter in a function that's started to get a little unwieldy.

src/coll/mod.rs
patrickfreed: done

src/test/spec/connection_stepdown.rs
```diff
@@ -32,39 +33,39 @@ fn run_test(name: &str, test: impl Fn(EventClient, Database, Collection)) {
...
-                    CreateCollectionOptions::builder()
-                        .write_concern(wc_majority)
-                        .build(),
-                ),
-            ),
-        )
-        .unwrap();
-
-    test(client, db, coll);
+    let _ = coll
```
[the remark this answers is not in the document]
patrickfreed: fixed

src/test/spec/mod.rs
patrickfreed: That's what I originally tried, but unfortunately it fails with the following message:
```
error[E0562]: `impl Trait` not allowed outside of function and inherent method return types
  --> src/test/spec/mod.rs:21:17
   |
21 |     F: Fn(T) -> impl Future<Output = ()>,
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^
```

patrickfreed:
Running the tests locally without the thread requirement seems to work. Running a [patch](https://evergreen.mongodb.com/version/5e8688589ccd4e391d09e931) right now to see if it's consistent. The locking on events has a higher probability of being a problem if we don't have backup threads I think, at least for CMAP.

src/test/spec/mod.rs
saghm: Oh well. Hopefully they eventually make it so this works.

saghm:
The patch you linked to looks green; once you push those changes, I think it's ready to merge

PR #160, opened 2020-04-06
RUST-215 Add sync API

patrickfreed:
[RUST-215](https://jira.mongodb.org/browse/RUST-215)

This PR adds the sync API to the driver, located in the `mongodb::sync` module. It also adds some basic tests of the sync API.


src/lib.rs
```diff
@@ -82,8 +82,8 @@ macro_rules! define_if_single_runtime_enabled {
     }
 }
 
-// In the case that neither tokio  or async-std is enabled, we want to disable all compiler errors
-// and warnings other than our custom ones.:w
+// In the case that neither tokio nor async-std is enabled, we want to disable all compiler errors
```
saghm, line 85: Good catch

src/lib.rs
```diff
@@ -82,8 +82,8 @@ macro_rules! define_if_single_runtime_enabled {
     }
 }
 
-// In the case that neither tokio  or async-std is enabled, we want to disable all compiler errors
-// and warnings other than our custom ones.:w
```
saghm, line 86: Wow, I can't believe I committed that. Oops...

src/runtime/mod.rs
```diff
@@ -18,6 +18,18 @@ use crate::{
     options::StreamAddress,
 };
 
+#[cfg(feature = "tokio-runtime")]
+lazy_static::lazy_static! {
+    static ref SYNC_RUNTIME: std::sync::Mutex<tokio::runtime::Runtime> = {
```
saghm, line 23: Calling this `SYNC_RUNTIME` feels a little weird; I know that it's only used in the sync API, but it sounds like it's saying its a synchronous runtime, which it's not. Maybe we could call this `GLOBAL_RUNTIME`?

src/runtime/mod.rs
```diff
@@ -18,6 +18,18 @@ use crate::{
...
+        let runtime = tokio::runtime::Builder::new()
+            .threaded_scheduler()
```
saghm, line 25: This defaults to using a thread per CPU core. I'm wondering if this is what we want, considering that the user might be using other sync wrappers on top of async implementations (e.g. trust-dns or reqwest). If each of those also starts a thread per CPU, would we be overloading it a bit? This feels like a bit of a prisoner's dilemma in terms of each library getting the CPU time it needs.

src/runtime/mod.rs
```diff
@@ -34,28 +46,30 @@ impl AsyncRuntime {
...
         match self {
             #[cfg(feature = "tokio-runtime")]
-            Self::Tokio => {
-                use tokio::runtime::Handle;
-
-                let handle = Handle::try_current().ok()?;
-
-                Some(AsyncJoinHandle::Tokio(handle.spawn(fut)))
-            }
+            Self::Tokio => match TokioCallingContext::current() {
```
saghm, line 58: Nice, I like this solution

src/sync/client.rs
```diff
@@ -0,0 +1,109 @@
...
+use super::Database;
+use crate::{
+    concern::{ReadConcern, WriteConcern},
+    error::Result,
+    options::{ClientOptions, DatabaseOptions, SelectionCriteria},
+    Client as AsyncClient,
+    RUNTIME,
+};
+
+/// This is the main entry point for the API. A `Client` is used to connect to a MongoDB cluster.
```
saghm, line 12: Do we want to document anything about how this wraps the async client? If so, how detailed should we get about how we start the runtime under the hood?

src/sync/coll.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// Runs an aggregation operation.
+    ///
+    /// See the documentation [here](https://docs.mongodb.com/manual/aggregation/) for more
+    /// information on aggregations.
+    pub fn aggregate(
+        &self,
+        pipeline: impl IntoIterator<Item = Document>,
+        options: impl Into<Option<AggregateOptions>>,
+    ) -> Result<Cursor> {
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
```
saghm, line 119: I think we can just pass the pipeline in directly, right? The async collection's `aggregate` method also accepts `impl IntoIterator<Item = Document>`.

src/sync/db.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// Runs an aggregation operation.
+    ///
+    /// See the documentation [here](https://docs.mongodb.com/manual/aggregation/) for more
+    /// information on aggregations.
+    pub fn aggregate(
+        &self,
+        pipeline: impl IntoIterator<Item = Document>,
+        options: impl Into<Option<AggregateOptions>>,
+    ) -> Result<Cursor> {
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
```
saghm, line 132: Ditto above about not needing to collect here

src/sync/coll.rs
patrickfreed: In order to pass it directly, we'd need to add the `Send` requirement to the `impl trait` clause in the sync API, which I figured we didn't want to do.

relevant compiler message:
```
error: future cannot be sent between threads safely
   --> src/sync/coll.rs:121:23
    |
116 |         pipeline: impl IntoIterator<Item = Document>,
    |                   ---------------------------------- help: consider further restricting this bound: `impl IntoIterator<Item = Document> + std::marker::Send`
...
121 |             .block_on(self.async_collection.aggregate(pipeline, options.into()))
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ future returned by `aggregate` is not `Send`
    |
    = help: within `impl core::future::future::Future`, the trait `std::marker::Send` is not implemented for `impl IntoIterator<Item = Document>`
note: future is not `Send` as this value is used across an await
```

src/sync/db.rs
patrickfreed: see comment above

src/runtime/mod.rs
saghm: 1000 threads, or 1000 processes? There's a bit of a difference, since in this case all the threads could be shared by the same process (i.e. their Rust application). In the case of your ~1000 processes, the single process containing the application might be getting only 1/1000 of the CPU time and then splitting that among the N threads, rather than the N threads each getting 1/1000 of the CPU time.

src/sync/coll.rs
saghm: Fair enough. I'd be surprised if people had containers of documents that were not `Send`, but like, if a user wants to use `cursor.map(|doc| doc.unwrap()` as their aggregation pipeline, I'm not going to stop them.

src/runtime/mod.rs
saghm: Works for me!

src/sync/client.rs
patrickfreed: Yeah that sounds better, updated. Added the link for mutex as well as other internal links that were missing (e.g. to `mongodb::Client`).

This also made me realize I completely omitted the docstring for `sync::Database`, so I've added that as well.

src/runtime/mod.rs
patrickfreed: Threads. I have around 250 or so processes running.

saghm: It looks like both [trust-dns](https://github.com/bluejekyll/trust-dns/blob/76a3776d8840b4915390f241cda94d1f12b0df73/crates/resolver/src/resolver.rs#L77-L79) and [reqwest](https://github.com/seanmonstar/reqwest/blob/71368104e360edfdba78fb7b8dd23b8274f19c57/src/blocking/client.rs#L696) use the basic scheduler; I feel like we should probably do the same unless we have a pretty strong reason not to.

src/sync/client.rs
saghm: Looks good!

src/runtime/mod.rs
patrickfreed: I would say that performance here is a pretty strong reason not to. We're doing a ton of concurrent heavy lifting, and constraining that to a single thread seems needlessly restrictive. Consider the 0.9.x driver: it spawns a monitoring thread per server plus a background thread per connection pool, which is a minimum of 2 threads regardless of deployment. 

From the tokio docs:
>Most applications should use the threaded scheduler, except in some niche use-cases, such as when running only a single thread is required.

I don't think our driver is a niche use-case or that running only on a single thread is required.

Another thing that might be worth thinking about is allowing the user to configure the amount of threads that we use. If they want to restrict it, then they can. That said, even if we go that route, I strongly believe that the default should not be 1.

saghm: I'm not super convinced by the 0.9.x driver precedent because that needs to use multiple threads in order to achieve concurrency, whereas our new synchronous client won't be doing nearly as much "concurrent heavy-lifting" because it still will still be using async tasks under the hood; in fact, we test the async client with only a single thread specifically because it's how we expect users will be using it. I think the correct way to think about the runtime here is "how many threads do we need to be able to run the internals of the driver (which are async)", and leave users to spawn additional threads for their own concurrent usage as needed; they're still free to spawn their own threads and share the client across them, in which case they won't be blocking any of the other driver functionality.

Note that the tokio docs says most "applications"; the driver is a library, not an application, so it's a bit harder to define a single correct value for all applications that the driver will be used with. Moreover, despite their documentation stating that, the tokio itself makes the default runtime single-threaded unless users opt-in to a feature that isn't enabled by default. From testing it locally, it seems like we don't even need that feature enabled in the driver anymore; it was only added so that we could run our tests with multiple threads while we were still blocking in parts of the driver.

My preference right now is still to enable only the basic scheduler and not letting it be configurable yet (since adding a configuration later is not a breaking API change, whereas if we add it now and want to remove it later, that would be a breaking change). Since I expect that you don't agree with this, I think we should probably take this discussion offline and circle back here once we all agree on a final decision.

.evergreen/run-async-std-tests.sh
```diff
@@ -4,3 +4,4 @@ set -o errexit
 
 . ~/.cargo/env
 RUST_BACKTRACE=1 cargo test --no-default-features --features async-std-runtime
+RUST_BACKTRACE=1 cargo test sync --no-default-features --features sync
```
patrickfreed, line 7: the tests are so minimal right now that I figured it wasn't worth a whole evergreen matrix, especially if we're only doing async-std for now.

patrickfreed:
Looks like this is passing all the Ubuntu tests again, so should be ready for a look.

src/runtime/mod.rs
saghm: Per offline discussion, we're making `sync` a separate feature that depends on `async-std` for now.

PR #161, opened 2020-04-10
RUST-104 Poll SRV records for mongos discovery

saghm:


src/client/options/mod.rs
```diff
@@ -304,6 +304,9 @@ pub struct ClientOptions {
     #[builder(default)]
     pub(crate) zlib_compression: Option<i32>,
 
+    #[builder(default)]
+    original_srv_hostname: Option<String>,
```
saghm, line 308: Although we already stored the original URI, it's more convenient to store the actual single hostname that we know exists in the case of SRV due to the rules about not having multiple hosts or a port.

src/client/options/mod.rs
```diff
@@ -492,12 +495,22 @@ impl From<ClientOptionsParser> for ClientOptions {
...
             original_uri: Some(parser.original_uri),
         }
     }
 }
 
 impl ClientOptions {
+    /// Creates a new ClientOptions with the `original_srv_hostname` field set to the testing value
+    /// used in the SRV tests.
+    #[cfg(test)]
+    pub(crate) fn new_srv() -> Self {
```
saghm, line 508: Rather than make the underlying field public, I added a test-only constructor that hardcodes the SRV domain that all the tests use. If necessary, we can update this later to take a parameter for the hostname.

src/client/options/mod.rs
```diff
@@ -560,6 +573,9 @@ impl ClientOptions {
                 .resolve_client_options(&options.hosts[0].hostname)
                 .await?;
 
+            // Save the original SRV hostname to allow mongos polling.
+            options.original_srv_hostname = Some(options.hosts[0].hostname.clone());
```
saghm, line 577: Cloning here isn't strictly necessary since we're about to overwrite the old hosts, but we'd have to use something like `into_iter().next().unwrap()` to avoid it, and I think the marginal performance cost is worth the readability.

src/sdam/description/topology/mod.rs
```diff
@@ -532,17 +566,24 @@ impl TopologyDescription {
...
-            let server = StreamAddress::parse(&server)?;
+    fn add_new_servers<'a>(&mut self, servers: impl Iterator<Item = &'a String>) -> Result<()> {
+        let servers: Result<Vec<_>> = servers.map(|server| StreamAddress::parse(server)).collect();
+
+        self.add_new_servers_from_addresses(servers?.iter());
+        Ok(())
+    }
 
+    /// Create a new ServerDescription for each address and add it to the topology.
+    fn add_new_servers_from_addresses<'a>(
```
saghm, line 577: Since I needed a subset of the functionality of `add_new_servers` for `sync_hosts`, I split it out into a separate method

src/sdam/monitor.rs
```diff
@@ -18,7 +18,7 @@ use crate::{
     RUNTIME,
 };
 
-const DEFAULT_HEARTBEAT_FREQUENCY: Duration = Duration::from_secs(10);
+pub(super) const DEFAULT_HEARTBEAT_FREQUENCY: Duration = Duration::from_secs(10);
```
saghm, line 21: We need to use the client's heartbeat frequency in the SRV polling implementation, and rather than having the monitor thread make the calculated value available back to the topology, it seemed simpler to just make this constant available to the other module and then redo the simply calculation.

src/sdam/srv_polling/mod.rs
```diff
@@ -0,0 +1,139 @@
+#[cfg(test)]
```
saghm, line 1: This module is based heavily on the SDAM monitor module.

src/sdam/srv_polling/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+            }
+            Ok(hosts) => hosts,
+            Err(err) => {
+                self.no_valid_hosts(Some(err));
+
+                return;
+            }
+        };
+
+        // TODO: Use TTL values to determine delay duration once `trust-dns-resolver`
```
saghm, line 91: As mentioned in the slack channel, trust-dns-resolver added the functionality to be able to inspect the TTL values of the SRV records right after their last release, but they haven't released since then, so we can't implement the optimization to hold off on future checks until the TTL has expired.

src/sdam/srv_polling/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+                self.update_hosts(hosts, topology, state).await;
+            }
+
+            RUNTIME
+                .delay_for(self.rescan_interval.unwrap_or(DEFAULT_RESCAN_SRV_INTERVAL))
+                .await;
+        }
+    }
+
+    pub(crate) async fn update_hosts(
```
saghm, line 71: The signature on this is a little weird; it requires both the Topology and the TopologyState so that it can perform all of the updates it needs without requiring the caller to do any work to update the topology, which makes implementing the spec tests easier. This method could just copy the TopologyState out internally, but since we already need to have copied it out to check that the topology type is either sharded or unknown, passing it in avoids needing to acquire the lock again.

src/srv.rs
```diff
@@ -39,7 +37,10 @@ impl SrvResolver {
         Ok(config)
     }
 
-    async fn get_srv_hosts(&self, original_hostname: &str) -> Result<Vec<StreamAddress>> {
+    pub(crate) async fn get_srv_hosts<'a>(
```
saghm, line 40: In order to properly implement the SRV polling spec, we need to keep the valid subset of records even if some of them are malformed, which differs from the initial seedlist spec, where a single error is allowed to cause the entire lookup to be considered failed. In order to facilitate this, I've changed the return value to iterate over `Result<StreamAddress>` instead of just shifting all of the internally errors to the outside `Result`.

src/sdam/srv_polling/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+        topology.update_state(diff, topology_state).await;
+    }
+
+    async fn lookup_hosts(&mut self) -> Result<Vec<StreamAddress>> {
+        let initial_hostname = self.initial_hostname.clone();
+        let resolver = self.get_or_create_srv_resolver().await?;
+        let mut new_hosts = Vec::new();
+
+        for host in resolver.get_srv_hosts(&initial_hostname).await? {
+            #[allow(clippy::single_match)]
```
saghm, line 105: Although we could just use an `if let` here, I figured we should just do the `match` now so that the TODO could be in a more correct place.

src/sdam/srv_polling/mod.rs
patrickfreed: sounds good, can we file a ticket for that and update this todo to reference it?

src/sdam/state/mod.rs
```diff
@@ -304,19 +379,42 @@ impl TopologyState {
...
+        self.sync_hosts(&hosts, options);
+
+        Ok(old_description.diff(&self.description))
+    }
+
+    /// Start/stop monitoring tasks and create/destroy connection pools based on the new and
+    /// removed servers in the topology description.
+    ///
+    /// This must **ONLY** be called on a copy of a TopologyState, not one that is stored in a
+    /// client. The `topology` parameter should contain a reference to the Topology that
```
patrickfreed, line 397: I think this line about the `topology` parameter is out of date now and can be removed. ditto for `update`.

src/sdam/srv_polling/test.rs
```diff
@@ -0,0 +1,91 @@
...
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn replace_all_dns_records() {
+    let hosts = vec![localhost_test_build_10gen(27019)];
+
+    run_test(Ok(hosts.clone()), hosts.into_iter().collect()).await;
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn timeout_error() {
```
patrickfreed, line 79: could you add some comments to each of the test cases describing what's going on? For instance, these last two look to have the same setup to me, but one results in a timeout error and the other in just no results.

saghm: I think you're getting the input and the output backwards; the first parameter is the result of the lookup (i.e. a list of hosts or an error that occurred during lookup), and the latter is the hosts expected in the topology after processing that result.

That said, I've added comments to each of them.

src/sdam/state/mod.rs
saghm: Done

src/sdam/srv_polling/test.rs
patrickfreed: ah ok, makes sense

PR #162, opened 2020-04-13
RUST-354 Fix handling of tailable cursors

saghm:


Cargo.toml
```diff
@@ -61,7 +61,7 @@ version = "1.0.98"
 features = ["derive"]
 
 [dependencies.tokio]
-version = "0.2.15"
+version = "0.2.18"
```
saghm, line 64: I updated this when debugging the issue in case it was a bug in tokio. It didn't turn out to fix anything, but I figured we might as well still update it as new versions come out until we do a release.

src/cursor.rs
```diff
@@ -0,0 +1,167 @@
+use std::{
```
saghm, line 1: Unfortunately git didn't recognize this as a move of the impatient module file, so it shows everything as new. I've pointed out the changes below.

src/cursor.rs
```diff
@@ -0,0 +1,167 @@
...
+
+#[derive(Debug)]
+pub struct Cursor {
+    client: Client,
+    get_more: GetMore,
+    exhausted: bool,
+    state: State,
+}
+
+type GetMoreFuture = BoxFuture<'static, Result<GetMoreResult>>;
```
saghm, line 41: At one point in the refactoring, I defined a function that needed this type, so I made an alias for it. Although it isn't use in any function signature anymore, I think using a type alias still makes it more readable.

src/cursor.rs
```diff
@@ -0,0 +1,167 @@
...
+                .await;
+        })
+    }
+}
+
+impl Stream for Cursor {
+    type Item = Result<Document>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
+        loop {
```
saghm, line 102: The body of this method is the most significant change from the previous implementation. The issue was caused by us not polling the new future we created for the new getMore operation after exhausting a buffer, which meant that the context would not be passed through to the underlying calls to poll on the TcpSocket, so the task wouldn't be woken up when the response was ready. More generally, it's a mistake to return `Poll::Pending` without polling the underlying future. Because the logic for handling the result of polling the future needs to be the same as in the case that the state is `Executing`, the simplest way to implement it is to wrap the state checking in a loop. In the case that the buffer is empty but the cursor is not exhausted, we start a new getMore operation and then continue looping, which will cause the future to be polled immediately.

src/test/spec/crud/aggregate.rs
```diff
@@ -62,9 +62,14 @@ async fn run_aggregate_test(test_file: TestFile) {
...
 
+            let results = cursor
+                .try_collect::<Vec<_>>()
+                .await
+                .expect(&test_case.description);
+
             assert_eq!(
                 outcome.result.unwrap_or_default(),
-                cursor.try_collect::<Vec<_>>().await.unwrap(),
+                results,
```
saghm, line 72: During my attempts to fix the issue, this test failed, and the use of `unwrap` made it difficult to determine which test was failing, so I updated it to use `expect`.

src/cursor.rs
```diff
@@ -0,0 +1,167 @@
...
+            let exhausted = self.exhausted;
```
patrickfreed, line 103: is this copy necessary, or can we just use `self.exhausted`?

src/test/cursor.rs
```diff
@@ -0,0 +1,86 @@
...
+        .unwrap();
+
+    let await_time = Duration::from_millis(500);
+
+    let mut cursor = coll
+        .find(
+            None,
+            FindOptions::builder()
+                .cursor_type(CursorType::TailableAwait)
+                .max_await_time(await_time)
```
patrickfreed, line 49: do we even want to allow users to customize this option anymore? It doesn't really have an effect on them anymore. Also, with client side operation timeout right around the corner, it could be soon deprecated anyways.

cc @mbroadst

src/test/cursor.rs
```diff
@@ -0,0 +1,86 @@
...
+        assert_eq!(
+            cursor.next().await.transpose().unwrap(),
+            Some(doc! { "_id": i })
+        );
+    }
+
+    let delay = RUNTIME.delay_for(await_time);
+    let next_doc = cursor.next();
+
+    let next_doc = match futures::future::select(Box::pin(delay), Box::pin(next_doc)).await {
```
patrickfreed, line 65: I think `RUNTIME.timeout` can be used for these.

src/test/cursor.rs
```diff
@@ -0,0 +1,86 @@
...
+    RUNTIME,
+};
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[function_name::named]
+async fn tailable_cursor() {
+    let _guard = LOCK.run_concurrently().await;
+
+    let client = TestClient::new().await;
```
patrickfreed, line 18: is it worth adding an `init_db_and_coll_with_options` for this sequence for future tests?

src/cursor.rs
saghm: Using `self.exhausted` requires mutably borrowing `self` again, which we can't do since we're borrowing `self.state`. Booleans are about as small as things get though, so I don't think it's much of issue.

src/test/cursor.rs
saghm: Do you mean specifically `TailableAwait`, or `CursorType` in general? The latter is still needed so the server knows not to close the cursor when the results are exhausted.

src/test/cursor.rs
saghm: `RUNTIME.timeout` doesn't give us back the non-completed future in the case of a timeout. The idea here is to keep reusing the future rather than calling `next` again so that we can test that it will complete with the right value once a new document is inserted.

src/test/cursor.rs
saghm: We could, I think I'd prefer a name that uses the word `create` somewhere in it, since explicitly creating the collection isn't something that `init_db_and_coll` does right now.

src/cursor.rs
patrickfreed: the `Executing` branch of the match statement uses `self.exhausted` though, which is why I was wondering. I just tried it out and it seems to compile using `self.exhausted` on line 144 which was previously the only usage of the copy. I'm not really concerned about the performance from copying it, just more about eliminating extra variables if we don't need them.

src/test/cursor.rs
patrickfreed: Yeah that sounds fine, do we want to make that change or just wait until a second case needs it?

src/test/cursor.rs
patrickfreed: I was referring to `max_await_time`.

src/cursor.rs
saghm: Interesting. At some point during my debugging, it was necessary, but I guess it's not anymore. I've updated it to get rid of that variable.

src/test/cursor.rs
saghm: I don't have a strong opinion either for or against doing it now, so I'll defer to whichever you prefer

src/test/cursor.rs
saghm: Ah, I misunderstood; sorry about that. I think it's reasonable to leave, since it determines the frequency of the `getMore` operations that the driver sends while waiting in a `Cursor::next` call, and some users might reasonably have preferences about that.

src/test/cursor.rs
patrickfreed: We're talking about it now, might as well do it while it's something we're thinking about.

src/test/cursor.rs
patrickfreed: Yeah I think for now it makes sense to leave it in, and only remove it if it gets deprecated as part of client side operation timeout.

src/test/cursor.rs
saghm: Added

PR #163, opened 2020-04-15
RUST-351 Add 'allowDiskUse' option to find command

isabelatkinson:


src/test/coll.rs
```diff
@@ -466,3 +466,109 @@ async fn empty_insert() {
...
+                }
+            }
+            _ => false,
+        })
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[function_name::named]
+async fn find_allow_disk_use() {
```
saghm, line 498: I think these tests could be abstracted to share most of their code. I'm thinking that you could write a function with the signature `fn allow_disk_use_test(options: FindOptions, expected_value: Option<bool>)` that would run the test with the given options and assert that there's a find command with the option as specified.

src/test/coll.rs
```diff
@@ -466,3 +466,109 @@ async fn empty_insert() {
...
+        .unwrap()
+        .iter()
+        .all(|event| match event {
+            CommandEvent::CommandStartedEvent(CommandStartedEvent {
+                command_name,
+                command,
+                ..
+            }) => {
+                if command_name == "find" {
+                    match command.get_bool("allowDiskUse") {
```
saghm, line 564: I think this should be slightly different; in the case that no option is specified, we want to assert that nothing is sent to the server. I think I'd do that by asserting that [`Document::contains_key("allowDiskUse")`](https://docs.rs/bson/0.14.1/bson/ordered/struct.OrderedDocument.html#method.contains_key) is false.

src/test/coll.rs
```diff
@@ -467,75 +467,32 @@ async fn empty_insert() {
...
-                }
-            }
-            _ => false,
-        })
-}
-
 #[cfg_attr(feature = "tokio-runtime", tokio::test)]
 #[cfg_attr(feature = "async-std-runtime", async_std::test)]
 #[function_name::named]
 async fn find_allow_disk_use() {
```
saghm, line 473: Since these test functions aren't using the `function_name!()` macro anymore, the `#[function_name::named]` annotation can be removed from each of them.

src/test/coll.rs
```diff
@@ -545,30 +502,26 @@ async fn find_allow_disk_use_not_specified() {
...
+        CommandEvent::CommandStartedEvent(CommandStartedEvent { command_name, .. }) => {
+            command_name == "find"
+        }
+        _ => false,
+    });
+
+    let event = iter.next().unwrap();
+    let allow_disk_use = match event {
+        CommandEvent::CommandStartedEvent(CommandStartedEvent { command, .. }) => {
+            match command.get_bool("allowDiskUse") {
```
saghm, line 518: Another way to do this would be to use the method [`Result::ok`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok), which converts `Ok(value)` to `Some(value)` and `Err(..)` to `None`

src/test/coll.rs
```diff
@@ -545,30 +502,26 @@ async fn find_allow_disk_use_not_specified() {
...
+                Ok(b) => Some(b),
+                Err(_) => None,
             }
-            _ => true,
-        }));
+        }
+        _ => None,
+    };
+    assert_eq!(allow_disk_use, expected_value);
+    assert_eq!(iter.count(), 0);
```
saghm, line 526: One thing that might not be obvious here is that if for some reason there are a large number of other events in the iterator, `count` will have to iterate over all of them in order to complete. A cheaper way of doing this would be to do `assert!(iter.next().is_none())`. I think it's fine to leave it as-is though, since this will give us an error message that tells us exactly how many extra find events there are rather than only the fact that there are some extras.

isabelatkinson: Ah makes sense, I originally had `assert_eq!(iter.next(), None)` but that wasn't compiling so I switched it over to use count, didn't realize there was an `is_none()` method. I'll leave it, I agree it would be useful to see how many extra events are created.

saghm: The reason that didn't work is that we haven't implemented the equality traits on the `CommandStartedEvent` type. In Rust, the `==` method is only defined for types which implement [`PartialEq`](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html), which can also be derived if all the contained types implement it. There's also [`Eq`](https://doc.rust-lang.org/std/cmp/trait.Eq.html), which can only be implemented on types that implement `PartialEq`; it doesn't add any new functionality, but it informs the compiler that the type implements full equality, which is can use to optimize things.

src/coll/options.rs
```diff
@@ -577,6 +577,11 @@ pub struct DistinctOptions {
 #[derive(Debug, Default, TypedBuilder, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct FindOptions {
+    /// Enables writing to temporary files. When set to true, the find operation can write data to
```
patrickfreed, line 580: I think we should clarify that this allows the _server_ to write to temporary files. I could imagine a user reading this as the driver buffering results on disk. We should also mention that this option is only supported in server versions 4.4+.

PR #164, opened 2020-04-17
minor: fix sync tests

saghm:
Some of the sync tests were failing on the waterfall. I think this was due to my not rebasing the sync changes onto my PRs that were in progress at the time (another casualty of merging through the Github UI...)


src/coll/mod.rs
```diff
@@ -46,9 +46,16 @@ const MAX_INSERT_DOCS_BYTES: usize = 16 * 1000 * 1000;
 ///
 /// ```rust
 /// # use bson::{bson, doc};
-/// # use mongodb::{Client, error::Result};
+/// # use mongodb::error::Result;
+/// # #[cfg(feature = "sync")]
```
patrickfreed, line 50: this doc example is in the async module, so I don't think we want to be using the sync client here. ditto for the others not in `src/sync`

saghm: Good call; I've updated the docs for `Collection` and `Database`, but I left the event handling code docs as-is pending our decision on RUST-340.

.evergreen/run-async-std-tests.sh
```diff
@@ -3,6 +3,13 @@
 set -o errexit
 
 . ~/.cargo/env
-RUST_BACKTRACE=1 cargo test --no-default-features --features async-std-runtime
-RUST_BACKTRACE=1 cargo test sync --no-default-features --features sync
-RUST_BACKTRACE=1 cargo test --doc sync --no-default-features --features sync
+
+if [ "$SINGLE_THREAD" = true ]; then
+	OPTIONS="-- --test-threads=1"
```
patrickfreed, line 8: we might also want to set `-j 1` here in case the test runner is still running them concurrently even with a single thread. Maybe that will help squash that last set of failures we're seeing.

saghm: The `j` flag doesn't affect anything besides building. From the `--help`:

```
The `--jobs` argument affects the building of the test executable but does
not affect how many jobs are used when running the tests. The default value
for the `--jobs` argument is the number of CPUs. If you want to control the
number of simultaneous running test cases, pass the `--test-threads` option
to the test binaries:

    cargo test -- --test-threads=1
```

patrickfreed:
There looks to be a lint failure in the code generated by `Derivative` or something?

saghm:
Strange! Upgrading the dependency to the latest version seems to have fixed it

PR #165, opened 2020-04-20
RUST-352 Support for 'authorizedDatabases' option

isabelatkinson:


src/test/client.rs
```diff
@@ -180,13 +181,72 @@ async fn list_database_names() {
...
+                &[AuthMechanism::ScramSha256],
+            )
+            .await
+            .unwrap();
+    }
+
+    for name in dbs {
+        let mut options = CLIENT_OPTIONS.clone();
+        let credential = Credential::builder()
+            .username(Some(format!("user_{}", name)))
```
saghm, line 227: Since the builder methods take `impl Into<Option<T>>` rather than just `Option<T>`, you can leave off the `Some(...)` and just put the value directly, e.g. `.username(format!("user_{}", name))`. This is because `Option<T>` implements `From<T>`.

src/test/client.rs
```diff
@@ -180,13 +181,72 @@ async fn list_database_names() {
...
+            .password(Some(String::from("pwd")))
+            .build();
+        options.credential = Some(credential);
+        let client = Client::with_options(options).unwrap();
+
+        let options = ListDatabasesOptions::builder()
+            .authorized_databases(true)
+            .build();
+        let result = client
+            .list_database_names(None, Some(options))
```
saghm, line 237: Similar to above, since the method takes `impl Into<Option<ListDatabaseOptions>>`, you can just put `.list_database_names(None, options)`. We define all of our option parameters this way to make things easier for users.

src/db/options.rs
```diff
@@ -136,3 +136,11 @@ pub struct ListCollectionsOptions {
     #[serde(serialize_with = "bson_util::serialize_batch_size", rename = "cursor")]
     pub batch_size: Option<u32>,
 }
+
+#[derive(Debug, Default, TypedBuilder, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ListDatabasesOptions {
+    // Determines which databases to return based on the user's access privileges.
```
patrickfreed, line 143: `//` (with two slashes) denotes a comment, whereas `///` (with three) denotes a docstring that will be rendered in the documentation. So let's add a slash here to make sure the latter happens.

Also, we should probably mention that this option is only supported in MongoDB 4.0.5+.

Lastly, let's add a docstring to the options type itself that links to the methods that use it. Check out other options structs for the format (e.g. [`CreateCollectionOptions`](https://github.com/mongodb/mongo-rust-driver/blob/master/src/db/options.rs#L31)).

PR #166, opened 2020-04-20
RUST-384 Flush connection read buffer before checking back into pool

saghm:


Lesiuk:
Unfortunately It's not enough. I will prepare test case which will reproduce dropping future in the middle of the read.

src/cmap/conn/mod.rs
```diff
@@ -207,9 +211,11 @@ impl Connection {
         request_id: impl Into<Option<i32>>,
     ) -> Result<CommandResponse> {
         let message = Message::with_command(command, request_id.into());
-        message.write_to(&mut self.stream).await?;
+        let request_id = message.write_to(&mut self.stream).await?;
+        self.pending_request_ids.push_back(request_id);
 
         let response_message = Message::read_from(&mut self.stream).await?;
+        self.pending_request_ids.pop_front();
```
mbroadst, line 218: My thought was that you would discard irrelevant results here, rather than in your drop implementation. I don't think you need to keep around the `pending_request_ids` in that case, rather just `read_from` until you get the id you're looking for or timeout. I think this "lazy" approach is easier to reason about, tracks less state, and has the added benefit of not doing I/O in your drop (maybe think of it as "only pay for the I/O you need").

saghm: The main reason I'm not super thrilled with doing it here is that it offloads the cost of reading the responses to the next operation. Ideally, if a user drops the future for an operation before reading the response, the next operation wouldn't have to pay any performance penalty for that. I think it's worth making the distinction this implementation doing do I/O in the drop implementation; it does it in the background task that the drop operation happened to have started. The actual code executed when the connection is dropped is still the same; the runtime will just schedule some extra work to be done at some point later without holding up the code path where the drop occurs.

That being said, cleaning it up before sending a new message does guarantee that there won't ever be more than one message stuck in the buffer waiting to be read, so the cost of reading it before sending a new message likely won't be that expensive compared to the network round trip that has to be done to actually perform the operation. I think there's a little bit of trickiness with regards to if the next operation's future happens to also be dropped, but before the buffer is fully flushed, so I'll add a bit more logic to handle that case as well.

src/test/client.rs
```diff
@@ -455,3 +456,42 @@ async fn saslprep_uri() {
...
+
+    let client = Client::with_options(options.clone()).unwrap();
+    let db = client.database("test");
+
+    db.collection("foo")
+        .insert_one(doc! { "x": 1 }, None)
+        .await
+        .unwrap();
+
+    let _: Result<_, _> = tokio::time::timeout(
```
Lesiuk, line 475: RUNTIME.timeout

saghm: Ah, good catch

src/test/temp.rs
```diff
@@ -0,0 +1,51 @@
+use std::time::Duration;
```
patrickfreed, line 1: this appears to be the same test as above, I think it got accidentally committed.

saghm: Oops, good catch

src/cmap/conn/mod.rs
```diff
@@ -323,6 +348,7 @@ impl From<DroppedConnectionState> for Connection {
             stream_description: state.stream_description.take(),
             ready_and_available_time: None,
             pool: None,
+            partial_message_state: state.partial_message_state.clone(),
```
patrickfreed, line 351: I think we can just move the partial message state here instead of cloning

patrickfreed: Ah makes sense. I think clippy would've caught it anyway if this clone was unnecessary.

src/cmap/conn/wire/message.rs
```diff
@@ -75,33 +75,41 @@ impl Message {
...
-    pub(crate) async fn read_from(reader: &mut AsyncStream) -> Result<Self> {
-        let header = Header::read_from(reader).await?;
-        let mut length_remaining = header.length - Header::LENGTH as i32;
-
+    pub(crate) async fn read_from(
+        reader: &mut AsyncStream,
+        partial_message_state: &mut PartialMessageState,
+    ) -> Result<Self> {
+        let length = reader.read_i32().await?;
+        partial_message_state.bytes_remaining = length as usize - Header::LENGTH;
```
patrickfreed, line 83: we're setting the bytes remaining before we actually performed the reads, and those reads may never occur if the future stops getting polled. I think we need to pass the partial message state down through `Head::read_from` as well, and make sure to update the bytes remaining with each read that we do.

src/cmap/conn/mod.rs
```diff
@@ -206,10 +216,23 @@ impl Connection {
...
+            let mut bytes = vec![0u8; self.partial_message_state.bytes_remaining];
+            self.stream.read_exact(&mut bytes).await?;
+            self.partial_message_state.bytes_remaining = 0;
+        }
+
         let message = Message::with_command(command, request_id.into());
         message.write_to(&mut self.stream).await?;
 
-        let response_message = Message::read_from(&mut self.stream).await?;
+        self.partial_message_state.needs_response = true;
```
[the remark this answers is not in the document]
patrickfreed: What happens if the future gets dropped midway through `write_to`? We'll probably need a way to flush the unsent data before starting to send new stuff, right?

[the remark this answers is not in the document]
saghm: If we send an only partially complete message to the server on a connection, then I'm not sure there's much we can do about that. I think the only two options to finish sending the message on the next write and then read and throw away the response or to just close the connection. My instinct is that an unfinished write will be a lot less common than an unfinished read, as there isn't going to be as much of a delay in writing the data as there is for waiting for the response to come back, so closing the connection in this case doesn't seem like quite as big a deal. I think there's also a case to be made that sending an unfinished message when the user stopped awaiting the future is a lot more likely to cause unexpected behavior for the user, since unlike reading the response of an already sent message, completing a sent message could have side effects on the database.

Because of this, I've updated the PR to drop the connection in the case of an unfinished write.

src/cmap/conn/mod.rs
```diff
@@ -262,13 +290,19 @@ impl Drop for Connection {
...
+                if let Some(strong_pool_ref) = weak_pool_ref.upgrade() {
+                    let dropped_connection_state = self.take();
+                    RUNTIME.execute(async move {
+                        strong_pool_ref
+                            .check_in(dropped_connection_state.into())
+                            .await;
+                    });
+                }
             } else {
                 self.close(ConnectionClosedReason::PoolClosed);
```
patrickfreed, line 307: this closed reason needs to be changed, maybe to `Error`, or perhaps to new closed reason case (is this okay with CMAP?). The `PoolClosed` one needs to move to an `else` on the upgrade attempt.

src/cmap/conn/wire/message.rs
```diff
@@ -75,33 +75,38 @@ impl Message {
...
         let flags = MessageFlags::from_bits_truncate(reader.read_u32().await?);
-        length_remaining -= std::mem::size_of::<u32>() as i32;
+        partial_message_state.bytes_remaining -= std::mem::size_of::<u32>();
 
         let mut count_reader = CountReader::new(reader);
         let mut sections = Vec::new();
 
-        while length_remaining - count_reader.bytes_read() as i32 > 4 {
+        while partial_message_state.bytes_remaining - count_reader.bytes_read() > 4 {
             sections.push(MessageSection::read(&mut count_reader).await?);
```
patrickfreed, line 91: the state needs to be propagated down here too. I wonder if this would be simpler if we just spawned a task via `RUNTIME.spawn` in `execute_operation` to guarantee everything got polled to completion. Doing all this partial reading and passing of state around seems to be pretty error-prone.

saghm: Can you elaborate on what you mean by spawning a task in execute operation? Do you mean running `send_command` in a background task and then using a channel to get the result back?

patrickfreed: yeah, except a channel wouldn't be needed. We could just use our [`AsyncJoinHandle`](https://github.com/mongodb/mongo-rust-driver/blob/master/src/runtime/join_handle.rs#L15).

patrickfreed: Thinking about it more, it would probably be safest to do the entirety of `execute_operation`'s body in a task, so that the various things it needs to do based on operation execution happen (like updating the cluster time, clearing the pool on certain errors, etc.)

saghm: I just tried to play around with this; it runs into some issues due to the fact that we can only spawn futures that are 'static, which won't be the case if we need to capture over any references. When trying to work around this, I ran into issues with the cursor's getMore operation, so I tried rebasing your sessions PR to see if the new cursor implementation avoided this issue, but this actually made things even more difficult due to the need to capture a reference to the session, which doesn't have the static lifetime.

saghm:
Hello @Lesiuk! Apologies for the long delay in action on this. After lots of discussion, we've come to the conclusion that simply dropping the connection in the case that an operation is unfinished due to the user not awaiting it to completion is probably the best approach. Although this does potentially cause a bit of a performance decrease due to a new connection needing to be created to replace the old one, it ensures that no extra work will be done in the case that a user completes an operation to completion or uses the built-in MongoDB timeout mechanisms (e.g. `serverSelectionTimeoutMS` or `maxTimeMS`), and it's by far the simplest solution that has the correct behavior in all edge cases. Since you had come up with this idea first in the pull request you submitted before, are you interested in submitting it again so we can merge it?

Lesiuk:
> Hello @Lesiuk! Apologies for the long delay in action on this. After lots of discussion, we've come to the conclusion that simply dropping the connection in the case that an operation is unfinished due to the user not awaiting it to completion is probably the best approach. Although this does potentially cause a bit of a performance decrease due to a new connection needing to be created to replace the old one, it ensures that no extra work will be done in the case that a user completes an operation to completion or uses the built-in MongoDB timeout mechanisms (e.g. `serverSelectionTimeoutMS` or `maxTimeMS`), and it's by far the simplest solution that has the correct behavior in all edge cases. Since you had come up with this idea first in the pull request you submitted before, are you interested in submitting it again so we can merge it?

Sure. I will prepare proper PR with tests. This PR I made before was more like a draft/poc.

saghm:
Sounds good; thanks!

patrickfreed:
Closing this now that a separate PR is being worked on (thanks @Lesiuk !).

PR #167, opened 2020-04-22
RUST-333 Improve invalid URI option error message

isabelatkinson:


src/client/options/mod.rs
```diff
@@ -1372,11 +1408,22 @@ impl ClientOptionsParser {
                 self.zlib_compression = Some(i);
             }
 
-            _ => {
-                return Err(ErrorKind::ArgumentError {
-                    message: "invalid option warning".to_string(),
+            other => {
+                let (jaro_winkler, option) = URI_OPTIONS.iter().fold((0.0, ""), |acc, option| {
```
saghm, line 1412: I think this can be written a bit more concisely by using [`Iterator::max_by`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max_by).

src/client/options/mod.rs
```diff
@@ -1372,11 +1408,22 @@ impl ClientOptionsParser {
...
+                    let jaro_winkler = jaro_winkler(option, other).abs();
+                    if jaro_winkler > acc.0 {
+                        return (jaro_winkler, option);
+                    }
+                    acc
+                });
+                let mut message = format!("{} is an invalid option", other);
+                if jaro_winkler >= 0.84 {
```
saghm, line 1420: I'm curious about how the value 0.84 was picked; was this just trial and error?

src/client/options/test.rs
```diff
@@ -332,3 +332,43 @@ async fn run_uri_options_spec_tests() {
...
+                None => assert_eq!(message, &format!("{} is an invalid option", option)),
+            };
+        }
+        Err(e) => panic!("expected ArgumentError, but got {:?}", e),
+    }
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn parse_unknown_options() {
```
saghm, line 360: I like these tests! The ones that give suggestions are all really good examples of things users might accidentally put in their URIs.

src/client/options/mod.rs
isabelatkinson: I re-implemented this work using `max_by` but I'm not sure if it's better than using fold. Here is what the new call would look like:
```
let option = URI_OPTIONS.iter().max_by(|o1, o2| {
    if jaro_winkler(other, o1).abs() > jaro_winkler(other, o2).abs() {
        return Ordering::Greater;
    } else {
        return Ordering::Less;
    }
});
let similarity = jaro_winkler(other, option.unwrap_or(&"")).abs();
```
A few downsides here: since the `f64` type doesn't implement `cmp()` the comparison logic needs to be done manually, and the score needs to be re-calculated at the end because `max_by` only returns the winning string. It does look a little cleaner, though. Would you consider this to be a stylistically better implementation?

src/client/options/mod.rs
isabelatkinson: Yep, mostly trial and error based on the strings in the test and some other ones I tried out, plus looking through some other examples of Jaro-Winkler being used.

src/client/options/mod.rs
saghm: Ah, I had forgotten that `f64` doesn't implement `cmp`. I think you can get around the recalulation by calling `map` before doing `max_by`, though.

src/client/options/test.rs
```diff
@@ -332,3 +332,43 @@ async fn run_uri_options_spec_tests() {
...
+    match ClientOptionsParser::parse(&format!("mongodb://host:27017/?{}=test", option))
+        .as_ref()
+        .map_err(|e| e.as_ref())
+    {
+        Ok(_) => panic!("expected error for option {}", option),
+        Err(ErrorKind::ArgumentError { message, .. }) => {
+            match suggestion {
+                Some(s) => assert_eq!(
+                    message,
+                    &format!(
```
patrickfreed, line 346: instead of checking for exact equality here, we might want to just check that the suggestion appears as a substring in the message. That way minor tweaks to this message won't break this test in the future.

For the `None` case, maybe we just pick a word we expect to be there in the future like "similar".

patrickfreed: "invalid" should appear regardless of whether we suggest something or not, and in the `None` case we want to verify that no suggestion is made. To that end, we may want to assert that the error message does _not_ contain a certain substring (e.g. "similar") that will only appear when something is suggested.

Also, that lint failure is unrelated to your changes and will go away once #164 is merged.

PR #169, opened 2020-04-30
Read preference and write concern changes from implicit sessions PR

patrickfreed:
This breaks off some changes from the implicit sessions PR to reduce the size of the diff. This PR adds support for determining if an operation is acknowledged (RUST-289) and fixes a bug where $readPreference wasn't being set against certain topologies (RUST-397).

### DIFF
--- a/src/concern/mod.rs
+++ b/src/concern/mod.rs
@@ -1,9 +1,12 @@
 //! Contains the types for read concerns and write concerns.
 
+#[cfg(test)]
+mod test;
+
 use std::time::Duration;
 
 use bson::doc;
-use serde::{Serialize, Serializer};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use serde_with::skip_serializing_none;
 use typed_builder::TypedBuilder;
 
@@ -72,7 +75,7 @@ impl Serialize for ReadConcern {
 /// See the documentation [here](https://docs.mongodb.com/manual/reference/write-concern/) for more
 /// information about write concerns.
 #[skip_serializing_none]
-#[derive(Clone, Debug, Default, PartialEq, TypedBuilder, Serialize)]
+#[derive(Clone, Debug, Default, PartialEq, TypedBuilder, Serialize, Deserialize)]
 pub struct WriteConcern {
     /// Requests acknowledgement that the operation has propagated to a specific number or variety
     /// of servers.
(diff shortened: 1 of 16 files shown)


src/concern.rs
```diff
@@ -152,6 +152,10 @@ impl Acknowledgment {
 }
 
 impl WriteConcern {
+    pub(crate) fn is_acknowledged(&self) -> bool {
+        self.w != Some(Acknowledgment::Nodes(0))
```
saghm, line 156: Per [the spec](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#unacknowledged-writeconcern), I think this also needs to check that `journal` is not true.

patrickfreed: Originally I didn't bother because `w: 0, j: true` was invalid, but outside of uri parsing, we were'nt actually validating any write concerns. To fix this, I went back to operations exposing the actual write concern and had `execute_operation` validate it. I also made `is_acknowledged` verify `j: false` for completeness.

Also, instead of maintaining a separate `is_acknowledged` for write concerns specified in `RunCommand`, I just added a `Deserialize` implementation to `WriteConcern` and leveraged that.

src/concern/mod.rs
```diff
@@ -152,8 +175,9 @@ impl Acknowledgment {
 }
 
 impl WriteConcern {
+    #[allow(dead_code)]
     pub(crate) fn is_acknowledged(&self) -> bool {
-        self.w != Some(Acknowledgment::Nodes(0))
+        self.w != Some(Acknowledgment::Nodes(0)) && self.journal != Some(false)
```
saghm, line 180: Should this be `||` rather than `&&`? `w=1, j=false` is acknowledged, but this expression would yield false

src/concern/mod.rs
```diff
@@ -152,8 +175,9 @@ impl Acknowledgment {
 }
 
 impl WriteConcern {
+    #[allow(dead_code)]
```
saghm, line 178: I don't think this is dead code; it gets called in `Operation::is_acknowledged`, which isn't marked as dead code

src/concern/mod.rs
patrickfreed: Whoops, fixed. Also added a test.

src/concern/mod.rs
patrickfreed: Yeah this one is a little weird. The compiled doesn't complain about `Operation::is_acknowledged` being unused (maybe because it's part of a trait definition?), but it does complain about this method.

src/concern/mod.rs
saghm: I'll be honest, I had to stare at this line for almost a full minute to convince myself that it did in fact need to be `||`. Once "not" gets involved, it's very hard to keep everything straight in your head.

src/concern/mod.rs
saghm: Hmm. Is there any reason to bother keeping them both around then, if neither is used?

saghm: Oh, wait, is it needed for sessions?

patrickfreed: Yep, they're just dead code right now because this is a separate PR. When I rebase the other one I can remove the ignore.

saghm: Awesome

PR #170, opened 2020-05-04
RUST-329 Update multi-threaded examples to use tasks

saghm:


src/coll/mod.rs
```diff
@@ -58,15 +62,12 @@ const MAX_INSERT_DOCS_BYTES: usize = 16 * 1000 * 1000;
 /// for i in 0..5 {
 ///     let coll_ref = coll.clone();
 ///
-///     std::thread::spawn(move || {
+///     task::spawn(async move {
 ///         // Perform operations with `coll_ref`. For example:
 ///         coll_ref.insert_one(doc! { "x": i }, None);
```
patrickfreed, line 67: we can now add an `.await` here

saghm: Ah, good catch

PR #173, opened 2020-05-07
RUST-408 Generate API documentation for the sync API on docs.rs

saghm:
With a bit of work, I was able to spin up a [local instance of docs.rs](https://github.com/rust-lang/docs.rs/blob/master/README.md#cli) to verify that this generates the sync API documentation. To see the generated documentation locally without needing to set up a docs.rs instance, you can run `cargo rustdoc -p mongodb --open -- --cfg docsrs`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,3 +77,6 @@ function_name = "0.2.0"
 pretty_assertions = "0.6.1"
 serde_json = "1.0.40"
 semver = "0.9.0"
+
+[package.metadata.docs.rs]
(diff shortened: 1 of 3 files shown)


saghm:
Updated; the experimental feature can only be used with nightly, but since it's gated to the `docsrs` configuration, that won't matter when building the driver normally (or when building the docs without that configuration). I re-tested it with the local docs.rs instance to verify that it builds the docs with nightly, and it does work.

PR #174, opened 2020-05-07
RUST-396 Parse responses with command errors into Errors in run_command

patrickfreed:
[RUST-396](https://jira.mongodb.org/browse/RUST-396)

This PR updates `run_command` to parse command errors from server responses and return them as `Error` rather than as part of the command response.

e.g.
```rust
// before
db.run_command(doc! { "asdfaf": true }, None).await; // Ok(Document({ "ok": 0, ... })

// after
db.run_command(doc! { "asdfaf": true }, None).await; // Err(ErrorKind::CommandError { ... })
```
Without this logic, it can be easy to ignore errors unintentionally, as is seen in the test fix included in this PR.

For completeness, I redid my survey of the other drivers, and there does not appear to be a consensus on this behavior:
- libmongoc - validates
- CXX - validates
- Swift - validates
- PHP - validates
- Ruby - result that can be validated by user
- Go - result that is implicitly validated in all its methods
- Java - no validation
- Python - optional validation, on by default
- C# - no validation
- Node - validates

So tallying that up, it seems 2 drivers do no validation, 2 do it optionally, and the rest validate.
I think that if we wanted to preserve the no-validation behavior, adding it as an option to `run_command` that is off by default would be reasonable.

### DIFF
--- a/src/test/client.rs
+++ b/src/test/client.rs
@@ -7,8 +7,7 @@ use crate::{
     error::{Error, ErrorKind},
     options::{
         auth::{AuthMechanism, Credential},
-        ClientOptions,
-        ListDatabasesOptions,
+        ClientOptions, ListDatabasesOptions,
     },
     selection_criteria::{ReadPreference, SelectionCriteria},
     test::{util::TestClient, CLIENT_OPTIONS, LOCK},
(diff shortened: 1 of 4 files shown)


src/test/client.rs
```diff
@@ -497,35 +497,6 @@ async fn saslprep_options() {
     auth_test_options("IX", "I\u{00AD}X", None, true).await;
     auth_test_options("\u{2168}", "IV", None, true).await;
     auth_test_options("\u{2168}", "I\u{00AD}V", None, true).await;
-}
-
-#[cfg_attr(feature = "tokio-runtime", tokio::test)]
```
saghm, line 502: I'm confused about why this was deleted; what does this test have to do with `run_command` errors?

patrickfreed: The setup created the same user twice, which was always failing. The assertions are preserved, just the duplicated setup was removed.

saghm: Okay, I see. I was confused by the fact that the two tests were merged into one.

PR #175, opened 2020-05-08
RUST-409 Add DriverInfo to client metadata

saghm:


src/client/options/mod.rs
```diff
@@ -528,6 +549,7 @@ impl From<ClientOptionsParser> for ClientOptions {
             socket_timeout: parser.socket_timeout,
             zlib_compression: parser.zlib_compression,
             direct_connection: parser.direct_connection,
+            driver_info: None,
```
saghm, line 552: Given that the purpose of this is for libraries wrapping the driver to add info about themselves, I don't think it's super likely that anyone will want to specify these options from the URI, so I didn't implement any URI options corresponding to `DriverInfo` fields.

src/cmap/establish/handshake.rs
```diff
@@ -1,99 +0,0 @@
-use bson::{doc, Document};
```
saghm, line 1: This was moved to `handshake/mod.rs`. Unfortunately git doesn't seem to have figured this out...

src/cmap/establish/handshake/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+use lazy_static::lazy_static;
+use os_info::{Type, Version};
+
+use crate::{
+    cmap::{options::ConnectionPoolOptions, Command, Connection, StreamDescription},
+    error::Result,
+    is_master::IsMasterReply,
+};
+
+#[derive(Clone, Debug)]
```
saghm, line 14: Rather than having to deal with setting nested values in a document, I figured it was easier to just use a struct for all of the client metadata and then convert to a document only once we've computed all the values.

src/cmap/establish/handshake/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+struct OsMetadata {
+    os_type: String,
+    name: Option<String>,
+    architecture: String,
+    version: Option<String>,
+}
+
+impl From<ClientMetadata> for Bson {
+    fn from(metadata: ClientMetadata) -> Self {
+        let mut metadata_doc = Document::new();
```
saghm, line 43: While we'd normally use `Serialize` and just propagate the errors from `bson::to_bson` for this, the place where `Handshaker::new` is actually called is often in an SDAM background task (as the `Establisher` gets created when a pool is created, which in turn is created when a new server is created, which only happens either at client creation or when a new server is discovered). Since we don't really have any way to handle a failure to create a connection pool there, I think it's cleaner to just manually write the infallible code for converting this to a document rather than either calling `unwrap` or encoding error-handling logic in the SDAM tasks where we errors won't actually ever occur.

In the longer term, it's worth looking into if there's any way for us to implement infallible conversion in the BSON library for types that are guaranteed to convert safely.

src/client/options/mod.rs
patrickfreed: should we mention this in the docstrings for the options?

src/cmap/establish/handshake/test.rs
```diff
@@ -0,0 +1,61 @@
+use bson::doc;
```
patrickfreed, line 1: these tests are good, but I think we should also have some integration tests w/ command monitoring so that we get "end to end" coverage so to speak.

src/client/options/mod.rs
saghm: Mention that this should only be specified by wrapping libraries, or that there's no URI options mapping to it?

src/cmap/establish/handshake/test.rs
saghm: I originally was going to implement them with that, but since command monitoring only happens in `execute_operation`, it doesn't happen during the handshake (which uses `send_command` directly).

src/client/options/mod.rs
patrickfreed: I meant the latter, and the docs in the PR already mention the former. Since [we direct users](https://github.com/mongodb/mongo-rust-driver/blob/master/src/client/mod.rs#L84) specifically to `ClientOptions::parse` (which does not mention this option), I don't actually think there's anything to do here though.

PR #176, opened 2020-05-11
RUST-392 Suppress "ns not found" errors

patrickfreed:
[RUST-392](https://jira.mongodb.org/browse/RUST-392)

This PR updates `Collection::drop` to ignore errors returned from the server when dropping a nonexistent collection.

### DIFF
--- a/src/operation/mod.rs
+++ b/src/operation/mod.rs
@@ -21,7 +21,15 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     cmap::{Command, CommandResponse, StreamDescription},
-    error::{BulkWriteError, BulkWriteFailure, ErrorKind, Result, WriteConcernError, WriteFailure},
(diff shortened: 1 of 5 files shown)


src/operation/mod.rs
```diff
@@ -57,6 +65,12 @@ pub(crate) trait Operation {
     /// Interprets the server response to the command.
     fn handle_response(&self, response: CommandResponse) -> Result<Self::O>;
 
+    /// Interpret an error encountered while sending the built command to the server, potentially
+    /// recovering.
+    fn handle_error(&self, error: Error) -> Result<Self::O> {
```
patrickfreed, line 70: Ironically, this amendment to `Operation` is similar to the `handles_command_errors` that I just removed, but I think it's a more general method that may prove to be useful in other cases in the future.

PR #178, opened 2020-05-14
RUST-261 Mark enums and options structs as non-exhaustible

saghm:


src/concern/mod.rs
```diff
@@ -103,6 +105,7 @@ pub struct WriteConcern {
 
 /// The type of the `w` field in a [`WriteConcern`](struct.WriteConcern.html).
 #[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
```
patrickfreed, line 108: This may not be required here, since it would be a breaking change technically for the server to introduce a new special "w" value, given that any arbitrary string could refer to custom write concern created by a user.

src/selection_criteria.rs
```diff
@@ -84,6 +85,7 @@ pub type Predicate = Arc<dyn Send + Sync + Fn(&ServerInfo) -> bool>;
 ///
 /// See the [MongoDB docs](https://docs.mongodb.com/manual/core/read-preference) for more details.
 #[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
 pub enum ReadPreference {
```
patrickfreed, line 89: can the cases themselves be marked as non-exhaustive? That way we could add a new filed to e.g. `Nearest` without a breaking change.

I don't think we can do it for `Primary`, although I think that one would only ever take a field if multi-master became a thing, and even then it might not be on `Primary`.

src/error.rs
```diff
@@ -92,6 +92,7 @@ impl std::ops::Deref for Error {
 
 /// The types of errors that can occur.
 #[derive(Debug, Error)]
+#[non_exhaustive]
 pub enum ErrorKind {
```
patrickfreed, line 96: ditto considering adding `non_exhaustive` to any enum variants here.

src/error.rs
```diff
@@ -340,6 +343,7 @@ pub struct WriteConcernError {
 /// An error that occurred during a write operation that wasn't due to being unable to satisfy a
 /// write concern.
 #[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
 pub struct WriteError {
```
patrickfreed, line 347: Is there a macro for disallowing initialization? If not, we may want to consider exposing these fields via methods instead, since any new fields we add to this would need to have a default implementation to avoid breaking user code, which for something like an error struct may not make sense.

src/sdam/description/server.rs
```diff
@@ -16,6 +16,7 @@ const DRIVER_MIN_WIRE_VERSION: i32 = 6;
 const DRIVER_MAX_WIRE_VERSION: i32 = 7;
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[non_exhaustive]
 pub enum ServerType {
```
patrickfreed, line 20: this one might be exhaustive. I imagine if the drivers or server introduced a new server type, it would break a lot of things.

src/concern/mod.rs
saghm: What about other types, like arrays or documents?

src/error.rs
saghm: Normally the solution is to just have private fields and define getters. If we're feeling especially lazy, we could use the [`getset` crate](https://crates.io/crates/getset).

src/sdam/description/server.rs
saghm: I wasn't sure about additional replica set types; there are already so many that I wouldn't be surprised if a new type was added. This is also only exposed because of the selection criteria predicate; I don't think it's too much of a burden to require predicates to be non-exhaustive with their matching of the server types.

src/selection_criteria.rs
saghm: The compiler doesn't complain, so I guess they can be. We _could_ do it for primary by defining it as `Primary {}`, but I don't think that's a good idea.

src/concern/mod.rs
saghm: Based on offline discussion, we'll leave this as-is

src/error.rs
patrickfreed: So rereading the `non_exhaustive` documentation, this macro actually makes it so outside crates cannot instantiate these types--they can only match on them. This works great for types like this, but not so great for our options types. We may have to only allow users to use the builder for creating options then. We also need to revisit the changes to BSON to make sure outside users can still create the things they need to.

src/sdam/description/server.rs
saghm: Leaving as-is per offline discussion

src/error.rs
saghm: Interesting; I'm not sure I like the idea of requiring the builder pattern for all options structs, as some of them are pretty small, making it overkill. I'll remove it from types we want users to be able to create for now until we reach a decision.

src/selection_criteria.rs
saghm: Left as-is per discussion about this making them non-constructible

src/error.rs
saghm: Done

src/error.rs
saghm: I left it on all enums (as a top-level attribute, not on the variants) since they can still be constructed even with it, as well as on all error- and event-related structs, since they only need to be emitted by the driver rather than created by users.

patrickfreed: So we actually don't need the getters and setters on the various error cases anymore, since `non_exhaustive` gives us the non-construction. I apologize for realizing this after you implemented all of them...

As for options, I'm not sure there's any other option (pun intended). If we allow users do do: 
```rust
let options = FooOptions { a: "hello".to_string() };
```
then any new field added to `FooOptions` will be a breaking change, which definitely won't work for us. We could add other methods to `FooOptions` that are constructor like, but I think the builder pattern is probably the best way to do that.

src/selection_criteria.rs
patrickfreed: I think we may need to take the struct with an enum field approach here to allow for things like hedged reads to be non-breaking. This has the unfortuante side effect of allowing users to specify invalid `ReadPreference`s like is possible with `WriteConcern`, though I don't really know of a way around this.

saghm: I've updated the ReadPreference API as discussed offline

src/error.rs
saghm: Getters removed and `non_exhaustive` added to all options structs as discussed

src/selection_criteria.rs
```diff
@@ -88,45 +90,43 @@ pub enum ReadPreference {
...
-    Nearest {
-        tag_sets: Option<Vec<TagSet>>,
-        max_staleness: Option<Duration>,
-    },
+    Nearest { options: ReadPreferenceOptions },
+}
+
+#[derive(Clone, Debug, Default, PartialEq, TypedBuilder)]
+#[non_exhaustive]
+pub struct ReadPreferenceOptions {
```
patrickfreed, line 109: could use a docsstring

saghm: Updated

PR #179, opened 2020-05-18
RUST-128 Retryable Reads

isabelatkinson:


src/error.rs
```diff
@@ -69,6 +71,43 @@ impl Error {
...
+                    return true;
+                }
+                false
+            }
+            None => false,
+        }
+    }
+
+    /// Whether an error originated in the driver
+    pub(crate) fn is_driver_error(&self) -> bool {
```
isabelatkinson, line 96: The retryable reads spec indicates that if the retry fails with an error that originates within the driver rather than from the server, then the original error should be returned -- this list could use a double-check to make sure that nothing is erroneously included/excluded

src/client/executor.rs
```diff
@@ -89,8 +86,50 @@ impl Client {
...
+                    || !err.is_read_retryable()
+                {
+                    return Err(err);
+                } else {
+                    err
+                }
+            }
+        };
+
+        let server = match self.select_server(op.selection_criteria()).await {
```
patrickfreed, line 113: I'm not sure if it's any clearer so feel free to not change this, but you can use `map_err` here in conjunction with the `?` operator as an alternative:
```rust
let server = self.select_server(op.selection_criteria()).await.map_err(|_| first_error)?;
```

src/client/executor.rs
```diff
@@ -99,17 +138,21 @@ impl Client {
                     .topology
                     .handle_post_handshake_error(err.clone(), conn, server)
                     .await;
-                Err(err)
+                if !err.clone().is_driver_error() {
```
patrickfreed, line 141: This `clone` doesn't seem necessary here.

src/error.rs
```diff
@@ -69,6 +71,43 @@ impl Error {
             _ => false,
         }
     }
+
+    /// Whether a read operation should be retried if this error occurs
+    pub(crate) fn is_read_retryable(&self) -> bool {
+        if self.is_network_error() {
+            return true;
+        }
+        let error_kind: &ErrorKind = &self.kind;
```
patrickfreed, line 80: This variable isn't used/necessary I think.

src/error.rs
```diff
@@ -69,6 +71,43 @@ impl Error {
...
+        match &self.kind.code_and_message() {
+            Some((code, message)) => {
+                if RETRYABLE_READ_CODES.contains(&code) {
+                    return true;
+                }
+                if message.contains("not master") || message.contains("node is recovering") {
```
patrickfreed, line 86: we actually have methods on `ErrorKind` for this already: `is_not_master` and `is_recovering`

src/error.rs
patrickfreed: I think this would be easier to implement / maintain in the future by listing the server errors rather than driver ones, since the number of server errors is much smaller. I think the only server errors as of right now are `WriteError`, `CommandError`, `BulkWriteError`, and `AuthenticationError`.

src/error.rs
```diff
@@ -69,6 +71,43 @@ impl Error {
...
+                false
+            }
+            None => false,
+        }
+    }
+
+    /// Whether an error originated in the driver
+    pub(crate) fn is_driver_error(&self) -> bool {
+        match self.kind.as_ref() {
+            ErrorKind::AddrParse(_) => true,
```
patrickfreed, line 98: for cases that evaluate to the same thing, you can use `|`, e.g.
```rust
match self.kind.as_ref() {
   ErrorKind::AddrParse(_) | ErrorKind::BsonDecode(_) => true,
   _ => false,
}
```

src/test/spec/json/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,667 @@
+===============
```
patrickfreed, line 1: I don't think we need to check in the entire retryable-reads spec into the repository, just the test files should be sufficient.

src/test/spec/retryable_reads.rs
```diff
@@ -0,0 +1,183 @@
...
+                options.retry_reads = Some(client_options.get_bool("retryReads").unwrap());
+            }
+            if TestClient::new().await.is_sharded() && test_case.use_multiple_mongoses != Some(true)
+            {
+                options.hosts = options.hosts.iter().cloned().take(1).collect();
+            }
+            let client = EventClient::with_options(options).await;
+
+            if let Some(ref run_on) = test_file.run_on {
+                let can_run_on = run_on.iter().any(|run_on| {
```
patrickfreed, line 54: this logic can be nicely abstracted to a method on `RunOn`, which will be useful for subsequent runners.

src/test/spec/retryable_reads.rs
```diff
@@ -0,0 +1,183 @@
...
+
+        for test_case in test_file.tests {
+            if let Some(skip_reason) = test_case.skip_reason {
+                println!("Skipping {}: {}", test_case.description, skip_reason);
+                continue;
+            }
+
+            let mut options = CLIENT_OPTIONS.clone();
+            if let Some(client_options) = test_case.client_options {
+                options.retry_reads = Some(client_options.get_bool("retryReads").unwrap());
```
patrickfreed, line 45: I'm a little concerned that we may accidentally omit other fields specified here. We may need a `pub(crate)` "merge" method for `ClientOptions` to ensure we get everything.

src/test/spec/retryable_reads.rs
```diff
@@ -0,0 +1,183 @@
...
+                        },
+                        None,
+                    )
+                    .await
+                    .unwrap();
+            }
+        }
+    }
+
+    run_spec_test(&["retryable-reads", "tests"], run_test).await;
```
patrickfreed, line 182: the spec tests can be moved up a directory so that this just has to specify "retryable-reads"

src/test/util/mod.rs
```diff
@@ -188,8 +194,8 @@ pub async fn drop_collection(coll: &Collection) {
 
 pub fn parse_version(version: &str) -> (u64, u64) {
```
patrickfreed, line 195: I think this method can be replaced to use `semver::Version`. We're using it already in `TestClient`.

src/test/util/test_event.rs
```diff
@@ -36,9 +36,19 @@ impl Matchable for TestEvent {
                     command: expected_command,
                 },
             ) => {
-                actual_command_name == expected_command_name
```
patrickfreed, line 39: Why did this need to be changed? It looks like the newer version could possibly have some false positives.

src/test/util/test_file.rs
```diff
@@ -0,0 +1,55 @@
+use crate::test::util::test_event::TestEvent;
```
patrickfreed, line 1: I think the test runner files introduced in this PR fit better somewhere like `src/test/spec/runner`, since they're specifically for spec tests, and they're also more than just extra utilities.

src/test/spec/retryable_reads.rs
```diff
@@ -0,0 +1,183 @@
...
+            let coll_name = match test_file.collection_name {
+                Some(ref coll_name) => coll_name.clone(),
+                None => String::from("coll"),
+            };
+
+            let coll = client.init_db_and_coll(&db_name, &coll_name).await;
+
+            if let Some(ref data) = test_file.data {
+                let data: Vec<Document> = data
+                    .as_array()
```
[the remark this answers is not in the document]
isabelatkinson: This is what I had originally, but ran into a problem with the GridFS tests that required a more generalized type. Even though the GridFS tests don't actually run, they need to be serialized into `TestFile`s in order to check if they contain an operation that should be skipped. GridFS tests represent their data as a [document](https://github.com/mongodb/specifications/blob/8b1d6e9ff4a7dfd2a47d3fcf6e0ee6b2d98291a7/source/retryable-reads/tests/gridfs-download-serverErrors.json#L19) whereas the rest of the tests have data as an [array of documents](https://github.com/mongodb/specifications/blob/8b1d6e9ff4a7dfd2a47d3fcf6e0ee6b2d98291a7/source/retryable-reads/tests/findOne.json#L19), so the type that represents data needs to be general enough to support both formats. I considered using an enum to represent the two types, but that complicated serialization so I abstracted it out to `Bson`.

src/error.rs
isabelatkinson: The [spec indicates](https://github.com/mongodb/specifications/blob/5979b36b16e37427d9d49c0825944a02ef665a10/source/retryable-reads/retryable-reads.rst#id7) that any error that contains either of these specific substrings should be considered a retryable error -- super nitpicky, but would it be an issue that `is_recovering` also allows for the String "not master or secondary"?

src/test/spec/json/retryable-reads/retryable-reads.rst
isabelatkinson: In a similar vein, is it necessary to include the `.yml` files alongside the `.json` since the runner doesn't use them?

src/test/spec/retryable_reads.rs
isabelatkinson: The only client option that's specified for these tests is retryReads so including more options isn't totally necessary here, although I considered this when I was originally writing a more generic v2 runner. I was planning on augmenting this to support more options when I implement the v2 runner (based on this runner), but I can also incorporate it now in this commit if you think that would be better!

src/test/util/test_event.rs
isabelatkinson: Some of the command monitoring expectations in the tests don't include all of the fields (e.g. [this one](https://github.com/mongodb/specifications/blob/8b1d6e9ff4a7dfd2a47d3fcf6e0ee6b2d98291a7/source/retryable-reads/tests/aggregate.json#L70) is missing `command_name`), so they need to be options to ignore the missing fields. The method on `EventClient` that returns all of the events to compare against the expectations already filters out any events that don't match the name of the operation though, so the comparison still occurs.

src/error.rs
patrickfreed: Nope that should be fine, since if "not master" itself is a substring of "not master or secondary", the union of our `is_recovering` and `is_not_master` methods is equivalent to the two substring checks you have here.

src/test/spec/retryable_reads.rs
patrickfreed: Leaving for the full v2 runner sounds good to me.

src/test/spec/retryable_reads.rs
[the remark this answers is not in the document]
patrickfreed: Ah good point. Then we can use an enum to model this and instruct serde how to deserialize it via the `untagged` attribute (https://serde.rs/enum-representations.html#untagged).

something like:
```rust
#[derive(Deserialize)]
#[serde(untagged)]
enum TestData {
    Single(Vec<Document>),
    Many(HashMap<String, Vec<Document>>)
}
```
Because Rust has such a strict type system, working with JSON/BSON can get _very_ boilerplate heavy unless you model the data precisely. The nice thing is that the more specific your models are, the more boilerplate goes away and is handled by `serde`.

src/test/util/test_event.rs
patrickfreed: The command monitoring tests are pretty old, so they don't use the unified/v2 test format. Because of that, I wouldn't make any changes in your new runner to accommodate them. Instead, just leave command monitoring as having a separate implementation so the new one can be left unhampered. If the command monitoring tests can borrow some stuff without having to make changes to the runner for them, then by all means share code though.

src/test/spec/retryable_reads.rs
```diff
@@ -0,0 +1,183 @@
...
+                    .database("admin")
+                    .run_command(fail_point.clone(), None)
+                    .await
+                    .unwrap();
+            }
+
+            let mut events: Vec<TestEvent> = Vec::new();
+            for operation in test_case.operations {
+                let test_operation: AnyTestOperation =
+                    bson::from_bson(Bson::Document(operation.as_document())).unwrap();
```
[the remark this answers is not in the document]
isabelatkinson: I tried this out, but it's a bit tricky since `TestCase` derives `Debug`. `AnyTestOperation` has a field `operation: Box<dyn TestOperation>`, so as far as I know it can't derive `Debug` since `TestOperation` is a trait which means not all of `AnyTestOperation`'s fields derive `Debug`. Is there a good workaround for this?

src/test/util/operation.rs
```diff
@@ -0,0 +1,652 @@
...
+#[async_trait]
+#[allow(where_clauses_object_safety)]
+pub trait TestOperation {
+    /// The command names to monitor as part of this test.
+    fn command_names(&self) -> &[&str];
+
+    // The linked issue causes a warning that cannot be suppressed when providing a default
+    // implementation for these functions.
+    // <https://github.com/rust-lang/rust/issues/51443>
+    async fn execute_on_collection(&self, collection: &Collection) -> Result<Option<Bson>>;
```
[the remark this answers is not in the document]
isabelatkinson: Yeah, I'll keep my eye out for a fix on this.

src/client/executor.rs
isabelatkinson: I tried this out, but it doesn't compile unless `first_error` is replaced with `first_error.clone()` in the closure to avoid the closure taking ownership of the error. I think I'll leave as-is to avoid an unnecessary clone.

src/test/util/mod.rs
isabelatkinson: I think we need the more flexible parser since the tests specify strings with and without patch versions (i.e. both "4.0" and "4.1.7") while the `Version::parse` function only takes strings with patch versions.

src/test/spec/retryable_reads.rs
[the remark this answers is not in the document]
saghm: > Because Rust has such a strict type system, working with JSON/BSON can get very boilerplate heavy unless you model the data precisely. The nice thing is that the more specific your models are, the more boilerplate goes away and is handled by serde

I'm totally going to steal this quote at some point

src/test/spec/retryable_reads.rs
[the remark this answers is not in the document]
saghm: Could you update the `operation` field to be `operation: Box<dyn TestOperation + Debug>`? Alternately, since this is only an internal API, we could define `TestOperation` so that it requires `Debug`, i.e. `trait TestOperation: Debug`.

src/test/util/mod.rs
saghm: Potential hacky workaround: if there is only one '.', append ".0" to the string and then pass to semver.

src/test/spec/json/retryable-reads/retryable-reads.rst
patrickfreed: I think we keep the .yml files around for the comments in them, though you're right that we don't really need them.

src/client/executor.rs
```diff
@@ -99,17 +138,21 @@ impl Client {
                     .topology
                     .handle_post_handshake_error(err.clone(), conn, server)
                     .await;
-                Err(err)
+                if !err.is_server_error() {
```
saghm, line 141: From looking at [this section of the spec](https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.rst#c-if-the-retry-attempt-fails), I think this should be checking for a retryable error as well, not just a server error:

>  If the retry failed due to another retryable error or some other error originating from the server, that error should be raised instead

src/test/spec/retryable_reads.rs
```diff
@@ -0,0 +1,174 @@
...
+                        client
+                            .run_database_operation(&test_operation, &db_name)
+                            .await
+                    }
+                    OperationObject::Collection => {
+                        client
+                            .run_collection_operation(&test_operation, &db_name, &coll_name)
+                            .await
+                    }
+                    OperationObject::GridfsBucket => panic!("unsupported operation type"),
```
saghm, line 116: Should we just skip this (i.e. `continue` to the next iteration) instead of erroring?

src/test/spec/json/retryable-reads/retryable-reads.rst
saghm: I personally like having the yaml files during development/while I'm debugging failed tests, as they're much easier for me to read.

src/test/spec/retryable_reads.rs
isabelatkinson: This branch should never be hit since the runner should `continue` if it finds "gridfs" in the operation names at the beginning of the function. If it were to execute it would indicate that something was going wrong in the runner, so I think it's more helpful to have it panic instead of fail silently.

saghm: Makes sense. I think it would be useful for the panic message to mention the operation name in this case.

saghm:
I think I'm ready to LGTM this, assuming CI passes after a rebase.

src/test/spec/retryable_reads.rs
```diff
@@ -0,0 +1,176 @@
...
+                        test_case.description,
+                        error,
+                        result
+                    );
+                }
+                if let Some(expected_result) = operation.result {
+                    let expected_result = match expected_result {
+                        Bson::Int32(n) => Bson::Int64(n as i64),
+                        _ => expected_result,
+                    };
```
isabelatkinson, line 136: [This update](https://github.com/mongodb/bson-rust/blob/cbcd60c850476cb6f1c17f221cc0be9f331f7249/src/extjson/de.rs#L188) broke the tests for operations that return an `i64` since the expected value in the json tests is always relatively small (i.e. less than `i32::MAX`), and was therefore being deserialized into a `Bson::Int32` and then compared against a `Bson::Int64`. This workaround just converts any expected result that's a `Bson::Int32` to a `Bson::Int64`, and should be fine to do in any case since there aren't any methods in the driver that return an `i32`.

saghm: I think the `match` module that Patrick wrote should be able to handle this: https://github.com/mongodb/mongo-rust-driver/blob/c4562c28d2e6b6a9330352fadb60e938ad6f5439/src/test/util/matchable.rs

src/test/spec/command_monitoring/mod.rs
```diff
@@ -67,6 +65,8 @@ async fn run_command_monitoring_test(test_file: TestFile) {
             }
         }
 
+        let _guard = LOCK.run_exclusively().await;
```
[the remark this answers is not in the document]
isabelatkinson: Oh good catch, not sure why that got moved. I'll revert

src/test/spec/runner/test_event.rs
```diff
@@ -0,0 +1,97 @@
...
+    bson::Document,
+    test::{CommandEvent, Matchable},
+};
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum TestEvent {
+    CommandStartedEvent {
+        command_name: Option<String>,
```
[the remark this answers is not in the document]
isabelatkinson: This still needs to be an option because some of the json tests don't specify a command/database name in the expected events, but they need to be serialized into `CommandStartedEvent`s.

src/test/spec/runner/test_event.rs
```diff
@@ -0,0 +1,97 @@
...
+                    command: expected_command,
+                },
+            ) => {
+                if actual_command_name.is_some()
+                    && expected_command_name.is_some()
+                    && actual_command_name != expected_command_name
+                {
+                    return false;
+                }
+                if actual_database_name.is_some()
```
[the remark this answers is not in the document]
isabelatkinson: Oh good point, `actual_database_name` should never be `None` anyway so this check is unnecessary.

src/test/util/mod.rs
isabelatkinson: This function is being called to parse the versions so they can be passed into `server_version_gt` or `server_version_lt` which both only compare against the major/minor versions, so using `Version::parse` to parse out the patch version won't add any extra accuracy. Maybe we should introduce an `Option<u64>` patch parameter to the gt/lt methods?

src/test/util/event.rs
```diff
@@ -183,6 +183,16 @@ impl EventClient {
         }
         panic!("could not find event for {} command", command_name);
     }
+
+    pub fn topology(&self) -> String {
```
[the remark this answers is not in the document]
isabelatkinson: Sorry, must've missed responding to this one; using a `String` here is a bit simpler because the topology types in the tests ("sharded", "replicaset", "single") don't correspond directly to the `TopologyType` types (`Single`, `ReplicaSetNoPrimary`, `ReplicaSetWithPrimary`, `Sharded`, `Unknown`), so there isn't a direct translation from "replicaset" in the test file into a `TopologyType`. I suppose I could just choose one of the replica set types to use for serializing into a `RunOn`/returning from this method? Or maybe introduce a separate test-specific `TopologyType` enum for the three options that the json provides.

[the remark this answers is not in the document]
patrickfreed: Hm, I see. Probably just sticking with strings is fine then.

src/test/util/mod.rs
patrickfreed: I think eventually the gt/lt methods can just be removed. `semver::Version` is `Ord`, so we can just use the regular comparison operators on the type itself. Also note that `TestClient.server_version` is public, so we can just do `parsed_version > client.server_version` etc.

PR #180, opened 2020-05-22
RUST-434 Rewrite ReadConcern as struct

patrickfreed:
[RUST-434](https://jira.mongodb.org/browse/RUST-434)

This PR reimplements a read concern as a struct (`ReadConcern`) with a single field `level` (`ReadConcernLevel` enum) both to better match with the spec and to allow for future compatibility. It also fixes a bug where every read concern parsed from a URI was being initialized as a `ReadConcernLevel::Custom`.

### DIFF
--- a/src/concern/mod.rs
+++ b/src/concern/mod.rs
@@ -20,8 +20,59 @@ use crate::{
 ///
 /// See the documentation [here](https://docs.mongodb.com/manual/reference/read-concern/) for more
 /// information about read concerns.
-#[derive(Clone, Debug)]
-pub enum ReadConcern {
+#[derive(Clone, Debug, Serialize, PartialEq)]
+#[non_exhaustive]
+pub struct ReadConcern {
+    /// The level of the read concern.
+    pub level: ReadConcernLevel,
+}
+
+impl ReadConcern {
+    /// Creates a read concern with level "majority".
+    /// See the specific documentation for this read concern level [here](https://docs.mongodb.com/manual/reference/read-concern-majority/).
+    pub fn majority() -> Self {
+        ReadConcernLevel::Majority.into()
+    }
+
+    /// Creates a read concern with level "local".
+    /// See the specific documentation for this read concern level [here](https://docs.mongodb.com/manual/reference/read-concern-local/).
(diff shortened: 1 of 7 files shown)


src/concern/mod.rs
```diff
@@ -39,34 +90,35 @@ pub enum ReadConcern {
     Custom(String),
 }
 
-impl PartialEq for ReadConcern {
```
patrickfreed, line 42: given that this `ParitalEq` implementation led to a bug in our URI parser, I wasn't inclined to keep it around. It may be mildly confusing for a user to see that `ReadConcernLevel::Custom("majority") != ReadConcernLevel::Majority`, but that should be rare given that `ReadConcern::custom` parses correct variant out.

src/concern/mod.rs
```diff
@@ -39,34 +90,35 @@ pub enum ReadConcern {
...
-    /// Gets the string representation of the `ReadConcern`.
-    pub fn as_str(&self) -> &str {
-        match *self {
-            ReadConcern::Local => "local",
-            ReadConcern::Majority => "majority",
-            ReadConcern::Linearizable => "linearizable",
-            ReadConcern::Available => "available",
-            ReadConcern::Custom(ref s) => s,
+    /// Gets the string representation of the `ReadConcernLevel`.
+    pub(crate) fn as_str(&self) -> &str {
```
patrickfreed, line 105: this didn't seem necessary as part of the public API. If anything, maybe a `Display` implementation would be preferable. I could add it back if we think users would miss it though.

src/concern/mod.rs
```diff
@@ -39,34 +90,35 @@ pub enum ReadConcern {
     Custom(String),
 }
 
-impl PartialEq for ReadConcern {
-    fn eq(&self, other: &Self) -> bool {
-        self.as_str() == other.as_str()
+impl ReadConcernLevel {
+    pub(crate) fn from_str(s: &str) -> Self {
```
patrickfreed, line 94: this may be useful for users, but I opted to keep it private until we decided that. When exposed publicly, it would probably be better as part of a `From<>` implementation.

PR #182, opened 2020-05-26
RUST-452 Fix tests with async-std 1.6.x

saghm:


src/test/mod.rs
```diff
@@ -14,15 +14,15 @@ pub(crate) use self::{
...
+use crate::options::ClientOptions;
 
 const MAX_POOL_SIZE: u32 = 100;
 
 lazy_static! {
     pub(crate) static ref CLIENT_OPTIONS: ClientOptions = {
         let uri = std::env::var("MONGODB_URI")
             .unwrap_or_else(|_| "mongodb://localhost:27017".to_string());
-        let mut options = RUNTIME.block_on(ClientOptions::parse(&uri)).unwrap();
+        let mut options = ClientOptions::parse_without_srv_resolution(&uri).unwrap();
```
yoshuawuyts, line 25: Neat!

Cargo.toml
```diff
@@ -49,7 +49,7 @@ webpki = "0.21.0"
 webpki-roots = "0.18.0"
 
 [dependencies.async-std]
-version = "~1.5.0"
+version = "1.6.2"
```
yoshuawuyts, line 52: :tada:

yoshuawuyts:
> Does 1.6 a really serious performance regression?

@patrickfreed We had some performance issues with the 1.6.x releases, as that was when we switched to a `smol`-based core. Those issues have since been resolved, and performance should be reliable.

yoshuawuyts:
I've run an example with `async-mongodb-session` and haven't experienced any timeouts or hangs so far. We did find another bug in our impl of the session store (https://github.com/yoshuawuyts/async-mongodb-session/issues/3), but that's most likely because of logic on our side than something caused by the mongo driver.

saghm:
@yoshuawuyts we plan to do a beta release of the driver tomorrow, so this change will be available to users then!

yoshuawuyts:
@saghm yay, exciting!

PR #184, opened 2020-05-26
RUST-439 Return errors when unacknowledged write concerns are passed in

isabelatkinson:


src/client/executor.rs
```diff
@@ -38,6 +38,13 @@ impl Client {
     /// an implicit session will be created if the operation and write concern are compatible with
     /// sessions.
     pub(crate) async fn execute_operation<T: Operation>(&self, op: T) -> Result<T::O> {
+        // TODO RUST-9: allow unacknowledged write concerns
```
saghm, line 41: Wow, that's a low ticket number

PR #186, opened 2020-05-27
Update the os_info crate version

DarkEld3r:
Additionally I have disabled the default features to (slightly) speedup compilation.


DarkEld3r:
I'm not sure if I'm able to see the reason of CI failures.

saghm:
They're not related to your patch; we had some issues with the async-std update, so we temporarily pinned to 1.5.x. Any chance you could rebase master, which should fix them?

DarkEld3r:
@saghm Done.

DarkEld3r:
It seems all checks have passed now. :green_circle:

PR #189, opened 2020-06-01
RUST-13 Add indexOptionDefaults to createCollections

isabelatkinson:


src/test/db.rs
```diff
@@ -318,3 +323,61 @@ async fn db_aggregate_disk_use() {
...
+async fn create_index_options_defaults_not_specified() {
+    index_option_defaults_test(None).await;
+}
+
+#[function_name::named]
+async fn index_option_defaults_test(defaults: Option<IndexOptionDefaults>) {
+    let _guard = LOCK.run_concurrently().await;
+
+    let mut options = CLIENT_OPTIONS.clone();
+    if TestClient::new().await.is_sharded() {
```
isabelatkinson, line 347: These tests originally failed on sharded clusters, so I stripped out the extra hosts similarly to retryable reads.

src/test/db.rs
```diff
@@ -318,3 +323,61 @@ async fn db_aggregate_disk_use() {
...
+    index_option_defaults_test(Some(defaults)).await;
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn create_index_options_defaults_not_specified() {
+    index_option_defaults_test(None).await;
+}
+
+#[function_name::named]
```
saghm, line 342: I think the two individually tests should pass in their function names to use rather than using the name of the shared function here. Otherwise, since the tests are running concurrently, they could race to create the same collection with different options.

src/test/db.rs
```diff
@@ -318,3 +323,61 @@ async fn db_aggregate_disk_use() {
...
+        .index_option_defaults(defaults.clone())
+        .build();
+    db.create_collection(function_name!(), options)
+        .await
+        .unwrap();
+    db.drop(None).await.unwrap();
+
+    let events = client.command_events.read().unwrap();
+    let mut iter = events.iter().filter(|event| match event {
+        CommandEvent::CommandStartedEvent(CommandStartedEvent { command_name, .. }) => {
```
saghm, line 363: Rather than doing the pattern match twice, I think using `https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map` above to return the command would be a little cleaner.

src/db/options.rs
```diff
@@ -90,6 +90,13 @@ pub struct CreateCollectionOptions {
     /// The write concern for the operation.   
     #[builder(default)]
     pub write_concern: Option<WriteConcern>,
+
+    /// The default configuration for indexes. This option accepts a `storageEngine` document in
```
patrickfreed, line 94: The first sentence could use some clarification. Maybe add " created on this collection, including the `_id` index."

Also, the second part documenting the storage engine option should go on the `storage_engine` option itself, rather than the field here.

src/db/options.rs
```diff
@@ -117,6 +124,12 @@ pub enum ValidationAction {
     Warn,
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct IndexOptionDefaults {
```
patrickfreed, line 129: the struct needs a docstring. it can be similar to the one in `CreateCollectionOptions`, while also mentioning that it should be used when creating collections.

src/test/db.rs
patrickfreed: What were the failures you were seeing? I don't think having multiple mongoses should cause problems here but I could be wrong.

isabelatkinson: [This run](https://evergreen.mongodb.com/version/5ed588f057e85a11b81f1688) failed on three sharded clusters when doing `db.createCollection()`, all with namespace not found. Stripping out the extra hosts seemed to do away with the problem, although I didn't investigate much further.

patrickfreed: So it looks like the error was happening in that patch because you were re-using the same database without calling [`flushRouterConfig`](https://docs.mongodb.com/manual/reference/command/dropDatabase/#replica-set-and-sharded-clusters) post-drop. Your latest patches have different databases per test, so they should work fine even using both mongoses.

isabelatkinson: Gotcha, I'll revert this change.

PR #190, opened 2020-06-02
RUST-372 Allow hinting the delete command

isabelatkinson:


src/test/util/event.rs
```diff
@@ -183,6 +183,24 @@ impl EventClient {
         }
         panic!("could not find event for {} command", command_name);
     }
+
+    /// Gets all of the command started events for a specified command name.
+    pub fn get_command_started_events(&self, command_name: &str) -> Vec<CommandStartedEvent> {
```
isabelatkinson, line 188: This pattern has come up in a few tests I've written, so I factored it out into a method on `EventClient` and refactored a test I wrote for an option I previously implemented to use it.

src/coll/options.rs
```diff
@@ -254,7 +254,7 @@ pub struct ReplaceOptions {
 /// [`Collection::delete_one`](../struct.Collection.html#method.delete_one) or
 /// [`Collection::delete_many`](../struct.Collection.html#method.delete_many) operation.
 #[serde_with::skip_serializing_none]
-#[derive(Debug, Default, TypedBuilder, Serialize)]
+#[derive(Clone, Debug, Default, TypedBuilder, Serialize)]
```
saghm, line 257: Good catch! I thought we were already deriving this, but I guess not

src/test/coll.rs
```diff
@@ -515,23 +523,16 @@ async fn allow_disk_use_test(options: FindOptions, expected_value: Option<bool>)
...
+    assert_eq!(events.len(), 1);
 
-    let event = iter.next().unwrap();
-    let allow_disk_use = match event {
-        CommandEvent::CommandStartedEvent(CommandStartedEvent { command, .. }) => {
-            command.get_bool("allowDiskUse").ok()
-        }
-        _ => None,
-    };
+    let allow_disk_use = events
```
saghm, line 529: You can also use `events[0]` instead of `events.first().unwrap()`

src/test/coll.rs
```diff
@@ -545,3 +546,91 @@ async fn ns_not_found_suppression() {
     coll.drop(None).await.expect("drop should not fail");
     coll.drop(None).await.expect("drop should not fail");
 }
+
+async fn delete_hint_test(options: Option<DeleteOptions>, name: &str) {
+    let _guard = LOCK.run_concurrently().await;
+
+    let client = EventClient::new().await;
+    let coll = client.database(name).collection(name);
+    let _ = coll.delete_many(doc! {}, options.clone()).await;
```
saghm, line 555: Patrick and I have started doing `let _: Result =` instead of `let _ =`. Otherwise, we might accidentally ignore something besides a result that's `must_use`, e.g. forgetting to await a future.

src/test/coll.rs
```diff
@@ -545,3 +546,91 @@ async fn ns_not_found_suppression() {
...
+
+    let events = client.get_command_started_events("delete");
+    assert_eq!(events.len(), 1);
+
+    let event_hint = events.first().unwrap().command.get("hint").cloned();
```
saghm, line 560: Ditto above about `events[0]`

src/test/coll.rs
```diff
@@ -545,3 +546,91 @@ async fn ns_not_found_suppression() {
...
+    let expected_hint = match options {
+        Some(options) => options.hint.map(|hint| hint.to_bson()),
```
saghm, line 562: I think this match can be replaced with `options.and_then(|opts| opts.hint).map(|hint| hint.to_bson())`.

src/coll/options.rs
```diff
@@ -268,12 +268,16 @@ pub struct DeleteOptions {
     /// The write concern for the operation.
     #[builder(default)]
     pub write_concern: Option<WriteConcern>,
+
+    /// The index to use for the operation.
```
patrickfreed, line 272: This and all the other hints should be documented as only being supported in MongoDB 4.4+.

src/test/coll.rs
```diff
@@ -545,3 +543,120 @@ async fn ns_not_found_suppression() {
...
+#[function_name::named]
+async fn delete_hint_not_specified() {
+    delete_hint_test(None, function_name!()).await;
+}
+
+async fn find_one_and_delete_hint_test(options: Option<FindOneAndDeleteOptions>, name: &str) {
+    let _guard = LOCK.run_concurrently().await;
+    let client = EventClient::new().await;
+
+    let req = VersionReq::parse("< 4.2").unwrap();
```
patrickfreed, line 594: I think this ought to be `"<=4.2"`, since this should fail in 4.2 as well. Kind of curious why this isn't causing errors right now.

isabelatkinson: I have `< 4.2` because the spec change you linked indicates that the driver should only raise an error (i.e. not send the command to the server) for servers < 4.2 -- "Older servers >= 4.2 will report an error for using this option." This test is only checking that the server received the command with the option specified, which is correct behavior for 4.2 even though the server will return an error. The test below (`find_one_and_delete_hint_server_version`) ensures that `find_one_and_delete` returns the correct thing for different server versions (i.e. driver error for < 4.2, server error for == 4.2, ok for 4.4+).

patrickfreed: Ah ok, makes sense :+1:

PR #191, opened 2020-06-03
update to latest bson version

saghm:


src/bson_util/async_encoding.rs
```diff
@@ -1,7 +1,7 @@
-use bson::Document;
 use futures::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 
 use crate::{
+    bson::Document,
```
saghm, line 4: As a sanity check to make sure the re-exporting of BSON worked, I changed all the imports in the driver to use BSON components from there rather than from the external package.

src/bson_util/mod.rs
```diff
@@ -61,12 +63,15 @@ pub(crate) fn update_document_check(update: &Document) -> Result<()> {
...
 }
 
-pub(crate) fn serialize_duration_as_i64_millis<S: Serializer>(
+pub(crate) fn serialize_duration_as_int_millis<S: Serializer>(
     val: &Option<Duration>,
     serializer: S,
 ) -> std::result::Result<S::Ok, S::Error> {
     match val {
-        Some(duration) => serializer.serialize_i64(duration.as_millis() as i64),
+        Some(duration) if duration.as_millis() > i32::MAX as u128 => {
```
saghm, line 71: Since our extjson parser now parses integers into the smallest width possible, I needed to update how we serialized durations for the purpose of tests. This shouldn't have any external impact on users, as the server is happy to read a timeout option as a 32-bit integer instead of a 64-bit one.

There are dozens of similar changes below; I didn't bother calling them all out.

src/sdam/description/topology/test/rtt.rs
```diff
@@ -21,13 +21,15 @@ pub struct TestFile {
...
     F(f64),
+    I(i32),
     S(String),
 }
 
 async fn run_test(test_file: TestFile) {
     let avg_rtt_ms = match test_file.avg_rtt_ms {
         AverageRtt::F(f) => Some(f),
         AverageRtt::S(ref s) if s == "NULL" => None,
+        AverageRtt::I(i) => Some(i as f64),
```
saghm, line 32: Our updated extjson parser requires the ".0" suffix to read in an integer as a double, so I added an i32 case here so that these tests would work.

src/test/spec/command_monitoring/operation.rs
```diff
@@ -108,28 +108,6 @@ impl TestOperation for DeleteOne {
     }
 }
 
-fn deserialize_i64_from_ext_json<'de, D>(
```
saghm, line 111: It was nice to get to remove this from our test suite

src/test/spec/mod.rs
```diff
@@ -43,6 +45,12 @@ where
         let json: Value =
             serde_json::from_reader(File::open(test_file_full_path.as_path()).unwrap()).unwrap();
 
-        run_test_file(bson::from_bson(Bson::from(json)).unwrap()).await
+        run_test_file(
```
saghm, line 48: I ran into some errors here due to our manual handling of extjson, so to help me trace them, I added logging for the exact file that we failed to parse in the case that it failed.

src/test/spec/read_write_concern/connection_string.rs
```diff
@@ -19,10 +18,6 @@ struct TestCase {
 }
 
 fn normalize_write_concern_doc(mut write_concern_doc: Document) -> Document {
-    if let Some(&Bson::I32(i)) = write_concern_doc.get("w") {
```
saghm, line 22: So nice not to have to do this anymore

src/test/spec/read_write_concern/document.rs
```diff
@@ -55,21 +55,6 @@ fn write_concern_from_document(write_concern_doc: Document) -> Option<WriteConce
     Some(write_concern)
 }
 
-fn normalize_write_concern_doc(write_concern_doc: Document) -> Document {
```
saghm, line 58: Soooooooooo nice

saghm:
I fixed the clippy complaints. I'll follow-up with another commit that updates the README and bumps the version number after this gets merged.

PR #193, opened 2020-06-04
RUST-170 Enable and configure TCP keepalive by default

isabelatkinson:


Cargo.toml
```diff
@@ -36,6 +36,7 @@ rand = "0.7.2"
 serde_with = "1.3.1"
 sha-1 = "0.8.1"
 sha2 = "0.8.0"
+socket2 = "0.3.8"
```
saghm, line 39: It looks like the latest version is 0.3.12; 0.3.8 appears to have come out two years ago. Does it still work if you use 0.3.12?

isabelatkinson: Oh whoops, I'll update that

PR #196, opened 2020-06-05
Make DB Module Public

danii:
Made DB module public so `mongodb::db::Database` can be passed around other crates more freely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ let mut client_options = ClientOptions::parse("mongodb://localhost:27017").await
 client_options.app_name = Some("My App".to_string());
 
 // Get a handle to the deployment.
-let client = Client::with_options(client_options)?;
+let client: Client = Client::with_options(client_options)?;
 
 // List the names of the databases in that deployment.
 for db_name in client.list_database_names(None, None).await? {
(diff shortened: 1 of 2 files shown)


saghm:
I'm not sure I understand the purpose of this pull request. `Database` is exported at the top-level `mongodb::Database`, and all of the options in `db::options` are exported in the `mongodb::options` module; nothing else in `mongodb::db` would be exported as a result of this change. Could you elaborate on what you're trying to do that this change would help?

Daniihh:
Oh, my bad, I didn't notice it was exported at the top level. I never saw anything in the README.md pointing to the fact that I was supposed to get the type from the top level, and just assumed the path that my linter showed me `mongodb::db::Database` would've worked fine. Since my main reason for creating this pull request was to be able to access the type, I'm gonna switch this pull request into modifying the README.md to better point out the top level types, if that's okay.

saghm:
Out of curiosity, did you look at the API documentation linked to at the top of the [example usage section](https://github.com/mongodb/mongo-rust-driver#example-usage)? Our assumption was that users would be able to find our through the API documentation that `Client`/`Database`/Collection`/`Cursor` are exported at the top-level, but if it's not obvious from the current state of the documentation, it might be worth calling it out explicitly in the prose higher up in the page.

Daniihh:
No, I just thought I'd be familiar enough with the library from using MongoDB libraries in other languages. Granted, I feel like if I had checked that documentation that it would be clear, making the README.md more verbose to point it out there may be worthwhile. (You just witnessed an overconfident guy skipping reading the documentation because he thought he knew how it would work from other libraries, it might happen again. 😂)

agolin95:
Tracked by https://jira.mongodb.org/browse/RUST-594

saghm:
After discussing with the team, we feel that this is sufficiently covered by existing documentation, so we're going to close this PR without merging

PR #200, opened 2020-06-17
Added optional bson u2i feature

jwillbold:
The BSON crate offered the "unsigned integer to integer (u2i)" auto conversion feature. Since the crate is now reexported, it would be nice to still be able to use this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,10 @@ default = ["tokio-runtime"]
 tokio-runtime = ["tokio/macros", "tokio/net", "tokio/rt", "tokio/time", "reqwest", "serde_bytes"]
 async-std-runtime = ["async-std", "async-std/attributes", "async-std-resolver", "tokio-util/compat"]
 sync = ["async-std-runtime"]
+# The bson/u2i feature enables automatic conversion from unsigned to signed types during
+# serialization. This feature is intended for use when serializing data types in third-party crates
+# whose implementation cannot be changed; otherwise, it is preferred to use the helper functions
+# provided in the bson::serde_helpers module.
 bson-u2i = ["bson/u2i"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


agolin95:
Tracked by https://jira.mongodb.org/browse/RUST-514

isabelatkinson:
Hi @jwillbold! https://github.com/mongodb/bson-rust/commit/e03e95e05a8f8da4d71f088c5e06c62488ba77ce includes helper functions to serialize unsigned numbers as signed numbers and will be part of the next release of the BSON library. You can use [Serde field attributes](https://serde.rs/field-attrs.html) to serialize your fields with these functions.

isabelatkinson:
@jwillbold Thanks for your feedback! We've decided to enable the feature based on your concerns. I'm going to add another commit to add some documentation containing guidance on when to use this feature and then tag in the rest of the team for review.

jwillbold:
Thanks!

Cargo.toml
```diff
@@ -24,6 +24,10 @@ default = ["tokio-runtime"]
 tokio-runtime = ["tokio/macros", "tokio/net", "tokio/rt", "tokio/time", "reqwest", "serde_bytes"]
 async-std-runtime = ["async-std", "async-std/attributes", "async-std-resolver", "tokio-util/compat"]
 sync = ["async-std-runtime"]
+# The bson/u2i feature enables automatic conversion from unsigned to signed types during
+# serialization. It is preferred to use the helper functions provided in the bson::serde_helpers
+# module, which cover most use cases for this feature.
```
saghm, line 29: I think it's worth specifically mentioning wanting to serialize/deserialize types from third-party crates (i.e. ones that you don't have access to changing the implementation of) as the intended use case for this feature. Otherwise, it might not be clear to users that it's expected in that scenario.

saghm (approved):

patrickfreed (approved):

PR #202, opened 2020-06-22
RUST-359 Implement MONGODB-AWS auth support

saghm:


src/client/auth/aws.rs
```diff
@@ -0,0 +1,402 @@
...
+            "x-amz-security-token;"
+        } else {
+            ""
+        };
+
+        // Generate the list of signed headers (either with or without the security token header).
+        #[rustfmt::skip]
+        let signed_headers = format!(
+            "\
+              content-type;\
```
markbenvenuto, line 228: These headers need to be in alphabetical order
https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html

src/client/auth/aws.rs
```diff
@@ -0,0 +1,402 @@
...
+        );
+
+        let nonce = base64::encode(server_nonce);
+
+        #[rustfmt::skip]
+		let request = format!(
+		    "\
+             POST\n\
+             /\n\
+ 	         Action=GetCallerIdentity&Version=2011-06-15\n\
```
markbenvenuto, line 245: The body is not part of the canonical request

See https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html

src/client/auth/aws.rs
```diff
@@ -0,0 +1,402 @@
...
+            let parts: Vec<_> = host.split('.').collect();
+            parts.get(1).copied().unwrap_or("us-east-1")
+        };
+
+        #[rustfmt::skip]
+		let string_to_sign = format!(
+			"\
+             AWS4-HMAC-SHA256\n\
+             {full_date}\n\
+			 {small_date}/{region}/iam/aws4_request\n\
```
markbenvenuto, line 280: This should be `sts`, not `iam`

src/client/auth/aws.rs
```diff
@@ -0,0 +1,402 @@
...
+			full_date = date_str,
+			small_date = small_date,
+			region = region,
+			hashed_request = hashed_request,
+		);
+
+        let first_hmac_key = format!("AWS4{}", self.secret_key);
+        let k_date = hmac(first_hmac_key, &small_date)?;
+        let k_region = hmac(k_date, region)?;
+        let k_service = hmac(k_region, "iam")?;
```
markbenvenuto, line 292: This should be `sts`, not `iam`

src/client/auth/aws.rs
```diff
@@ -0,0 +1,402 @@
...
+        let k_signing = hmac(k_service, "aws4_request")?;
+
+        let signature_bytes = hmac(k_signing, string_to_sign)?;
+        let signature = hex::encode(signature_bytes);
+
+        #[rustfmt::skip]
+		let auth_header = format!(
+	        "\
+             Authorization: AWS-4-HMAC-SHA-256 \
+             Credential={access_key}/{small_date}/{region}/iam/aws4_request, \
```
markbenvenuto, line 302: This should be `sts`, not `iam`

src/client/auth/mod.rs
```diff
@@ -165,6 +189,15 @@ impl FromStr for AuthMechanism {
             MONGODB_X509_STR => Ok(AuthMechanism::MongoDbX509),
             GSSAPI_STR => Ok(AuthMechanism::Gssapi),
             PLAIN_STR => Ok(AuthMechanism::Plain),
+
+            #[cfg(feature = "tokio-runtime")]
+            MONGODB_AWS_STR => Ok(AuthMechanism::MongoDbAws),
+            #[cfg(not(feature = "tokio-runtime"))]
+            MONGODB_AWS_STR => Err(ErrorKind::ArgumentError {
```
saghm, line 196: I figured a custom error message for this case would mitigate potential confusion from users in the case that they try to use AWS auth with async-std.

src/client/auth/mod.rs
```diff
@@ -262,12 +299,16 @@ impl Credential {
         };
 
         // Authenticate according to the chosen mechanism.
-        mechanism.authenticate_stream(conn, self).await
+        mechanism.authenticate_stream(conn, self, http_client).await
     }
 }
 
-pub(crate) fn generate_nonce() -> String {
+pub(crate) fn generate_nonce_bytes() -> Vec<u8> {
```
saghm, line 306: We need the nonce as bytes in AWS auth, so I pulled this from `generate_nonce` into a separate method.

src/client/auth/sasl.rs
```diff
@@ -0,0 +1,98 @@
+use crate::{
+    bson::{doc, spec::BinarySubtype, Binary, Bson, Document},
+    bson_util,
+    client::auth::AuthMechanism,
+    error::{Error, Result},
+};
+
+/// Encapsulates the command building of a `saslStart` command.
```
saghm, line 8: Since there's a bit of duplicate functionality between the auth mechanisms in terms of building the `saslStart` and `saslContinue` commands and reading the server responses, I pulled it out from `scram` into a separate module. These are essentially mini-operations, although I didn't use the `Operation` trait since they won't ever be sent through `execute_operation`, making a lot of the `Operation` functionality overkill for them.

src/client/auth/scram.rs
```diff
@@ -142,12 +146,8 @@ impl ScramVersion {
         // Normal SCRAM implementations would cease here. The following round trip is MongoDB
         // implementation specific and just consists of a client no-op followed by a server no-op
         // with "done: true".
-        let noop = doc! {
-            "saslContinue": 1,
-            "conversationId": server_final.conversation_id().clone(),
-            "payload": Bson::Binary(Binary { subtype: BinarySubtype::Generic, bytes: Vec::new() })
-        };
-        let command = Command::new("saslContinue".into(), source.into(), noop);
+        let noop = SaslContinue::new(server_final.conversation_id().clone(), Vec::new());
```
saghm, line 149: The changes to this module are just to use the abstracted `sasl` module functionality that was pulled out.

src/runtime/http.rs
```diff
@@ -0,0 +1,84 @@
+#[cfg(feature = "tokio-runtime")]
+use reqwest::{Method, Response};
+#[cfg(feature = "tokio-runtime")]
+use serde::Deserialize;
+
+#[cfg(feature = "tokio-runtime")]
+use crate::error::Result;
+
+#[derive(Clone, Debug, Default)]
```
saghm, line 9: Rather than only defining `HttpClient` in the case that the tokio runtime is used, I instead make it an empty struct in non-tokio cases, which allows us to create it and pass it around without conditional compilation with no cost to async-std users. This helps minimize the amount of runtime-specific code we have to write outside this module.

src/sdam/state/mod.rs
```diff
@@ -107,6 +113,7 @@ impl Topology {
         let mut topology_state = TopologyState {
             description,
             servers: Default::default(),
+            http_client: Default::default(),
```
saghm, line 116: I [checked with the people who develop hyper](https://github.com/hyperium/hyper/issues/2228) (the HTTP library that reqwest wraps) to ensure that creating a client and then never using it doesn't incur any runtime penalties, which luckily is the case. This means that instead of passing around an Option and then populating it lazily (which would require a lock, since we can't otherwise modify an Option that's cloned across multiple threads), we can eagerly create the client, and if the client doesn't use AWS auth, we don't pay any costs (e.g. extra threads spun up) other than the memory to allocate it.

src/client/auth/aws.rs
```diff
@@ -0,0 +1,414 @@
...
+
+    /// Computes the signed authorization header for the credentials to send to the server in a sasl
+    /// payload.
+    fn compute_authorization_header(
+        &self,
+        date: DateTime<Utc>,
+        host: &str,
+        server_nonce: &[u8],
+    ) -> Result<String> {
+        let date_str = date.format("%Y%m%dT%H%M%SZ").to_string();
```
patrickfreed, line 212: similarly, since this format is used in a couple places, I think a constant for the format string or a function that produces the formatted date would be useful.

src/client/auth/aws.rs
```diff
@@ -0,0 +1,414 @@
...
+            let parts: Vec<_> = host.split('.').collect();
+            parts.get(1).copied().unwrap_or("us-east-1")
+        };
+
+        #[rustfmt::skip]
+        let string_to_sign = format!(
+            "\
+             AWS4-HMAC-SHA256\n\
+             {full_date}\n\
+             {small_date}/{region}/sts/aws4_request\n\
```
patrickfreed, line 289: I see in the spec that it uses `sts` in the example conversation as part of this scope, but is there anywhere in the prose where it actually says to use it over like `iam`? It's a little confusing because in one of the psuedocode examples it does use `iam`, plus the auth mechanism used to be called MONGODB-IAM.

src/client/auth/aws.rs
```diff
@@ -0,0 +1,414 @@
...
+
+    let mut client_second_payload_bytes = Vec::new();
+    client_second_payload.to_writer(&mut client_second_payload_bytes)?;
+
+    let sasl_continue = SaslContinue::new(
+        server_first.conversation_id.clone(),
+        client_second_payload_bytes,
+    );
+
+    let client_second = Command::new(
```
patrickfreed, line 81: ditto making this `sasl_continue.into_command(source.clone())`

src/client/auth/aws.rs
```diff
@@ -0,0 +1,414 @@
...
+impl ServerFirst {
+    /// Parses the response of the `saslStart` command.
+    fn parse(response: Document) -> Result<Self> {
+        let SaslResponse {
+            conversation_id,
+            payload,
+            done,
+        } = SaslResponse::parse("MONGODB-AWS", response)?;
+
+        let mut payload_document = Document::from_reader(&mut payload.as_slice())?;
```
patrickfreed, line 351: I think this could be a good opportunity to leverage `serde` to avoid all the type and presence checking boilerplate here. We'd have to map any errors to use `Error::invalid_authentication_response` as part of that.

src/error.rs
```diff
@@ -212,6 +216,14 @@ pub enum ErrorKind {
         file_path: String,
     },
 
+    #[cfg(feature = "tokio-runtime")]
+    #[error(display = "{}", _0)]
+    ReqwestError(#[error(source)] reqwest::Error),
```
patrickfreed, line 221: since these two errors only ever occur in the context of aws auth, I wonder if it would be better to just return them as authentication errors containing their error descriptions, rather than the full error structs themselves. This is similar to how we repackage any connection errors that occur during the authentication process as auth errors. 

Also, do we need to perform any sort of redaction of the contents of these errors?

src/error.rs
```diff
@@ -167,6 +167,10 @@ pub enum ErrorKind {
     #[error(display = "Command failed {}", _0)]
     CommandError(CommandError),
 
+    /// Wrapper around `cryto_mac::InvalidKeyLength`.
+    #[error(display = "{}", _0)]
+    CryptoMacInvalidKeyLength(#[error(source)] crypto_mac::InvalidKeyLength),
```
patrickfreed, line 172: similarly, should we just be returning this under authentication errors?

src/client/auth/aws.rs
```diff
@@ -0,0 +1,414 @@
...
+            region = region,
+            signed_headers = signed_headers,
+            signature = signature
+        );
+
+        Ok(auth_header)
+    }
+}
+
+fn hmac(key: impl AsRef<[u8]>, input: impl AsRef<[u8]>) -> Result<impl AsRef<[u8]>> {
```
patrickfreed, line 326: theres already a `ScramVersion::hmac` method in `scram.rs`. Would it be possible to share code between the two so that we don't have separate hmac functions for different auth mechanisms?

.evergreen/aws-ecs-test/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "aws-ecs-test"
+version = "0.1.0"
+authors = ["Saghm Rossi <saghmrossi@gmail.com>"]
+edition = "2018"
+
+[dependencies]
+tokio = "0.2.21"
```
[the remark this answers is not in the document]
patrickfreed: I cant seem to find that discussion, do you remember what the rationale was?

src/client/auth/aws.rs
patrickfreed: Yeah looking at the scram code it does seem to avoid `serde`. I honestly don't have any idea why; maybe I just wasn't that familiar with it back then or something. Whatever the reason, current me thinks that `serde` would be the way to go for this kind of stuff.

src/client/auth/mod.rs
```diff
@@ -262,12 +300,23 @@ impl Credential {
...
+}
+
+pub(crate) fn generate_nonce() -> String {
+    let result = generate_nonce_bytes();
     base64::encode(result.as_slice())
 }
+
+fn mac<M: Mac>(key: &[u8], input: &[u8]) -> Result<impl AsRef<[u8]>> {
+    let mut mac =
+        M::new_varkey(key).or_else(|_| Err(Error::unknown_authentication_error("SCRAM")))?;
```
patrickfreed, line 319: so this always will emit an error for SCRAM, but it may be used with other auth mechanisms.

src/client/auth/aws.rs
```diff
@@ -0,0 +1,421 @@
...
+
+    /// Obtains credentials from the ECS endpoint.
+    async fn get_from_ecs(relative_uri: String, http_client: &HttpClient) -> Result<Self> {
+        // Use the local IP address that AWS uses for ECS agents.
+        let uri = format!("http://{}/{}", AWS_ECS_IP, relative_uri);
+
+        http_client
+            .get_and_deserialize_json(&uri, None)
+            .await
+            .or_else(|_| Err(Error::unknown_authentication_error("SCRAM")))
```
patrickfreed, line 181: here and a few places below use "SCRAM" instead of "MONGODB_AWS" (or whatever string we're using)

src/client/auth/aws.rs
patrickfreed: does hmac 0.8.x not work for us?

src/client/auth/mod.rs
saghm: Oops, should have paid more attention to my copypasta

src/client/auth/aws.rs
saghm: The 0.8.x API seems to be designed in way that's not very conducive to making functions generic over the type of Hmac. I think it's possible to get working, but it's a pretty big hassle. They changed the API to split all of the stuff in `Mac` to be half a dozen or so different traits, and they all depend on each other in different ways. Upgrading `hmac` to 0.8.1 without changing the code yields 115 compiler errors. Adding `NewMac` (the trait that now contains `new` and `new_varkey`) bounds to all the parameters currently requiring `Mac` increases that to 116 instead of getting rid of any errors. I tried adding all the traits that are required and got up to 171 compiler errors. The changes are complex enough that I think it's probably a bit out of scope for this ticket. We can file another one and decide how much effort we want to put into it; it honestly might end up being less work to just duplicate a bit of code instead of making it generic. 

More generally, we probably want to figure out some sort of inform policy for when we want to update dependencies; a decent number of them have minor or bugfix releases out newer than the version we specify, so it might be worth deciding "we update all dependencies to the latest bugfix release on the minor version we're using whenever we do a bugfix release of the driver, and we update minor versions whenever we do a minor version release of the driver" or something like that. Right now, we're just updating only when there's a security bug or we need a new feature, and we probably don't want to be constantly updating all of our dependencies as soon as they publish new releases, but there's probably a middle ground between those that would balance the costs of churn with the risks of using outdated versions.

patrickfreed: Yeah defining such a policy sounds like a good idea. Once we have that in place we can revisit updating `hmac`.

src/client/auth/mod.rs
patrickfreed: FYI you haven't pushed any commits yet to address this (just in case you forgot, no rush if you haven't gotten to it).

saghm: Sorry about that; just pushed now

.evergreen/config.yml
```diff
@@ -310,6 +531,51 @@ tasks:
...
+    # "4.4" explicitly left off to keep this out of the generic matrix
+    tags: ["aws-auth"]
+    commands:
+      - func: "bootstrap mongo-orchestration"
+        vars:
+          ORCHESTRATION_FILE: "auth-aws.json"
+          MONGODB_VERSION: "4.4"
+          AUTH: "auth"
+          TOPOLOGY: "server"
+      - func: "add aws auth variables to file"
```
patrickfreed, line 571: does the test binary need to be recompiled to get the new environment variables? I remember something about that being true for the driver but I'm not sure.

.evergreen/aws-ecs-test/Cargo.toml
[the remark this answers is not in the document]
patrickfreed: Based on offline conversation, we're holding off on implementing it for `async-std` until a suitable HTTP library is made.

.evergreen/config.yml
saghm: Nope, not anymore. That was because we used the [`option_env!](https://doc.rust-lang.org/std/macro.option_env.html) macro, which has a nicer API in that it returns an option rather than a Result but has the downside of resolving the variable at compile time. We now use [`std::env:var`](https://doc.rust-lang.org/std/env/fn.var.html), which annoyingly returns a Result, but it resolves the variable at runtime. See [here](https://github.com/mongodb/mongo-rust-driver/blob/5970f98c145ddcdf0988b6dfe2b684f10c8ddbdc/src/test/mod.rs#L23) for where we use it.

src/client/auth/aws.rs
patrickfreed: So I think updating this to use serde is the only remaining thing for me before LGTM.

src/client/auth/mod.rs
```diff
@@ -262,12 +300,23 @@ impl Credential {
...
-        mechanism.authenticate_stream(conn, self).await
+        mechanism.authenticate_stream(conn, self, http_client).await
     }
 }
 
-pub(crate) fn generate_nonce() -> String {
+pub(crate) fn generate_nonce_bytes() -> Vec<u8> {
     let mut rng = rand::thread_rng();
-    let result: Vec<u8> = (0..32).map(|_| rng.gen()).collect();
+    (0..32).map(|_| rng.gen()).collect()
```
markbenvenuto, line 309: Would rng.fill_bytes() not be faster?

src/client/auth/aws.rs
markbenvenuto: The server uses IDL since I trust a machine generated parser more then a user-written one.

src/client/auth/aws.rs
```diff
@@ -0,0 +1,423 @@
...
+            .await
+            .or_else(|_| Err(Error::unknown_authentication_error("MONGODB-AWS")))
+    }
+
+    /// Obtains temporary credentials for an EC2 instance to use for authentication.
+    async fn get_from_ec2(http_client: &HttpClient) -> Result<Self> {
+        let temporary_token = http_client
+            .put_and_read_string(
+                &format!("http://{}/latest/api/token", AWS_EC2_IP),
+                &[("X-aws-ec2-metadata-token-ttl-seconds", "30")],
```
markbenvenuto, line 189: The server got bit by the Expect header since curl sends it by default. I assume since all tests pass that you did not hit something similar.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect

saghm: Nope, I didn't run into anything related to the `Expect` header. What exactly happened that caused an issue with the server?

PR #206, opened 2020-06-24
minor: remove unused variables

saghm:
It appears that we accidentally never removed the project-wide annotation to suppress unused variable warnings, so I removed it and fixed all the issues. Fortunately, almost all of these were not indicators of bugs; the one exception was a very minor bug in handshake metadata logic for libraries wrapping the driver, where we accidentally appended the platform name to the driver's version string instead of the platform string (but with the proper delimiter, so this likely didn't break anything).

### DIFF
--- a/src/event/cmap.rs
+++ b/src/event/cmap.rs
@@ -230,41 +230,41 @@ pub struct ConnectionCheckedInEvent {
 pub trait CmapEventHandler: Send + Sync {
     /// A [`Client`](../../struct.Client.html) will call this method on each registered handler
     /// whenever a connection pool is created.
-    fn handle_pool_created_event(&self, event: PoolCreatedEvent) {}
+    fn handle_pool_created_event(&self, _event: PoolCreatedEvent) {}
 
     /// A [`Client`](../../struct.Client.html) will call this method on each registered handler
     /// whenever a connection pool is cleared.
-    fn handle_pool_cleared_event(&self, event: PoolClearedEvent) {}
+    fn handle_pool_cleared_event(&self, _event: PoolClearedEvent) {}
 
     /// A [`Client`](../../struct.Client.html) will call this method on each registered handler
     /// whenever a connection pool is cleared.
-    fn handle_pool_closed_event(&self, event: PoolClosedEvent) {}
+    fn handle_pool_closed_event(&self, _event: PoolClosedEvent) {}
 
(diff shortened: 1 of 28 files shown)


src/sdam/monitor.rs
```diff
@@ -55,13 +52,13 @@ impl Monitor {
             .heartbeat_freq
             .unwrap_or(DEFAULT_HEARTBEAT_FREQUENCY);
 
-        while let Some(server) = self.server.upgrade() {
+        while let Some(_server) = self.server.upgrade() {
```
saghm, line 55: We want to make sure we can upgrade the weak reference to the server to a strong reference, since being unable to indicates that the server was dropped from the cluster, but we don't actually need the server itself for the body of the loop.

src/sdam/state/mod.rs
```diff
@@ -254,8 +254,6 @@ impl Topology {
         // references to the same instances though, since each is wrapped in an `Arc`.
         let mut state_clone = self.state.read().await.clone();
 
-        let old_description = state_clone.description.clone();
```
saghm, line 257: We call [`update_state`](https://github.com/mongodb/mongo-rust-driver/blob/6dc30284f139ed098a90d908762b84dfaeceec5d/src/sdam/state/mod.rs#L297-L328) to do the work of replacing the topology state, which handles [replacing the description and  starting the new monitoring threads](https://github.com/mongodb/mongo-rust-driver/blob/6dc30284f139ed098a90d908762b84dfaeceec5d/src/sdam/state/mod.rs#L297-L328) without needing the old description, so this dead code doesn't represent any missing logic, but rather an unnecessary variable.

src/event/cmap.rs
```diff
@@ -230,41 +230,41 @@ pub struct ConnectionCheckedInEvent {
 pub trait CmapEventHandler: Send + Sync {
     /// A [`Client`](../../struct.Client.html) will call this method on each registered handler
     /// whenever a connection pool is created.
-    fn handle_pool_created_event(&self, event: PoolCreatedEvent) {}
+    fn handle_pool_created_event(&self, _event: PoolCreatedEvent) {}
```
utkmehta, line 233: Question: Why not just use _ instead of _event?

saghm: These are methods on a public trait, so I think it's useful to have names for the parameters to ease readability for users looking at the docs.

PR #207, opened 2020-06-25
RUST-376 Added test to verify if hints have been added to update properly

utkmehta:
This PR adds a test to make sure that hints get added to updates properly


utkmehta:
> Are these tests not already implemented in #203 ?

Those tests are for findAndModify functions. This PR is for testing hint in update operations

PR #208, opened 2020-06-26
RUST-302 Implement v2 spec test runner

isabelatkinson:


src/test/spec/runner/test_file.rs
```diff
@@ -37,8 +39,8 @@ pub struct RunOn {
 impl RunOn {
     pub fn can_run_on(&self, client: &EventClient) -> bool {
         if let Some(ref min_version) = self.min_server_version {
-            let req = VersionReq::parse(&format!(">= {}", &min_version)).unwrap();
-            if !req.matches(&client.server_version) {
+            let req = VersionReq::parse(&format!("< {}", &min_version)).unwrap();
```
isabelatkinson, line 42: Quirk of using `VersionReq`: a release candidate version is not considered to be greater than a lower non-release candidate version (e.g. `4.4.0-rc11` is not >= `4.2`). Flipping the logic here fixes that issue.

src/test/util/matchable.rs
```diff
@@ -59,6 +59,20 @@ impl Matchable for Bson {
 impl Matchable for Document {
     fn content_matches(&self, expected: &Document) -> bool {
         for (k, v) in expected.iter() {
+            if k == "upsertedCount" {
```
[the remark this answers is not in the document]
saghm: A more concise way would be to use the fact that `Option` is an iterator that always yields zero or one things; you could do `v == actual_v.iter().count()`

src/test/spec/runner/test_file.rs
isabelatkinson: This is linked in the source code for `semver` -- the logic makes sense but doesn't seem as applicable to the driver. Is there any reason we'd need to know if a server version is prerelease? If not, maybe it would be better to strip out the tag to avoid accidentally skipping the latest server version in future tests. https://docs.npmjs.com/misc/semver#prerelease-tags

src/test/util/matchable.rs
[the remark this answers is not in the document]
isabelatkinson: I don't think that would work since there's a difference between `Some(Bson::Null)` and `None` here -- `None` indicates that the field doesn't exist, which means the documents don't match, whereas `Some(Bson::Null)` corresponds to `upsertedCount` being 0.

src/test/spec/runner/test_file.rs
saghm: From grepping the specifications repo for `rc0` through `rc10`, there doesn't seem to be any usage of it in any JSON or YAML files. There are a few things about `listCollections` with certain rc versions in the spec itself, but those are for versions below 3.6, so we don't need to worry about them. There's also a mention of an rc in the auth spec, but since it's not in any of the tests, we don't need to worry about it here. Based on all that, I think it's probably fine to strip it out for now. We can revisit if there ever ends up being any spec tests that use prelease tags.

src/test/util/matchable.rs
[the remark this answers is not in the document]
saghm: Oh, good point; I didn't think of that

src/test/coll.rs
```diff
@@ -660,3 +660,59 @@ async fn find_one_and_delete_hint_server_version() {
...
+        .unwrap();
+
+    let error = client
+        .database(function_name!())
+        .collection(function_name!())
+        .insert_one(doc! { "x": 1 }, None)
+        .await
+        .expect_err("insert should fail");
+
+    let _expected = ErrorKind::WriteError(WriteFailure::WriteConcernError(WriteConcernError {
```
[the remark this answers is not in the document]
isabelatkinson: Ah good catch (both you and the linter)

src/client/options/mod.rs
```diff
@@ -702,6 +714,87 @@ impl ClientOptions {
         }
         Ok(())
     }
+
+    /// Merges these options with another set of ClientOptions (excluding hosts). If a field is
+    /// present on both options, favors the option present on self.
+    #[cfg(test)]
+    pub(crate) fn merge(&mut self, other: ClientOptions) {
```
patrickfreed, line 721: this could possibly just be replaced with a usage of the `resolve_options` macro we have defined in `src/options.rs`. The macro expects the second argument to be an `Option` though, so I'm not sure if the current mutable approach would work. You could instead return a new options struct that's a result of the merging.

Check out `src/coll/mod.rs` for some usages of it.

src/coll/options.rs
```diff
@@ -477,7 +489,8 @@ pub struct AggregateOptions {
     #[builder(default)]
     #[serde(
         serialize_with = "serialize_duration_as_int_millis",
-        rename = "maxTimeMS"
+        rename(serialize = "maxTimeMS"),
```
patrickfreed, line 492: I think this needs to be deserialized from `maxTimeMS` also

src/test/spec/runner/mod.rs
```diff
@@ -2,8 +2,220 @@ mod operation;
...
+    "listCollectionObjects",
+    "listDatabaseObjects",
+    "listIndexNames",
+    "listIndexes",
+    "mapReduce",
+    "watch",
+];
+
+pub async fn run_v2_test(test_file: TestFile) {
+    let has_skipped_op = test_file.tests.iter().any(|test_case| {
```
patrickfreed, line 35: this can be done on a test case level, right? I dont think we want to skip an entire test file because one of the tests had an invalid operation.

src/test/spec/runner/mod.rs
```diff
@@ -2,8 +2,220 @@ mod operation;
...
+                                .await
+                        } else {
+                            client.init_db_and_coll(&db_name, &coll_name).await
+                        };
+                        coll.insert_many(data.clone(), None)
+                            .await
+                            .expect(&test_case.description);
+                    }
+                }
+                TestData::Many(_) => panic!("{}: invalid data format", &test_case.description),
```
patrickfreed, line 106: why is this unimplemented?

src/test/util/matchable.rs
patrickfreed, line 62: According to the [crud spec](https://github.com/mongodb/specifications/tree/master/source/crud/tests#optional-fields-in-expected-result-objects), if we don't include an upsertedCount filed in our result, we can just skip verifying it in the test runner.

src/client/options/mod.rs
isabelatkinson: I think this would work, but it would still require hard-coding the fields to merge since the third param to the macro is a list of options to check for. I could just pass in a list of all the options present on `ClientOptions`, but that's a pretty long list and is not super maintainable if more options get added in the future. What are your thoughts?

isabelatkinson: I presume by the existence of `serde` that there's some way to iterate over the fields of a struct which is what I'd actually like to do here, but I'm not well-versed enough in macros to tell if that's feasible -- some googling made it seem like it would be tricky.

patrickfreed: I was imagining it would be a giant list actually. The code as written is basically a giant list itself, just with some extra stuff for checking if things are `None` or not. I agree that either one isn't really maintainable, but without a macro for enumerating all the fields, I'm not sure how else to implement this.

src/coll/options.rs
isabelatkinson: I was generally avoiding assuming what the casing should be for these fields since they're not actually present in any of the crud spec tests, but from grepping the spec repo it seems like this one is consistently maxTimeMS elsewhere -- I'll update.

src/test/spec/runner/mod.rs
isabelatkinson: This is the data format used for gridfs tests -- since we don't have gridfs implemented, all those tests should skip which means reaching this branch would be an error in the test runner.

patrickfreed: Ah makes sense. I remember implementing this when doing retryable reads for Swift, but I guess I never verified whether it was necessary or not (we skip gridfs there too).

src/client/options/mod.rs
saghm: To do this automatically with a macro, you'd likely have to write a procedural macro (which is more complex than macros like `resolve_options!` or `doc!`; they need to be written in a separate crate). I think that's probably more work than it's worth; since this is test code, I'm fine with hard-coding and then updating whenever we need to.

src/test/spec/runner/mod.rs
isabelatkinson: oh good catch -- this was actually skipping over an operation that needed to be implemented

src/client/options/mod.rs
isabelatkinson: agreed, I'll update

src/selection_criteria.rs
```diff
@@ -84,7 +86,7 @@ pub type Predicate = Arc<dyn Send + Sync + Fn(&ServerInfo) -> bool>;
 /// option and will be sent to the server as an integer number of seconds.
 ///
 /// See the [MongoDB docs](https://docs.mongodb.com/manual/core/read-preference) for more details.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, Deserialize, PartialEq)]
```
patrickfreed, line 89: This also won't work, but for more frustrating reasons. According to the [server selection specification](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#read-preference), a read preference is supposed to look like this:
```json
"readPreference": { "mode": "primaryPreferred", "tag_sets": [] }
```
In the [spec tests](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/transaction-options.json#L1102) however, sometimes what's actually provided is just a string that only contains the mode, and [in other cases](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/read-pref.json#L114), the correct format is supplied. Ideally, the specs would respect the correct shape or at least be consistent in the representation they do use, but unfortunately neither is the case. To resolve this issue, I think you'll need to implement a custom `Deserialize` implementation for `ReadPreference`, one that first attempts to deserialize a string and, failing that, attempts to deserialize a struct with the `mode`, `tag_sets`, and `maxStalenessSeconds` fields per the spec.

Given that all of this is only necessary for the transactions runner as far as I can tell, I think leaving a TODO with a follow up ticket would be sufficient. What do you think @saghm ?

src/test/coll.rs
```diff
@@ -720,3 +721,77 @@ async fn find_one_and_delete_hint_server_version() {
         assert!(res.is_ok());
     }
 }
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
```
patrickfreed, line 725: this change looks like it's from @utkmehta 's changes, I think the rebase might not have happened properly if it's being included in the changes here.

src/selection_criteria.rs
saghm: Yeah, I'm fine with leaving this until later

isabelatkinson: https://jira.mongodb.org/browse/RUST-495

src/test/coll.rs
isabelatkinson: I actually wrote this test, it's included as a prose test in the crud spec https://github.com/mongodb/specifications/tree/master/source/crud/tests#errinfo-is-propagated

patrickfreed: Oh okay, well then I think we can just remove this one then. It's almost identical to the [one already implemented](https://github.com/mongodb/mongo-rust-driver/blob/master/src/test/coll.rs#L31).

PR #209, opened 2020-06-30
RUST-481 Add Windows to CI

saghm:


saghm:
There are some timeouts for replica sets with auth/TLS enabled with async-std, but I think these are general flakiness like we've seen before with other variants (as replica sets with TLS are the most constrained clusters), so I don't think we need to block merging this on figuring those out

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -37,7 +40,12 @@ async fn run() {
             return;
         }
 
-        let result = ClientOptions::parse(&test_file.uri).await;
+        let result = if cfg!(target_os = "windows") {
+            ClientOptions::parse_with_resolver_config(&test_file.uri, ResolverConfig::cloudflare())
```
patrickfreed, line 44: I thought these tests relied on the DNS records set on the system, meaning using the cloudflare config should cause it to fail. They seem to be passing though, which makes me concerned for my implementation of the config.

.evergreen/config.yml
```diff
@@ -200,6 +200,9 @@ functions:
             mv $i.new $i
           done
 
+          # Copy client certificate because symlinks do not work on Windows.
+          cp ${DRIVERS_TOOLS}/.evergreen/x509gen/client.pem ${MONGO_ORCHESTRATION_HOME}/lib/client.pem
```
patrickfreed, line 204: do other drivers not need to do this? why is this specific to us?

src/test/spec/initial_dns_seedlist_discovery.rs
saghm: I'm pretty sure that putting something in /etc/hosts (or whatever the equivalent is on Windows) will still resolve even if you don't use the system resolver. Either way though, these records do exist; see [an SRV lookup of one of the domains on some random DNS lookup site I found when googling](https://dnschecker.org/all-dns-records-of-domain.php?query=_mongodb._tcp.test5.test.build.10gen.cc&rtype=SRV). You might be thinking of the `ldaptest.10gen.cc` records, which I think are not published.

patrickfreed: Oh wow, okay, so the SRV records do resolve. I wonder if the "localhost.test.build.10gen.cc" will too, though. I guess we'll find out in RUST-489.

.evergreen/config.yml
saghm: Drivers that test on windows do; I coped from Python [here](https://github.com/mongodb/mongo-python-driver/blob/master/.evergreen/config.yml#L642-L643).

One thing I found out when looking for this after hitting an error is that surprisingly few drivers actually test on Windows.

saghm:
I think it's worth mentioning that we test on Linux, Mac, and Windows in CI. I don't think there's any reason to say we support/don't support something; the driver probably works on FreeBSD too, but we just don't test it so we have no way of knowing if something breaks.

saghm:
Oops, I just updated my Github email like 15 minutes ago; I'll commit amend with the new email now.

patrickfreed:
>I think it's worth mentioning that we test on Linux, Mac, and Windows in CI.

Sounds good, could you update the CI section to say this?

saghm:
I also just grudgingly added my old email back to my profile as a secondary address just to avoid having my entire git history look like it wasn't done by me.

> Sounds good, could you update the CI section to say this?

Will do

saghm:
@patrickfreed I've updated the README

saghm:
They're also timeouts, just internal to the tests rather than evergreen timeouts. See the code [here](https://github.com/mongodb/mongo-rust-driver/blob/66e06d5dc2bdb7b627009169afb51df85b5a5034/src/cmap/test/mod.rs#L134-L143) leading up to the assertion failures. Basically, we need to assert that the `Drop` implementation on `ConnectionPool` has run, but since we clean up the pool in a spawned async task, we have no guarantee about when that task will complete. The tests wait a certain duration, but it's never a guarantee.

PR #211, opened 2020-06-30
RUST-370 Ensuring that the WriteConcernError errInfo object is propagated

utkmehta:
This PR introduces a new field errInfo in WriteConcernError and tests verifies that the provenance field is propagated from the server to the client

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,7 +7,7 @@ use lazy_static::lazy_static;
 use serde::Deserialize;
 use time::OutOfRangeError;
 
-use crate::options::StreamAddress;
+use crate::{bson::Document, options::StreamAddress};
 
 lazy_static! {
     static ref RECOVERING_CODES: Vec<i32> = vec![11600, 11602, 13436, 189, 91];
@@ -381,6 +381,10 @@ pub struct WriteConcernError {
     /// A description of the error that occurred.
     #[serde(rename = "errmsg")]
     pub message: String,
+
+    /// A document identifying the write concern setting related to the error.
+    #[serde(rename = "errInfo")]
+    pub details: Option<Document>,
 }
 
 /// An error that occurred during a write operation that wasn't due to being unable to satisfy a
--- a/src/operation/delete/test.rs
+++ b/src/operation/delete/test.rs
@@ -167,7 +167,14 @@ async fn handle_write_concern_failure() {
         "writeConcernError": {
             "code": 456,
             "codeName": "wcError",
-            "errmsg": "some message"
+            "errmsg": "some message",
+            "errInfo": {
+                "writeConcern": {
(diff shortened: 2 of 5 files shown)


src/test/coll.rs
```diff
@@ -25,6 +25,65 @@ use crate::{
...
+                        "provenance": "clientSupplied"
+                    }
+                    }
+                }
+                },
+                "mode": { "times": 1 }
+            },
+            None,
+        )
+        .await;
```
saghm, line 65: The clippy warning in CI is referring to this line; we're running a command and returning a Result, but we're not using it. Since this command need to be successfully for the test to work, you should `unwrap` here.

src/operation/delete/test.rs
```diff
@@ -180,6 +180,7 @@ async fn handle_write_concern_failure() {
                 code: 456,
                 code_name: "wcError".to_string(),
                 message: "some message".to_string(),
+                err_info: None,
```
isabelatkinson, line 183: Instead of having `err_info` be `None` here, it would be useful to add an `errInfo` field to the `CommandResponse` document above and make sure it's properly included in the error -- ditto for the other tests below.

utkmehta: yep makes sense. fixed

utkmehta:
> From looking at CI, this test seems to fail on 3.6 replica sets. Can you try it out locally to verify this? My guess is that `err_info` might not get returned on 3.6. Assuming that's the case, you can just require 4.0+ for this test.

Yes I added a check and the tests work now

src/error.rs
```diff
@@ -381,6 +381,9 @@ pub struct WriteConcernError {
     /// A description of the error that occurred.
     #[serde(rename = "errmsg")]
     pub message: String,
+
+    #[serde(rename = "errInfo")]
+    pub err_info: Option<Document>,
```
patrickfreed, line 386: Oh I actually do have one comment. According to the [spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#error-handling), this field is meant to be named `details`.

utkmehta: fixed

src/error.rs
```diff
@@ -381,6 +381,9 @@ pub struct WriteConcernError {
     /// A description of the error that occurred.
     #[serde(rename = "errmsg")]
     pub message: String,
+
+    #[serde(rename = "errInfo")]
+    pub details: Option<Document>,
```
patrickfreed, line 386: Since this field is public, we should document what its contents might contain. 

From the CRUD spec:

>A document identifying the write concern setting related to the error.

utkmehta: not sure how I missed that haha. Thank you! Fixed.

PR #212, opened 2020-07-10
Designed the user facing ChangeStream API

utkmehta:
This PR is the layout of the API interface for changeStreams

### DIFF
--- a/src/change_stream/document.rs
+++ b/src/change_stream/document.rs
@@ -1,0 +1,97 @@
+use crate::coll::Namespace;
+use bson::{Bson, Document};
+use serde::{Deserialize, Serialize};
+
+/// An opaque token used for resuming an interrupted `ChangeStream`.
+///
+/// When starting a new change stream, `start_after` and `resume_after` fields on
+/// `ChangeStreamOptions` can be specified with instances of `ResumeToken`.
+///
+/// See the documentation
+/// [here](https://docs.mongodb.com/manual/changeStreams/#change-stream-resume-token) for more
+/// information on resume tokens.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct ResumeToken(pub(crate) Bson);
+
+/// A `ChangeStreamEventDocument` represents a [change event]
+/// (https://docs.mongodb.com/manual/reference/change-events/) in the associated change stream.
+/// Instances of `ChangeStreamEventDocument` are returned from calls to `ChangeStream::next`.
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ChangeStreamEventDocument {
+    /// An opaque token for use when resuming an interrupted `ChangeStream`.
+    ///
+    /// See the documentation
+    /// [here](https://docs.mongodb.com/manual/changeStreams/#change-stream-resume-token) for
+    /// more information on resume tokens.
+    ///
+    /// Also see the documentation on [resuming a change
+    /// stream](https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream).
+    #[serde(rename = "_id")]
+    pub id: ResumeToken,
+
+    /// Describes the type of operation represented in this change notification.
+    pub operation_type: OperationType,
+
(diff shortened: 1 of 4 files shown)


src/change_stream/document.rs
```diff
@@ -0,0 +1,95 @@
...
+/// `ChangeStreamOptions` can be specified with instances of `ResumeToken`.
+///
+/// See the documentation
+/// [here](https://docs.mongodb.com/manual/changeStreams/#change-stream-resume-token) for more
+/// information on resume tokens.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct ResumeToken(pub(crate) Bson);
+
+/// A `ChangeStreamEventDocument` represents a [change event]
+/// (https://docs.mongodb.com/manual/reference/change-events/) in the associated change stream.
```
saghm, line 15: I don't think a markdown link will work when broken across multiple comment lines like this

src/change_stream/document.rs
```diff
@@ -0,0 +1,95 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct ChangeStreamEventDocument {
+    /// An opaque token for use when resuming an interrupted `ChangeStream`.
+    ///
+    /// See the documentation
+    /// [here](https://docs.mongodb.com/manual/changeStreams/#change-stream-resume-token) for
+    /// more information on resume tokens.
+    ///
+    /// Also see the documentation on [resuming a change
+    /// stream](https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream).
```
saghm, line 27: Ditto above about splitting the link

src/change_stream/document.rs
```diff
@@ -0,0 +1,95 @@
...
+    #[serde(rename = "_id")]
+    pub id: ResumeToken,
+
+    /// Describes the type of operation represented in this change notification.
+    pub operation_type: OperationType,
+
+    /// Contains two fields: "db" and "coll" containing the database and collection name in which
+    /// the change happened.
+    pub ns: Option<Document>,
```
saghm, line 36: We can actually update this to use the `NameSpace` type that we define: https://docs.rs/mongodb/1.0.0/mongodb/struct.Namespace.html

src/change_stream/document.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    /// An array of field names that were removed from the `Document`.
+    pub removed_fields: Vec<String>,
+}
+
+/// The operation type represented in a given change notification.
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub enum OperationType {
+    /// See https://docs.mongodb.com/manual/reference/change-events/#insert-event.
```
saghm, line 72: We should make these markdown links so that they can be clicked in the rendered documentation

src/change_stream/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+    fn aggregate(
+        &self,
+        pipeline: Vec<Document>,
+        options: Option<AggregateOptions>,
+    ) -> Result<Cursor> {
+        todo!();
+    }
+}
+
+impl<'a, T> Stream for ChangeStreamTail<'a, T>
```
saghm, line 131: Since we don't use `Tail` in our API anymore, this `impl` block can be removed

src/change_stream/document.rs
utkmehta: I don't think they're actually split. locally they aren't. github probably just renders them in that way

src/change_stream/mod.rs
utkmehta: fixed

src/change_stream/document.rs
utkmehta: fixed

src/change_stream/document.rs
utkmehta: fixed

src/change_stream/mod.rs
utkmehta: fixed

PR #214, opened 2020-07-11
RUST-384 Close connection which was dropped during command execution

Lesiuk:


saghm:
Thanks for the PR! It looks like the rustfmt lint check failed; could you add the nightly version of rustfmt (`rustup component add rustfmt --toolchain nightly`) and run `rustfmt +nightly --unstable-features src/lib.rs`? You can verify locally that it worked by running the `.evergreen/check-rustfmt.sh` script from the root of the repo.

src/event/cmap.rs
```diff
@@ -117,6 +117,9 @@ pub enum ConnectionClosedReason {
     /// An error occurred while using the connection.
     Error,
 
+    /// The connection was dropped during read or write
+    Dropped,
```
saghm, line 121: I think that adding a new reason here is the right choice, but I'm curious about whether we think it's compliant with the CMAP spec. @patrickfreed @isabelatkinson @mbroadst any thoughts on this?

patrickfreed: The other options don't seem to line up with what's going on, but CMAP also doesn't have any guidance for adding new reasons. We _could_ piggy back off of error here if it were important to match the spec exactly, but I have a feeling adding a new case would actually be a better API.

mbroadst: a new reason seems fine to me

Lesiuk:
@saghm Reformatted and verified locally using check-rustfmt.sh script.

src/cmap/conn/mod.rs
```diff
@@ -61,6 +61,7 @@ pub(crate) struct Connection {
     /// currently checked into the pool, this will be None.
     pub(super) pool: Option<Weak<ConnectionPoolInner>>,
 
+    commmand_executing: bool,
```
patrickfreed, line 64: for future readers of the code, could we have a docstring indicating the purpose of this value / why it's necessary?

src/cmap/conn/mod.rs
```diff
@@ -206,11 +208,17 @@ impl Connection {
...
         request_id: impl Into<Option<i32>>,
     ) -> Result<CommandResponse> {
+        self.commmand_executing = true;
+
         let message = Message::with_command(command, request_id.into());
         message.write_to(&mut self.stream).await?;
 
         let response_message = Message::read_from(&mut self.stream).await?;
-        CommandResponse::new(self.address.clone(), response_message)
+        let command_response = CommandResponse::new(self.address.clone(), response_message);
```
patrickfreed, line 217: do we need to split it up like this? I think we should be fine just setting `command_executing` to false right after the reading is done and just keep returning the `CommandResponse` construction.

src/event/cmap.rs
```diff
@@ -117,6 +117,9 @@ pub enum ConnectionClosedReason {
     /// An error occurred while using the connection.
     Error,
 
+    /// The connection was dropped during read or write
```
patrickfreed, line 120: nitpick, but can we add a period to the end of this doctring?

src/cmap/conn/mod.rs
saghm: Added

src/cmap/conn/mod.rs
saghm: I've moved the assignments of `self.command_executing` to be immediately before the I/O starts and after it completes. This technically shouldn't matter due to there not being any other instances of `await` between the existing locations of the assignments, but I agree that it's easier to read this way.

src/event/cmap.rs
saghm: Added

src/cmap/conn/mod.rs
```diff
@@ -61,6 +61,11 @@ pub(crate) struct Connection {
     /// currently checked into the pool, this will be None.
     pub(super) pool: Option<Weak<ConnectionPoolInner>>,
 
+    /// Whether or not a command is currently being run on this connection. This is set to `true`
+    /// right before sending bytes to the server and set back to `false` once a full response has
+    /// been read.
+    commmand_executing: bool,
```
isabelatkinson, line 67: nit: there's an extra 'm' in command

saghm:
I just merged this; thanks again for identifying the bug and contributing the fix!

bcortier-devolutions:
Do you have an eta for next crate.io release?
Currently `1.1.1-beta` is published but it doesn't contain this fix. A `1.1.1-beta2` would be good too, but this fix is really critical. The driver can't be used to build an async web server in its current state since this bug is arising too frequently. Using `patch` in Cargo.toml is an option, but most users won't be necessarily aware of the fix.

extrawurst:
@bcortier-devolutions I agree it is very essential, also I found another one related to this that popped up when switching to master: https://github.com/mongodb/mongo-rust-driver/pull/234

bcortier-devolutions:
@extrawurst wow, thank you for the pointer on your PR. We're currently working with a pinned version.

PR #217, opened 2020-07-22
RUST-483 Allow alternate DNS resolver configs

patrickfreed:
RUST-483

This PR allows users to resolve `mongodb+srv` connection strings using alternate DNS resolver configurations. This helps users avoid the [known performance issue in trust-dns-resolver on Windows](https://github.com/bluejekyll/trust-dns#resolver) when using the system configuration (issue seen in RUST-467).

I opted to only include a helper on `ClientOptions` and not one on `Client` itself since this seemed like a very specific issue that was strictly configuration related.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ This repository contains the officially supported MongoDB Rust driver, a client 
         - [Inserting documents into a collection](#inserting-documents-into-a-collection)
         - [Finding documents in a collection](#finding-documents-in-a-collection)
     - [Using the sync API](#using-the-sync-api)
+- [Atlas note](#atlas-note)
+- [Windows DNS note](#windows-dns-note)
 - [Bug Reporting / Feature Requests](#bug-reporting--feature-requests)
 - [Contributing](#contributing)
 - [Running the tests](#running-the-tests)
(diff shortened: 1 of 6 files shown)


src/client/options/mod.rs
```diff
@@ -639,13 +650,38 @@ impl ClientOptions {
...
+    /// returned.
+    ///
+    /// In the case that "mongodb+srv" is used, SRV and TXT record lookups will be done using the
+    /// provided `ResolverConfig` as part of this method.
+    ///
+    /// The format of a MongoDB connection string is described [here](https://docs.mongodb.com/manual/reference/connection-string/#connection-string-formats).
+    ///
+    /// See the docstring on `ClientOptions::parse` for information on how the various URI options
+    /// map to fields on `ClientOptions`.
+    pub async fn parse_with_resolver_config(
```
saghm, line 667: Good call on this; I hadn't thought of the fact that there would need to be a separate parse method until I started reviewing, so I'm pleased to see you already handled it

README.md
```diff
@@ -182,6 +184,21 @@ for result in cursor {
 
 Currently, the driver has issues connecting to Atlas tiers above M2 unless the server version is at least 4.2. We're working on fixing this, but in the meantime, a workaround is to upgrade your cluster to 4.2. The driver has no known issues with either M0 or M2 instances.
 
+## Windows DNS note
+
+On Windows, there is a known issue in the `trust-dns-resolver` crate, which the driver uses to perform DNS lookups, that causes severe performance degradation in resolvers that use the system configuration. Since the driver uses the system configuration by default, users are recommended to specify an alternate resolver configuration on Windows until that issue is resolved. This is only has an effect when connecting to deployments using a `mongodb+srv` connection string.
```
isabelatkinson, line 189: there's an extraneous "is" at the beginning of the last sentence

patrickfreed: fixed

PR #218, opened 2020-07-22
RUST-418 Add test to verify that $readPreference isn't sent to standa…

saghm:
…lones

### DIFF
--- a/src/test/coll.rs
+++ b/src/test/coll.rs
@@ -12,9 +12,12 @@ use crate::{
         AggregateOptions,
         DeleteOptions,
         FindOneAndDeleteOptions,
+        FindOneOptions,
         FindOptions,
         Hint,
         InsertManyOptions,
+        ReadPreference,
+        SelectionCriteria,
         UpdateOptions,
     },
     results::DeleteResult,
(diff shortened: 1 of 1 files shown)


src/test/coll.rs
```diff
@@ -720,3 +723,34 @@ async fn find_one_and_delete_hint_server_version() {
         assert!(res.is_ok());
     }
 }
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[function_name::named]
+async fn no_read_preference_to_standalone() {
```
saghm, line 730: We already have the correct behavior required by the drivers ticket, so I implemented a test that verifies it so we don't regress

src/test/coll.rs
```diff
@@ -720,3 +723,34 @@ async fn find_one_and_delete_hint_server_version() {
...
+    client
+        .database(function_name!())
+        .collection(function_name!())
+        .find_one(None, options)
+        .await
+        .unwrap();
+
+    let command_started = client.get_successful_command_execution("find").0;
+
+    dbg!(&command_started);
```
patrickfreed, line 754: this debug print can be removed.

saghm: Oops, good catch

saghm:
Hmm, evergreen doesn't seem to be running for this patch now. I'm going to hold off on merging for a bit in case evergreen is just backed up or something.

saghm:
Evergreen is back up and CI passed, so I'll merge now

PR #219, opened 2020-07-22
RUST-51 Retryable Writes

isabelatkinson:
https://jira.mongodb.org/browse/RUST-51

spec: https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst
test spec: https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.rst

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -196,10 +196,7 @@ impl Client {
                     .get_error_with_retryable_write_label(&conn, &op, err)
                     .await?;
 
-                if err.is_server_error()
-                    || err.is_read_retryable()
-                    || err.is_write_retryable()
-                {
+                if err.is_server_error() || err.is_read_retryable() || err.is_write_retryable() {
                     Err(err)
                 } else {
                     Err(first_error)
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -131,6 +131,9 @@ impl Client {
                     }
                 }
 
+                // For a pre-4.4 connection, an error label should be added to any write-retryable
+                // error as long as the retry_writes client option is not set to false. For a 4.4+
+                // connection, a label should be added only to non-command errors.
                 let err = if self.inner.options.retry_writes != Some(false)
                     && op.retryability() == Retryability::Write
                 {
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -121,6 +121,8 @@ impl Client {
                     .handle_post_handshake_error(err.clone(), &conn, server)
                     .await;
 
+                // Retryable writes are only supported by storage engines with document-level
+                // locking, so users need to disable retryable writes if using mmapv1.
                 if let ErrorKind::CommandError(ref err) = err.kind.as_ref() {
                     if err.code == 20 && err.message.starts_with("Transaction numbers") {
                         let mut err = err.clone();
@@ -153,7 +155,7 @@ impl Client {
                     || (self.inner.options.retry_writes != Some(false)
                         && op.retryability() == Retryability::Write
                         && op.is_acknowledged()
-                        && err.labels().contains(&"RetryableWriteError".to_string())
+                        && err.contains_label("RetryableWriteError")
                         && conn.stream_description()?.supports_retryable_writes())
                 {
                     err
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -93,18 +93,7 @@ impl Client {
             }
         };
 
-        let retryability = match op.retryability() {
-            Retryability::Read if self.inner.options.retry_reads != Some(false) => {
-                Retryability::Read
-            }
-            Retryability::Write
-                if self.inner.options.retry_writes != Some(false)
-                    && conn.stream_description()?.supports_retryable_writes() =>
-            {
-                Retryability::Write
-            }
-            _ => Retryability::None,
-        };
+        let retryability = self.get_retryability(&conn, &op).await?;
 
         let txn_number = match session {
             Some(ref mut session) => {
@@ -149,9 +138,10 @@ impl Client {
                 // For a pre-4.4 connection, an error label should be added to any write-retryable
                 // error as long as the retry_writes client option is not set to false. For a 4.4+
                 // connection, a label should be added only to network errors.
-                let err = self
-                    .get_error_with_retryable_write_label(&conn, &op, err)
-                    .await?;
+                let err = match retryability {
+                    Retryability::Write => get_error_with_retryable_write_label(&conn, err).await?,
+                    _ => err,
+                };
 
                 // TODO RUST-90: Do not retry read if session is in a transaction
                 if retryability == Retryability::Read && err.is_read_retryable()
@@ -181,12 +171,11 @@ impl Client {
                 return Err(first_error);
             }
         };
-
-        if op.retryability() == Retryability::Write
-            && !conn.stream_description()?.supports_retryable_writes()
-        {
+        if !conn.stream_description()?.supports_retryable_writes() {
             return Err(first_error);
         }
+
+        let retryability = self.get_retryability(&conn, &op).await?;
 
         match self
             .execute_operation_on_connection(&op, &mut conn, &mut session, txn_number)
@@ -199,9 +188,10 @@ impl Client {
                     .handle_post_handshake_error(err.clone(), &conn, server)
                     .await;
 
-                let err = self
-                    .get_error_with_retryable_write_label(&conn, &op, err)
-                    .await?;
+                let err = match retryability {
+                    Retryability::Write => get_error_with_retryable_write_label(&conn, err).await?,
+                    _ => err,
+                };
 
                 if err.is_server_error() || err.is_read_retryable() || err.is_write_retryable() {
                     Err(err)
@@ -381,25 +371,36 @@ impl Client {
         }
     }
 
-    /// Returns an Error with a "RetryableWriteError" label added if necessary. On a pre-4.4
-    /// connection, a label should be added to any write-retryable error. On a 4.4+ connection, a
-    /// label should only be added to network errors. Regardless of server version, a label should
-    /// only be added if the `retry_writes` client option is not set to `false`.
-    async fn get_error_with_retryable_write_label<T: Operation>(
+    /// Returns the retryability level for the execution of this operation.
+    async fn get_retryability<T: Operation>(
         &self,
         conn: &Connection,
         op: &T,
-        err: Error,
-    ) -> Result<Error> {
-        if self.inner.options.retry_writes != Some(false)
-            && op.retryability() == Retryability::Write
-        {
-            if let Some(max_wire_version) = conn.stream_description()?.max_wire_version {
-                if err.should_add_retryable_write_label(max_wire_version) {
-                    return Ok(err.with_label("RetryableWriteError"));
-                }
+    ) -> Result<Retryability> {
+        match op.retryability() {
+            Retryability::Read if self.inner.options.retry_reads != Some(false) => {
+                Ok(Retryability::Read)
             }
+            Retryability::Write
+                if self.inner.options.retry_writes != Some(false)
+                    && conn.stream_description()?.supports_retryable_writes() =>
+            {
+                Ok(Retryability::Write)
+            }
+            _ => Ok(Retryability::None),
         }
-        Ok(err)
     }
+}
+
+/// Returns an Error with a "RetryableWriteError" label added if necessary. On a pre-4.4
+/// connection, a label should be added to any write-retryable error. On a 4.4+ connection, a
+/// label should only be added to network errors. Regardless of server version, a label should
+/// only be added if the `retry_writes` client option is not set to `false`.
+async fn get_error_with_retryable_write_label(conn: &Connection, err: Error) -> Result<Error> {
+    if let Some(max_wire_version) = conn.stream_description()?.max_wire_version {
+        if err.should_add_retryable_write_label(max_wire_version) {
+            return Ok(err.with_label("RetryableWriteError"));
+        }
+    }
+    Ok(err)
 }
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -8,7 +8,7 @@ use time::PreciseTime;
 use crate::{
     bson::Document,
     cmap::Connection,
-    error::{ErrorKind, Result},
+    error::{Error, ErrorKind, Result},
     event::command::{CommandFailedEvent, CommandStartedEvent, CommandSucceededEvent},
     operation::Operation,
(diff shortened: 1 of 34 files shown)


techbelle:
🎉  yay

src/test/spec/retryable_writes/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+        .run_command(doc! { "serverStatus": 1 }, None)
+        .await
+        .unwrap();
+    let name = server_status
+        .get_document("storageEngine")
+        .unwrap()
+        .get_str("name")
+        .unwrap();
+    if name != "mmapv1" {
+        return;
```
isabelatkinson, line 302: AFAIK this test is always returning here on Evergreen since wiredTiger is being used, but the test passes locally when `--storagEngine=mmapv1` is passed to mongod. Is there anything we should do to test this in CI?

src/test/spec/runner/operation.rs
```diff
@@ -261,8 +271,11 @@ impl TestOperation for InsertMany {
         let result = collection
             .insert_many(self.documents.clone(), self.options.clone())
             .await?;
-        let result = bson::to_bson(&result)?;
-        Ok(Some(result))
+        let mut ids = Document::new();
+        for (id, bson) in result.inserted_ids {
+            ids.insert(id.to_string(), bson);
+        }
```
isabelatkinson, line 277: This is where I ran into the issue of deserializing an unsigned integer -- creating the document manually as a workaround fixed the problem

src/test/spec/retryable_writes/mod.rs
saghm: We could add an mmapv1 task, but honestly it's probably not worth the effort (and evergreen cost). If a user ends up reporting a bug with mmapv1, we can revisit, but I don't expect that we have many mmapv1 users (if any at all).

src/test/spec/runner/operation.rs
saghm: Hmm, the ids being used are unsigned integers?

src/test/util/event.rs
```diff
@@ -143,6 +143,55 @@ impl EventClient {
         }
     }
 
+    pub async fn merge_options(
```
saghm, line 146: I have a moderate preference for calling this something like `with_additional_options`; `merge_options` doesn't quite sound like the name of a constructor to me, since `merge` sounds like verb that's just used to combine options, not create something else entirely

src/test/util/event.rs
```diff
@@ -143,6 +143,55 @@ impl EventClient {
...
+        use_multiple_mongoses: Option<bool>,
+    ) -> Self {
+        let mut options = match options {
+            Some(mut options) => {
+                options.hosts = CLIENT_OPTIONS.hosts.clone();
+                merge_options!(
+                    CLIENT_OPTIONS.clone(),
+                    &mut options,
+                    [
+                        app_name,
```
saghm, line 157: Is it worth defining an additional macro for cases like this where we want the entirety of options? If we ever need to do this somewhere else, it would be nice to only have to maintain the list in one place.

isabelatkinson: Is there another use case you can think of? Everywhere I've needed to do this has been in the context of merging the test options with options in spec tests which is taken care of by this method

src/error.rs
```diff
@@ -101,6 +123,76 @@ impl Error {
...
+        }
+    }
+
+    pub(crate) fn with_retryable_write_label(self, add_to_command_error: bool) -> Self {
+        if add_to_command_error && self.is_write_retryable() {
+            self.with_label("RetryableWriteError".to_string())
+        } else {
+            match self.kind.as_ref() {
+                ErrorKind::CommandError(_) => self,
+                _ => {
```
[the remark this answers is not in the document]
saghm: My reading of the spec is that the label shouldn't be added command errors in that case, but I don't think it means to add them in all other cases. I think the idea is that we need to add the label manually pre-4.4, but anything that comes from the server 4.4+ will already have it if it needs it, so we shouldn't add it ourselves.

src/test/spec/retryable_writes/mod.rs
```diff
@@ -314,3 +314,45 @@ async fn mmapv1_error_raised() {
...
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn label_not_added_second_read_error() {
+    let _guard = LOCK.run_exclusively().await;
+    assert!(label_not_added(true, 2).await);
+}
+
+#[function_name::named]
+async fn label_not_added(retry_reads: bool, times: i32) -> bool {
+    let options = ClientOptions::builder().retry_reads(retry_reads).build();
```
saghm, line 334: I think retryable writes should be enabled for these tests too? Otherwise, `self.inner.options.retry_writes != Some(false) && op.retryability() == Retryability::Write` will be false due to the first check, and we won't be testing whether labels are added to reads when retryable writes are enabled.

src/test/util/event.rs
saghm: Fair enough. We can hold off until/unless we run into another case where we need it.

src/test/spec/runner/operation.rs
saghm: Oh, I see; the ids are signed, but the keys int the map are `usize`, so they can't be serialized. Another workaround would be to map the keys, e.g. `result.inserted_ids.into_iter().map(|(k, v)|, (k.to_string(), v)).collect()`.

src/test/spec/retryable_writes/mod.rs
saghm: Ah, right, I forgot they're enabled by default

src/test/spec/retryable_writes/mod.rs
```diff
@@ -319,21 +320,41 @@ async fn mmapv1_error_raised() {
 #[cfg_attr(feature = "async-std-runtime", async_std::test)]
 async fn label_not_added_first_read_error() {
     let _guard = LOCK.run_exclusively().await;
-    assert!(label_not_added(false, 1).await);
+
+    let options = ClientOptions::builder().retry_reads(false).build();
+    let client = TestClient::with_additional_options(Some(options), false).await;
+
+    let req = VersionReq::parse(">=4.0").unwrap();
```
saghm, line 327: Could you add a comment here explaining why this check is needed? I'm guessing it's something to do with when the server started adding the label to errors, but I'm not sure.

Also, would it be cleaner to just put the skip check in the helper function rather than duplicating it in each of these tests?

src/test/spec/runner/operation.rs
```diff
@@ -271,10 +271,12 @@ impl TestOperation for InsertMany {
         let result = collection
             .insert_many(self.documents.clone(), self.options.clone())
             .await?;
-        let mut ids = Document::new();
-        for (id, bson) in result.inserted_ids {
-            ids.insert(id.to_string(), bson);
-        }
+        let ids: HashMap<String, Bson> = result
```
saghm, line 274: If you don't end up using `result` again, you can use `into_iter()` instead of `iter()` and avoid needing to clone `v`

src/error.rs
```diff
@@ -91,6 +102,17 @@ impl Error {
         }
     }
 
+    // Whether a write operation should be retried if this error occurs
+    pub(crate) fn is_write_retryable(&self) -> bool {
```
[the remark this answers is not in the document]
patrickfreed: Or--rather than making this private, tweak it to just check for the error label and introduce a new `should_add_retryable_write_label` or something that is private with this methods current body. The situation I'm trying to avoid with this is where we have different methods for determining if an error is retryable (e.g. in executor, we both check for the label and invoke this method in different places).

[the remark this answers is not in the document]
isabelatkinson: Makes sense. I updated this to be `should_add_retryable_write_label` and added a different `is_write_retryable` to check for the label presence.

src/error.rs
```diff
@@ -101,6 +123,80 @@ impl Error {
...
+            _ => {
+                self.labels.push(label);
+                self
+            }
+        }
+    }
+
+    /// Returns a new Error with a "RetryableWriteError" label added if the Error is write
+    /// retryable. Only adds the label to command errors if add_to_command_error is true.
+    pub(crate) fn with_retryable_write_label(self, add_to_command_error: bool) -> Self {
```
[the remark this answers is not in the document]
isabelatkinson: This method is removed in the new refactor.

src/test/spec/retryable_writes/mod.rs
```diff
@@ -0,0 +1,374 @@
...
+async fn run_spec_tests() {
+    async fn run_test(test_file: TestFile) {
+        for test_case in test_file.tests {
+            if test_case.operation.name == "bulkWrite" {
+                continue;
+            }
+
+            if test_case.description
+                == "InsertOne fails after multiple retryable writeConcernErrors"
+            {
```
isabelatkinson, line 27: [This](https://github.com/mongodb/specifications/blob/192753e0ee9e7dd8ea30c10d055e119c7db1b99d/source/retryable-writes/tests/insertOne-serverErrors.json#L957) is the test that is causing issues with the error labels. The configured failpoint does not contain a label; however, the test results expect a RetryableWriteError label. This originally caused me to believe that the driver should be adding a RetryableWriteError label to write concern errors from 4.4+ servers, but Jeremy & Emily indicated that the server itself should be adding the label to the error from the configured failpoint on 4.4+ (more info [here](https://github.com/mongodb/mongo/wiki/The-%22failCommand%22-fail-point) under the "errorLabels" bullet point). However, it appears some servers are not doing this (verified in the shell and by printing out the error received from the server in the driver before driver label addition). We are currently investigating why this issue is occurring, but in the meantime I've skipped the test.

src/client/executor.rs
```diff
@@ -103,16 +118,47 @@ impl Client {
...
+                    err.with_retryable_write_label(add_to_command_error)
                 } else {
                     err
+                };
+
+                // TODO RUST-90: Do not retry read if session is in a transaction
+                if (self.inner.options.retry_reads != Some(false)
+                    && op.retryability() == Retryability::Read
+                    && err.is_read_retryable())
+                    || (self.inner.options.retry_writes != Some(false)
```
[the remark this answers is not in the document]
patrickfreed: Alternatively, we could just perform this check right after we select a server to determine what sort of retryability we're dealing with here. You can skip checking for acknowledged I think because we already error for that earlier in the operation execution lifecycle.
e.g.
```rust
let retryability = match op.retryability() {
    Retryability::Read if self.inner.optins.retry_reads != Some(false) => Retryablity::Read,
    Retryability::Write if self.inner.options.retry_writes != Some(false) && conn.stream_description()?.supports_retryable_writes() => Retryabilty::Write,
    _ => Retryability::None
}
```
Then here, the condition would look like:
```rust
if retryability == Retryability::Read && err.is_read_retryable() 
    || retryability == Retryability::Write && err.is_write_retryable() 
{ 
    ... 
}
```
Which I think is a little more readable. This could also be repeated below to guard the invocation of `get_error_with_retryable_write_label`.

src/error.rs
```diff
@@ -101,6 +133,66 @@ impl Error {
...
+            ErrorKind::BulkWriteError(err) => match err.write_concern_error {
+                Some(ref err) => &err.labels,
+                None => &self.labels,
+            },
+            _ => &self.labels,
+        }
+    }
+
+    pub fn contains_label<T: AsRef<str>>(&self, label: T) -> bool {
+        self.labels().contains(&label.as_ref().to_string())
```
patrickfreed, line 154: instead of allocating the string, we could just map over references here.

src/error.rs
```diff
@@ -101,6 +133,66 @@ impl Error {
...
+            },
+            ErrorKind::BulkWriteError(err) => match err.write_concern_error {
+                Some(ref err) => &err.labels,
+                None => &self.labels,
+            },
+            _ => &self.labels,
+        }
+    }
+
+    pub fn contains_label<T: AsRef<str>>(&self, label: T) -> bool {
```
patrickfreed, line 153: needs a docstring

src/client/executor.rs
```diff
@@ -103,16 +118,47 @@ impl Client {
...
+                // connection, a label should be added only to non-command errors.
+                let err = if self.inner.options.retry_writes != Some(false)
+                    && op.retryability() == Retryability::Write
                 {
-                    return Err(err);
+                    let add_to_command_error = conn
+                        .stream_description()?
+                        .max_wire_version
+                        .map_or(false, |version| version <= 8);
+                    err.with_retryable_write_label(add_to_command_error)
```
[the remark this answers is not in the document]
patrickfreed: Since we'll be cloning anyways, we can use the "clone on write" behavior of [`make_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut).

src/client/executor.rs
```diff
@@ -93,8 +93,25 @@ impl Client {
             }
         };
 
+        let retryability = self.get_retryability(&conn, &op).await?;
+
+        let txn_number = match session {
+            Some(ref mut session) => {
+                if self.inner.options.retry_writes != Some(false)
```
patrickfreed, line 100: we can replace this with a check on `retryability`

src/test/spec/retryable_writes/mod.rs
```diff
@@ -0,0 +1,374 @@
...
+                let can_run_on = run_on.iter().any(|run_on| run_on.can_run_on(&client));
+                if !can_run_on {
+                    println!("Skipping {}", test_case.description);
+                    continue;
+                }
+            }
+
+            let mut db_name = test_case.description.replace('$', "%").replace(' ', "_");
+            // database names must have fewer than 64 characters
+            db_name.truncate(63);
```
patrickfreed, line 47: could we abstract this replace and truncate logic to a helper function? We use it in a multiple places in the tests I think.

src/test/spec/retryable_writes/mod.rs
```diff
@@ -0,0 +1,374 @@
...
+
+    let client = EventClient::new().await;
+
+    if !(client.is_replica_set() || client.is_sharded()) {
+        return;
+    }
+
+    let coll = client.init_db_and_coll(function_name!(), "coll").await;
+
+    let excludes = |command_name: &str| -> bool {
```
patrickfreed, line 183: can we name this to `excludesTxnNumber`? Also, why is the txnNumber excluded from these operations?

src/client/executor.rs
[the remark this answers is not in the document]
patrickfreed: It's probably more straightforward to not rely on that actually, so we can just leave this as-is.

PR #221, opened 2020-07-27
RUST-520 Implement 'watch' operation

utkmehta:
This PR implements the watch operation and the watch functions for client/coll/database

### DIFF
--- a/src/change_stream/mod.rs
+++ b/src/change_stream/mod.rs
@@ -13,13 +13,16 @@ use crate::{
         document::{ChangeStreamEventDocument, ResumeToken},
         options::ChangeStreamOptions,
     },
+    client::ClientSession,
+    cursor::CursorSpecification,
     error::Result,
     options::AggregateOptions,
-    selection_criteria::ReadPreference,
+    selection_criteria::SelectionCriteria,
     Client,
     Collection,
     Cursor,
     Database,
+    Namespace,
 };
 
 /// A `ChangeStream` streams the ongoing changes of its associated collection, database or
@@ -72,6 +75,7 @@ use crate::{
 ///
 /// See the documentation [here](https://docs.mongodb.com/manual/changeStreams) for more
 /// details. Also see the documentation on [usage recommendations](https://docs.mongodb.com/manual/administration/change-streams-production-recommendations/).
+#[derive(Debug)]
 pub struct ChangeStream {
     /// The cursor to iterate over `ChangeStreamEventDocument` instances.
     cursor: Cursor,
@@ -93,9 +97,9 @@ pub struct ChangeStream {
     /// The options provided to the initial `$changeStream` stage.
     options: Option<ChangeStreamOptions>,
 
-    /// The read preference for the initial `$changeStream` aggregation, used for server selection
-    /// during an automatic resume.
-    read_preference: Option<ReadPreference>,
+    /// The SelectionCriteria for the initial `$changeStream` aggregation, used for server
+    /// selection during an automatic resume.
+    selection_criteria: Option<SelectionCriteria>,
 
     /// Whether or not the change stream has attempted a resume, used to attempt a resume only
     /// once.
@@ -107,6 +111,25 @@ pub struct ChangeStream {
 }
 
 impl ChangeStream {
+    pub(crate) fn new(
+        spec: ChangeStreamSpecification,
+        session: Option<ClientSession>,
+        client: Client,
+    ) -> Self {
+        let cursor = Cursor::new(client.clone(), spec.cursor_spec, session);
+        Self {
+            cursor,
+            pipeline: spec.pipeline,
+            client,
+            target: spec.target,
+            resume_token: None,
+            options: spec.options.clone(),
+            selection_criteria: spec.options.and_then(|opts| opts.selection_criteria),
+            resume_attempted: false,
+            document_returned: false,
+        }
+    }
+
     /// Returns the cached resume token that can be used to resume after the most recently returned
     /// change.
     ///
@@ -116,13 +139,18 @@ impl ChangeStream {
     pub fn resume_token(&self) -> Option<ResumeToken> {
         todo!();
     }
+
+    #[cfg(test)]
+    pub fn options(&self) -> Option<ChangeStreamOptions> {
+        self.options.clone()
+    }
 }
 
 #[derive(Debug, Clone)]
 pub(crate) enum ChangeStreamTarget {
-    Collection(Collection),
-    Database(Database),
-    Cluster(Database),
+    Collection(Namespace),
+    Database(String),
+    Cluster(String),
 }
 
 impl Stream for ChangeStream {
@@ -130,5 +158,38 @@ impl Stream for ChangeStream {
 
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
         todo!();
+    }
+}
+
+// Specification used to create a new `ChangeStream`.
+#[derive(Debug)]
+pub(crate) struct ChangeStreamSpecification {
+    /// Specification used to create a cursor wrapped by the `ChangeStream`.
+    pub(crate) cursor_spec: CursorSpecification,
+
+    /// The pipeline of stages to append to an initial `$changeStream` stage.
+    pub(crate) pipeline: Vec<Document>,
+
+    /// The options provided to the initial `$changeStream` stage.
+    pub(crate) options: Option<ChangeStreamOptions>,
+
+    /// The original target of the change stream, used for re-issuing the aggregation during
+    /// an automatic resume.
+    pub(crate) target: ChangeStreamTarget,
+}
+
+impl ChangeStreamSpecification {
+    pub(crate) fn new(
+        cursor_spec: CursorSpecification,
+        pipeline: Vec<Document>,
+        options: Option<ChangeStreamOptions>,
+        target: ChangeStreamTarget,
+    ) -> Self {
+        Self {
+            cursor_spec,
+            pipeline,
+            options,
+            target,
+        }
     }
 }
--- a/src/change_stream/options.rs
+++ b/src/change_stream/options.rs
@@ -6,7 +6,9 @@ use typed_builder::TypedBuilder;
 use crate::{
     change_stream::document::ResumeToken,
     collation::Collation,
+    concern::ReadConcern,
     options::AggregateOptions,
+    selection_criteria::SelectionCriteria,
 };
 
 #[derive(Clone, Debug, Default, Deserialize, Serialize, TypedBuilder)]
@@ -66,10 +68,35 @@ pub struct ChangeStreamOptions {
     #[builder(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub start_after: Option<ResumeToken>,
+
+    /// The criteria used to select a server for this operation.
+    ///
+    /// If none is specified, the selection criteria defined on the object executing this operation
+    /// will be used.
+    #[builder(default)]
+    #[serde(skip)]
+    #[serde(rename = "readPreference")]
+    pub selection_criteria: Option<SelectionCriteria>,
+
+    /// The `ReadConcern` to be used by the `ChangeStream`.
+    ///
+    /// `ChangeStreams` support only “majority” `ReadConcern` or no `ReadConcern`.
+    /// If a non-"majority" `ReadConcern` is specified on or inherited by the `Client`, `Database`,
+    /// or `Collection` that will be used to create the `ChangeStream`, then a "majority"
+    /// `ReadConcern` MUST be specified here.
+    ///
+    /// For MongoDB 4.2+, `ChangeStreams` are supported even if "majority" read concerns are
+    /// disabled server side.
+    /// See the official MongoDB documentation for more information on when ChangeStreams are
+    /// available: [here]("https://docs.mongodb.com/manual/changeStreams/#availability")
+    #[builder(default)]
+    #[serde(skip)]
+    pub read_concern: Option<ReadConcern>,
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[non_exhaustive]
+#[serde(rename_all = "camelCase")]
 /// Describes the modes for configuring the `full_document` field of a
 /// [`ChangeStreamEventDocument`](../document/struct.ChangeStreamEventDocument.html)
 pub enum FullDocumentType {
@@ -80,4 +107,15 @@ pub enum FullDocumentType {
 
     /// User-defined other types for forward compatibility.
     Other(String),
+}
+
+impl ChangeStreamOptions {
+    pub(crate) fn get_aggregation_options(&self) -> AggregateOptions {
+        AggregateOptions::builder()
+            .batch_size(self.batch_size.map(|batch_size| batch_size as u32))
+            .collation(self.collation.as_ref().map(Clone::clone))
+            .max_await_time(self.max_await_time)
+            .read_concern(self.read_concern.clone())
+            .build()
+    }
 }
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -12,12 +12,12 @@ use time::PreciseTime;
 use crate::options::StreamAddress;
 use crate::{
     bson::{Bson, Document},
-    change_stream::{options::ChangeStreamOptions, ChangeStream},
+    change_stream::{options::ChangeStreamOptions, ChangeStream, ChangeStreamTarget},
     concern::{ReadConcern, WriteConcern},
     db::Database,
     error::{ErrorKind, Result},
     event::command::CommandEventHandler,
-    operation::ListDatabases,
+    operation::{ListDatabases, Watch},
     options::{
         ClientOptions,
         DatabaseOptions,
@@ -288,10 +288,24 @@ impl Client {
         }
     }
 
+    /// A helper method to start a `ChangeStream`. Used by watch methods in client, coll, and db.
+    pub(crate) async fn start_change_stream(
+        &self,
+        pipeline: impl IntoIterator<Item = Document>,
+        options: Option<ChangeStreamOptions>,
+        target: ChangeStreamTarget,
+    ) -> Result<ChangeStream> {
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        let watch = Watch::new(target, pipeline.clone(), options.clone())?;
+        let (change_stream_spec, session) = self.execute_cursor_operation(watch).await?;
+        Ok(ChangeStream::new(change_stream_spec, session, self.clone()))
+    }
+
     /// Starts a new [`ChangeStream`](change_stream/struct.ChangeStream.html) that receives events
     /// for all changes in the cluster. The stream does not observe changes from system
-    /// collections or the "config", "local" or "admin" databases. Note that this method
-    /// (`watch` on a cluster) is only supported in MongoDB 4.0 or greater.
+    /// collections or the "config", "local" or "admin" databases.
+    ///
+    /// Note that this method (`watch` on a cluster) is only supported in MongoDB 4.0 or greater.
     ///
     /// See the documentation [here](https://docs.mongodb.com/manual/changeStreams/) on change
     /// streams.
@@ -307,8 +321,15 @@ impl Client {
     pub async fn watch(
         &self,
         pipeline: impl IntoIterator<Item = Document>,
-        options: Option<ChangeStreamOptions>,
+        options: impl Into<Option<ChangeStreamOptions>>,
     ) -> Result<ChangeStream> {
-        todo!();
+        let mut options = options.into();
+        resolve_options!(self, options, [read_concern, selection_criteria]);
+        self.start_change_stream(
+            pipeline,
+            options,
+            ChangeStreamTarget::Cluster("admin".to_string()),
+        )
+        .await
     }
 }
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -10,7 +10,7 @@ use self::options::*;
 use crate::{
     bson::{doc, Bson, Document},
     bson_util,
-    change_stream::{options::ChangeStreamOptions, ChangeStream},
+    change_stream::{options::ChangeStreamOptions, ChangeStream, ChangeStreamTarget},
     concern::{ReadConcern, WriteConcern},
     error::{convert_bulk_errors, BulkWriteError, BulkWriteFailure, ErrorKind, Result},
     operation::{
@@ -599,9 +599,18 @@ impl Collection {
     pub async fn watch(
         &self,
         pipeline: impl IntoIterator<Item = Document>,
-        options: Option<ChangeStreamOptions>,
+        options: impl Into<Option<ChangeStreamOptions>>,
     ) -> Result<ChangeStream> {
-        todo!();
+        let mut options = options.into();
+        resolve_options!(self, options, [read_concern, selection_criteria]);
+        let client = self.client();
+        client
+            .start_change_stream(
+                pipeline,
+                options,
+                ChangeStreamTarget::Collection(self.namespace()),
+            )
+            .await
     }
 }
 
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -6,7 +6,7 @@ use futures::stream::TryStreamExt;
 
 use crate::{
     bson::{Bson, Document},
-    change_stream::{options::ChangeStreamOptions, ChangeStream},
+    change_stream::{options::ChangeStreamOptions, ChangeStream, ChangeStreamTarget},
     concern::{ReadConcern, WriteConcern},
     cursor::Cursor,
     error::{ErrorKind, Result},
@@ -266,6 +266,8 @@ impl Database {
     /// for all changes in this database. The stream does not observe changes from system
     /// collections and cannot be started on "config", "local" or "admin" databases.
     ///
+    /// Note that this method (`watch` on a database) is only supported in MongoDB 4.0 or greater.
+    ///
     /// See the documentation [here](https://docs.mongodb.com/manual/changeStreams/) on change
     /// streams.
     ///
@@ -280,8 +282,17 @@ impl Database {
     pub async fn watch(
         &self,
         pipeline: impl IntoIterator<Item = Document>,
-        options: Option<ChangeStreamOptions>,
+        options: impl Into<Option<ChangeStreamOptions>>,
     ) -> Result<ChangeStream> {
-        todo!();
+        let mut options = options.into();
+        resolve_options!(self, options, [read_concern, selection_criteria]);
+        let client = self.client();
+        client
+            .start_change_stream(
+                pipeline,
+                options,
+                ChangeStreamTarget::Database(self.name().to_string()),
+            )
+            .await
     }
 }
--- a/src/operation/watch/mod.rs
+++ b/src/operation/watch/mod.rs
@@ -1,0 +1,78 @@
+#[cfg(test)]
+mod test;
+
+use super::aggregate::{Aggregate, AggregateTarget};
+
+use crate::{
+    bson::{doc, Document},
+    change_stream::{ChangeStreamSpecification, ChangeStreamTarget},
+    cmap::{Command, CommandResponse, StreamDescription},
+    error::Result,
+    operation::{append_options, Operation},
+    options::ChangeStreamOptions,
+};
+
+#[derive(Debug)]
+pub(crate) struct Watch {
+    target: ChangeStreamTarget,
+    pipeline: Vec<Document>,
+    options: Option<ChangeStreamOptions>,
+    aggregate_operation: Aggregate,
+}
+
+impl Watch {
+    pub(crate) fn new(
+        target: impl Into<ChangeStreamTarget>,
+        pipeline: impl IntoIterator<Item = Document>,
+        options: Option<ChangeStreamOptions>,
+    ) -> Result<Self> {
+        let target: ChangeStreamTarget = target.into();
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+
+        let mut options_doc: Document = Document::new();
+        append_options(&mut options_doc, options.as_ref())?;
+        let aggregate_target = match &target {
+            ChangeStreamTarget::Collection(namespace) => {
+                AggregateTarget::Collection(namespace.clone())
+            }
+            ChangeStreamTarget::Database(database) => AggregateTarget::Database(database.clone()),
+            ChangeStreamTarget::Cluster(database) => {
+                options_doc.insert("allChangesForCluster", true);
+                AggregateTarget::Database(database.clone())
+            }
+        };
+        let mut aggregate_pipeline = vec![doc! { "$changeStream": options_doc }];
+        aggregate_pipeline.extend(pipeline.clone());
+
+        let aggregate_operation = Aggregate::new(
+            aggregate_target,
+            aggregate_pipeline,
+            options.as_ref().map(|opts| opts.get_aggregation_options()),
+        );
+        Ok(Self {
+            target,
+            pipeline: pipeline.into_iter().collect(),
+            options,
+            aggregate_operation,
+        })
+    }
+}
+
+impl Operation for Watch {
+    type O = ChangeStreamSpecification;
+    const NAME: &'static str = "watch";
+
+    fn build(&self, description: &StreamDescription) -> Result<Command> {
+        self.aggregate_operation.build(description)
+    }
+
+    fn handle_response(&self, response: CommandResponse) -> Result<Self::O> {
+        let cursor_spec = self.aggregate_operation.handle_response(response)?;
+        Ok(ChangeStreamSpecification::new(
+            cursor_spec,
+            self.pipeline.clone(),
+            self.options.clone(),
+            self.target.clone(),
+        ))
+    }
+}
--- a/src/operation/watch/test.rs
+++ b/src/operation/watch/test.rs
@@ -1,0 +1,239 @@
+use std::time::Duration;
+
+use crate::{
+    bson::{doc, Document},
+    bson_util,
+    change_stream::ChangeStreamTarget,
+    cmap::{CommandResponse, StreamDescription},
+    concern::{ReadConcern, ReadConcernLevel},
+    operation::{Operation, Watch},
+    options::{
+        ChangeStreamOptions,
+        FullDocumentType,
+        ReadPreference,
+        SelectionCriteria,
+        StreamAddress,
+    },
+    Namespace,
+};
+
+fn build_test(
+    target: impl Into<ChangeStreamTarget>,
+    pipeline: Vec<Document>,
+    options: Option<ChangeStreamOptions>,
+    mut expected_body: Document,
+) {
+    let watch = Watch::new(target.into(), pipeline, options).unwrap();
+
+    let mut cmd = watch.build(&StreamDescription::new_testing()).unwrap();
+
+    assert_eq!(cmd.name.as_str(), "aggregate");
+
+    bson_util::sort_document(&mut expected_body);
+    bson_util::sort_document(&mut cmd.body);
+
+    assert_eq!(cmd.body, expected_body);
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn build() {
+    let pipeline = vec![doc! { "$match": { "x": 3 } }];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .read_concern(ReadConcern::from(ReadConcernLevel::Available))
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+
+    let expected_body = doc! {
+        "aggregate": 1,
+        "cursor": { "batchSize": 10 },
+        "pipeline": [
+            { "$changeStream": { "fullDocument": "updateLookup" } },
+            { "$match": { "x": 3 } },
+        ],
+        "readConcern": { "level": "available" },
+    };
+
+    build_test(
+        ChangeStreamTarget::Database("test_db".to_string()),
+        pipeline,
+        Some(options),
+        expected_body,
+    );
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn build_collection() {
+    let ns = Namespace {
+        db: "test_db".to_string(),
+        coll: "test_coll".to_string(),
+    };
+
+    let pipeline = vec![doc! { "$match": { "x": 3 } }];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .read_concern(ReadConcern::from(ReadConcernLevel::Available))
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+
+    let expected_body = doc! {
+        "aggregate": "test_coll",
+        "cursor": {},
+        "pipeline": [
+            { "$changeStream": { "fullDocument": "updateLookup" } },
+            { "$match": { "x": 3 } }
+        ],
+        "readConcern": { "level": "available" },
+    };
+
+    build_test(
+        ChangeStreamTarget::Collection(ns),
+        pipeline,
+        Some(options),
+        expected_body,
+    );
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn build_cluster() {
+    let pipeline = vec![doc! { "$match": { "x": 3 } }];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .read_concern(ReadConcern::from(ReadConcernLevel::Available))
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+
+    let expected_body = doc! {
+        "aggregate": 1,
+        "cursor": {},
+        "pipeline": [
+            { "$changeStream": { "fullDocument": "updateLookup", "allChangesForCluster": true } },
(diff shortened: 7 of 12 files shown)


src/operation/watch/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+                Self {
+                    target,
+                    pipeline: pipeline.into_iter().collect(),
+                    options,
+                    aggregate_operation,
+                }
+            }
+            ChangeStreamTarget::Cluster(database) => {
+                let aggregate_target = AggregateTarget::Database(database.name().to_string());
+                doc.get_document_mut("$changeStream")
```
saghm, line 69: I think it would be cleaner to do this check before putting the $changeStream document into the pipeline:

```rust
let mut options_doc = Document::new();
append_options(&mut options_doc, &options)?;

if let ChangeStreamTarget::Cluster(..) = target {
	options_doc.insert("allChangesForCluster", true);
}

let mut aggregate_pipeline = vec![options_doc];
aggregate_pipeline.extend(pipeline.clone());
```

src/operation/watch/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+impl Operation for Watch {
+    type O = ChangeStreamSpecification;
+    const NAME: &'static str = "watch";
+
+    fn build(&self, description: &StreamDescription) -> Result<Command> {
+        self.aggregate_operation.build(description)
+    }
+
+    fn handle_response(&self, response: CommandResponse) -> Result<Self::O> {
+        self.aggregate_operation.handle_response(response)?;
```
saghm, line 98: The CursorSpecification needs to be saved in the ChangeStreamSpecification

src/change_stream/mod.rs
```diff
@@ -132,3 +135,22 @@ impl Stream for ChangeStream {
...
 }
+
+// Specification used to create a new cursor.
+#[derive(Debug)]
+pub(crate) struct ChangeStreamSpecification {
+    /// The pipeline of stages to append to an initial `$changeStream` stage.
+    pub(crate) pipeline: Vec<Document>,
+
+    /// The options provided to the initial `$changeStream` stage.
+    pub(crate) options: Option<ChangeStreamOptions>,
```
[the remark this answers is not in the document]
utkmehta: fixed

src/operation/watch/mod.rs
utkmehta: fixed

src/operation/watch/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+impl Watch {
+    // Create an aggregate command
+    pub(crate) fn new(
+        target: impl Into<ChangeStreamTarget>,
+        pipeline: impl IntoIterator<Item = Document>,
+        options: Option<ChangeStreamOptions>,
+    ) -> Self {
+        let mut doc: Document = Document::new();
+        match &options {
+            Some(opts) => doc.insert("$changeStream", bson::to_bson(&opts).unwrap()),
```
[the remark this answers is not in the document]
utkmehta: fixed

src/operation/watch/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+            None => doc.insert("$changeStream", doc! {}),
+        };
+        let mut aggregate_pipeline = Vec::new();
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        let target: ChangeStreamTarget = target.into();
+        match &target {
```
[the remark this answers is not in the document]
utkmehta: fixed

src/operation/watch/mod.rs
utkmehta: fixed

src/change_stream/mod.rs
```diff
@@ -93,9 +95,11 @@ pub struct ChangeStream {
     /// The options provided to the initial `$changeStream` stage.
     options: Option<ChangeStreamOptions>,
 
-    /// The read preference for the initial `$changeStream` aggregation, used for server selection
-    /// during an automatic resume.
-    read_preference: Option<ReadPreference>,
+    /// The criteria used to select a server for this operation.
```
saghm, line 98: I think this comment should be left as it was in the previous commit, just with "read preference" replaced with "selection criteria".

src/change_stream/mod.rs
```diff
@@ -107,6 +111,28 @@ pub struct ChangeStream {
 }
 
 impl ChangeStream {
+    pub(crate) fn new(
+        mut cursor: Cursor,
```
saghm, line 115: This should just take a `ChangeStreamSpecification` rather than taking the individual fields of it passed in separately

src/client/mod.rs
```diff
@@ -309,6 +310,27 @@ impl Client {
         pipeline: impl IntoIterator<Item = Document>,
         options: Option<ChangeStreamOptions>,
     ) -> Result<ChangeStream> {
-        todo!();
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        let mut options = options;
```
saghm, line 314: Instead of rebinding `options` as mut, the parameter can just be defined as mut, i.e. `mut options: Option<ChangeStreamOptions>`

src/coll/mod.rs
```diff
@@ -601,7 +602,29 @@ impl Collection {
         pipeline: impl IntoIterator<Item = Document>,
         options: Option<ChangeStreamOptions>,
     ) -> Result<ChangeStream> {
-        todo!();
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        let mut options = options;
```
saghm, line 606: Ditto above about marking the parameter as `mut`

src/db/mod.rs
```diff
@@ -282,6 +282,28 @@ impl Database {
         pipeline: impl IntoIterator<Item = Document>,
         options: Option<ChangeStreamOptions>,
     ) -> Result<ChangeStream> {
-        todo!();
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        let mut options = options;
```
saghm, line 286: Ditto above about marking the parameter as `mut`

src/client/mod.rs
```diff
@@ -309,6 +310,27 @@ impl Client {
...
+
+        let watch = Watch::new(
+            ChangeStreamTarget::Cluster(self.database("admin")),
+            pipeline.clone(),
+            options.clone(),
+        )?;
+        let cursor = self
+            .execute_cursor_operation(watch)
+            .await
+            .map(|(spec, session)| Cursor::new(self.clone(), spec.cursor_spec, session))?;
```
saghm, line 325: I'm a bit confused by what's being done here; `execute_cursor_operation` is being called with a `Watch` operation and returns a `ChangeStreamSpecification`, which contains the pipeline, the options, and the cursor spec, but then the `CursorSpecification` is being taken out and the `ChangeStreamSpecification` is thrown away. As it's written right now, the `ChangeStreamSpecification` type isn't being used for anything. I think it would make more sense to have `ChangeStream::new` take in a `ChangeStreamSpecification` the same way that `Cursor::new` takes in a `CursorSpecification`, and then the `ChangeStream` can construct the `Cursor` itself.

src/operation/watch/test.rs
```diff
@@ -0,0 +1,246 @@
...
+    Namespace,
+};
+
+fn build_test(
+    target: impl Into<ChangeStreamTarget>,
+    pipeline: Vec<Document>,
+    options: Option<ChangeStreamOptions>,
+    mut expected_body: Document,
+) {
+    let target = target.into();
```
saghm, line 27: `target` isn't used elsewhere, so it's fine to just pass `target.into()` directly as an argument to `Watch::new()` without making a separate binding

src/operation/watch/test.rs
```diff
@@ -0,0 +1,246 @@
...
+    bson_util::sort_document(&mut expected_body);
+    bson_util::sort_document(&mut cmd.body);
+
+    assert_eq!(cmd.body, expected_body);
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn build() {
+    let client = TestClient::new().await;
```
saghm, line 44: The build tests don't do anything with the database, so they shouldn't create a client and don't need to check the topology. The topology only needs to be checked the integration tests (i.e. the ones that connect to the database by calling `watch`)

src/change_stream/mod.rs
utkmehta: fixed

src/client/mod.rs
utkmehta: fixed

src/coll/mod.rs
utkmehta: fixed

src/db/mod.rs
utkmehta: fixed

src/operation/watch/test.rs
utkmehta: fixed

src/operation/watch/test.rs
utkmehta: changed the ChangeStream target to not include an actual database/collection. fixed.

src/change_stream/mod.rs
saghm: This doesn't seem to be fixed in the latest commit; the comment is still, the same as before. It should be the same as it was before this pull request (i.e. the left side of the diff in this view).

src/client/mod.rs
saghm: This still doesn't seem to be fixed in the latest commit.

src/coll/mod.rs
saghm: This doesn't seem to be fixed in the latest commit.

src/db/mod.rs
saghm: This doesn't seem to be fixed in the latest commit.

src/operation/watch/test.rs
saghm: This doesn't seem to be fixed in the latest commit.

src/operation/watch/test.rs
saghm: This doesn't seem fixed in the latest commit.

src/test/client.rs
```diff
@@ -500,3 +507,21 @@ async fn saslprep() {
...
+    }
+    let pipeline = vec![doc! { "$match": { "x": 3 }}];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+    let result = client.watch(pipeline, Some(options)).await;
+    assert!(result.is_ok());
```
saghm, line 526: This test seems to be failing in CI, but it's hard to tell while due to the fact that this assertion only prints that `result.is_ok()` isn't true. Please update the above call to be `client.watch(pipeline, options).await.unwrap()` so that the full error will be printed in the case of a failure.

src/test/coll.rs
```diff
@@ -753,3 +757,29 @@ async fn no_read_preference_to_standalone() {
...
+
+    let pipeline = vec![doc! { "$match": { "x": 3 }}];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+    let result = coll.watch(pipeline, Some(options)).await;
+    assert!(result.is_ok());
```
saghm, line 784: Ditto above about unwrapping instead of asserting

src/test/db.rs
```diff
@@ -362,3 +370,24 @@ async fn index_option_defaults_test(defaults: Option<IndexOptionDefaults>, name:
...
+    let db = client.database("test_db");
+    let pipeline = vec![doc! { "$match": { "x": 3 }}];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+    let result = db.watch(pipeline, Some(options)).await;
+    assert!(result.is_ok());
```
saghm, line 392: Ditto above about unwrapping instead of asserting

src/change_stream/mod.rs
utkmehta: didn't push earlier. fixed now

src/change_stream/mod.rs
utkmehta: didn't push earlier. fixed now

src/client/mod.rs
utkmehta: didn't push earlier. fixed now

src/coll/mod.rs
utkmehta: didn't push earlier. fixed now

src/db/mod.rs
utkmehta: didn't push earlier. fixed now

src/client/mod.rs
utkmehta: fixed

src/operation/watch/test.rs
utkmehta: didn't push earlier. fixed now

src/operation/watch/test.rs
utkmehta: didn't push earlier. fixed now

src/test/client.rs
utkmehta: fixed

src/test/coll.rs
utkmehta: fixed

src/test/db.rs
utkmehta: fixed

src/change_stream/mod.rs
```diff
@@ -95,7 +99,7 @@ pub struct ChangeStream {
 
     /// The read preference for the initial `$changeStream` aggregation, used for server selection
```
saghm, line 100: As I mentioned in my first comment about this, "read preference" does need to be replaced with "SelectionCriteria"; the rest of the comment should remain the same

src/change_stream/mod.rs
```diff
@@ -107,6 +111,25 @@ pub struct ChangeStream {
...
+    ) -> Self {
+        let cursor = Cursor::new(client.clone(), change_streams_spec.cursor_spec, session);
+        Self {
+            cursor,
+            pipeline: change_streams_spec.pipeline,
+            client,
+            target: change_streams_spec.target,
+            resume_token: None,
+            options: change_streams_spec.options.clone(),
+            selection_criteria: change_streams_spec.options.unwrap().selection_criteria,
```
saghm, line 127: We definitely do not want to unwrap here; if the user doesn't specify any options, this will panic and crash the entire application. Instead, use `and_then` to return the selection criteria if there is one or `None` if there either the options or criteria are `None`:

```rust
selection_critiera: change_streams_spec.options.and_then(|opts| opts.selection_critiera),`
```

src/client/mod.rs
```diff
@@ -307,8 +307,17 @@ impl Client {
...
-        todo!();
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        resolve_options!(self, options, [selection_criteria]);
+
+        let watch = Watch::new(
+            ChangeStreamTarget::Cluster("admin".to_string()),
+            pipeline.clone(),
+            options.clone(),
+        )?;
+        let result = self.execute_cursor_operation(watch).await?;
```
saghm, line 320: I think it's cleaner to use destructuring to bind the return values separately:

```rust
let (change_stream_spec, session) = self.execute_cursor_operation(watch).await?;
Ok(ChangeStream::new(change_stream_spec, session, self.clone())
```

src/coll/mod.rs
```diff
@@ -599,9 +600,19 @@ impl Collection {
...
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        resolve_options!(self, options, [selection_criteria]);
+        let client = self.client();
+        let watch = Watch::new(
+            ChangeStreamTarget::Collection(self.namespace()),
+            pipeline.clone(),
+            options.clone(),
+        )?;
+
+        let result = client.execute_cursor_operation(watch).await?;
```
saghm, line 614: Ditto above about destructuring

src/db/mod.rs
```diff
@@ -280,8 +280,18 @@ impl Database {
...
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        resolve_options!(self, options, [selection_criteria]);
+        let client = self.client();
+        let watch = Watch::new(
+            ChangeStreamTarget::Database(self.name().to_string()),
+            pipeline.clone(),
+            options.clone(),
+        )?;
+
+        let result = client.execute_cursor_operation(watch).await?;
```
saghm, line 294: Ditto above about destructuring

src/operation/watch/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+                AggregateTarget::Database(database.clone())
+            }
+        };
+        let mut aggregate_pipeline = vec![doc! { "$changeStream": options_doc }];
+        aggregate_pipeline.extend(pipeline.clone());
+
+        let aggregate_operation = Aggregate::new(
+            aggregate_target,
+            aggregate_pipeline,
+            options.clone().map(|opts| opts.get_aggregation_options()),
```
saghm, line 46: This should use `options.as_ref()` instead of `options.clone()`; `get_aggregation_options` takes `&self` rather than `self` and clones the data it needs, so there's no need to clone it twice

src/test/client.rs
```diff
@@ -500,3 +507,20 @@ async fn saslprep() {
...
+        return;
+    }
+    let pipeline = vec![doc! { "$match": { "x": 3 }}];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+    let result = client.watch(pipeline, Some(options)).await.unwrap();
```
saghm, line 525: I think this will cause a lint error due to `result` being unused. Since we don't actually need the change stream for this test until later, we can just skip binding it

src/test/coll.rs
```diff
@@ -753,3 +757,28 @@ async fn no_read_preference_to_standalone() {
...
+        .await;
+
+    let pipeline = vec![doc! { "$match": { "x": 3 }}];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+    let result = coll.watch(pipeline, Some(options)).await.unwrap();
```
saghm, line 783: Ditto above about fixing potential lint issue

src/test/db.rs
```diff
@@ -362,3 +370,23 @@ async fn index_option_defaults_test(defaults: Option<IndexOptionDefaults>, name:
...
+    }
+    let db = client.database("test_db");
+    let pipeline = vec![doc! { "$match": { "x": 3 }}];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+    let result = db.watch(pipeline, Some(options)).await.unwrap();
```
saghm, line 391: Ditto above about fixing potential lint issue

src/change_stream/mod.rs
utkmehta: fixed

src/change_stream/mod.rs
utkmehta: fixed

src/client/mod.rs
utkmehta: fixed

src/coll/mod.rs
utkmehta: fixed

src/db/mod.rs
utkmehta: fixed

src/operation/watch/mod.rs
utkmehta: fixed

src/test/client.rs
utkmehta: fixed

src/test/coll.rs
utkmehta: fixed

src/test/db.rs
utkmehta: fixed

src/operation/watch/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+        target: impl Into<ChangeStreamTarget>,
+        pipeline: impl IntoIterator<Item = Document>,
+        options: Option<ChangeStreamOptions>,
+    ) -> Result<Self> {
+        let target: ChangeStreamTarget = target.into();
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+
+        let mut options_doc: Document = Document::new();
+        append_options(&mut options_doc, options.as_ref())?;
+        let aggregate_target = match &target {
```
saghm, line 30: Instead of defining `target`, I think this can just be `let aggregate_target = match target.into() { ... };`

saghm:
Okay, I think this is ready for the full team to review @patrickfreed @isabelatkinson

src/change_stream/mod.rs
```diff
@@ -107,6 +111,27 @@ pub struct ChangeStream {
 }
 
 impl ChangeStream {
+    pub(crate) fn new(
+        change_streams_spec: ChangeStreamSpecification,
```
patrickfreed, line 115: nit: it's a little redundant to call this a `change_streams_spec`, can we just call it `spec`?

src/change_stream/mod.rs
```diff
@@ -132,3 +157,36 @@ impl Stream for ChangeStream {
         todo!();
     }
 }
+
+// Specification used to create a new changeStream.
```
patrickfreed, line 161: can we rewrite "changeStream" as `ChangeStream` (with backticks) to emphasize its the type we're referring to?

src/client/mod.rs
```diff
@@ -307,8 +307,17 @@ impl Client {
...
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        resolve_options!(self, options, [selection_criteria]);
+
+        let watch = Watch::new(
+            ChangeStreamTarget::Cluster("admin".to_string()),
+            pipeline.clone(),
+            options.clone(),
+        )?;
+        let (change_stream_spec, session) = self.execute_cursor_operation(watch).await?;
+        Ok(ChangeStream::new(change_stream_spec, session, self.clone()))
```
patrickfreed, line 321: given that this implementation is identical across the three `watch` implementations, I think we should abstract it to a single helper that can be called with the various options and an appropriate value for `ChangeStreamTarget`.

src/operation/watch/test.rs
```diff
@@ -0,0 +1,221 @@
...
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+
+    let expected_body = doc! {
+        "aggregate": 1,
+        "cursor": {"batchSize": 10},
+        "pipeline": [doc!{"$changeStream": {"fullDocument": "UpdateLookup"}}, doc!{ "$match": { "x": 3 }}]
```
patrickfreed, line 51: this should be "updateLookup" with a lower-case "u".

src/operation/watch/test.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    bson_util::sort_document(&mut expected_body);
+    bson_util::sort_document(&mut cmd.body);
+
+    assert_eq!(cmd.body, expected_body);
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn build() {
```
patrickfreed, line 39: can we update the spacing in all these `doc!` invocations to be consistent? (i.e. spaces after every `[` or `{` and before every `]` or `}`). You don't need them after at the beginning and end of `vec!` invocations though, and rustfmt will enforce that I think.

src/operation/watch/test.rs
```diff
@@ -0,0 +1,221 @@
...
+        "pipeline": [doc!{"$changeStream": {}}, doc!{ "$match": { "x": 3 }}]
+    };
+
+    let result = watch.handle_response(CommandResponse::with_document_and_address(
+        address.clone(),
+        response.clone(),
+    ));
+    assert!(result.is_ok());
+
+    let change_streams_spec = result.unwrap();
```
patrickfreed, line 171: can we drop the "s" from `change_streams_spec`?

src/test/client.rs
```diff
@@ -500,3 +506,19 @@ async fn saslprep() {
     auth_test_uri("%E2%85%A8", "IV", None, true).await;
     auth_test_uri("%E2%85%A8", "I%C2%ADV", None, true).await;
 }
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn watch_test() {
+    let client = TestClient::new().await;
```
patrickfreed, line 513: this test needs to acquire the lock

src/test/coll.rs
```diff
@@ -753,3 +756,27 @@ async fn no_read_preference_to_standalone() {
...
 }
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[function_name::named]
+async fn watch_test() {
+    let _guard = LOCK.run_concurrently().await;
+
+    let client = EventClient::new().await;
+    if client.server_version_lt(4, 0) || client.is_standalone() {
```
patrickfreed, line 767: change streams were introdued in 3.6, is there any reason we're version gating here?

src/operation/watch/mod.rs
saghm: I missed that `target` is used on line 49; marking this as resolved

src/change_stream/mod.rs
utkmehta: fixed

src/change_stream/mod.rs
utkmehta: fixed

src/client/mod.rs
utkmehta: fixed

src/operation/watch/test.rs
utkmehta: fixed

src/operation/watch/test.rs
utkmehta: fixed

src/operation/watch/test.rs
utkmehta: fixed

src/test/client.rs
utkmehta: fixed

src/test/coll.rs
utkmehta: my bad, collections work on 3.6+. fixed

src/client/mod.rs
```diff
@@ -288,6 +288,19 @@ impl Client {
         }
     }
 
+    pub(crate) async fn watch_helper(
```
patrickfreed, line 291: rather than calling this simply `watch_helper`, can we give it a more descriptive name? something like `start_change_stream` or something. Also, can we add a docstring for it? Doesn't have to be super long but it should explain why this method exists.

src/client/mod.rs
```diff
@@ -288,6 +288,19 @@ impl Client {
...
+        mut options: Option<ChangeStreamOptions>,
+        target: ChangeStreamTarget,
+    ) -> Result<ChangeStream> {
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        resolve_options!(self, options, [selection_criteria]);
+
+        let watch = Watch::new(target, pipeline.clone(), options.clone())?;
+        let (change_stream_spec, session) = self.execute_cursor_operation(watch).await?;
+        Ok(ChangeStream::new(change_stream_spec, session, self.clone()))
+    }
```
patrickfreed, line 303: add a newline after this brace to give docstring below some space

src/operation/watch/test.rs
patrickfreed: There are still many in this file that need proper spacing. Also it's not necessary to invoke `doc!` from within itself.
e.g. on line 51:
```rust
"pipeline": [doc! {"$changeStream": {"fullDocument": "updateLookup"}}, doc! { "$match": { "x": 3 }}]
// should be
"pipeline": [{ "$changeStream": { "fullDocument": "UpdateLookup" } }, { "$match": { "x": 3 } }]
```

These often will look a lot better if you add line breaks.
```rust
"pipeline": [
    { "$changeStream": { "fullDocument": "UpdateLookup" } },
    { "$match": { "x": 3 } }
]
```
Check out [`src/operation/find/test.rs`](https://github.com/mongodb/mongo-rust-driver/blob/master/src/operation/find/test.rs) for examples of some well-formatted `doc!` invocations.

src/client/mod.rs
utkmehta: fixed

src/client/mod.rs
utkmehta: fixed

src/operation/watch/test.rs
```diff
@@ -0,0 +1,232 @@
...
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+
+    let expected_body = doc! {
+        "aggregate": 1,
+        "cursor": {"batchSize": 10},
+        "pipeline": [
+            {"$changeStream": {"fullDocument": "updateLookup"}},
```
isabelatkinson, line 52: re Patrick's comment above, we also tend to add a space after `{` and before `}`, so this should be 
`{ "$changeStream": { "fullDocument": "updateLookup" } }` (and ditto elsewhere)

src/client/mod.rs
```diff
@@ -288,6 +288,21 @@ impl Client {
...
 
+    /// A helper method to start a `ChangeStream`. Used by watch methods in client, coll, and db.
+    pub(crate) async fn start_change_stream(
+        &self,
+        pipeline: impl IntoIterator<Item = Document>,
+        mut options: Option<ChangeStreamOptions>,
+        target: ChangeStreamTarget,
+    ) -> Result<ChangeStream> {
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        resolve_options!(self, options, [selection_criteria]);
```
patrickfreed, line 299: We should also accept a read concern here via the options like we do for other read operations. According to the spec, we must document that only the default (i.e. no read concern) or "majority" read concern levels are supported in 3.6 and 4.0, and that starting in 4.2, any read concern can be specified.

From the spec:
>The initial implementation of change streams on the server requires a “majority” read concern or no read concern. Drivers MUST document this requirement. Drivers SHALL NOT throw an exception if any other read concern is specified, but instead should depend on the server to return an error.

src/operation/watch/mod.rs
```diff
@@ -0,0 +1,73 @@
+#[cfg(test)]
+mod test;
+use super::aggregate::{Aggregate, AggregateTarget};
```
patrickfreed, line 3: can we add line breaks between the various blocks in this file? (e.g. between the `mod` declaration and the `use`, between the `use` and the `struct`, etc.)

src/operation/watch/test.rs
patrickfreed: The example I referred to still does not appear to be completely fixed, in addition to several other instances in this file. Refer to the examples I proved above and Isabel's comment for more information on what we're looking for.

src/test/coll.rs
```diff
@@ -753,3 +756,25 @@ async fn no_read_preference_to_standalone() {
...
+#[function_name::named]
+async fn watch_test() {
+    let _guard = LOCK.run_concurrently().await;
+
+    let client = EventClient::new().await;
+    if client.is_standalone() {
+        return;
+    }
+
+    let coll = client.database(function_name!()).collection("coll");
```
patrickfreed, line 771: I think this will still error on 3.6 if the collection doesn't exist; I think it needs to be explicitly created. Also, to ensure we don't invoke `create_collection` with an already existing collection, be sure to still use `init_db_and_coll` which drops it first. Lastly, we typically use the function name for both the database and the collection.

src/test/db.rs
```diff
@@ -362,3 +369,22 @@ async fn index_option_defaults_test(defaults: Option<IndexOptionDefaults>, name:
...
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn watch_test() {
+    let _guard = LOCK.run_concurrently().await;
+
+    let client = EventClient::new().await;
+    if client.server_version_lt(4, 0) || client.is_standalone() {
+        return;
+    }
+    let db = client.database("test_db");
```
patrickfreed, line 382: ditto here use the function name for this.

src/change_stream/mod.rs
```diff
@@ -132,3 +155,36 @@ impl Stream for ChangeStream {
         todo!();
     }
 }
+
+// Specification used to create a new `ChangeStream`.
+#[derive(Debug)]
+pub(crate) struct ChangeStreamSpecification {
+    /// Specification used to create a new cursor.
```
patrickfreed, line 162: this docstring could be a little more specific, e.g. "...used to create the cursor wrapped by the `ChangeStream`."

src/client/mod.rs
```diff
@@ -307,8 +322,13 @@ impl Client {
     pub async fn watch(
         &self,
         pipeline: impl IntoIterator<Item = Document>,
-        options: Option<ChangeStreamOptions>,
+        mut options: Option<ChangeStreamOptions>,
```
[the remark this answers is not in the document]
utkmehta: makes sense. fixed

src/coll/mod.rs
```diff
@@ -599,9 +599,16 @@ impl Collection {
     pub async fn watch(
         &self,
         pipeline: impl IntoIterator<Item = Document>,
-        options: Option<ChangeStreamOptions>,
+        mut options: Option<ChangeStreamOptions>,
```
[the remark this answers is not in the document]
utkmehta: fixed

src/db/mod.rs
```diff
@@ -280,8 +280,15 @@ impl Database {
     pub async fn watch(
         &self,
         pipeline: impl IntoIterator<Item = Document>,
-        options: Option<ChangeStreamOptions>,
+        mut options: Option<ChangeStreamOptions>,
```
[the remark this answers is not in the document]
utkmehta: fixed

src/client/mod.rs
utkmehta: added

src/operation/watch/mod.rs
utkmehta: fixed

src/test/db.rs
utkmehta: fixed

src/change_stream/mod.rs
utkmehta: fixed

src/operation/watch/test.rs
utkmehta: I think its fixed everywhere else now!

src/test/coll.rs
patrickfreed: We still need to invoke `create_collection` here otherwise this test will continue to fail ([example from latest run of this PR](https://evergreen.mongodb.com/task_log_raw/mongo_rust_driver_stable_tests__async_runtime~async_std_auth_and_tls~noauth_and_notls_os~macos_10.14_test_3.6_sharded_cluster_patch_3ca7a68a50022f20592cc44175bd3938bab8a117_5f249f370ae6060c2dc2d803_20_07_31_22_46_16/0?type=T#L1262))

src/test/db.rs
patrickfreed: I'm now realizing that this test has the same name as the other watch tests, so there will be some collisions. Can you rename each of the tests to be unique?

src/client/mod.rs
patrickfreed: The requisite documentation still needs to be added.

src/change_stream/options.rs
```diff
@@ -66,10 +68,26 @@ pub struct ChangeStreamOptions {
...
+    ///
+    /// If none is specified, the selection criteria defined on the object executing this operation
+    /// will be used.
+    #[builder(default)]
+    #[serde(skip)]
+    #[serde(rename = "readPreference")]
+    pub selection_criteria: Option<SelectionCriteria>,
+
+    /// Specifies the default read concern for operations performed on the Client. See the
+    /// ReadConcern type documentation for more details.
```
patrickfreed, line 82: Can we wrap `Client` and `ReadConcern` in backticks to signal they're types?

src/test/coll.rs
utkmehta: fixed

src/test/db.rs
utkmehta: makes sense, fixed

src/change_stream/options.rs
utkmehta: yes, fixed

src/client/mod.rs
utkmehta: added!

src/change_stream/options.rs
patrickfreed: The main portion of the docstring doesn't really apply here--it looks like it's the same text that was used for `read_concern` on `ClientOptions`. We should rewrite this to be specific to change streams, e.g. "The read concern to be used by the `ChangeStream`."

Also, can we add a newline between the opening paragraph and the subsequent lines about which levels are supported?

src/client/mod.rs
```diff
@@ -288,6 +288,21 @@ impl Client {
...
 
+    /// A helper method to start a `ChangeStream`. Used by watch methods in client, coll, and db.
+    pub(crate) async fn start_change_stream(
+        &self,
+        pipeline: impl IntoIterator<Item = Document>,
+        mut options: Option<ChangeStreamOptions>,
+        target: ChangeStreamTarget,
+    ) -> Result<ChangeStream> {
+        let pipeline: Vec<Document> = pipeline.into_iter().collect();
+        resolve_options!(self, options, [read_concern, selection_criteria]);
```
patrickfreed, line 299: Hm, I actually think we need to invoke this particular line in the individual `watch` helpers themselves, since otherwise it will always resolve using the options specified on the `Client` rather than the `Database` or `Collection` that `watch` may have been called from. Once you move this line to those places, can you add a test verifying the `ReadConcern` and `SelectionCriteria` are inherited properly?

src/test/db.rs
```diff
@@ -362,3 +370,24 @@ async fn index_option_defaults_test(defaults: Option<IndexOptionDefaults>, name:
...
+
+    let client = EventClient::new().await;
+    if client.server_version_lt(4, 0) || client.is_standalone() {
+        return;
+    }
+    let db = client.database(function_name!());
+    let pipeline = vec![doc! { "$match": { "x": 3 } }];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .read_concern(ReadConcern::from(ReadConcernLevel::Available))
```
patrickfreed, line 388: this test is specifying a read concern other than `Majority` on versions that include 4.0, so this test _should_ be failing. It appears to not be, though, meaning the read concern is likely not getting serialized properly. This needs investigation.

src/operation/watch/test.rs
```diff
@@ -0,0 +1,236 @@
...
+    let pipeline = vec![doc! { "$match": { "x": 3 } }];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .read_concern(ReadConcern::from(ReadConcernLevel::Available))
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+
+    let expected_body = doc! {
```
patrickfreed, line 50: regarding my comment below, this `expected_body` should contain a `readConcern` field (since a read concern is specified above) but it doesn't.

src/change_stream/options.rs
utkmehta: fixed

src/client/mod.rs
utkmehta: fixed

src/operation/watch/test.rs
utkmehta: fixed

src/client/mod.rs
```diff
@@ -307,8 +320,15 @@ impl Client {
     pub async fn watch(
         &self,
         pipeline: impl IntoIterator<Item = Document>,
-        options: Option<ChangeStreamOptions>,
+        options: impl Into<Option<ChangeStreamOptions>>,
     ) -> Result<ChangeStream> {
-        todo!();
+        let options = options.into();
+        resolve_options!(self, options.clone(), [read_concern, selection_criteria]);
```
patrickfreed, line 326: we don't want to clone here since we want to mutate the options we already have. This applies to other invocations of `resolve_options` too.

src/test/coll.rs
```diff
@@ -787,3 +788,43 @@ async fn watch_test_coll() {
...
+    let coll = client
+        .init_db_and_coll(function_name!(), function_name!())
+        .await;
+
+    let coll_options = CollectionOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .read_concern(ReadConcern::from(ReadConcernLevel::Available))
+        .build();
+
+    client
```
patrickfreed, line 812: the result of this line is unused, so the options set on it will not be included in the final change stream, causing this test to fail. Also, we need to create the collection in this test like we do in the other change stream tests.

I would encourage you to test these locally first, since waiting for evergreen patches can take some time, and doing so also saves on round trips of code review.

src/client/mod.rs
utkmehta: fixed

src/test/coll.rs
utkmehta: fixed

PR #222, opened 2020-07-29
minor: make AuthMechanism non-exhaustive

saghm:
I realized right after merging RUST-359 that I forgot to make the change that was the whole reason we delayed merging it in the first place, so here it is

### DIFF
--- a/src/client/auth/mod.rs
+++ b/src/client/auth/mod.rs
@@ -81,7 +81,8 @@ pub enum AuthMechanism {
     /// Assume Role request, or temporary AWS IAM credentials assigned to an EC2 instance or ECS
     /// task.
     ///
-    /// Note: This mechanism is not currently only supported with the tokio runtime.
+    /// Note: Only server versions 4.4+ suppport AWS authentication. Additionally, the driver only
+    /// supports AWS authentication with the tokio runtime.
     #[cfg(feature = "tokio-runtime")]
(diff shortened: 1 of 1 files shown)


saghm:
I've added a doc comment. Interestingly, I couldn't find any official documentation to link to for it; it might not be available until 4.4 is GA

patrickfreed:
Sorry one more nit, can we include that it is only supported in 4.4+?

saghm:
Oops, good catch; updated

src/client/auth/mod.rs
```diff
@@ -75,6 +76,13 @@ pub enum AuthMechanism {
     /// Note: This mechanism is not currently supported by this driver but will be in the future.
     Plain,
 
+    /// MONGODB-AWS authenticates using AWS IAM credentials (an access key ID and a secret access
+    /// key), temporary AWS IAM credentials obtained from an AWS Security Token Service (STS)
+    /// Assume Role request, or temporary AWS IAM credentials assigned to an EC2 instance or ECS
+    /// task.
+    ///
+    /// Note: Only server versions 4.4+ suppport AWS authentication. Additionally, the driver only
```
isabelatkinson, line 84: nit: there's an extra p in support

saghm: Oops, not sure how that happened! Fixed

PR #224, opened 2020-07-30
RUST-478 Add connection snippet examples

isabelatkinson:
This PR adds an examples directory to the driver with connection snippets.

### DIFF
--- a/examples/connection_snippets.rs
+++ b/examples/connection_snippets.rs
@@ -29,6 +29,7 @@ async fn main() -> Result<()> {
 
 #[cfg(feature = "async-std-runtime")]
 #[async_std::main]
+#[cfg(not(feature = "sync"))]
 async fn main() -> Result<()> {
     use mongodb::{options::ClientOptions, Client};
     let client_options = ClientOptions::parse(CONNECTION_URI).await?;
@@ -36,10 +37,3 @@ async fn main() -> Result<()> {
     let database = client.database("test");
     Ok(())
 }
-
-#[cfg(all(
-    not(feature = "sync"),
-    not(feature = "tokio-runtime"),
-    not(feature = "async-std-runtime")
-))]
-fn main() {}
--- a/examples/connection_snippets.rs
+++ b/examples/connection_snippets.rs
@@ -4,35 +4,35 @@
 
 extern crate mongodb;
 
-use mongodb::error::Result;
+#[cfg(feature = "tokio-runtime")]
+use tokio::main;
 
-const CONNECTION_URI: &str =
-    "mongodb+srv://<username>:<password>@<cluster-url>/<dbname>?w=majority";
+#[cfg(feature = "async-std-runtime")]
+use async_std::main;
+
+use mongodb::error::Result;
 
 #[cfg(feature = "sync")]
 fn main() -> Result<()> {
     use mongodb::sync::Client;
-    let client = Client::with_uri_str(CONNECTION_URI)?;
+    let client = Client::with_uri_str(
+        "mongodb+srv://<username>:<password>@<cluster-url>/<dbname>?w=majority",
+    )?;
     let database = client.database("test");
     Ok(())
 }
 
-#[cfg(feature = "tokio-runtime")]
-#[tokio::main]
+#[cfg(not(feature = "sync"))]
+#[crate::main]
 async fn main() -> Result<()> {
+    // CONNECTION EXAMPLE STARTS HERE
     use mongodb::{options::ClientOptions, Client};
-    let client_options = ClientOptions::parse(CONNECTION_URI).await?;
+    let client_options = ClientOptions::parse(
+        "mongodb+srv://<username>:<password>@<cluster-url>/<dbname>?w=majority",
+    )
+    .await?;
     let client = Client::with_options(client_options)?;
     let database = client.database("test");
-    Ok(())
-}
-
-#[cfg(all(feature = "async-std-runtime", not(feature = "sync")))]
-#[async_std::main]
-async fn main() -> Result<()> {
-    use mongodb::{options::ClientOptions, Client};
-    let client_options = ClientOptions::parse(CONNECTION_URI).await?;
-    let client = Client::with_options(client_options)?;
-    let database = client.database("test");
+    // CONNECTION EXAMPLE ENDS HERE
     Ok(())
 }
--- a/examples/connection_snippets.rs
+++ b/examples/connection_snippets.rs
@@ -27,9 +27,8 @@ async fn main() -> Result<()> {
     Ok(())
 }
 
-#[cfg(feature = "async-std-runtime")]
+#[cfg(all(feature = "async-std-runtime", not(feature = "sync")))]
 #[async_std::main]
-#[cfg(not(feature = "sync"))]
 async fn main() -> Result<()> {
     use mongodb::{options::ClientOptions, Client};
     let client_options = ClientOptions::parse(CONNECTION_URI).await?;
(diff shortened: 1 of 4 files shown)


examples/connection_snippets.rs
```diff
@@ -0,0 +1,45 @@
...
+#[tokio::main]
+async fn main() -> Result<()> {
+    use mongodb::{options::ClientOptions, Client};
+    let client_options = ClientOptions::parse(CONNECTION_URI).await?;
+    let client = Client::with_options(client_options)?;
+    let database = client.database("test");
+    Ok(())
+}
+
+#[cfg(feature = "async-std-runtime")]
```
saghm, line 30: I think this also needs `not(feature = "sync")`, since the `sync` feature also has `async-std-runtime`. I suspect that's what's causing the failure on evergreen

saghm:
Hmm, there still seems to be an issue with the lint task, although I honestly can't tell if that's a transient failure or just more shenanigans with the feature configuration for examples

isabelatkinson:
> Hmm, there still seems to be an issue with the lint task, although I honestly can't tell if that's a transient failure or just more shenanigans with the feature configuration for examples

Seems like a fluke, I ran another patch and the lint passed; it's also passing locally.

saghm:
Awesome! It's possible the evergreen UI might let you restart it, although maybe not since there are some issues right now

isabelatkinson:
lint is now passing after a restart

examples/connection_snippets.rs
```diff
@@ -0,0 +1,38 @@
...
+fn main() -> Result<()> {
+    use mongodb::sync::Client;
+    let client = Client::with_uri_str(CONNECTION_URI)?;
+    let database = client.database("test");
+    Ok(())
+}
+
+#[cfg(feature = "tokio-runtime")]
+#[tokio::main]
+async fn main() -> Result<()> {
```
patrickfreed, line 22: for Swift they requested that we only do the async API, so we probably want to do the same here. And we should probably just stick to tokio to keep things simple.

examples/connection_snippets.rs
```diff
@@ -0,0 +1,38 @@
...
+#[cfg(feature = "sync")]
+fn main() -> Result<()> {
+    use mongodb::sync::Client;
+    let client = Client::with_uri_str(CONNECTION_URI)?;
+    let database = client.database("test");
+    Ok(())
+}
+
+#[cfg(feature = "tokio-runtime")]
+#[tokio::main]
```
patrickfreed, line 21: typically in docs examples, we surround the sections we want the other teams to copy/paste with comments so they know exactly what should be taken. I think we should do that here too. In order to actually connect, users need the `#[tokio::main]` thing, so we should include the entirety of the `main` declaration, but not the `cfg` part. If we put comments between the two macros does that mess anything up?

e.g.
```rust
#[cfg(feature = "tokio-runtime")]
// CONNECTION EXAMPLE STARTS HERE
#[tokio::main]
async fn main() -> Result<()> { ... }
// CONNECTION EXAMPLE ENDS HERE
```

examples/connection_snippets.rs
```diff
@@ -0,0 +1,38 @@
+#![allow(dead_code)]
+#![allow(unused_imports)]
+#![allow(unused_variables)]
+
+extern crate mongodb;
+
+use mongodb::error::Result;
+
+const CONNECTION_URI: &str =
```
patrickfreed, line 9: since the Atlas team will just be copy/pasting this example, I think we actually need to have this string literal in the examples themselves, though I do agree that in "real" code we should use a constant.

examples/connection_snippets.rs
saghm: Maybe we can split each feature into a separate example file?

isabelatkinson: That was my first approach, but in order for the examples to compile there needs to be a `main` function present for each feature, since these examples need to be able to be run. Is there a way to enable/disable an entire file based on features?

isabelatkinson: I suppose I could also just include an empty main for the compiler's sake

saghm: I think the main function itself should be identical for tokio and async-std, right? The only difference will be using `tokio::main` versus `async_std::main`. Maybe we could cheat a little and conditionally import whichever `main` we need and then define a single async main:

```rust
#[cfg(feature = "tokio-runtime")]
use tokio::main;

#[cfg(feature = "async-std-runtime")]
use async_std::main;

#[cfg(not(feature = "sync"))]
#[main]
async fn main() -> Result<()> {
   ...
}
```

examples/connection_snippets.rs
```diff
@@ -0,0 +1,38 @@
...
+        "mongodb+srv://<username>:<password>@<cluster-url>/<dbname>?w=majority",
+    )?;
+    let database = client.database("test");
+    Ok(())
+}
+
+#[cfg(not(feature = "sync"))]
+#[crate::main]
+async fn main() -> Result<()> {
+    // CONNECTION EXAMPLE STARTS HERE
```
isabelatkinson, line 28: Combined the two async mains here. The example still requires there be a sync main so that everything compiles with the sync feature flag, but I added comments to indicate where the example snippet is.

examples/connection_snippets.rs
isabelatkinson: The other code snippets in Atlas only include the lines to create a client/database, so I added the comments in line with that; let me know if you think we should do it differently

src/test/connection_snippets.rs
```diff
@@ -0,0 +1,20 @@
+#![allow(dead_code)]
+#![allow(unused_imports)]
+#![allow(unused_variables)]
+
+use crate::error::Result;
+use crate as mongodb;
```
isabelatkinson, line 6: hacky aliasing so that the example uses `mongodb` instead of `crate`

src/test/connection_snippets.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#[cfg(feature = "tokio-runtime")]
+async fn connect() -> Result<()> {
```
isabelatkinson, line 9: This will fail if we actually run it since the URI has placeholders

src/test/connection_snippets.rs
saghm: Earlier in the driver's development, we moved all of the external tests from the `tests` directory to the `test` submodule so that we could share code between our internal and external tests. However, since we're writing this test as a way to document usage of the driver, we know we won't ever be using any internal or test-only features of the driver, so this might be a good candidate to put in the `tests` directory at the top-level of the repo, in which case we wouldn't need to do the aliasing.

src/test/connection_snippets.rs
saghm: I think this might cause some lint warnings due to stuff not being used; could you take a look at the CI output and annotate the function accordingly to suppress the warnings?

patrickfreed, line 9: can we add a `#[tokio::main]` macro here and include that in the example? Otherwise users won't be able to copy/paste this snippet and have it start working. Also, does splitting up the various attributes with `///` break anything?

examples/connection_snippets.rs
patrickfreed: Oh sorry, I didn't see your response here before commenting in the other thread. I think we need to deviate a little from what other languages do here, since in Rust the `.await`s won't work unless called in the context of a `#[tokio::main]` annotated function, and the goal here is to provide copy/pastable code for new users.

src/test/connection_snippets.rs
isabelatkinson: looks like it was just formatting, fixed

src/test/connection_snippets.rs
isabelatkinson: makes sense, I'll move it

PR #230, opened 2020-08-06
RUST-525 Implement and Test Stream for ChangeStreams

utkmehta:
This PR implements and tests Stream (without resumability)

### DIFF
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -615,7 +615,7 @@ impl Collection {
 }
 
 /// A struct modeling the canonical name for a collection in MongoDB.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Deserialize, PartialEq)]
 pub struct Namespace {
     /// The name of the database associated with this namespace.
     pub db: String,
@@ -632,16 +632,8 @@ impl Namespace {
             coll: String::new(),
         }
     }
-}
 
-impl fmt::Display for Namespace {
-    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        write!(fmt, "{}.{}", self.db, self.coll)
-    }
-}
-
-impl<'de> Deserialize<'de> for Namespace {
-    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    pub(crate) fn deserialize_ns<'de, D>(deserializer: D) -> std::result::Result<Self, D::Error>
     where
         D: Deserializer<'de>,
     {
@@ -658,5 +650,11 @@ impl<'de> Deserialize<'de> for Namespace {
             }),
             _ => Err(D::Error::custom("Missing one or more fields in namespace")),
         }
+    }
+}
+
+impl fmt::Display for Namespace {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        write!(fmt, "{}.{}", self.db, self.coll)
     }
 }
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -3,6 +3,7 @@ pub mod options;
 use std::sync::Arc;
 
 use futures::stream::TryStreamExt;
+use serde::{Deserialize, Deserializer};
 
 use crate::{
     bson::{Bson, Document},
@@ -294,5 +295,20 @@ impl Database {
                 ChangeStreamTarget::Database(self.name().to_string()),
             )
             .await
+    }
+
+    pub(crate) fn deserialize_str_from_map<'de, D>(
+        deserializer: D,
+    ) -> std::result::Result<String, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        struct Db {
+            db: String,
+        }
+
+        let db_object: Db = Deserialize::deserialize(deserializer)?;
+        Ok(db_object.db)
     }
 }
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -3,7 +3,6 @@ pub mod options;
 use std::sync::Arc;
 
 use futures::stream::TryStreamExt;
-use serde::{Deserialize, Deserializer};
 
 use crate::{
     bson::{Bson, Document},
@@ -295,20 +294,5 @@ impl Database {
                 ChangeStreamTarget::Database(self.name().to_string()),
             )
             .await
-    }
-
-    pub(crate) fn deserialize_str_from_map<'de, D>(
-        deserializer: D,
-    ) -> std::result::Result<String, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        #[derive(Deserialize)]
-        struct Db {
-            db: String,
-        }
-
-        let db_object: Db = Deserialize::deserialize(deserializer)?;
-        Ok(db_object.db)
     }
 }
--- a/src/test/change_stream.rs
+++ b/src/test/change_stream.rs
@@ -1,0 +1,711 @@
+use std::time::Duration;
+
+use futures::StreamExt;
+
+use crate::{
+    bson::doc,
+    change_stream::document::{ChangeStreamEventSource, OperationType, UpdateDescription},
+    options::{ChangeStreamOptions, FullDocumentType, UpdateOptions},
+    selection_criteria::{ReadPreference, SelectionCriteria},
+    test::{util::drop_collection, TestClient, LOCK},
+    Namespace,
+    RUNTIME,
+};
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn empty_change_stream() {
+    let _guard = LOCK.run_concurrently().await;
+
+    let client = TestClient::new().await;
+    if client.server_version_lt(4, 0) || client.is_standalone() {
+        return;
+    }
+    let pipeline = vec![doc! { "$match": { "x": 3 } }];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .build();
+    let mut change_stream = client.watch(pipeline, Some(options)).await.unwrap();
+    let result = RUNTIME
+        .timeout(Duration::from_millis(5), change_stream.next())
+        .await;
+    assert!(!result.is_ok());
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[function_name::named]
+async fn change_stream_insert_one() {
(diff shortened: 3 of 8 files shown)


saghm:
From looking at the CI, it seems that this branch isn't compiling. Can you fix the issues?

src/db/mod.rs
```diff
@@ -295,4 +296,19 @@ impl Database {
             )
             .await
     }
+
+    pub(crate) fn deserialize_str_from_map<'de, D>(
```
saghm, line 300: This doesn't have anything to do with the database type, so I don't think it should be inside the `Database` impl block. Please move it to the `bson_util` module as a top-level function.

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
+    let _guard = LOCK.run_concurrently().await;
+
+    let client = TestClient::new().await;
+    if client.server_version_lt(4, 0) || client.is_standalone() {
+        return;
+    }
+    let pipeline = vec![doc! { "$match": { "x": 3 } }];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .max_await_time(Duration::from_millis(5))
```
saghm, line 27: 5 ms is very short for a maxAwaitTimeMS; in practice, this will mean the change stream will send a getMore to the server as soon as it gets back an empty response, which is too frequent. This should be something higher (probably at least 500 ms).

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
+    RUNTIME.spawn(async move {
+        let docs = vec![
+            doc! { "_id": 1, "title": "1984", "author": "George Orwell" },
+            doc! { "_id": 2, "title": "Animal Farm", "author": "George Orwell" },
+            doc! { "_id": 3, "title": "The Great Gatsby", "author": "F. Scott Fitzgerald" },
+        ];
+        coll_ref.insert_many(docs, None).await.unwrap();
+    });
+
+    for id in 1..4 {
```
saghm, line 120: I think it would be a bit more clear to use an inclusive range here, i.e. `1..=3`

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
+            .insert_one(original_doc.clone(), None)
+            .await
+            .unwrap();
+        coll_ref
+            .update_one(original_doc, doc! { "$set": { "x": 5 } }, None)
+            .await
+            .unwrap();
+    });
+
+    let _insert_document = change_stream.next().await.unwrap().unwrap();
```
saghm, line 172: If there isn't any need to check this document in the test, it's probably cleaner to just insert it before creating the change stream rather than skipping over this without testing.

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
+            upsert: Some(true),
+            ..Default::default()
+        };
+        coll_ref
+            .update_one(filter, doc! { "$set": { "_id": 2, "x": 5 } }, options)
+            .await
+            .unwrap();
+    });
+
+    let _insert_document = change_stream.next().await.unwrap().unwrap();
```
saghm, line 286: Ditto above about inserting before watching

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
+        let filter = doc! { "x": { "$gt": 0 } };
+
+        coll_ref
+            .update_many(filter, doc! { "$set": { "y": 3 } }, None)
+            .await
+            .unwrap();
+    });
+
+    for _i in 0..2 {
+        let _insert_document = change_stream.next().await.unwrap().unwrap();
```
saghm, line 351: Ditto above about inserting before watching

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
+            .unwrap();
+
+        let filter = doc! { "x": 0 };
+        coll_ref
+            .replace_one(filter, doc! { "x": 1 }, None)
+            .await
+            .unwrap();
+    });
+
+    let _insert_document = change_stream.next().await.unwrap().unwrap();
```
saghm, line 440: Ditto above about inserting before watching

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
+    if client.server_version_lt(4, 0) || client.is_standalone() {
+        return;
+    }
+    let pipeline = vec![doc! { "$match": { "x": 3 } }];
+    let options = ChangeStreamOptions::builder()
+        .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+        .max_await_time(Duration::from_millis(5))
+        .batch_size(10)
+        .build();
+    let mut change_stream = client.watch(pipeline, Some(options)).await.unwrap();
```
saghm, line 30: Since the `watch` helpers take `impl Into<Option<ChangeStreamOptions>>`, there's no need to use `Some` here; `T` implements `Into<Option<T>>` for all `T`.

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
+        .database(function_name!())
+        .create_collection(function_name!(), None)
+        .await
+        .unwrap();
+
+    let change_stream_options = ChangeStreamOptions::builder()
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+
+    let mut change_stream = coll.watch(None, Some(change_stream_options)).await.unwrap();
```
saghm, line 217: Ditto above about omitting `Some`

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
```
saghm, line 332: Ditto above about omitting `Some`

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
```
saghm, line 424: Ditto above about omitting `Some`

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
```
saghm, line 486: Ditto above about omitting `Some`

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
+        coll_ref
+            .insert_one(doc! { "_id": 0, "x": 0 }, None)
+            .await
+            .unwrap();
+
+        let filter = doc! { "x": 0 };
+        coll_ref.delete_one(filter, None).await.unwrap();
+    });
+
+    let _insert_document = change_stream.next().await.unwrap().unwrap();
```
saghm, line 499: Ditto above about inserting before watching

src/test/change_stream.rs
```diff
@@ -0,0 +1,711 @@
...
+        .database(function_name!())
+        .create_collection(function_name!(), None)
+        .await
+        .unwrap();
+
+    let change_stream_options = ChangeStreamOptions::builder()
+        .full_document(FullDocumentType::UpdateLookup)
+        .build();
+
+    let mut change_stream = coll.watch(None, Some(change_stream_options)).await.unwrap();
```
saghm, line 542: Ditto above about omitting `Some`

src/db/mod.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

src/test/change_stream.rs
utkmehta: fixed

saghm:
Some of the tests seem to be failing on 3.6 now that they're enabled

src/coll/mod.rs
```diff
@@ -632,16 +632,8 @@ impl Namespace {
...
     }
-}
-
-impl fmt::Display for Namespace {
-    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        write!(fmt, "{}.{}", self.db, self.coll)
-    }
-}
 
-impl<'de> Deserialize<'de> for Namespace {
```
patrickfreed, line 643: I worry that this change could potentially break user's code that were relying on it. What about just augmenting this implementation to support both formats? @saghm

src/test/change_stream.rs
```diff
@@ -0,0 +1,694 @@
...
+        .unwrap();
+
+    let mut change_stream = coll.watch(None, None).await.unwrap();
+
+    RUNTIME.spawn(async move {
+        client
+            .database("admin")
+            .run_command(
+                doc! {
+                    "renameCollection": format!("{}.{}", function_name!().to_string(), function_name!().to_string()),
```
patrickfreed, line 595: are all these `to_string` calls necessary?

utkmehta: fixed

src/test/change_stream.rs
```diff
@@ -0,0 +1,694 @@
...
+
+    let client = TestClient::new().await;
+    if client.is_standalone() {
+        return;
+    }
+
+    let coll = client
+        .init_db_and_coll(function_name!(), function_name!())
+        .await;
+    client
```
isabelatkinson, line 56: Is it necessary to call `create_collection` here? Ditto other tests below.

saghm:
We're deprioritizing implementing change streams for the time being, so closing this out for now

chdsbd:
Is there an issue tracking change stream support?

agolin95:
Hi @chdsbd, here is a jira filter list of Rust issues relating to change streams: https://jira.mongodb.org/issues/?jql=project%20%3D%20rust%20and%20summary%20~%20%22change%20streams%22%20and%20status%20!%3D%20closed

at the moment we don't currently have any of this work scheduled, but will update these tickets when we pick it up.

PR #232, opened 2020-08-12
minor: AWS auth spec test fixes

saghm:


src/client/options/mod.rs
```diff
@@ -834,6 +834,19 @@ fn validate_userinfo(s: &str, userinfo_type: &str) -> Result<()> {
         }
         .into());
     }
+
+    // All instances of '%' in the username must be part of an percent-encoded substring. This means
```
saghm, line 838: Previously, we had a check that would error if a percent character was present in the decoded username from the URI. This let us pass [this connection string spec test](https://github.com/mongodb/specifications/blob/4f89b79b6be51d6a61ae4196d07a0d6940d80be5/source/connection-string/tests/invalid-uris.yml#L236). However, [this auth spec test](https://github.com/mongodb/specifications/blob/4f89b79b6be51d6a61ae4196d07a0d6940d80be5/source/auth/tests/connection-string.yml#L348) contains an encoded percent sign in a valid URI. The restriction given in the [connection string spec](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#userinfo-optional) states that "If the user information contains an at-sign ("@"), more than one colon (":"), or a percent-sign ("%") that does not match the rules for "pct-encoded", then an exception MUST be thrown informing the user that the username and password must be URL encoded", but says nothing about disallowing an encoded percent sign, so I removed the previous check and added one here that checks for any `%` that isn't followed by two hex digits.

src/client/options/mod.rs
```diff
@@ -1018,12 +1031,6 @@ impl ClientOptionsParser {
             let mut credential = options.credential.get_or_insert_with(Default::default);
             validate_userinfo(u, "username")?;
             let decoded_u = percent_decode(u, "username must be URL encoded")?;
-            if decoded_u.chars().any(|c| c == '%') {
```
saghm, line 1021: This was the previous check mentioned above

src/test/spec/auth.rs
```diff
@@ -50,13 +50,7 @@ async fn run_auth_test(test_file: TestFile) {
...
         test_case.description = test_case.description.replace('$', "%");
 
-        let skipped_mechanisms = [
-            "GSSAPI",
-            "MONGODB-X509",
-            "PLAIN",
-            "MONGODB-CR",
-            "MONGODB-AWS",
-        ];
+        let skipped_mechanisms = ["GSSAPI", "MONGODB-X509", "PLAIN", "MONGODB-CR"];
```
saghm, line 53: For some reason there wasn't any TODO for AWS auth even though it was listed in this array.

src/test/spec/auth.rs
```diff
@@ -84,7 +78,11 @@ async fn run_auth_test(test_file: TestFile) {
                     None => assert!(options.credential.is_none(), "{}", test_case.description),
                 }
             }
-            Err(_) => assert!(!test_case.valid, "{}", test_case.description),
+            Err(e) => assert!(
```
saghm, line 81: While debugging, I found that it helped to see the actual error that occurred when we didn't expect one.

src/client/auth/mod.rs
```diff
@@ -122,6 +122,18 @@ impl AuthMechanism {
                 };
                 Ok(())
             }
+            #[cfg(feature = "tokio-runtime")]
+            AuthMechanism::MongoDbAws => {
+                if credential.username.is_some() && credential.password.is_none() {
+                    return Err(ErrorKind::ArgumentError {
+                        message: "Username cannot be provided with password for MONGODB-AWS \
```
isabelatkinson, line 129: I think this should be "without password"

src/client/auth/mod.rs
```diff
@@ -122,6 +122,18 @@ impl AuthMechanism {
...
+                                  authentication"
+                            .to_string(),
+                    }
+                    .into());
+                }
+                Ok(())
+            }
```
isabelatkinson, line 136: Is an AWS credential without a username valid?

src/client/auth/mod.rs
saghm: Oops, good catch; updated

src/client/auth/mod.rs
saghm: Yes, as long as there's also no password. In addition to the user specifying them, the driver also can obtain them through [environmental variables or ECS/EC2 endpoints](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#obtaining-credentials).

src/test/spec/auth.rs
```diff
@@ -50,13 +50,7 @@ async fn run_auth_test(test_file: TestFile) {
     for mut test_case in test_file.tests {
         test_case.description = test_case.description.replace('$', "%");
 
-        let skipped_mechanisms = [
-            "GSSAPI",
-            "MONGODB-X509",
-            "PLAIN",
-            "MONGODB-CR",
-            "MONGODB-AWS",
```
isabelatkinson, line 58: I forgot to check CI before, but it looks like this test is always failing on async-std since AWS is only enabled on tokio

saghm: Ah, right. Fixed.

src/client/options/mod.rs
patrickfreed: Will `percent_encoding::percent_decode_str` not error already if an unescaped "%" is included in the input? We call this in our `percent_decode` function which is invoked right after `validate_userinfo` is.

saghm: No, I don't think it will: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ef5d67a3353011e4bb59b98e568b9861

PR #234, opened 2020-08-13
RUST-530 Decrement connection count when connection is dropped due to unfinished operation

extrawurst:
fixes https://jira.mongodb.org/browse/RUST-530

related to https://jira.mongodb.org/browse/RUST-384
the dropped connections that are rendered useless were basically just forgotten and the connection pool was never notified that it has to consider it gone. so by having a very small pool size limit this quickly renders the entire driver blocked and unable to do anymore DB requests.

### DIFF
--- a/src/client/auth/mod.rs
+++ b/src/client/auth/mod.rs
@@ -122,6 +122,18 @@ impl AuthMechanism {
                 };
                 Ok(())
             }
+            #[cfg(feature = "tokio-runtime")]
+            AuthMechanism::MongoDbAws => {
+                if credential.username.is_some() && credential.password.is_none() {
+                    return Err(ErrorKind::ArgumentError {
+                        message: "Username cannot be provided without password for MONGODB-AWS \
+                                  authentication"
+                            .to_string(),
+                    }
+                    .into());
+                }
+                Ok(())
+            }
             _ => Ok(()),
         }
     }
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -834,6 +834,19 @@ fn validate_userinfo(s: &str, userinfo_type: &str) -> Result<()> {
         }
         .into());
     }
+
+    // All instances of '%' in the username must be part of an percent-encoded substring. This means
+    // that there must be two hexidecimal digits following any '%' in the username.
+    if s.split('%')
+        .skip(1)
+        .any(|part| part.len() < 2 || part[0..2].chars().any(|c| !c.is_ascii_hexdigit()))
+    {
+        return Err(ErrorKind::ArgumentError {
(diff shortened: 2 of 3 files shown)


src/cmap/mod.rs
```diff
@@ -300,6 +300,14 @@ impl ConnectionPoolInner {
         }
     }
 
+    async fn dropped(&self, conn: Connection) {
```
patrickfreed, line 303: From looking at the ["Connection Monitoring and Pooling Specification" (CMAP)](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst), it states that the `check_in` method should be [in charge of closing connections that are "perished"](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#checking-in-a-connection). That spec also [defines "perished"](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection) to mean a number of conditions, among them "Driver-Side timeout".

Given this, I think we can consolidate this logic into the `check_in` method, and simply add another check to the `if else` chain there for connections that are still executing. Then we can remove this method and also simplify `Connection`'s `drop` implementation a little bit to just call `check_in` if the pool still exists.

extrawurst: fine by me, I am not familiar with the pool logic. feel free to change accordingly. I really just needed this to be fixed and did whatever I needed to do 🙈

extrawurst: @patrickfreed or we merge this, someone smarter than me with the necessary domain knowledge of the driver adds a repro to the testing suit and *then* makes those optimisations that you mention. the last week with hassling with this in production gave me some serious **ocd**

patrickfreed: I can just go ahead and make those changes now. Just wanted to give you the opportunity to do so if you wished :+1:

saghm:
Apologies about closing this! I was attempting to my local branch with rebased history to `extrawurst/master`, but I accidentally overwrote it with my local master instead of my local branch, which Github apparently interpreted as a request to close the PR, which now prevents me from pushing to the fork again. Since we're trying to get a release out tonight with this fix, I pushed the changes manually to Github; you can see extrawurst's commit [here](https://github.com/mongodb/mongo-rust-driver/commit/4d4e0070ff42059a1af8c2f7e017b2f0bdd9950b) and Patrick's follow-up [here](https://github.com/mongodb/mongo-rust-driver/commit/8f26bfcfb8d8eb5d0fc4dc91ea446e831758c3a2).

PR #188, opened 2020-05-30
RUST-273: Implement Create indexes

Stargateur:
https://jira.mongodb.org/browse/RUST-458

PR #235, opened 2020-08-17
RUST-527 - Implementation of gridfs

Stargateur:
This is a work in progress implementation of [gridfs](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst). The jira ticket [RUST-527](https://jira.mongodb.org/browse/RUST-527). Any help/remark/critic are always welcome. I will update this according to my progress:

- [x] Basic Structure
- [x] Add features to avoid include it in crate that doesn't need it `gridfs`
- [x] Require to be able to create index, https://github.com/mongodb/mongo-rust-driver/pull/188
- [ ] Should we implement MD5 ?
- [ ] ?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ default = ["tokio-runtime"]
 tokio-runtime = ["tokio/dns", "tokio/macros", "tokio/rt-core", "tokio/tcp", "tokio/rt-threaded", "tokio/time", "reqwest", "serde_bytes"]
 async-std-runtime = ["async-std", "async-std/attributes"]
 sync = ["async-std-runtime"]
+gridfs = []
 
 [dependencies]
 async-trait = "0.1.24"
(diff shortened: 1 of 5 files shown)


Zageron:
Is this stale? Does this driver support gridfs?

kmahar:
Hey @Zageron! St this time the driver does not support GridFS. This PR has been inactive for a while but we've left it open in case the original author or anyone else is interesting in resuming work on it!

To respond to a couple of the points from the original PR -
* The driver has an index management API now which was added in version 2.0.0.
* We should omit MD5 support in a Rust implementation as per the GridFS specification it has been deprecated and should not be supported in new implementations (see [Design Rationale](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#design-rationale) section.)

kmahar:
we currently have implementation of GridFS in progress (so far a sketch of the public API has been merged in #688), so I am going to close out this pull request. thanks for the initial work on this @Stargateur!

PR #237, opened 2020-08-18
RUST-489 Check hosts field in initial dns discovery tests

isabelatkinson:


src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -68,44 +80,66 @@ async fn run() {
...
+            if requires_tls {
+                let tls_options = TlsOptions::builder()
+                    .allow_invalid_certificates(true)
+                    .build();
+                options_with_tls.tls = Some(Tls::Enabled(tls_options));
+            }
+            let mut client = TestClient::with_options(Some(options_with_tls)).await;
+            let mut actual_hosts = client.server_info.hosts.unwrap();
+            if let Some(ref mut arbiters) = client.server_info.arbiters {
+                actual_hosts.append(arbiters);
```
isabelatkinson, line 123: localhost:27019 consistently shows up in the arbiters list rather than the host list in the NoTLS variants on CI

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -40,6 +42,16 @@ async fn run() {
             return;
         }
 
+        // "encoded-userinfo-and-db.json" specifies a database name with a question mark which is
```
saghm, line 45: Wow, I'm kind of surprised that's allowed at all

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -68,44 +80,66 @@ async fn run() {
...
-            });
+        // "txt-record-with-overridden-ssl-option.json" requires SSL be disabled; see DRIVERS-1324.
+        let requires_tls = match test_file.options {
+            Some(ref options) => options.ssl,
+            None => true,
+        };
+        let client = TestClient::new().await;
+        if requires_tls == client.options.tls_options().is_some()
+            && client.is_replica_set()
+            && client.options.repl_set_name.as_deref() == Some("repl0")
```
saghm, line 91: Is the purpose of this check just to skip the tests if the user spun up a replica set with a different name?

isabelatkinson: Yep, the replica set name specified in the SRV records/spec test URIs is "repl0" so creating a client with the options from the spec test only works when the replica set is named "repl0"

src/test/spec/initial_dns_seedlist_discovery.rs
saghm: Looks like that's due to the configs in drivers-evergreen-tools; the basic replica set has [one arbiter](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/3be7d5197facf4f41b56cbda4a346d4e69a8ac62/.evergreen/orchestration/configs/replica_sets/basic.json#L35-L43), but the auth-ssl one [doesn't seem to have any](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/3be7d5197facf4f41b56cbda4a346d4e69a8ac62/.evergreen/orchestration/configs/replica_sets/auth-ssl.json).

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -68,44 +80,66 @@ async fn run() {
...
+        {
+            // If the connection URI provides authentication information, manually create the user
+            // before connecting.
+            if let Some(ParsedOptions {
+                user: Some(ref user),
+                password: Some(ref pwd),
+                ref db,
+            }) = test_file.parsed_options
+            {
+                client
```
patrickfreed, line 101: Authentication isn't required for monitoring threads to discover the topology, so I don't think it's necessary that we create these users.

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -68,44 +80,66 @@ async fn run() {
...
-                replica_set: options.repl_set_name.clone(),
-                auth_source: options
-                    .credential
-                    .as_ref()
-                    .and_then(|cred| cred.source.clone()),
-                ssl: options.tls_options().is_some(),
-            };
-
-            assert_eq!(&actual_options, resolved_options);
+            // When an `authSource` is provided without any other authentication information, we do
```
patrickfreed, line 133: I filed DRIVERS-796 a while back to improve this and it was subsequently closed, but clearly it's still a problem for us. I added a comment to it to try to re-open it so we can properly assert here.

src/test/spec/initial_dns_seedlist_discovery.rs
saghm: The test seems to specifically be about having a URL-encoded user and database. Maybe it can use a different character that isn't prohibited anywhere, though.

src/test/spec/initial_dns_seedlist_discovery.rs
isabelatkinson: `TestClient::new()` sends a few commands via `execute_operation_on_session` which goes through the authentication process for URIs that specify a username/password pair; commenting out this step causes a SCRAM failure.

PR #238, opened 2020-08-19
RUST-361 Implement OCSP test harness

patrickfreed:
RUST-361

This PR implements the OCSP testing harness. OCSP client verification is not currently implemented in `rustls` (see: https://github.com/ctz/rustls/issues/31), so any test which requires the driver to fail creating a connection will not pass.

According to the spec, we are required to document that such verification is not implemented. I didn't write this documentation yet since I wasn't sure if we were going to go ahead and try to implement these features anyways. If we think that won't be for a while then I can go ahead and write it now.

Also, since the tests will always be failing, I was wondering if we should just park this on a separate branch or comment them out until we start implementing OCSP verification.


.evergreen/run-ocsp-test.sh
```diff
@@ -0,0 +1,36 @@
...
+#                            given configuration.
+
+echo "Running OCSP test"
+
+# show test output
+set -x
+
+set -o errexit
+
+if [[ "$MONGODB_URI" == *"?"* ]]; then
```
patrickfreed, line 25: this logic didn't exactly line up with `generate-uri.sh` so I just duplicated some of it here instead of making the former more general. My bash is pretty bad so I figured it wasn't worth it, but I'd be happy to consolidate if we do think it's worth it.

src/test/spec/ocsp.rs
```diff
@@ -0,0 +1,47 @@
...
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn run() {
+    let _guard = LOCK.run_exclusively().await;
+
+    let should_succeed = std::env::var("OCSP_TLS_SHOULD_SUCCEED")
+        .unwrap()
+        .to_lowercase();
+
+    let mut options = CLIENT_OPTIONS.clone();
+    let mut tls_options: TlsOptions = options.tls.clone().unwrap().options().cloned().unwrap();
```
patrickfreed, line 20: this process is quite awkward with the current API since we have to go through `Option` -> `Tls` -> `TlsOptions`. I'm not sure what the answer is so I just added a `#[cfg(test)]` accessor and figured we could try to improve this API later.

.evergreen/config.yml
```diff
@@ -339,6 +339,101 @@ functions:
...
+          cd ${DRIVERS_TOOLS}/.evergreen/ocsp
+          ${PYTHON} -m virtualenv ./venv
+          ./venv/${VENV_BIN_DIR}/python -m pip install -r mock-ocsp-responder-requirements.txt
+    - command: shell.exec
+      params:
+        background: true
+        script: |
+          cd ${DRIVERS_TOOLS}/.evergreen/ocsp
+
+          ./venv/${VENV_BIN_DIR}/python ocsp_mock.py \
```
saghm, line 370: I assume these tasks are mostly copied from another driver's config?

.evergreen/config.yml
```diff
@@ -671,6 +1090,10 @@ axes:
         display_name: "latest"
         variables:
            MONGODB_VERSION: "latest"
+      - id: "4.4"
```
saghm, line 1093: Hmm, I thought this was already added; are you fully rebased with the latest master?

.evergreen/config.yml
```diff
@@ -736,17 +1159,28 @@ axes:
       - id: ubuntu-18.04
         display_name: "Ubuntu 18.04"
         run_on: ubuntu1804-test
+        variables:
+          PYTHON: "/opt/mongodbtoolchain/v3/bin/python"
+          VENV_BIN_DIR: "bin"
       - id: ubuntu-16.04
         display_name: "Ubuntu 16.04"
         run_on: ubuntu1604-test
+        variables:
```
saghm, line 1168: We aren't actually using this on 16.04 anywhere in this file, right?

.evergreen/run-ocsp-test.sh
saghm: Honestly, I consider needing to add the TLS options kind of a hack anyways; drivers-evergreen-tools should be outputting the TLS options, but it hasn't been updated to since the URI options spec was written.

src/client/options/mod.rs
```diff
@@ -435,6 +435,16 @@ pub enum Tls {
     Disabled,
 }
 
+impl Tls {
+    #[cfg(test)]
+    pub fn options(&self) -> Option<&TlsOptions> {
```
saghm, line 440: Is there a reason you couldn't use `ClientOptions::tls_options` (or modify it to return a reference, if you wanted to avoid the clone)?

src/test/spec/ocsp.rs
saghm: Is there a reason you couldn't use `ClientOptions::tls_options` (or modify it to return a reference, if you wanted to avoid the clone)? I originally added it for pretty much this exact type of use case

src/client/options/mod.rs
patrickfreed: Oh I didn't see that--yeah that would work nicely here, updated.

src/test/spec/ocsp.rs
patrickfreed: nope, just didn't know that method existed. updated to use that.

.evergreen/config.yml
patrickfreed: They are basically copies of the go driver's tasks, though with some very minor modifications to how python is invoked and whether there is a default value for `$VENV_BIN_DIR`.

.evergreen/config.yml
patrickfreed: Yep, this axis was actually not used by our testing at all (until now). Instead, we have separate task definitions for each of the different server versions. The OCSP ones introduce a ton of new tasks per version so it made sense for them to leverage the axis instead of being bundled together, at least from the `config.yml`'s perspective.

.evergreen/config.yml
patrickfreed: nope. I was wondering if we ought to remove this and run everything on 18.04, since currently we have a bit of a mix.

.evergreen/config.yml
saghm: Ah, right, I misread what this axis was

.evergreen/config.yml
saghm: I believe that the only existing place we use 18.04 (before this PR) is with the AWS auth tests, as it's required for some of the container tests. We use 16.04 for the rest of the variants because there are no 3.6 binaries for 18.04, and it's easier to just use the same Ubuntu version for all of them.

patrickfreed:
Updated to check for the environment variable first, and also commented them out in the `config.yml`.

PR #239, opened 2020-08-19
RUST-366 Support shorter SCRAM conversation

isabelatkinson:


src/client/auth/test.rs
```diff
@@ -70,3 +72,37 @@ async fn negotiate_mangled() {
...
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn sasl_first_options_specified() {
+    scram_sasl_first_options(AuthMechanism::ScramSha1);
+    scram_sasl_first_options(AuthMechanism::ScramSha256);
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg(feature = "tokio-runtime")]
+async fn sasl_first_options_not_specified() {
```
saghm, line 101: Now that X.509 is merged, could you update this to use that instead of AWS so we don't have to make this test tokio-specific?

PR #242, opened 2020-08-28
RUST-544 Ensure correct termination of monitoring tasks

saghm:


src/client/mod.rs
```diff
@@ -82,6 +82,12 @@ struct ClientInner {
     session_pool: ServerSessionPool,
 }
 
+impl Drop for ClientInner {
```
saghm, line 85: We mark the topology as closed when the ClientInner is dropped rather than the Client since there can be multiple copies of the Client

src/lib.rs
```diff
@@ -79,6 +79,7 @@
     )
 )]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(test, type_length_limit = "6000000")]
```
saghm, line 82: This is an unfortunate necessity on Rust 1.46

src/sdam/monitor.rs
```diff
@@ -52,7 +52,11 @@ impl Monitor {
             .heartbeat_freq
             .unwrap_or(DEFAULT_HEARTBEAT_FREQUENCY);
 
-        while let Some(_server) = self.server.upgrade() {
+        while self.topology.is_alive() {
+            if self.server.upgrade().is_none() {
```
saghm, line 56: We still want to break from the loop if the server can't be upgraded (since that means it's removed from the topology), but we need an additional check to see if the Topology is still alive. This is a little strange since we could potentially fail to upgrade the topology immediately after, but that's not a huge deal, as it just means that the Topology was closed between the beginning of the loop and the upgrade, so we're still fine with breaking out of the loop.

src/sdam/srv_polling/mod.rs
```diff
@@ -59,14 +59,21 @@ impl SrvPollingMonitor {
...
+
             let state = topology.clone_state().await;
 
             if state.is_sharded() || state.is_unknown() {
                 let hosts = self.lookup_hosts().await;
-                self.update_hosts(hosts, topology, state).await;
+                self.update_hosts(hosts, topology.clone(), state).await;
             }
 
+            std::mem::drop(topology);
```
saghm, line 75: Unlike in the SDAM monitoring tasks, we don't ever need to wake up early due to messages from server selection, so we don't need the topology again until the next loop. As a slight optimization, we can drop the topology early and potentially help the other background tasks exit a little more quickly.

src/sdam/state/mod.rs
```diff
@@ -47,6 +51,7 @@ pub(crate) struct WeakTopology {
 /// manager and the client options.
 #[derive(Clone, Debug)]
 struct Common {
+    is_alive: Arc<AtomicBool>,
```
saghm, line 54: Even though AtomicBool is atomic, we still need to use an Arc so that changing it on one copy changes it on all the others.

src/sdam/monitor.rs
patrickfreed: I agree that we need to still try to upgrade the topology, but why do we need to do it twice? Can we not just use the upgrade attempt where we keep the topology around?

saghm: I'm not sure I understand what you're asking. `WeakTopology::is_alive` doesn't upgrade it, and the first call to `upgrade` here is on the server, not the topology. We already had separate calls to upgrade for the server and topology; see the left side of the diff.

patrickfreed: Oh, I misread, I thought this was upgrading the topology not the server. That said, I don't think there could be a case where a topology upgrade succeeds but the server upgrade does not, given the topology stores instances of `Arc<Server>`. That said, having the extra layer of redundancy probably doesn't hurt here.

saghm: If the server has been removed from the cluster, then we need to terminate the monitoring thread even if the topology is still around, so this check isn't redundant

patrickfreed: Ah, true. I'll refrain from further suggestions on this, lol.

PR #243, opened 2020-09-01
RUST-371 Implement speculative authentication

saghm:


.evergreen/run-x509-tests.sh
```diff
@@ -13,4 +13,4 @@ export SUBJECT=${SUBJECT#"subject="}
 # Remove any leading or trailing whitespace
 export SUBJECT=`echo "$SUBJECT" | awk '{$1=$1;print}'`
 
-RUST_BACKTRACE=1 MONGO_X509_USER="$SUBJECT" cargo test x509_auth
+RUST_BACKTRACE=1 MONGO_X509_USER="$SUBJECT" cargo test x509
```
saghm, line 16: Since I added a speculative auth test for X.509, I broadened the filter for this script to run any test with `x509` in the name

src/client/auth/mod.rs
```diff
@@ -108,6 +108,13 @@ impl AuthMechanism {
         }
     }
 
+    pub(crate) fn is_scram(&self) -> bool {
```
saghm, line 111: We need to add an extra field to the speculative auth document we send to the server when SCRAM is used, which makes having this helper useful

src/client/auth/mod.rs
```diff
@@ -185,6 +192,43 @@ impl AuthMechanism {
         }
     }
 
+    /// Constructs the first message to be sent to the server as part of the authentication
+    /// handshake, which can be used for speculative authentication.
+    pub(crate) fn build_client_first(
```
saghm, line 197: The strategy I took to implement speculative auth was to separate out the building (and sending) of the first round of the SCRAM and X.509 auth handshakes (as they're the only mechanisms that support this feature). To that end, I defined a separate message to build and return the first client message in the handshake so that it can be put in the handshake.

src/client/auth/mod.rs
```diff
@@ -349,6 +407,42 @@ impl Credential {
     }
 }
 
+/// Contains the first client message sent as part of the authentication handshake.
+pub(crate) enum ClientFirst {
+    Scram(ScramVersion, scram::ClientFirst),
+    X509(Command),
```
saghm, line 413: Since the X.509 handshake is fairly simple, we don't have much of a need for a structured type for this, so I just used Command.

src/client/auth/sasl.rs
```diff
@@ -67,9 +67,11 @@ impl SaslContinue {
 
 /// Validates that a `saslStart` or `saslContinue` command response is successful.
 fn validate_command_success(auth_mechanism: &str, response: &Document) -> Result<()> {
-    let ok = response
-        .get("ok")
-        .ok_or_else(|| Error::invalid_authentication_response(auth_mechanism))?;
+    let ok = match response.get("ok") {
```
saghm, line 70: The server doesn't send back `"ok": 1` in the speculative auth response document, so the simplest solution is to just treat a lack of any `"ok"` field as the same as `"ok": 1`. I left in the check that if the field is present, it must be `1` (or `1.0`).

src/client/auth/scram.rs
```diff
@@ -57,19 +57,32 @@ struct CacheEntry {
 }
 
 /// The versions of SCRAM supported by the driver (classified according to hash function used).
-#[derive(Hash, Eq, PartialEq, Clone)]
+#[derive(Debug, Hash, Eq, PartialEq, Clone)]
```
saghm, line 60: While testing this, I found it useful to have Debug on a struct that contained this

src/client/auth/scram.rs
```diff
@@ -57,19 +57,32 @@ struct CacheEntry {
...
 pub(crate) enum ScramVersion {
     Sha1,
     Sha256,
 }
 
+/// Contains the authentication info derived from a Credential.
+pub(crate) struct ClientAuthInfo<'a> {
```
saghm, line 67: I wanted to abstract out the code that parses and validates these fields, and having a struct seemed better than a 3-tuple

src/client/auth/scram.rs
```diff
@@ -92,15 +105,61 @@ impl ScramVersion {
...
+            server_first,
+        })
+    }
+
+    /// Perform SCRAM authentication for a given stream.
+    pub(crate) async fn authenticate_stream(
+        &self,
+        conn: &mut Connection,
+        credential: &Credential,
+        first_round: impl Into<Option<FirstRound>>,
```
saghm, line 145: I updated the handshakes to optionally take a FirstRound struct that contained both the first client message (which was sent as part of the handshake) as well as the server's response to the speculative authentication attempt. If this is present, the handshake will just skip the first round trip and proceed in the same fashion as it would if it had performed the handshake normally.

src/client/auth/scram.rs
```diff
@@ -304,18 +363,22 @@ fn parse_kvp(str: &str, expected_key: char) -> Result<String> {
...
+pub(crate) struct ClientFirst {
     source: String,
 
     message: String,
 
     gs2_header: Range<usize>,
 
     bare: Range<usize>,
+
+    nonce: String,
```
saghm, line 376: One quirk in adapting the existing handshake algorithm to support speculative auth is that the nonce is generated before the `authenticate_stream` function is called. Since the `ClientFirst` message already needs to be passed in, the simplest way to pass in the nonce is just to piggyback it on the ClientFirst. In terms of space efficiency, this is identical to passing it in as a separate argument, and since the only ways ClientFirst is used is to pass into this module, it's not present in any cases where it's not being used.

src/client/auth/scram.rs
```diff
@@ -304,18 +363,22 @@ fn parse_kvp(str: &str, expected_key: char) -> Result<String> {
...
 }
 
 impl ClientFirst {
-    fn new(source: &str, username: &str, nonce: &str) -> Self {
+    fn new(source: &str, username: &str) -> Self {
+        let nonce = auth::generate_nonce();
```
saghm, line 381: Since the nonce can't just be generated in the same scope where it's used like it was before, there isn't a strong need for the caller of this method to keep ownership of it, so to simplify things I moved the generation inside the method.

src/client/auth/scram.rs
```diff
@@ -346,7 +410,7 @@ impl ClientFirst {
         &self.message[..]
     }
 
-    fn to_command(&self, scram: &ScramVersion) -> Command {
+    pub(super) fn to_command(&self, scram: &ScramVersion) -> Command {
```
saghm, line 413: This method needs to be called by the handshake module now

src/client/auth/x509.rs
```diff
@@ -1,14 +1,12 @@
...
     error::{Error, Result},
     options::Credential,
 };
 
-pub(super) async fn authenticate_stream(
-    conn: &mut Connection,
-    credential: &Credential,
-) -> Result<()> {
+/// Constructs the first client message in the X.509 handshake.
+pub(crate) fn build_client_first(credential: &Credential) -> Command {
```
saghm, line 9: Similar to SCRAM, I made a separate method to build the first client message that can be used for speculative auth. Since we don't need to inspect the contents of this later like we need to with the first SCRAM message, I didn't bother defining a separate type this.

src/client/auth/x509.rs
```diff
@@ -18,13 +16,31 @@ pub(super) async fn authenticate_stream(
...
+) -> Result<()> {
+    let server_response = match server_first.into() {
+        Some(server_first) => server_first,
+        None => send_client_first(conn, credential).await?,
+    };
 
-    if !server_response.is_success()
-        || server_response.raw_response.get_str("dbname") != Ok("$external")
-    {
+    if server_response.raw_response.get_str("dbname") != Ok("$external") {
```
saghm, line 43: Due to the lack of `"ok": 1` in the server's speculative response, the cleanest solution is to just rely on the "dbname" field alone to verify that authentication succeeded, as it won't be present in an `"ok": 0` response

src/cmap/conn/command.rs
```diff
@@ -70,7 +70,6 @@ pub(crate) struct CommandResponse {
 }
 
 impl CommandResponse {
-    #[cfg(test)]
     pub(crate) fn with_document_and_address(source: StreamAddress, doc: Document) -> Self {
```
saghm, line 73: This method is useful in the speculative auth handshake, so I removed the conditional compilation attribute

src/cmap/establish/handshake/mod.rs
```diff
@@ -184,18 +194,82 @@ impl Handshaker {
...
+                    let command_response = CommandResponse::with_document_and_address(
+                        conn.address().clone(),
+                        server_first,
+                    );
+
+                    client_first.into_first_round(command_response)
+                })
+        });
 
-        // TODO RUST-192: Calculate round trip time.
```
saghm, line 191: This TODO should have been done a while ago, so I just did it here. We were properly doing this in the monitoring checks, so it's not a huge deal that we didn't have it here, but it's worth having given that this is the round trip that will be exposed to users who use a predicate SelectionCriteria.

src/lib.rs
```diff
@@ -79,7 +79,7 @@
     )
 )]
 #![cfg_attr(docsrs, feature(doc_cfg))]
-#![cfg_attr(test, type_length_limit = "6000000")]
+#![cfg_attr(test, type_length_limit = "10000000")]
```
saghm, line 82: This is going to be annoying going forward...

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -102,7 +102,7 @@ async fn run() {
                 client
                     .drop_and_create_user(
                         user,
-                        pwd,
+                        pwd.as_str(),
```
saghm, line 105: Unfortunately, even though `&String` can coerce to `&str` and `&str` implements `Into<Option<&str>>`, the compiler doesn't accept `&String` as an `impl Into<Option<&str>>`. I propose a new meme format called "Rust World Problems" (a la "first world problems") to refer to situations like this.

patrickfreed (commented):
Looking really good. The logic all looks right to me; I mostly have small comments about naming things. I had heard this was annoying to integrate into a lot of other drivers' auth flows, but it looks like you've been able to make it quite natural in ours, nice job!

src/client/auth/scram.rs
```diff
@@ -57,19 +57,32 @@ struct CacheEntry {
...
-#[derive(Hash, Eq, PartialEq, Clone)]
+#[derive(Debug, Hash, Eq, PartialEq, Clone)]
 pub(crate) enum ScramVersion {
     Sha1,
     Sha256,
 }
 
+/// Contains the authentication info derived from a Credential.
+pub(crate) struct ClientAuthInfo<'a> {
+    pub(crate) username: &'a str,
```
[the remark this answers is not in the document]
saghm: The auth info is all derived from the Credential, which is passed in by reference to the methods below. To take ownership, we'd have to copy them, and while it's not a huge deal since they should be relatively small, the usage of lifetimes here doesn't complicate things too much, so I figured I would just do the efficient thing.

src/cmap/establish/handshake/mod.rs
```diff
@@ -184,18 +194,82 @@ impl Handshaker {
...
+        None => return Ok(None),
+    };
+
+    let mut client_first_doc = client_first.to_document();
+
+    // Since the first round of the SCRAM handshake isn't being sent as an OP_MSG to the server, the
+    // database containing the user needs to be specified in the document. This isn't an issue for
+    // X.509 due to it always being used with the $external database.
+    if auth_mechanism.is_scram() {
+        client_first_doc.insert("db", credential.resolved_source());
```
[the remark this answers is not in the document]
patrickfreed: Yeah, or also have a separate `ScramVersion::build_speculative_client_first` that is `pub(super)` which calls through to the private one that takes the boolean parameter.

In a similar vein, could we rename `AuthMechanism::build_client_first` method to `AuthMechanism::build_speculative_client_first`? 

These more specific names help differentiate between the ones used internally by the individual mechanisms and the ones needed for speculative auth, which was somewhat difficult for me the first time I looked at this code.

src/client/auth/mod.rs
```diff
@@ -349,6 +407,42 @@ impl Credential {
...
+        }
+    }
+}
+
+/// Contains the complete first round of the authentication handshake, including the client message
+/// and the server response.
+#[derive(Debug)]
+pub(crate) enum FirstRound {
+    Scram(ScramVersion, scram::FirstRound),
+    X509(CommandResponse),
```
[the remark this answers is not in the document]
patrickfreed: That logic can still remain internal to each of the authentication submodules though. I don't think invoking `ServerFirst::parse(doc)` is really any different than invoking `ClientFirst::to_command` (which we do in `set_speculative_auth_info`) in terms of caring about structure or breaking encapsulation.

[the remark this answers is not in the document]
saghm: I think parsing the response is different because it's fallible; it seems a little strange to be failing in the handshake due to the shape of the response rather than in the regular auth flow.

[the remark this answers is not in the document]
patrickfreed: Ok, well at any rate, a  `Document` would also be fine by me here. I just wanted to avoid "emulating" a `CommandResponse`.

src/cmap/establish/test.rs
```diff
@@ -0,0 +1,141 @@
...
+    // We expect that the server will return a response with the `speculativeAuthenticate` field if
+    // and only if it's new enough to support it.
+    assert_eq!(first_round.is_some(), client.server_version_gte(4, 4));
+
+    // Regardless of whether the server supports our speculative authentication attempt, we should
+    // be able to successfully authenticate after the handshake.
+    credential
+        .authenticate_stream(&mut conn, &Default::default(), first_round)
+        .await
+        .unwrap();
```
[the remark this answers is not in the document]
saghm: Updated

src/cmap/establish/handshake/mod.rs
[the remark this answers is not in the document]
saghm: Just pushed the changes you suggested

src/client/auth/mod.rs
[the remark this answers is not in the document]
saghm: Fair enough; I'll change it to Document

src/cmap/establish/test.rs
```diff
@@ -0,0 +1,155 @@
...
+        .build();
+
+    speculative_auth_test(&client, credential, "root", function_name!()).await;
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[function_name::named]
+async fn speculative_auth_scram_sha_256() {
+    let _guard = LOCK.run_concurrently().await;
```
saghm, line 110: I realized that the X.509 test was the only one that conflicted with anything; the SCRAM tests all use different users, so I've switched them to run concurrently

patrickfreed (approved):
LGTM, great job with this!

isabelatkinson (approved):

PR #246, opened 2020-09-02
RUST-526 Exclude unnecessary files

saghm:


Cargo.toml
```diff
@@ -10,6 +10,14 @@ readme = "README.md"
 name = "mongodb"
 version = "1.1.1"
 
+exclude = [
+    "etc/**",
+    "rustfmt.toml",
+    ".evergreen/**",
+    ".gitignore",
+    "src/test/**"
```
isabelatkinson, line 18: Should this also exclude the top-level `tests` directory?

saghm: Oops, I forgot we had one in the driver now that we added the connection snippets. I'll add it to the list here!

isabelatkinson (approved):

patrickfreed (approved):

PR #248, opened 2020-09-09
minor: allow secondaries in command following speculative auth test

saghm:


src/cmap/establish/test.rs
```diff
@@ -59,9 +59,12 @@ async fn speculative_auth_test(
         .await
         .unwrap();
 
-    let command = Command::new(
+    let command = Command::new_read(
```
saghm, line 62: This command was sometimes failing with notMaster errors

src/cmap/establish/test.rs
```diff
@@ -74,7 +77,7 @@ async fn speculative_auth_test(
 #[cfg_attr(feature = "async-std-runtime", async_std::test)]
 #[function_name::named]
 async fn speculative_auth_default() {
-    let _guard = LOCK.run_concurrently().await;
+    let _guard = LOCK.run_exclusively().await;
```
saghm, line 80: Because these tests run on a connection directly and don't go through server selection, they were occasionally running operations against a server that was not ready to receive operations due to other tests going on at the same time. Running these exclusively seems to have fixed that issue.

isabelatkinson (approved):

PR #249, opened 2020-09-10
RUST-485 Unified test runner

isabelatkinson:
This PR implements the unified test runner.


src/test/spec/unified_runner/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+        if let Some(_expect_events) = test_case.expect_events {
+            // TODO implement the spec for command monitoring
+        }
+
+        for fail_point in fail_points {
+            client
+                .database("admin")
+                .run_command(fail_point, None)
+                .await
+                .unwrap();
```
isabelatkinson, line 186: Disable any fail points that were set when running the operations.

src/test/spec/unified_runner/operation.rs
```diff
@@ -0,0 +1,801 @@
...
+    // <https://github.com/rust-lang/rust/issues/51443>
+    async fn execute_on_collection(&self, collection: &Collection) -> Result<Option<Bson>>;
+
+    async fn execute_on_client(&self, client: &EventClient) -> Result<Option<Bson>>;
+
+    async fn execute_on_database(&self, database: &Database) -> Result<Option<Bson>>;
+}
+
+#[derive(Debug)]
+pub struct EntityOperation {
```
isabelatkinson, line 52: This file is a copy of `operation.rs` from the v2 runner -- the only difference is the definition of `EntityOperation` here to accommodate for the new fields defined in the unified runner. I didn't bother trying to make them share code since I anticipate we'll be removing that runner as we port tests over to the unified runner anyway.

src/test/spec/unified_runner/test_event.rs
```diff
@@ -0,0 +1,93 @@
+use crate::{
+    bson::Document,
+    test::{CommandEvent, Matchable},
+};
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub enum TestEvent {
```
isabelatkinson, line 9: Ditto above comment -- this file is a copy of `test_event.rs` with the only change being that test events are camelCased in the new runner format rather than snake_cased.

src/test/util/event.rs
```diff
@@ -164,6 +164,32 @@ impl EventClient {
         EventClient::with_options(options).await
     }
 
+    pub async fn unified_with_additional_options(
```
isabelatkinson: `use_multiple_mongoses` is handled a bit differently in the unified runner (see line 278 in the spec) which required a new method for merging test options with `CLIENT_OPTIONS`.

src/test/spec/unified_runner/entity.rs
```diff
@@ -0,0 +1,55 @@
+use crate::{bson::Bson, test::util::EventClient, Collection, Database};
+
+pub enum Entity {
+    Client(EventClient),
+    Database(Database),
+    Collection(Collection),
+    Result(Option<Bson>),
```
saghm, line 7: Is there a reason to have `None` be a Result instead of a separate variant, e.g. `Result(Bson)` and `None`?

src/test/spec/unified_runner/matcher.rs
```diff
@@ -0,0 +1,235 @@
+use crate::bson::{doc, spec::ElementType, Bson};
+
+pub fn results_match(actual: Option<&Bson>, expected: &Bson) -> bool {
+    match expected {
+        Bson::Document(expected_doc) => {
+            if let Some(special_op) = expected_doc.iter().next() {
```
saghm, line 6: Instead of using `iter()` and then `.0`, you can use `expected_doc.keys()`

src/test/spec/unified_runner/matcher.rs
```diff
@@ -0,0 +1,235 @@
...
+            "int" => actual.element_type() == ElementType::Int32,
+            "timestamp" => actual.element_type() == ElementType::Timestamp,
+            "long" => actual.element_type() == ElementType::Int64,
+            "decimal" => actual.element_type() == ElementType::Decimal128,
+            "minKey" => actual.element_type() == ElementType::MinKey,
+            "maxKey" => actual.element_type() == ElementType::MaxKey,
+            other => panic!("unrecognized type: {}", other),
+        },
+        Bson::Int32(n) => match n {
+            1 => actual.element_type() == ElementType::Double,
```
saghm, line 90: ElementType implements `From<u8>`; using that should eliminate most of these branches

src/test/spec/unified_runner/matcher.rs
```diff
@@ -0,0 +1,235 @@
...
+            true
+        }
+        _ => match actual {
+            Some(actual) => actual == expected,
+            None => false,
+        },
+    }
+}
+
+fn special_operator_matches(special_op: (&String, &Bson), actual: Option<&Bson>) -> bool {
```
saghm, line 46: Instead of using `.0` and `.1`, you can destructure in the parameter list, e.g. `fn special_operator_matches((key, value): (&String, &Bson), ...)`

src/test/spec/unified_runner/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+                    }
+                } else if let Some(expect_error) = operation.expect_error {
+                    let error = result.unwrap_err();
+                    expect_error.verify_result(error);
+                }
+            }
+        }
+
+        if let Some(_expect_events) = test_case.expect_events {
+            // TODO implement the spec for command monitoring
```
saghm, line 178: What's the rationale for not doing this yet?

src/test/spec/unified_runner/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+                insert_initial_data(data, &client).await;
+            }
+        }
+
+        let mut entities: HashMap<String, Entity> = HashMap::new();
+        if let Some(ref create_entities) = test_file.create_entities {
+            populate_entity_map(&mut entities, create_entities).await;
+        }
+
+        let mut fail_points: Vec<Document> = Vec::new();
```
saghm, line 99: Could we name this something that indicates that this contains the commands to disable the failpoints, e.g. `failpoint_disable_commands`?

src/test/spec/unified_runner/matcher.rs
isabelatkinson: It's still useful to get the entire doc from the iterator since the value gets passed into `special_operator_matches`, but I'll update this to de-structure into `(key, value)`.

src/test/spec/unified_runner/matcher.rs
isabelatkinson: Oh cool, I'll update to use that.

src/test/spec/unified_runner/mod.rs
isabelatkinson: I'm working on it right now; we discussed in the team channel that it would be better to get team feedback on the general syntax/structure of the POCs sooner rather than later, so in the interest of getting that feedback now I moved this into review while implementing this piece.

src/test/spec/unified_runner/entity.rs
isabelatkinson: Hm, I hadn't thought about doing that. The purpose of this variant is to store return values from operations as entities so that they can be used later on, so it makes the most sense to me to have one variant to represent that instead making a differentiation between an operation that returns a Bson value and an operation that returns `None`. Since we don't have change streams in the driver, though, it's hard to know what the most convenient representation would be for actually using these values since they're mainly intended for change stream tests.

src/test/spec/unified_runner/matcher.rs
```diff
@@ -86,30 +86,20 @@ fn type_matches(types: &Bson, actual: &Bson) -> bool {
...
-            18 => actual.element_type() == ElementType::Int64,
-            19 => actual.element_type() == ElementType::Decimal128,
-            -1 => actual.element_type() == ElementType::MinKey,
-            127 => actual.element_type() == ElementType::MaxKey,
-            other => panic!("unrecognized type: {}", other),
-        },
+        Bson::Int32(n) => {
+            let n = *n;
+            // -1 cannot be converted to a u8
+            if n == -1 {
```
saghm, line 92: Now that I'm looking at this again, I'm surprised by this check. Is there any reason why the unified spec test format can't just use 0xFF (or 255, if hex isn't supported)? I don't think it should make anything less difficult for languages without strict integer types, and it will make things easier for languages like us that do.

src/test/spec/unified_runner/mod.rs
saghm: Ah, okay, sounds good

src/test/spec/unified_runner/entity.rs
saghm: I might just not be understanding the context for this variant correctly, but I'd think that we'd generally want to execute different code paths when we have `Some(bson_val)` and None. If it's None, I imagine if it's `Some(bson_val)`, we'd presumably be doing something with it, and if it's None, we wouldn't be doing much with it at all. To put it another way, any time we have an Option<Bson>, is the first thing we're going to do with it be checking if it's Some or None and then taking a separate code path? If so, then there might not be much benefit to having them be the same variant.

isabelatkinson: [Q&A on the spec PR](https://github.com/mongodb/specifications/pull/846#discussion_r486568386), a re-read of the `saveResultAsEntity` section made me question when we'd use this functionality in the first place.

From the discussion here, it appears that storing these results is not useful in the current iteration of the spec. I see what you're saying, and if there were some operation introduced in the future it would probably be along the lines of what you described, so I'm open to splitting this out into two variants and can revisit the the design if needed if there is an option introduced that uses them in the future.

src/test/spec/unified_runner/matcher.rs
jmikola: For context, -1 is only used for consistency with https://docs.mongodb.com/manual/reference/operator/query/type/. I imagine u8 is being used here because the rest of the values happen to fall in the range of 1...127, but it'd probably be more straightforward to use a signed 8-bit integer if that's possible.

saghm: Unfortunately we can't change the BSON library itself, as that could break user code. I'm honestly pretty surprised that a signed integer is used anywhere; I've never thought of BSON type tags as signed integers, and I'm not really sure why anyone would ever prefer to do that.

kmahar: FWIW in Swift we also used a `UInt8` as the raw type for for our `BSONType` enum, with 0xFF representing the MinKey case: https://github.com/mongodb/swift-bson/blob/b8d953ae93951e7bee6cde4f7cc6774211378982/Sources/SwiftBSON/BSONValue.swift#L95

we expose this via a `bsonType` property on each BSON value.

this is for consistency with the BSON spec itself and the values it uses for encoded type prefixes. I know libbson similarly has a [`bson_type_t`](http://mongoc.org/libbson/current/bson_type_t.html) enum which uses 0xFF for MinKey.

we could work around this by special-casing as Isabel does here, but I'm curious what the benefit is of going with consistency with the query operator over with how a number of drivers already represent BSON type via integer?

jmikola: I didn't that this was interfacing with your BSON library, and it never occurred to me that the numbers in https://docs.mongodb.com/manual/reference/operator/query/type/ were related to the type codes in http://bsonspec.org/spec.html. In that case, I'd rather remove support for a numeric code in `$$type` and stick to strings.

saghm: That sounds good to me! It will make the spec tests more human-readable for sure

jmikola: https://github.com/mongodb/specifications/pull/846/commits/5480885e819e8affe154b38954e3ae09f36caa5e

isabelatkinson: Removed the `Int32` case here.

src/test/spec/unified_runner/test_file.rs
```diff
@@ -0,0 +1,254 @@
...
+                assert!(error.labels().contains(&label));
+            }
+        }
+        if let Some(error_labels_omit) = self.error_labels_omit {
+            for label in error_labels_omit {
+                assert!(!error.labels().contains(&label));
+            }
+        }
+        if self.expected_result.is_some() {
+            panic!("Bulk write not implemented");
```
patrickfreed, line 251: We do have bulk write by means of `insert_many`. We do not, however, report the partial results of the insert many when it fails (RUST-260), so we don't need to verify `expected_result` matches. Perhaps replace this `if`/`panic` combo with a `TODO: RUST-260` comment?

src/test/spec/unified_runner/matcher.rs
```diff
@@ -0,0 +1,197 @@
...
+pub fn results_match(actual: Option<&Bson>, expected: &Bson) -> bool {
+    match expected {
+        Bson::Document(expected_doc) => {
+            if let Some((key, value)) = expected_doc.iter().next() {
+                if key.starts_with("$$") {
+                    return special_operator_matches((key, value), actual);
+                }
+            }
+            let actual = match actual {
+                Some(actual) => actual,
```
patrickfreed, line 12: maybe we can change this case to `Some(Bson::Document(actual))` and the other case to `_`. This let's us avoid the subsequent unwrap / panic.

src/test/spec/unified_runner/matcher.rs
```diff
@@ -0,0 +1,197 @@
...
+            for (key, value) in expected_doc {
+                if !results_match(actual_doc.get(key), value) {
+                    return false;
+                }
+            }
+            true
+        }
+        Bson::Array(expected_array) => {
+            let actual_array = actual.unwrap().as_array().unwrap();
+            if expected_array.len() != actual_array.len() {
```
patrickfreed, line 27: can the actual array not be a superset of the expected array?

src/test/spec/unified_runner/entity.rs
```diff
@@ -0,0 +1,85 @@
...
+        ignore_events: Option<Vec<String>>,
+    ) -> Self {
+        Self::Client(Client {
+            client,
+            observe_events,
+            ignore_events,
+        })
+    }
+
+    pub fn as_client(&self) -> &EventClient {
```
patrickfreed, line 51: I think it would be more starightforward to return the `Client` type from above here. This would allow us to get rid of the below getters that are client specific. For ease of use, we could also implement `Deref<Target=EventClient>` on the `Client` type.

src/test/spec/unified_runner/entity.rs
```diff
@@ -0,0 +1,85 @@
...
+
+pub enum Entity {
+    Client(Client),
+    Database(Database),
+    Collection(Collection),
+    Result(Bson),
+    None,
+}
+
+pub struct Client {
```
patrickfreed, line 11: to disambiguate from the various other clients we defined, perhaps it would be better to name this `ClientEntity`?

src/test/spec/unified_runner/test_file.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+    Version::parse(&schema_version).map_err(|e| serde::de::Error::custom(format!("{}", e)))
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RunOnRequirement {
+    pub min_server_version: Option<String>,
+    pub max_server_version: Option<String>,
+    pub topology: Option<Vec<String>>,
```
patrickfreed, line 54: these could be better modeled via an enum

src/test/spec/unified_runner/test_file.rs
```diff
@@ -0,0 +1,254 @@
...
+    if schema_version.split('.').count() == 2 {
+        schema_version.push_str(".0");
+    }
+    Version::parse(&schema_version).map_err(|e| serde::de::Error::custom(format!("{}", e)))
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RunOnRequirement {
+    pub min_server_version: Option<String>,
```
patrickfreed, line 52: do these fields need to be `pub`? If this type will be exclusively accessed via `can_run_on`, then the `pub` visibility can be removed.

src/test/spec/unified_runner/test_file.rs
```diff
@@ -0,0 +1,254 @@
...
+    Session(Session),
+    Bucket(Bucket),
+    Stream(Stream),
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Client {
+    pub id: String,
+    pub uri_options: Option<ClientOptions>,
```
patrickfreed, line 95: the derived `Deserialize` won't map perfectly here, since this field corresponds to a list of URI options, several of which might combine to single options on `ClientOptions` (e.g. `w` is a URI option, but it is only a component of the `WriteConcern` option).

We may need a custom deserializer for this field which joins all the URI option pairs into a URI itself and parses that or something.

src/test/spec/unified_runner/operation.rs
```diff
@@ -0,0 +1,801 @@
...
+    operation: Box<dyn TestOperation>,
+    pub name: String,
+    pub object: String,
```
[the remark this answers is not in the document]
isabelatkinson: That won't quite work since the `object` field is the name of a specific entity in the entity map to call the operation on (e.g. something like "client1" or "database5").

src/test/spec/unified_runner/entity.rs
isabelatkinson: Oh that's a cool trick, I originally did it this way because it seemed inconvenient to need to reference the inner field but deref made it work

src/test/spec/unified_runner/operation.rs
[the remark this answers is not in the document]
patrickfreed: What about the following:
```rust
enum OperationObject {
    Entity { id: String },
    TestRunner
}
```
This could ensure we don't accidentally use the "testRunner" string as an entity id.

src/test/spec/unified_runner/entity.rs
patrickfreed: Yeah it's definitely a nifty feature. It's commonly used by pointer types like `Arc` to make them more ergonomic.

src/test/util/mod.rs
```diff
@@ -256,14 +257,15 @@ impl TestClient {
...
-            String::from("sharded")
+            Topology::Sharded
         } else if self.is_replica_set() {
-            String::from("replicaset")
+            Topology::ReplicaSet
         } else {
-            String::from("single")
+            Topology::Single
         }
+        // TODO parse out sharded replica set
```
patrickfreed, line 268: In 3.6+, shards must be replica sets, and since we only support 3.6+ perhaps we only need the single `Sharded` case.

saghm: Are you sure about that? I'm pretty sure I've made a sharded cluster with singleton shards with a recent server version.

patrickfreed: To be honest, I thought I had too, but according to the docs you can't do that in 3.6: https://docs.mongodb.com/manual/core/sharded-cluster-components/. It's possible the docs are incorrect though.

saghm: I just tried it with 4.4, and it works fine. I guess the docs are wrong.

src/test/spec/unified_runner/test_file.rs
```diff
@@ -0,0 +1,323 @@
...
+}
+
+fn default_uri() -> String {
+    std::env::var("MONGODB_URI").unwrap_or_else(|_| "mongodb://localhost:27017/".to_string())
+}
+
+fn deserialize_uri_options_to_uri_string<'de, D>(
+    deserializer: D,
+) -> std::result::Result<String, D::Error>
+where
```
isabelatkinson, line 124: `ClientOptions::parse` can't be used here since it's async, so instead the URI options are deserialized into a URI. This function takes care of merging the options in the existing URI with the test options (which was previously done when creating the client), so when the client gets created it now just needs to pass in this URI to `ClientOptions::parse`.

src/test/spec/unified_runner/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+        }
+
+        let mut entities: HashMap<String, Entity> = HashMap::new();
+        if let Some(ref create_entities) = test_file.create_entities {
+            populate_entity_map(&mut entities, create_entities).await;
+        }
+
+        let mut failpoint_disable_commands: Vec<Document> = Vec::new();
+        for operation in test_case.operations {
+            if operation.object.as_str() == "testRunner" {
```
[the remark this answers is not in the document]
isabelatkinson: Updated to add an `execute_on_test_runner` method.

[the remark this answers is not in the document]
patrickfreed: We're still switching on the type of object when executing the operation, which I don't believe is technically correct. There doesn't seem to be anything in the spec that precludes the "Special Test Operations" (i.e. the test runner ones) from making use of the `exectResult`, `expectError`, or `saveResultAsEntity` fields (or any shared fields in the future, for that matter), so ideally we'd aim to take the same code path for executing both types of operations, regardless of their target, for the sake of forwards compatibility. Right now only the entity operations have logic for using those fields.

To that end, I think instead of separately calling `entity.execute_operation` and `operation.execute_on_test_runner`, we could have a single `Operation::execute(&self, runner: &mut TestRunner) -> Result<Option<Bson>>` that is called regardless of operation type. Then we can remove the early branching on `OperationObject` in the runner and leave that internal to the implementation of `Operation::execute`.

I imagine `TestRunner` could look something like:
```rust
struct TestRunner {
    internal_client: TestClient,
    entities: HashMap<String, Entity>,
    failpoint_disable_commands: Vec<Document>,
}
```
This would also allow operations like `TargetedFailPoint` to update `failpoint_disable_commands` directly, rather than returning it from `execute_on_test_runner`, which I think would be too brittle in the face of future changes.

What do you think?

patrickfreed (commented):
I just realized I forgot to submit the comments I wrote yesterday :facepalm:, sorry for the delay!

src/client/session/test.rs
```diff
@@ -247,11 +247,11 @@ async fn cluster_time_in_commands() {
         options.heartbeat_freq = Some(Duration::from_secs(1000));
         let client = EventClient::with_options(options).await;
 
-        operation(client.clone())
+        operation(client.clone_inner())
```
isabelatkinson, line 250: Implementing `Clone` on `TestClient` broke the deref that was cloning the inner client directly here, so I added `clone_inner` to do it manually.

src/test/spec/unified_runner/operation.rs
```diff
@@ -30,30 +30,17 @@ use crate::{
...
-
-    async fn execute_on_test_runner(
+    async fn execute<'a>(
         &self,
-        _client: &TestClient,
-        _entities: &HashMap<String, Entity>,
-    ) -> Option<Document>;
+        object: &OperationObject,
+        test_runner: &'a mut TestRunner,
+    ) -> Result<Option<Entity>>;
```
isabelatkinson, line 43: Consolidated everything into one `execute` that takes in a `TestRunner` function per @patrickfreed's suggestion; this worked out nicely in cutting down the code duplication required due to issue with providing default implementations.

src/test/spec/unified_runner/entity.rs
```diff
@@ -1,15 +1,29 @@
...
 pub enum Entity {
     Client(ClientEntity),
     Database(Database),
     Collection(Collection),
-    Result(Bson),
+    Bson(Bson),
     None,
 }
 
+impl fmt::Debug for Entity {
```
patrickfreed: Why not use the derived `Debug` impl?

src/test/spec/unified_runner/mod.rs
```diff
@@ -36,6 +36,109 @@ lazy_static! {
...
+            self.internal_client
+                .database("admin")
+                .run_command(disable_command.clone(), None)
+                .await
+                .unwrap();
+        }
+    }
+
+    pub async fn populate_entity_map(&mut self, create_entities: &[TestFileEntity]) {
+        for entity in create_entities {
```
patrickfreed, line 79: I don't think this will really be a problem, but we probably should clear out the entity map before populating it to ensure each run has completely new state.

patrickfreed (commented):
Looks really good! I only have a few minor comments.

src/test/spec/unified_runner/entity.rs
isabelatkinson: I think this is from an old commit range; I deleted this in the most recent commit

patrickfreed: Whoops, my bad.

src/client/session/test.rs
patrickfreed: I actually just ran into this myself, and I think a simpler solution would be to just update these tests to use `EventClient` instead of `Client` for the generic parameter `F`. I think they were only accidentally relying on this cloning-into-`Client` behavior anyways.

Cargo.toml
```diff
@@ -39,6 +39,7 @@ hex = "0.4.0"
 hmac = "~0.7.1"
 lazy_static = "1.4.0"
 md-5 = "0.8.0"
+num = "0.3.0"
```
isabelatkinson, line 42: This dependency adds functionality to compare numbers of different types that might result in lossy conversion (as required [here](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#flexible-numeric-comparisons)).

src/test/spec/unified_runner/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+            &test_file.schema_version
+        );
+    }
+
+    let mut test_runner = TestRunner::new().await;
+
+    if let Some(requirements) = test_file.run_on_requirements {
+        let mut can_run_on = false;
+        for requirement in requirements {
+            if requirement.can_run_on(&test_runner.internal_client).await {
```
isabelatkinson, line 199: I previously had this using the iterator `any` method but changing `can_run_on` to be `async` broke it -- is there a better way to do this?

src/test/spec/unified_runner/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+                test_runner.insert_initial_data(data).await;
+            }
+        }
+
+        if let Some(ref create_entities) = test_file.create_entities {
+            test_runner.populate_entity_map(create_entities).await;
+        }
+
+        for operation in test_case.operations {
+            match operation.object {
```
isabelatkinson, line 253: The spec was [updated](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#operation) to disallow the save/expect fields on test runner operations, so I reverted this back to using two different cases.

isabelatkinson:
@saghm @patrickfreed this PR is ready for review again

Cargo.toml
saghm: Is just just needed for tests? If so, we can put it in `dev-dependencies` so that cargo can skip downloading/compiling it for users of the driver.

src/client/session/test.rs
```diff
@@ -319,7 +319,7 @@ async fn cluster_time_in_commands() {
 async fn session_usage() {
     let _guard: RwLockReadGuard<()> = LOCK.run_concurrently().await;
 
-    let client = TestClient::new().await;
+    let client = EventClient::new().await;
```
saghm, line 322: Why was this changed?

src/test/spec/mod.rs
```diff
@@ -63,6 +67,11 @@ where
         let json: Value =
             serde_json::from_reader(File::open(test_file_full_path.as_path()).unwrap()).unwrap();
 
+        println!(
```
saghm, line 70: Was this left in by accident?

src/test/spec/unified_runner/mod.rs
saghm: The only way I can think of is to fold into a `Future<Output=bool>` and then `await`, but I think the way you have it here is cleaner.

src/test/spec/unified_runner/test_file.rs
```diff
@@ -0,0 +1,379 @@
...
+                return false;
+            }
+        }
+        true
+    }
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub enum Entity {
```
saghm, line 97: It's a little confusing that there are two things called `Entity`. Could they be named something more specific to make it clear which is which, e.g. `EntityEntry` for the one stored in the map?

src/test/spec/unified_runner/test_file.rs
```diff
@@ -0,0 +1,379 @@
...
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct Stream {
+    pub id: String,
+    pub hex_bytes: String,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct CollectionOrDatabaseOptions {
+    // TODO properly implement Deserialize for ReadConcern
```
saghm, line 213: Do we have a ticket for this, or will it be done as part of this work?

src/test/spec/unified_runner/test_file.rs
```diff
@@ -0,0 +1,379 @@
...
+    pub read_concern: Option<ReadConcern>,
+    #[serde(rename = "readPreference")]
+    pub selection_criteria: Option<SelectionCriteria>,
+    pub write_concern: Option<WriteConcern>,
+}
+
+impl CollectionOrDatabaseOptions {
+    pub fn as_database_options(&self) -> DatabaseOptions {
+        let mut database_options = DatabaseOptions::builder().build();
```
saghm, line 222: Is there a reason you didn't just use the builder API here? i.e.

```rust
DatabaseOptions::builder()
    .read_concern(self.read_concern.clone())
    .selection_critiera(self.selection_criteria.clone())
    .write_concern(self.write_concern.clone())
    .build()
```

Since this is within the driver, you could also just use a struct literal if you prefer that.

saghm (commented):
I have a few style comments, but overall this looks really good!

src/coll/options.rs
```diff
@@ -49,6 +49,20 @@ pub enum ReturnDocument {
     Before,
 }
 
+impl<'de> Deserialize<'de> for ReturnDocument {
```
patrickfreed, line 52: I think this could be achieved via the derived `Deserialize` implementation and `#[serde(rename_all="lowercase")]`

src/test/spec/json/unified-runner-examples/poc-crud.json
```diff
@@ -0,0 +1,446 @@
+{
```
patrickfreed, line 1: do we still need this example files? or are they now included in their respective spec folders (e.g. crud)?

src/test/spec/unified_runner/matcher.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+}
+
+fn numbers_match(actual: &Bson, expected: &Bson) -> bool {
+    if actual.element_type() == expected.element_type() {
+        return actual == expected;
+    }
+
+    match (actual, expected) {
+        (Bson::Int32(actual), Bson::Int64(expected)) => i32_matches_i64(*actual, *expected),
```
patrickfreed, line 94: because each of these pairs always includes an integer, we can use `bson_util::get_int` to reduce some of the boilerplate here.

patrickfreed (commented):
Did a quick pass and had a few minor questions / suggestions, will do a more thorough review tomorrow.

src/test/spec/json/unified-runner-examples/poc-crud.json
isabelatkinson: Yep, these are the only tests that have been written so far for what we have implemented in the driver, so until the rest of the tests are ported we need to keep these around to ensure the runner is working.

src/coll/options.rs
saghm: This is a little tedious, but it's possible to specify multiple alternate names with [`alias`](https://serde.rs/field-attrs.html#alias). I don't think we'd want to use it for names longer than one word, but for this case it might be a reasonable choice.

src/test/spec/mod.rs
saghm: Sounds good. Could you add a comment in the code explaining this?

src/test/spec/unified_runner/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+                    } else if let Some(expect_error) = operation.expect_error {
+                        let error = result
+                            .expect_err(&format!("{} should return an error", operation.name));
+                        expect_error.verify_result(error);
+                    }
+                }
+            }
+        }
+
+        test_runner.disable_fail_points().await;
```
patrickfreed, line 309: If a test panics before this is run, the failpoints won't be disabled. To deal with that, we can use the (newly introduced since this PR was opened) `FailPoint` and `FailPointGuard` types by storing the `FailPointGuard`s instead of just the disable-failpoint-command documents. Their `Drop` implementations (which disable the failpoints) should get invoked even in the event of a panic. In the normal case, we could just drop them all manually via `disable_fail_points`.

As part of that, we'd also need a `Deserialize` implementation for `FailPoint`, but since that type is just a wrapper around a document that should be pretty easy. Also, when using the `FailPointGuard` type, the tokio tests need to use the threaded scheduler since the `Drop` implementations use `RUNTIME.block_on` ([example](https://github.com/mongodb/mongo-rust-driver/blob/master/src/test/spec/retryable_reads.rs#L5)).

patrickfreed (commented):
Overall looks really good, just have a few comments on the matching bit and on failpoints.

src/test/util/failpoint.rs
```diff
@@ -54,21 +61,39 @@ pub struct FailPointGuard {
     failpoint_name: String,
 }
 
+impl FailPointGuard {
+    pub async fn disable(&self) {
```
patrickfreed, line 65: Typically guard types are just dropped to invoke their "disable" functionality, and any direct methods on them like this are actually to disable the _guard_, not to disable or release whatever the guard is holding on to (e.g. see [`disarm`](https://docs.rs/futures-intrusive/0.3.1/futures_intrusive/sync/struct.GenericSemaphoreReleaser.html#method.disarm) on the `futures_intrusive` semaphore releaser type). Given that, I'd suggest just going back to only a `Drop` implementation and doing `fail_point_guards.clear()` in the test runner to drop all the fail point guards rather than iterating over them and manually disabling their failpoints.

src/error.rs
```diff
@@ -1,5 +1,5 @@
 //! Contains the `Error` and `Result` types that `mongodb` uses.
-
+//
```
patrickfreed, line 2: i think this was accidentally left in

isabelatkinson: deleted

src/test/util/failpoint.rs
isabelatkinson: Ah makes sense, I didn't realize `clear` would call `drop` -- I'll revert.

patrickfreed (approved):
LGTM! great job with this, especially with handling all the moving parts.

BTW: the test failures are due to some DNS changes by the build team and are unrelated. I ran a patch with the master branch to confirm and got them there as well.

saghm (approved):

PR #250, opened 2020-09-11
RUST-533 Add sync ClientOptions::parse

saghm:
As part of this work, I ensured that the documentation would render both the sync and async versions of the method. Unfortunately I could not figure out a way to get it to show the `feature="sync"` tag on the sync versions of the methods added, but I did add info in the doc comments for all of the sync/async methods that are duplicated explaining when they would be present and linking to their corresponding method. You can view what the docs will look like (and try out the links) by running `cargo +nightly rustdoc -p mongodb --open -- --cfg docsrs` with this branch checked out.

### DIFF
--- a/src/client/auth/mod.rs
+++ b/src/client/auth/mod.rs
@@ -319,7 +319,7 @@ pub struct Credential {
 }
 
 impl Credential {
-    #[cfg(test)]
+    #[cfg(all(test, not(feature = "sync")))]
     pub(crate) fn into_document(mut self) -> Document {
         use crate::bson::Bson;
 
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -95,7 +95,7 @@ impl Client {
     /// See the documentation on
     /// [`ClientOptions::parse`](options/struct.ClientOptions.html#method.parse) for more details.
     pub async fn with_uri_str(uri: &str) -> Result<Self> {
-        let options = ClientOptions::parse(uri).await?;
+        let options = ClientOptions::parse_uri(uri, None).await?;
 
         Client::with_options(options)
     }
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -293,7 +293,7 @@ impl Client {
         }
     }
 
-    #[cfg(test)]
+    #[cfg(all(test, not(feature = "sync")))]
     pub(crate) async fn get_hosts(&self) -> Vec<String> {
         let servers = self.inner.topology.servers().await;
         servers
--- a/src/sync/test.rs
+++ b/src/sync/test.rs
@@ -1,7 +1,17 @@
+use pretty_assertions::assert_eq;
+
 use crate::{
     bson::{doc, Document},
     error::{CommandError, ErrorKind, Result},
-    options::{Acknowledgment, CollectionOptions, DatabaseOptions, FindOptions, WriteConcern},
+    options::{
+        Acknowledgment,
+        ClientOptions,
+        CollectionOptions,
+        DatabaseOptions,
+        FindOptions,
+        StreamAddress,
+        WriteConcern,
+    },
     sync::{Client, Collection},
     test::CLIENT_OPTIONS,
 };
@@ -19,6 +29,23 @@ pub fn drop_collection(coll: &Collection) {
             e.unwrap();
         }
     };
+}
+
+#[test]
+fn client_options() {
+    let mut options = ClientOptions::parse("mongodb://localhost:27017/").unwrap();
+
+    options.original_uri.take();
+
+    assert_eq!(
+        options,
+        ClientOptions::builder()
+            .hosts(vec![StreamAddress {
+                hostname: "localhost".into(),
+                port: Some(27017)
+            }])
+            .build()
+    );
 }
 
 #[test]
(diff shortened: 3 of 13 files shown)


src/client/auth/mod.rs
```diff
@@ -319,7 +319,7 @@ pub struct Credential {
 }
 
 impl Credential {
-    #[cfg(test)]
+    #[cfg(all(test, not(feature = "sync")))]
```
saghm, line 322: The connection string and URI options tests make heavy use of `ClientOptions::parse`. Rather than adding conditional compilation to the tests themselves when needed, it was simpler to just disable those tests for the `sync` feature, as we [only run the `sync` module tests](https://github.com/mongodb/mongo-rust-driver/blob/b0e59f9fde2e5bc3df3217dec05ffbca553d1b8e/.evergreen/run-async-std-tests.sh#L15) with the `sync` feature anyhow.

This method (and a few others) are only used in the connection string/URI options tests, so I updated the conditional compilation to omit them if the sync feature is enabled.

src/client/mod.rs
```diff
@@ -95,7 +95,7 @@ impl Client {
     /// See the documentation on
     /// [`ClientOptions::parse`](options/struct.ClientOptions.html#method.parse) for more details.
     pub async fn with_uri_str(uri: &str) -> Result<Self> {
-        let options = ClientOptions::parse(uri).await?;
+        let options = ClientOptions::parse_uri(uri, None).await?;
```
saghm, line 98: Since both the sync and async versions of `ClientOptions::parse` both are thin wrappers around `ClientOptions::parse_uri`, calling that directly from within the driver seems simpler than adding conditional compilation everywhere.

src/sync/test.rs
```diff
@@ -1,7 +1,17 @@
...
     bson::{doc, Document},
     error::{CommandError, ErrorKind, Result},
-    options::{Acknowledgment, CollectionOptions, DatabaseOptions, FindOptions, WriteConcern},
+    options::{
+        Acknowledgment,
+        ClientOptions,
+        CollectionOptions,
+        DatabaseOptions,
+        FindOptions,
+        StreamAddress,
```
saghm, line 12: The difficulty I had in parsing this to figure out that the difference was `StreamAddress` being added makes me wonder if it's worth updating our rustfmt config so that it always puts each item being imported on a separate line.

src/test/atlas_connectivity.rs
```diff
@@ -13,13 +13,9 @@ async fn run_test(uri_env_var: &str, resolver_config: Option<ResolverConfig>) {
...
 
     let uri_string = uri.to_string_lossy();
-    let options = match resolver_config {
-        Some(resolver_config) => {
-            ClientOptions::parse_with_resolver_config(uri_string.as_ref(), resolver_config).await
-        }
-        None => ClientOptions::parse(uri_string.as_ref()).await,
-    }
-    .expect("uri parsing should succeed");
+    let options = ClientOptions::parse_uri(uri_string.as_ref(), resolver_config)
```
saghm, line 16: In this case, using `parse_uri` is actually way more readable than calling the public methods; either that method didn't exist when I originally wrote these tests, or I forgot that it did.

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -52,13 +52,14 @@ async fn run() {
             }
         }
 
-        let result = if cfg!(target_os = "windows") {
-            ClientOptions::parse_with_resolver_config(&test_file.uri, ResolverConfig::cloudflare())
-                .await
+        let config = if cfg!(target_os = "windows") {
```
saghm, line 55: Passing in the Option<ResolverConfig>` to `parse_uri` is cleaner here too IMO

src/test/atlas_connectivity.rs
patrickfreed: I'd personally prefer if the tests used the public API as much as possible, even if they are just thin wrappers. If we're not running the async tests as part of the sync tests, why not just conditionally compile the entire test module? That way we can continue to use the public `parse` without per-test conditional compilation.

saghm: Updated to disable the module when sync is enabled

saghm: Turned out that didn't quite work due to test module functionality being used outside of the module; I've pushed again with more selective disabling

src/sync/test.rs
isabelatkinson: I'm in favor, I always find that format much easier to read

isabelatkinson (commented):
It looks like there are CI failures on async-std -- maybe there are still some issues with the conditional compilation?

src/sync/test.rs
saghm: Any objections @patrickfreed?

saghm:
Nope, turns out it's the type length limit hitting us again...

src/sync/test.rs
patrickfreed: sgtm

patrickfreed (commented):
Code and docs LGTM, though it seems we still have issues with async-std.

patrickfreed (approved):

isabelatkinson (approved):

PR #251, opened 2020-09-12
RUST-542 Release lock on ConnectionPool while establishing connections

patrickfreed:
RUST-542

This PR fixes the `check_out` and background thread implementations to release the lock on the pool while connections are being established. This will allow the driver to create multiple connections concurrently, which may lead to performance improvements.

This PR also updates the terminology used in the driver to match the changes made in SPEC-1412. (https://github.com/mongodb/specifications/pull/848)


src/cmap/background.rs
```diff
@@ -24,25 +27,24 @@ pub(crate) fn start_background_task(pool: Weak<ConnectionPoolInner>) {
...
+    pool.inner.remove_perished_connections().await;
+    pool.inner.ensure_min_connections().await;
 }
 
-impl ConnectionPool {
+impl ConnectionPoolInner {
     /// Iterate over the connections and remove any that are stale or idle.
     async fn remove_perished_connections(&self) {
-        let mut connection_manager = self.inner.connection_manager.lock().await;
+        let mut available_connections = self.available_connections.lock().await;
```
patrickfreed, line 37: I imagine this process will be really quick, so I just left it as holding the lock for the duration.

src/cmap/background.rs
```diff
@@ -53,13 +55,31 @@ impl ConnectionPool {
...
+                            // between the count check and our attempt to reserve a spot. In that
+                            // case we just return early because those threads that did enter the
+                            // queue will ensure min_pool_size is fulfilled.
+                            return;
+                        }
+                        Some(handle) => handle,
+                    };
+
+                    let connection = self.create_pending_connection();
+                    match self.establish_connection(connection).await {
```
patrickfreed, line 78: this could be improved to be done in the background as well (with respect to other establishments required to meet minPoolSize), but I figured it probably didn't buy us much.

src/cmap/mod.rs
```diff
@@ -51,144 +58,35 @@ impl From<Arc<ConnectionPoolInner>> for ConnectionPool {
     }
 }
 
-/// A struct used to manage the creation, closing, and storage of connections for a
-/// `ConnectionPool`.
-#[derive(Debug)]
-struct ConnectionManager {
```
patrickfreed, line 57: I ended up removing just removing this type and consolidating everything back onto `ConnectionPoolInner`, with the `Vec` of available connections being the only thing guarded by the lock now. This is basically how it originally was before I added the `ConnectionManager`. The initial goal of `ConnectionManager` was to simplify the considerations around concurrency and state updates of the pool, but it ended up being too restrictive to be useful. I also think that now that the spec has been updated to be more clear about these state changes, the simplicity that `ConnectionManager` got us isn't as worthwhile. (e.g. use of `PendingConnection` helps describe these state changes at the type level).

src/cmap/test/event.rs
```diff
@@ -1,29 +1,27 @@
...
 
 #[derive(Clone, Debug)]
 pub struct EventHandler {
-    ignore: Vec<String>,
-    pub events: Arc<RwLock<Vec<Event>>>,
+    pub events: Arc<RwLock<VecDeque<Event>>>,
 }
 
 impl EventHandler {
-    pub fn new(ignore: Vec<String>) -> Self {
```
patrickfreed, line 14: According to the CMAP spec:
>It is important to note that the ignore list is used for calculating actualEvents, but is NOT used for the waitForEvent command

This was causing us to fail some of the newer spec tests, so I fixed it to only do this at the end.

src/cmap/mod.rs
```diff
@@ -405,13 +380,15 @@ impl Drop for ConnectionPoolInner {
...
-        let connection_manager = self.connection_manager.clone();
+        let available_connections =
+            std::mem::replace(&mut self.available_connections, Mutex::new(Vec::new()));
         let event_handler = self.event_handler.clone();
 
         RUNTIME.execute(async move {
-            let mut connection_manager = connection_manager.lock().await;
-            while let Some(connection) = connection_manager.checked_in_connections.pop() {
-                connection_manager.close_connection(connection, ConnectionClosedReason::PoolClosed);
+            // this lock attempt will always succeed.
```
saghm, line 388: Can we add "immediately" or something to this comment to indicate that the lock won't be held by anything else? "succeed" alone sounds like it's staying it won't return an error, which is evident from looking at it

src/cmap/test/event.rs
saghm: Oh, interesting, I hadn't realized that

saghm (commented):
I haven't finished reviewing the whole thing yet, but I have some feedback on the first part I've looked at, so I figured I'd post it now rather than delaying until tomorrow

src/cmap/background.rs
patrickfreed: Good idea! Updated to quit early when encountering a non-idle, non-stale connection. Also updated to release the lock between iterations.

src/cmap/mod.rs
```diff
@@ -388,14 +293,84 @@ impl ConnectionPoolInner {
...
 
+        // drop the lock before we create a new connection to allow other connections to stop
+        // waiting.
+        drop(available_connections_lock);
+
         // There are no connections in the pool, so open a new one.
-        let connection = connection_manager.create_connection().await?;
-        wait_queue_handle.disarm();
-        Ok(connection)
+        let pending_connection = self.create_pending_connection();
```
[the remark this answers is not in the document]
patrickfreed: I'm not sure it really matters, so I just opted to drop the lock as soon as possible. Did you have a case in mind where creating it after would cause issues?

src/cmap/mod.rs
patrickfreed: hah, good point, fixed.

src/cmap/background.rs
```diff
@@ -24,27 +27,46 @@ pub(crate) fn start_background_task(pool: Weak<ConnectionPoolInner>) {
...
+                    ConnectionClosedReason::Stale,
+                ))
+            } else if connection.is_idle(self.max_idle_time) {
+                Some((
+                    available_connections.remove(0),
+                    ConnectionClosedReason::Idle,
+                ))
             } else {
-                i += 1;
+                None
```
saghm, line 58: I think you could simplify this by breaking here; then you wouldn't need `close_request` to be an Option.

src/cmap/mod.rs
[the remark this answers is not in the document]
saghm: If the lock is dropped beforehand, theoretically maxPoolSize could be reached between when the lock is dropped and when the pending connection is created, leading us to exceed maxPoolSize. More generally, changing the total connection count without holding the lock will potentially be susceptible to races.

src/cmap/test/event.rs
```diff
@@ -1,29 +1,27 @@
...
 
 use serde::{de::Unexpected, Deserialize, Deserializer};
 
 use crate::{event::cmap::*, options::StreamAddress};
 
 #[derive(Clone, Debug)]
 pub struct EventHandler {
-    ignore: Vec<String>,
-    pub events: Arc<RwLock<Vec<Event>>>,
+    pub events: Arc<RwLock<VecDeque<Event>>>,
```
saghm, line 12: Is there a reason this was switched from Vec to VecDeque? I can't find any changes below that seem like they'd require it.

src/cmap/test/integration.rs
```diff
@@ -58,3 +63,84 @@ async fn acquire_connection_and_send_command() {
...
 }
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn concurrent_connections() {
+    let _guard = LOCK.run_exclusively().await;
+
+    let client = TestClient::new().await;
+    let version = VersionReq::parse(">= 4.2.9").unwrap();
+    // blockConnection failpoint option only added in 4.4.
```
saghm, line 74: Based on this comment, shouldn't it be checking `>= 4.4.0`? 4.4 is already released, so we don't have to worry about 4.3.x or rc's anymore, and assuming the failpoint option isn't being backported to 4.2.x, this will fail if/when 4.2.10 comes out.

src/cmap/test/integration.rs
```diff
@@ -58,3 +63,84 @@ async fn acquire_connection_and_send_command() {
...
+    let client_options = CLIENT_OPTIONS.clone();
+    let mut options = ConnectionPoolOptions::from_client_options(&client_options);
+    options.event_handler = Some(handler.clone() as Arc<dyn crate::cmap::CmapEventHandler>);
+    let pool = ConnectionPool::new(
+        CLIENT_OPTIONS.hosts[0].clone(),
+        Default::default(),
+        Some(options),
+    );
+
+    let mut tasks: Vec<AsyncJoinHandle<()>> = Default::default();
```
saghm, line 104: I think you can do this using `map` on the range; just don't put `move` in front of the closure being passed in, and then `pool` will only be borrowed, allowing it to be cloned each invocation.

src/cmap/test/integration.rs
```diff
@@ -58,3 +63,84 @@ async fn acquire_connection_and_send_command() {
...
+        tasks.push(
+            RUNTIME
+                .spawn(async move {
+                    pool_clone.check_out().await.unwrap();
+                })
+                .unwrap(),
+        );
+    }
+
+    while let Some(handle) = tasks.pop() {
```
saghm, line 116: I think [`join_all`](https://docs.rs/futures/0.3.5/futures/future/fn.join_all.html) would be best here.

src/cmap/test/integration.rs
```diff
@@ -58,3 +63,84 @@ async fn acquire_connection_and_send_command() {
...
+        let _ = handle.await;
+    }
+
+    // ensure all three ConnectionCreatedEvents were emitted before one ConnectionReadyEvent.
+    let mut events = handler.events.write().unwrap();
+    let mut consecutive_creations = 0;
+    while let Some(event) = events.pop_front() {
```
saghm, line 123: You can use a for loop with [`drain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain) here.

src/cmap/background.rs
```diff
@@ -53,13 +75,31 @@ impl ConnectionPool {
...
+                            // queue will ensure min_pool_size is fulfilled.
+                            return;
+                        }
+                        Some(handle) => handle,
+                    };
+
+                    let connection = self.create_pending_connection();
+                    match self.establish_connection(connection).await {
+                        Ok(mut connection) => {
+                            let mut available_connections = self.available_connections.lock().await;
```
saghm, line 100: Similar to my comment below about accessing the connection count without holding the lock, I'm starting to wonder if we're opening ourselves up to races in the way that we're using the connection count. We check if we're below minPoolSize above, but we're not holding the lock in the meantime. By the time we actually add the connection to the pool, we've taken the time to not only create the connection but perform the handshake and potentially authenticate as well, so it seems not unlikely that the connection count could have changed between now and then. While it's not necessarily a problem if we populate past minPoolSize, I think we should be checking maxPoolSize while holding the lock before to ensure we don't go over it when adding the connection to the set of available ones.

saghm (commented):
I've finished reviewing the entire PR now; here are the rest of the comments

src/cmap/background.rs
patrickfreed: done

src/cmap/background.rs
patrickfreed: Similar situation here to the other one: we're holding a wait queue handle so we're guaranteed to not exceed maxPoolSize. We definitely could go over minPoolSize, but I didn't really see that as an issue.

src/cmap/test/event.rs
patrickfreed: It's to allow for quicker popping from the front when iterating over the events in order. In fact, we do the same when we purge connections, so I've updated the `available_connections` to be a `VecDeque` as well.

src/cmap/test/integration.rs
patrickfreed: Whoops, this comment is incorrect. It was added in 4.2.9 (see: https://jira.mongodb.org/browse/SERVER-41070). Updated the comment to match the version requirement.

src/cmap/test/integration.rs
patrickfreed: yep, done.

src/cmap/test/integration.rs
patrickfreed: done

src/cmap/test/integration.rs
patrickfreed: fixed.

src/cmap/test/event.rs
saghm: This isn't necessary for the events anymore now that you switched to using `drain` though, right?

patrickfreed: We don't actually even need the `drain` there, we can just iterate over the events without removing any. Reverted to regular `Vec`.

saghm (approved):
Nice work on this!

isabelatkinson (approved):

PR #253, opened 2020-09-23
RUST-197 Add plain authentication

saghm:


src/test/client.rs
```diff
@@ -602,3 +602,49 @@ async fn x509_auth() {
         .await
         .unwrap();
 }
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[function_name::named]
+async fn plain_auth() {
+    let _guard: RwLockReadGuard<_> = LOCK.run_concurrently().await;
```
saghm, line 610: This test is based on the "Testing LDAP" page in our wiki. It uses the credentials listed as being safe to share externally. However, since the domain we connect to is only reachable from within the office or the VPN, the test skips unless a specific environmental variable is set.

patrickfreed (approved):
one minor nit in the config otherwise LGTM

.evergreen/config.yml
```diff
@@ -1236,6 +1251,18 @@ buildvariants:
...
   tasks:
     - "test-x509-auth"
+
+- matrix_name: "plain-auth"
+  matrix_spec:
+    os:
+      - ubuntu-16.04
+      - macos-10.14
+      - windows-64-vs2017-small
+    auth-and-tls: "auth-and-tls"
```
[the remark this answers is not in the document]
saghm: Ah, right. Sorry for the copypasta

src/client/auth/mod.rs
```diff
@@ -140,6 +139,23 @@ impl AuthMechanism {
 
                 Ok(())
             }
+            AuthMechanism::Plain => {
+                if credential.username.is_none() {
```
isabelatkinson, line 143: The spec indicates that the username needs to be non-zero length -- is that something we should be validating here as well?

saghm: Oh, I missed that; we should definitely be validating it. Good catch!

isabelatkinson (approved):

PR #254, opened 2020-09-24
RUST-325 Implement hedged read support

saghm:


src/selection_criteria.rs
```diff
@@ -126,6 +126,30 @@ pub struct ReadPreferenceOptions {
     /// specified for an operation, the operation will return an error.
     #[builder(default)]
     pub max_staleness: Option<Duration>,
+
+    /// Specifies hedging behavior for reads.
+    ///
+    /// See the [MongoDB docs](https://docs.mongodb.com/manual/core/read-preference-hedge-option/) for more details.
+    #[builder(default)]
+    pub hedge: Option<HedgedRead>,
```
saghm, line 134: Based on discussion back when this option was originally added to the spec, I believe the consensus is that we need to be future-compatible with any new options that might be added to the document value under the `hedge` field. To that end, I made a struct called `HedgedRead` (although I welcome feedback on the name, as I'm not super thrilled with anything I came up with) and marked it as non-exhaustive.

src/selection_criteria.rs
```diff
@@ -126,6 +126,30 @@ pub struct ReadPreferenceOptions {
...
+#[non_exhaustive]
+pub struct HedgedRead {
+    /// Whether or not to allow reads from a sharded cluster to be "hedged" across two replica
+    /// set members per shard, with the results from the first response received back from either
+    /// being returned.
+    pub enabled: bool,
+}
+
+impl HedgedRead {
+    pub fn with_enabled(enabled: bool) -> Self {
```
saghm, line 150: Given that there's currently only a single field and it's required, forcing users to do `HedgedRead::builder().enabled(true).build()` felt like a bit too much, so I added a constructor here. Given that any future fields added to the `hedge` document will not be required due to backwards compatibility, I think this is a safe API to add.

src/selection_criteria.rs
```diff
@@ -216,9 +250,34 @@ impl ReadPreference {
...
+
         doc
     }
 }
 
 /// A read preference tag set. See the documentation [here](https://docs.mongodb.com/manual/tutorial/configure-replica-set-tag-sets/) for more details.
 pub type TagSet = HashMap<String, String>;
+
+#[cfg(test)]
+mod test {
```
saghm, line 265: No new spec tests were added as part of the work for the hedged read option, so I just made a minimal test to ensure that the hedge document is correctly added when serializing the ReadPreference. I didn't think it was necessary to add any integration tests (e.g. with command monitoring) due to the fact that `ReadPreference::into_document` is really the only part of this module that's used for the other components that would be tested by that.

src/selection_criteria.rs
isabelatkinson: The best thing I could come up with is `HedgedReadOptions`, but I'm not sure if that's any better since there's only one at the moment.

src/selection_criteria.rs
patrickfreed: I wonder if separate `HedgedRead::enabled` and `HeadgedRead::disabled` would be more clear. `HedgedRead::with_enabled(false)` just seems kind of confusing to me. `HedgedRead::with_enabled(true)` also seems a bit odd to me to be honest.

e.g.
```rust
let options = ReadPreferenceOptions::builder()
    .hedge(HedgedRead::with_enabled(false))
    .build();
// vs
let options = ReadPreferenceOptions::builder()
    .hedge(HedgedRead::disabled())
    .build();
```
and
```rust
let options = ReadPreferenceOptions::builder()
    .hedge(HedgedRead::with_enabled(true))
    .build();
// vs
let options = ReadPreferenceOptions::builder()
    .hedge(HedgedRead::enabled())
    .build();
```

Alternatively, we could go for `HedgedReadOptions` as @isabelatkinson suggests. In that case I think `with_enabled` is slightly less confusing, since it's more obvious that it's an option, albeit a required one.

src/selection_criteria.rs
```diff
@@ -126,6 +126,30 @@ pub struct ReadPreferenceOptions {
     /// specified for an operation, the operation will return an error.
     #[builder(default)]
     pub max_staleness: Option<Duration>,
+
+    /// Specifies hedging behavior for reads.
```
patrickfreed, line 130: I think we could include a little more contextual info here so that users won't always need to go to the external page. Specifically, I think we ought to mention that this only applies to sharded clusters in 4.4+ and that it is automatically enabled for "nearest" read preferences.

src/selection_criteria.rs
saghm: I'm not necessarily opposed to having separate constructors; I opted not to do that initially due to the fact that it might seem weird to call it "enabled" if other options exist later as well. It does seem like `HedgedReadOptions` is the best course, so I'll change it to that.

src/selection_criteria.rs
saghm: Updated

patrickfreed (approved):

isabelatkinson (approved):

PR #255, opened 2020-09-29
RUST-561 Accept serializable data types as arguments to insert methods

isabelatkinson:


src/coll/mod.rs
```diff
@@ -111,9 +118,25 @@ impl Collection {
...
+    pub fn clone_with_type<U>(&self) -> Collection<U>
+    where
+        U: Serialize,
+    {
+        let mut options = CollectionOptions::builder().build();
+        options.selection_criteria = self.inner.selection_criteria.clone();
+        options.read_concern = self.inner.read_concern.clone();
+        options.write_concern = self.inner.write_concern.clone();
+
+        self.inner
```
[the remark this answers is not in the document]
saghm: Interesting! Another way to work around this would be to give a separate generic parameter to the `new` method, e.g. `fn new<U: Serialize>(...) -> Collection<U>`. That way, it wouldn't be tied to returning the same generic parameter.

src/coll/mod.rs
```diff
@@ -414,23 +439,27 @@ impl Collection {
...
     /// [here](https://docs.mongodb.com/manual/core/retryable-writes/) for more information on
     /// retryable writes.
     pub async fn insert_many(
         &self,
-        docs: impl IntoIterator<Item = Document>,
+        docs: impl IntoIterator<Item = T>,
         options: impl Into<Option<InsertManyOptions>>,
     ) -> Result<InsertManyResult> {
+        let mut translated_docs = Vec::new();
+        for doc in docs {
```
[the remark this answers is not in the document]
saghm: Usually the idiomatic way to to this is to map to `Iterator<Item = Result<T>>` and then collect into `Result<Vec<T>>`. Since `Result` implements FromIterator as well as `Vec`, `collect` (which uses `FromIterator` under the hood) is able to collect into a `Result`. See [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6f913bd68ce6833f2f7f0b4cb46549cc) for an example.

src/test/coll.rs
```diff
@@ -754,3 +756,158 @@ async fn no_read_preference_to_standalone() {
...
+    T: Serialize + DeserializeOwned + Clone + PartialEq + Debug,
+{
+    coll.insert_one(insert_data.clone(), None).await.unwrap();
+    let result = coll
+        .find_one(to_document(&insert_data).unwrap(), None)
+        .await
+        .unwrap();
+    match result {
+        Some(actual) => {
+            let actual: T = from_document(actual).unwrap();
```
saghm, line 807: This part will presumably have to be updated in the next PR, right?

src/coll/mod.rs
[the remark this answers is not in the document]
isabelatkinson: oh good idea, i'll update

src/test/coll.rs
isabelatkinson: yep, this step won't be necessary once returning `T` is implemented

patrickfreed (approved):

src/test/util/mod.rs
```diff
@@ -172,6 +172,15 @@ impl TestClient {
         coll
     }
 
+    pub async fn init_db_and_typed_coll<T>(&self, db_name: &str, coll_name: &str) -> Collection<T>
+    where
+        T: Serialize,
+    {
+        let coll: Collection<T> = self.database(db_name).collection_with_type(coll_name);
```
saghm, line 179: Interesting, does the compiler require the type annotation on `coll` here?

saghm (commented):
LGTM mod the change about making `new` generic over a different parameter

src/test/util/mod.rs
isabelatkinson: oh that must've been a leftover from a previous implementation, it's not necessary

isabelatkinson:
> LGTM mod the change about making `new` generic over a different parameter

I was able to fix this by updating the return type of `clone_with_type` to `Collection<U>` instead of `Self` -- the current implementation calls `Collection::new`.

saghm (approved):
Great job on this!

PR #256, opened 2020-09-29
RUST-539 Create CRUD examples for documentation

saghm:
The examples for us to implement are given in [this comment*](https://jira.mongodb.org/browse/DRIVERS-356?focusedCommentId=1521532&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1521532) on the DRIVERS ticket. The examples themselves are rendered in the various subpages of [this tutorial](https://docs.mongodb.com/manual/crud/), which has separate tabs for viewing the examples in each driver's language. I heavily followed [my footsteps](https://github.com/mongodb/mongo-cxx-driver/commit/201505d7bfbb9e3e7509340bcd8a9c006bb0000a) from my carefree youth when I was on the C++ driver, including making these examples a test that can be executing to ensure that the examples were written correctly. (Ironically, the C++ examples don't seem to have ever been added to the docs page).

* sorry non-MongoDB folks who might be reading this; this link won't work for you

### DIFF
--- a/src/test/documentation_examples.rs
+++ b/src/test/documentation_examples.rs
@@ -312,7 +312,7 @@ async fn query_embedded_documents_examples(collection: &Collection) -> Result<()
         .find(
             doc! {
                 "size": {
-                    "h":14,
+                    "h": 14,
                     "w": 21,
                     "uom": "cm",
                 },
--- a/src/test/documentation_examples.rs
+++ b/src/test/documentation_examples.rs
@@ -1,0 +1,1385 @@
+use futures::TryStreamExt;
+
+use crate::{
+    bson::{doc, Bson},
+    error::Result,
+    options::FindOptions,
+    test::TestClient,
+    Collection,
+};
+
+macro_rules! assert_coll_count {
+    ($coll:expr, $expected:expr) => {
+        assert_eq!($coll.count_documents(None, None).await.unwrap(), $expected);
+    };
+}
+
+macro_rules! assert_cursor_count {
+    ($cursor:expr, $expected:expr) => {{
+        let docs: Vec<_> = $cursor.try_collect().await.unwrap();
+        assert_eq!(docs.len(), $expected);
+    }};
+}
+
+macro_rules! run_on_each_doc {
+    ($cursor:expr, $name:ident, $check:block) => {{
+        let mut cursor = $cursor;
+
+        while let Some($name) = cursor.try_next().await.unwrap() $check;
+    }};
+}
+
+async fn insert_examples(collection: &Collection) -> Result<()> {
+    collection.drop(None).await.unwrap();
+
+    // Start Example 1
+    collection
+        .insert_one(
+            doc! {
+                "item": "canvas",
+                "qty": 100,
+                "tags": ["cotton"],
+                "size": {
+                    "h": 28,
+                    "w": 35.5,
+                    "uom": "cm",
+                }
+            },
+            None,
+        )
+        .await?;
+    // End Example 1
+
+    assert_coll_count!(collection, 1);
+
+    // Start Example 2
+    let cursor = collection.find(doc! { "item": "canvas" }, None).await?;
+    // End Example 2
+
+    assert_cursor_count!(cursor, 1);
+
+    // Start Example 3
+    let docs = vec![
+        doc! {
+            "item": "journal",
+            "qty": 25,
+            "tags": ["blank", "red"],
+            "size": {
+                "h": 14,
+                "w": 21,
+                "uom": "cm"
+            }
+        },
+        doc! {
+            "item": "mat",
+            "qty": 85,
+            "tags": ["gray"],
+            "size":{
+                "h": 27.9,
+                "w": 35.5,
+                "uom": "cm"
+            }
+        },
+        doc! {
+            "item": "mousepad",
+            "qty": 25,
+            "tags": ["gel", "blue"],
+            "size": {
+                "h": 19,
+                "w": 22.85,
+                "uom": "cm"
+            }
+        },
+    ];
+
+    collection.insert_many(docs, None).await?;
+    // End Example 3
+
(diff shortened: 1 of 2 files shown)


src/test/documentation_examples.rs
```diff
@@ -0,0 +1,1385 @@
...
+
+use crate::{
+    bson::{doc, Bson},
+    error::Result,
+    options::FindOptions,
+    test::TestClient,
+    Collection,
+};
+
+macro_rules! assert_coll_count {
```
saghm, line 11: I originally wrote these as functions, but then I kept forgetting to `await` calls I made to them, so I instead just made them macros.

src/test/documentation_examples.rs
```diff
@@ -0,0 +1,1385 @@
...
+    ($coll:expr, $expected:expr) => {
+        assert_eq!($coll.count_documents(None, None).await.unwrap(), $expected);
+    };
+}
+
+macro_rules! assert_cursor_count {
+    ($cursor:expr, $expected:expr) => {{
+        let docs: Vec<_> = $cursor.try_collect().await.unwrap();
```
saghm, line 19: Unfortunately, `StreamExt` doesn't have `count` method like [`Iterator` does](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count). A long time back I tried to convince Lucio that there should be one, but I was sadly unsuccessful.

src/test/documentation_examples.rs
```diff
@@ -0,0 +1,1385 @@
...
+        assert_eq!(docs.len(), $expected);
+    }};
+}
+
+macro_rules! run_on_each_doc {
+    ($cursor:expr, $name:ident, $check:block) => {{
+        let mut cursor = $cursor;
+
+        while let Some($name) = cursor.try_next().await.unwrap() $check;
```
saghm, line 28: I originally wrote this to take lambda as an expression rather than a name and a block and then called the lambda inside the loop, but it sometimes ran into type inference issues on the type of the parameter, and I couldn't figure out why it was only happening sometimes. Using a name and then a block doesn't seem to cause these issues, and it ends up generating code closer to what I'd actually write without a macro, so I figured this was fine.

src/test/documentation_examples.rs
```diff
@@ -0,0 +1,1385 @@
...
+    let cursor = collection
+        .find(
+            doc! {
+                "status": "A",
+                "$or": [
+                    {
+                        "qty": { "$lt": 30 },
+                    },
+                    {
+                        "item": { "$regex": "^p" },
```
saghm, line 234: Based on discussion I found on the [Python ticket](https://jira.mongodb.org/browse/PYTHON-1250) for these examples, it seems like this is explicitly supposed to be the `$regex` query operator, not the BSON type, so I left it like this.

(The fact that the operator doesn't actually require there to be an underlying BSON type is one of the several reasons that I wish BSON didn't have a native regex type...)

src/test/documentation_examples.rs
```diff
@@ -0,0 +1,1385 @@
...
+
+    assert_coll_count!(collection, 3);
+
+    // Start Example 58
+    collection.delete_one(doc! { "status": "D" }, None).await?;
+    // End Example 58
+
+    assert_coll_count!(collection, 2);
+
+    // Start Example 56
```
saghm, line 1358: This example is explicitly out of order because deleting all the documents in the collection makes trying to delete a subset of them superfluous. See the discussion on the DRIVERS ticket for verification that the docs team doesn't require the examples to be in order.

src/test/documentation_examples.rs
```diff
@@ -0,0 +1,1385 @@
...
+    // End Example 14
+
+    assert_coll_count!(collection, 5);
+
+    // Start Example 15
+    let cursor = collection
+        .find(
+            doc! {
+                "size": {
+                    "h":14,
```
patrickfreed, line 315: need a space here after the `:`

patrickfreed (commented):
Examples look good to me bar one nit.

Should we consider holding off on these until @isabelatkinson 's serde work is complete? Some of these examples, particularly the insert ones, would be greatly served by using models, and given this is the tutorial perhaps we want to demonstrate what we consider to be best practices.

src/test/documentation_examples.rs
```diff
@@ -0,0 +1,1385 @@
...
+    // Start Example 56
+    collection.delete_many(doc! {}, None).await?;
+    // End Example 56
+
+    assert_coll_count!(collection, 0);
+
+    Ok(())
+}
+
+#[tokio::test]
```
patrickfreed, line 1367: these are explcitly tokio tests right now, causing the async-std builds to fail.

patrickfreed (commented):
Just noticed the CI is red, looks like async-std issues and lint.

src/test/documentation_examples.rs
saghm: Oops, that was left over from when I originally had this in the `examples` directory rather than a test.

src/test/documentation_examples.rs
saghm: Oops, good catch

saghm:
I'm not sure if we should be showcasing the serde functionality in these examples. From looking at the existing examples, none of the other drivers seem to have provided examples like that, even in languages where that's a core part of the API (e.g. Go). I think it might be confusing for only the Rust examples to be using structs instead of documents; users might incorrectly infer that they're required to use structs instead of documents. Additionally, the examples don't give much of a change to show how to construct the collections, so users might get confused and try to pass structs to the default `Collection` type instead of one parametrized with the correct type. While we could go the route that languages like Python went and show construction of a client from the database in each example, this still has the issues of potentially misleading users into thinking that's the only way to do it and our examples sticking out as doing something quite different from what the others are doing.

patrickfreed (approved):

isabelatkinson (approved):

PR #258, opened 2020-09-30
RUST-565 Properly report connection closures due to error

patrickfreed:
RUST-565

This PR fixes a bug where connection closure events emitted due to errors never contained the `ConnectionClosedReason::Error` as a reason. This also fixes a bug where `ok: 0` responses from the initial handshake were ignored.

For convenience in testing, this PR also adds a `FailPoint` type that faciliates the enabling and cleanup of fail points.

In order to test this, RUST-304 and RUST-568 are also fixed in this PR.


src/test/spec/retryable_reads.rs
```diff
@@ -2,7 +2,7 @@ use tokio::sync::RwLockWriteGuard;
 
 use crate::test::{run_spec_test, run_v2_test, LOCK};
 
-#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "tokio-runtime", tokio::test(threaded_scheduler))]
```
patrickfreed, line 5: Now that we're actually waiting on the `watch` channels instead of immediately returning their latest values, this test started failing with "resource is busy" errors. It wasn't immediately clear what that meant, but I did notice that the number of threads running while retryable reads ran was exploding into the thousands before the test failed, at which point it dropped to what it was before the test started. Running these tests with the threaded scheduler did not create nearly as many threads and the tests pass, so for now I think we need to leave this test as threaded. It's also possible that https://github.com/tokio-rs/tokio/issues/2637 is related.

src/test/spec/crud_v1/aggregate.rs
```diff
@@ -19,6 +19,7 @@ struct Arguments {
 
 #[function_name::named]
 async fn run_aggregate_test(test_file: TestFile) {
+    let _guard: RwLockReadGuard<()> = LOCK.run_concurrently().await;
```
patrickfreed, line 22: Because we're setting failpoints in tests, acquiring the lock needs to be done before any i/o is performed whatsoever (e.g. in `TestClient::new` or in the background threads of a client). To that end, I updated many tests who were in violation of this, since they were interfering with the failpoints set in the tests required for this work.

src/cmap/conn/mod.rs
```diff
@@ -187,6 +193,11 @@ impl Connection {
         self.command_executing
     }
 
+    /// Checks if the connection experienced a network error and should be closed.
+    pub(super) fn is_errored(&self) -> bool {
```
saghm, line 197: I'm not super fond of this name; what do you think of "has_errored"? I know that "is" is generally used for boolean predicates, but I think "has" is pretty clear that it's checking for a condition as well

src/cmap/establish/handshake/mod.rs
```diff
@@ -201,6 +201,7 @@ impl Handshaker {
         let response = conn.send_command(command, None).await?;
         let end_time = PreciseTime::now();
 
+        response.validate()?;
```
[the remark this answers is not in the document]
saghm: Good catch!

src/cmap/mod.rs
```diff
@@ -216,8 +212,9 @@ impl ConnectionPoolInner {
 
         let mut available_connections = self.available_connections.lock().await;
 
-        // Close the connection if it's stale.
-        if conn.is_stale(self.generation.load(Ordering::SeqCst)) {
+        if conn.is_errored() {
```
saghm, line 215: I assume this check occurs because connections which error when handshaking in the background thread aren't eagerly closed?

src/cmap/test/event.rs
```diff
@@ -64,8 +81,36 @@ impl CmapEventHandler for EventHandler {
     }
 }
 
+pub struct EventSubscriber<'a> {
+    _handler: &'a EventHandler,
```
saghm, line 85: Could you add a doc comment explaining why `_handler` is here? My assumption is that it's to catch bugs where a subscriber outlives the handler at compile time, but I'm not positive

src/cmap/test/integration.rs
```diff
@@ -137,3 +138,73 @@ async fn concurrent_connections() {
         .await
         .expect("disabling fail point should succeed");
 }
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test(threaded_scheduler))]
```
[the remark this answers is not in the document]
saghm: I wonder if it's possible to take this a step further, e.g. automatically acquire the lock exclusively when we create a failpoint. This would presumably require us not to acquire the lock manually in the tests to avoid deadlocks, but maybe we could get around that by _always_ acquring a read lock and then passing that into the constructor for Failpoint (which then drops it and acquires a write lock instead). There wouldn't be any way to get the read lock back in the test function after the Failpoint is dropped, but I'm guessing there isn't really anywherte we'd need that.

src/cmap/test/mod.rs
```diff
@@ -216,20 +209,44 @@ impl Operation {
...
-                    if let Some(pool) = state.pool.read().await.deref() {
-                        pool.check_in(conn).await;
-                    }
+                    let id = conn.id;
+                    // connections are checked in via tasks spawned in their drop implementation,
+                    // they are not checked in explicitly.
+                    drop(conn);
+
+                    // wait for event to be emitted to ensure check in has completed.
+                    subscriber
```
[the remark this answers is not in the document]
saghm: Awesome

src/test/spec/retryable_reads.rs
saghm: Is it worth filing a ticket to investigate this more later?

src/cmap/conn/mod.rs
patrickfreed: fixed

src/cmap/mod.rs
patrickfreed: This path is taken when a connection gets a network error after the handshake succeeds actually (e.g. during operation execution).

src/cmap/test/event.rs
patrickfreed: Yep, done.

src/cmap/test/integration.rs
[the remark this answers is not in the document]
patrickfreed: In a lot of cases we need to check the server version before we enable failpoints, so we definitely need to acquire at least a read lock beforehand. Also, we may need to repeatedly enable failpoints over the duration of a test (e.g. in a  spec test), so I don't think we can use the one that upgrades the lock either. I think we'll just have to rely on our diligence until we come up with a macro for automating the lock acquisition or something.

src/test/spec/retryable_reads.rs
patrickfreed: Yeah I think so. I did a little preliminary digging to see if the linked tokio issue was related, but there have been enough breaking changes in tokio to make that investigation non-trivial, so I just didn't bother. Filed RUST-581 for follow on investigation.

src/cmap/test/integration.rs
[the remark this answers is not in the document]
saghm: Fair enough

src/cmap/mod.rs
saghm: Ah, right, I see now

saghm (approved):

isabelatkinson (approved):

PR #259, opened 2020-10-09
RUST-556 POC of maxConnecting

patrickfreed:
RUST-556

This PR implements a POC of the maxConnecting requirement of the "Avoiding Connection Storms" project. In doing so, it also re-architects the connection pool to use a channel based WaitQueue rather than semaphore based one. This was done to allow for easier and more effective maintenance of the wait queue now that we need to support arbitrary conditions for exiting the queue (e.g. maxConnecting). 

I realize that a rewrite of this magnitude typically goes through a design phase, so I apologize for suddenly dropping it like this. I decided it was the preferable choice for this POC largely because a counting semaphore wait queue was frustrating to extend to use in this case, and that this lockless + channel-based pattern is a [much more common aproach in the async world for accessing shared state](https://docs.rs/tokio/0.2.22/tokio/sync/index.html#mpsc-channel) than our previous lock heavy one. Maintaining the old pool and adding support for maxConnecting would introduce even _more_ locking while also producing a suboptimal solution (impossible or difficult to ensure fairness with multiple semaphores/locks in this case), so this seemed like an good time for a rewrite. Additionally, this is a proof-of-concept after all, so I figured it would be a good time to proof-of-concept what an idiomatic async pool would look like too. I think that I should have opted for a pool like this back when I originally converted the old pool to async, but I was lacking in experience with async patterns at the time. Ironically, the original sync pool with its queue of condvars was somewhat similar to this one.

## Overview of new pool

I'll give a brief overview of the layout of the new pool here to make digesting the actual code a little easier. We have the same clonable `ConnectionPool` type as before with the same functionality; however, when we create a new pool, instead of creating a `ConnectionPoolInner` type that is arc'd and wrapping that, we start a worker task (`ConnectionPoolWorker`) that opens up a few different channels, and we store senders to those channels (`ConnectionRequester`, `PoolManager`) on the `ConnectionPool`. The worker task is in charge of all changes and access to the connection pool state; to do anything with the pool, a you need to send a request via one of the two senders mentioned before. To check out a connection for example, a thread sends a `ConnectionRequest` via a `ConnectionRequester` and waits for the worker to respond. To perform any other pool interactions (e.g. clearing, checking in), a thread uses the `PoolManager` to send the appropriate requests. The `ConnectionRequesters` act like strong references to the pool in that they keep the worker alive until they are all dropped, whereas `PoolManagers` are like weak references and can still be in scope when the worker ceases execution. A `ConnectionPool` instance holds onto both a `ConnectionRequester` and a `PoolManager`, but a `Connection` for example only holds on to a `PoolManager` in order to check itself back in. Instead of having a separate background task for ensuring minPoolSize, the worker task just takes care of that.

In summary:
- `ConnectionPoolInner` gone, replaced with a background worker task `ConnectionPoolWorker` that has exclusive access to pool state
- `ConnectionPoolWorker` listens on a few channels for incoming requests, responding to them as necessary
- Threads can check out connections via `ConnectionRequester`s. The `ConnectionPoolWorker` quits once all the requesters are dropped
- Threads can modify the pool state via `PoolManager`s. These do _not_ keep the worker running and may no-op if outlive the worker.
- `Connection`s use `PoolManager`s to check themselves back in

### Some notes
Given this architecture, it was trivially easy to introduce maxConnecting: we simply don't poll the `ConnectionRequestReceiver` when the pool is empty and has too many pending connections. The actual changes required to implement that part are all in the last few commits or so. Additionally, we now have a context which always has exclusive access to the pool state, making it much easier to reason about. Lastly, the `drop` implementations of the various connection types have been simplified with the removal of `ConnectionPoolInner` and `DroppedConnectionState`.


src/cmap/test/mod.rs
```diff
@@ -18,15 +18,18 @@ use crate::{
...
     RUNTIME,
 };
+use bson::doc;
 
 const TEST_DESCRIPTIONS_TO_SKIP: &[&str] = &[
     "must destroy checked in connection if pool has been closed",
     "must throw error if checkOut is called on a closed pool",
 ];
 
+const EVENT_DELAY: Duration = Duration::from_millis(500);
```
patrickfreed, line 31: This is just a temporary solution until the CMAP spec test changes in #258 go in. I wanted to avoid duplicating them here and just opted for the easy solution. Even with this generous delay, we still sometimes don't see the events on Windows with async-std, which is a bit concerning. The failures should go away once the aforementioned changes are merged in though.

patrickfreed (commented):
The clippy failures seem unrelated to this PR. The other failures are either due to the max files issue or should be fixed in #258 I think.

src/cmap/connection_requester.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// Returns an error if it takes longer than wait_queue_timeout before either a connection is
+    /// received or an establishment begins.
+    pub(super) async fn request(&self, wait_queue_timeout: Option<Duration>) -> Result<Connection> {
+        let (sender, receiver) = oneshot::channel();
+
+        // this only errors if the receiver end is dropped, which can't happen because
+        // we own a handle to the worker, keeping it alive.
+        self.sender.send(sender).unwrap();
+
+        // similarly, the receiver only returns an error if the sender is dropped, which
```
saghm, line 49: Could you move this comment to the call to `receiver.await.unwrap()`? It confused me at first because I thought it was referring to the timeout error

src/cmap/test/mod.rs
```diff
@@ -191,40 +194,37 @@ impl Operation {
...
                 }
                 Operation::CheckIn { connection } => {
                     let conn = state.connections.write().await.remove(&connection).unwrap();
+                    drop(conn);
 
-                    if let Some(pool) = state.pool.read().await.deref() {
-                        pool.check_in(conn).await;
-                    }
+                    // give some time for drop to happen.
+                    RUNTIME.delay_for(Duration::from_millis(500)).await;
```
saghm: Is there a way we can use channels here to avoid having to sleep and hope that the background task runs fast enough?

src/runtime/join_handle.rs
```diff
@@ -23,28 +21,20 @@ pub(crate) enum AsyncJoinHandle<T> {
 }
 
 impl<T> Future for AsyncJoinHandle<T> {
-    // tokio wraps the Output of its JoinHandle in a Result, while async-std does not.
-    // In order to standardize on this without us panicking, we too wrap the output in a Result.
-    type Output = Result<T>;
+    // tokio wraps the Output of its JoinHandle in a Result that conatins an error if the task
+    // panicked, while async-std does not.
+    //
+    // Given that async-std will panic or abort the task in this scenario, there is not a
```
saghm, line 27: Was this change made to help with debugging?

src/cmap/mod.rs
```diff
@@ -207,28 +84,12 @@ impl ConnectionPoolInner {
...
-            self.close_connection(conn, ConnectionClosedReason::Dropped)
-        } else {
-            available_connections.push_back(conn);
-        }
-
-        self.wait_queue.wake_front();
-    }
-
-    async fn check_out(&self) -> Result<Connection> {
+    /// Checks out a connection from the pool. This method will block until this thread is at the
```
saghm, line 87: Should we say "yield" instead of "block" since this is async?

src/cmap/worker.rs
```diff
@@ -0,0 +1,494 @@
...
+impl HandleListener {
+    fn new() -> (Self, PoolWorkerHandle) {
+        let (sender, receiver) = mpsc::channel(1);
+        (Self { receiver }, PoolWorkerHandle { sender })
+    }
+
+    /// Listen until all handles are dropped.
+    /// This will not return until all handles are dropped, so make sure to only poll this via
+    /// select or with a timeout.
+    async fn listen(&mut self) {
```
saghm, line 514: Could we name something that conveys its usage a bit more clearly, e.g. "wait_for_handle_drop"?

src/cmap/worker.rs
```diff
@@ -0,0 +1,494 @@
...
+            // If there's an outstanding request and we can serve it, do so. Otherwise poll
+            // for a new request
+            let task = if self.unfinished_check_out.is_some()
+                && self.can_service_connection_request()
+            {
+                PoolTask::CheckOut(self.unfinished_check_out.take().unwrap())
+            } else {
+                tokio::select! {
+                    Some(result_sender) = self.request_receiver.recv(),
+                                        if self.unfinished_check_out.is_none() => PoolTask::CheckOut(result_sender),
```
saghm: I could be wrong, but this looks like if the pool is empty, checkouts will be delayed until the next tick of the maintenance interval populates it. If so, this should be fixed so that a new connection created immediately. Am I understanding this correctly?

src/cmap/manager.rs
```diff
@@ -0,0 +1,68 @@
...
+#[derive(Debug)]
+pub(super) enum PoolManagementRequest {
+    Clear,
+    Populate(Connection),
+    CheckIn(Connection),
+    HandleConnectionFailed(Error),
+}
+
+impl PoolManagementRequest {
+    fn unwrap_check_in(self) -> Connection {
```
saghm: It's not clear to me where the `unwrap_` methdods are used (for this type or the other request types). Are they just used in tests, and if so, can we annotate them as such?

saghm (commented):
I've reviewed the refactor of the pool and added comments for it. I'll come back and do a second review of the updates for maxConnecting later this week

kmahar:
I’m going to mainly focus my review on the design you describe and the big picture here, rather than the code. I trust that Isabel and Sam can confirm your code does what you say it does 🙂 

I mainly just have questions about the overall architecture. I don't think I necessarily know enough about CMAP or what an idiomatic async pool would look like to have strong opinions, so it might be worthwhile having Matt weigh in on this since he is familiar with the past pool designs, and also with maxConnecting, but I will defer to your judgement on whether that is needed.

> We have the same clonable `ConnectionPool` type

This isn’t that important but just curious, IIUC something being clonable means you can make a deep copy of it, right? Why do we need that capability for our connection pool?

So what I understand so far is that (I think) there is a 1:1 pool-worker relationship. Is there also a single manager and a single requester per pool? I thought so from the overview, but then this sentence confused me, because it sounds like there are multiple requesters which a single worker is paying attention to:

> The ConnectionPoolWorker quits once all the requesters are dropped

Also, at what point is a requester dropped?

saghm:
> IIUC something being clonable means you can make a deep copy of it, right? Why do we need that capability for our connection pool?

Generally, yes, this is correct. However, Rust's mechanism for reference counted types is to define a type that implements the same Clone trait as other clonable types but with semantics that share the underlying data. In this case, we utilize those types for all of the mutable state in the connection pool to make it act as if there's a shallow copy. We use the same strategy for Client, Database, and Collection types to allow users to make copies to send to different threads/tasks.

kmahar:
@saghm Ahh I see, that makes sense.

Tangential question: it makes sense to me that a client would be a reference type, since you want to be able to share all of its underlying resources between objects that reference it. But why make databases and collections reference types, too? Would doing so mean that copying them ends up deep copying everything they store references to (such clients)? 

I'm comparing this to Swift where if you copy a value type - say `MongoDatabase` which has a property that is a reference type - say `MongoClient` - the copy of the database will automatically refer to the same client as the original database referred to.

saghm:
You raise a good point here; I don't think we technically need to make the Database and Collection reference counted right now since the only mutable data they have (the internals of the underlying client) are already reference counted, so it doesn't actually make any semantic difference from the users perspective. This is all internal implementation right now though, so the only real effect of making Database and Collection reference counted right now is that it causes the small amount of data specific to the Database and Collection types (their names, read preferences, and read/write concerns) allocated on the heap rather than the stack.

kmahar:
Gotcha, thanks for the explanation!

src/cmap/test/mod.rs
patrickfreed: Definitely, that is what is being done in #258. I didn't want to duplicate the changes from that in this PR though, so I just added heavy wait times. Once that PR goes in I'll rebase this one to use the channel approach.

src/cmap/worker.rs
patrickfreed: I don't think that is the case. If a check out request is received and the pool is empty, this branch will be taken and `check_out` will be called, which conditionally creates a connection as necessary.

patrickfreed (commented):
>So what I understand so far is that (I think) there is a 1:1 pool-worker relationship. Is there also a single manager and a single requester per pool? I thought so from the overview, but then this sentence confused me, because it sounds like there are multiple requesters which a single worker is paying attention to. Also, at what point is a requester dropped?

This is partially explained by the cloning discussion above. While there is a 1:1 relationship between requesters and `ConnectionPools`, there can be many requesters for a single worker because each instance of `ConnectionPool` has one and can be cloned. A requester is dropped when the `ConnectionPool` instance that owns it is dropped. `ConnectionPools` also contain managers, so there can be many managers per worker, and since checked out `Connections` also have managers, there isn't a 1:1 relationship between pools and managers for a given worker.

src/cmap/test/mod.rs
saghm: Sounds good!

Cargo.toml
```diff
@@ -85,7 +85,7 @@ optional = true
 
 [dependencies.tokio]
 version = "0.2.18"
-features = ["io-util", "sync"]
+features = ["io-util", "sync", "macros"]
```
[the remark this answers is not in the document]
saghm: The most objective way to measure would be to compare the binary size of a crate with and without the macros added. I agree that they likely won't make a huge difference, and in the case of users who use tokio, they're likely going to end up including it anyways to be able to use `tokio::main`.

src/runtime/join_handle.rs
saghm: Were we ever returning an error to user that we're now panicking on, or is this just internal?

src/cmap/worker.rs
saghm: Yep, that's perfect

src/cmap/manager.rs
saghm: Can you walk me through why we know these won't panic in the cases that they're called? There are a lot of moving pieces, so I'm not sure I fully understand

saghm: Never mind, I think I understand now

src/cmap/worker.rs
saghm: Okay, when reading all this over the second time, I think I understand what's going on better.

Cargo.toml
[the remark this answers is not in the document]
patrickfreed: The binary size seems to differ by 1K on my machine when building with async-std, so it doesn't seem to be an issue.

src/runtime/join_handle.rs
patrickfreed: Nope, we only ever actually used the join handle in the CMAP spec tests, and we were always unwrapping them there anyways. The driver mostly used `execute` before this PR which discarded the result of the async task.

Cargo.toml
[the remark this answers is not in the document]
saghm: Yeah, I don't think it's an issue. If users really need a minimal binary, they can strip it, which will likely remove part of that 1K since we aren't using every macro (along with large parts of our other dependencies as well).

saghm:
The lint failures are likely due to the new minor version of Rust that came out since you started this PR; Isabel merged a PR to fix them last week, so if you rebase with master, I think they should go away

src/cmap/worker.rs
```diff
@@ -0,0 +1,506 @@
...
+        (manager, connection_requester)
+    }
+
+    /// Run the worker thread, listening on the various receivers until all handles have been
+    /// dropped. Once all handles are dropped, the pool will close any available connections and
+    /// emit a pool closed event.
+    async fn execute(mut self) {
+        let mut maintenance_interval = RUNTIME.interval(Duration::from_millis(500));
+
+        loop {
```
patrickfreed, line 166: I moved the unfinished request cache to the receiver itself, which made this loop/select a little simpler.

patrickfreed:
Just rebased, hopefully that fixes them.

src/cmap/worker.rs
saghm: Awesome, this does make it a lot easier to follow!

saghm (approved):
All the questions I had about this have been answered, and the fully green build is super nice, so I think this is ready for my LGTM!

isabelatkinson (approved):

kmahar:
@patrickfreed Thanks for the explanation, that makes sense. Design SGTM then.

kmahar (approved):

PR #261, opened 2020-10-15
RUST-562 Return deserializable data types from find methods

isabelatkinson:


src/test/coll.rs
```diff
@@ -908,6 +901,43 @@ async fn typed_replace_one() {
...
 
     let result = coll.find_one(doc! { "x": 2 }, None).await.unwrap().unwrap();
-    let result: UserType = from_document(result).unwrap();
     assert_eq!(result, replacement);
 }
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[function_name::named]
+async fn typed_returns() {
```
isabelatkinson, line 910: The previously written methods already test that `find`, `find_one`, and `find_one_and_replace` deserialize properly so I just added tests for `find_one_and_update` and `find_one_and_delete`.

src/db/mod.rs
```diff
@@ -211,7 +211,7 @@ impl Database {
     ) -> Result<Vec<String>> {
         let list_collections =
             ListCollections::new(self.name().to_string(), filter.into(), true, None);
-        let cursor = self
+        let cursor: Cursor<Document> = self
```
saghm, line 214: I assume the reason this is needed is that `self` is a `Database<T>` here, and `T` isn't necessarily `Document`? I want to make sure that this isn't something that could happen with existing user code.

src/test/coll.rs
saghm: I think it would still be worth having tests that use the collection type without a parameter (i.e. with T = Document implicit) to make sure that things will still work fine for users.

saghm (commented):
I have one clarification question and one small suggestion for additional tests, but other than that, this looks really good!

src/db/mod.rs
isabelatkinson: The reason is because `Cursor::new` is now generic over `T`, so a `T` needs to be specified when it gets called. `Cursor::new` is `pub(crate)`, though, so it won't be an issue for existing code.

src/test/coll.rs
isabelatkinson: Is it necessary to add new tests for that? It seems that functionality is already covered by the above tests/crud spec tests.

src/db/mod.rs
saghm: Ah, right. Thanks for the clarification

src/coll/mod.rs
```diff
@@ -381,12 +386,17 @@ where
...
-    ) -> Result<Option<Document>> {
+    ) -> Result<Option<T>> {
         let mut options = options.into();
         resolve_options!(self, options, [write_concern]);
 
         let op = FindAndModify::with_delete(self.namespace(), filter, options);
-        self.client().execute_operation(op).await
+        let result = self.client().execute_operation(op).await;
+
+        result.and_then(|opt| {
```
patrickfreed, line 396: Rather than repeating this in each of the `find_one_and_x` method bodies, it seems like we ought to handle this in `FindAndModify::handle_response`. We'd have to make `FindAndModify` generic as part of that.

src/operation/find_and_modify/mod.rs
```diff
@@ -67,12 +86,20 @@ impl FindAndModify {
...
     }
 }
 
-impl Operation for FindAndModify {
-    type O = Option<Document>;
+impl<T> Operation for FindAndModify<T>
+where
+    T: Serialize + DeserializeOwned + Unpin + Debug,
+{
+    type O = Option<T>;
```
isabelatkinson, line 102: The type `O` for `Operation` requires `Debug`, which necessitates the addition of `Debug` as a requirement for everything upstream. I figured this isn't too onerous for users since they probably already have `Debug` implemented on their types and/or can just derive the trait.

patrickfreed: Hm yeah, that should probably be okay. Honestly I kind of wish everything derived `Debug` by default in Rust. I forget the reasoning for that not being the case.

patrickfreed (approved):

src/test/coll.rs
saghm: Good point!

saghm (approved):

univerz:
any chance to parametrize methods instead of whole `Collection`? when porting projects i ran into "projection" datatypes - adding `Serialize` is fast solution (most of the time), but it's nice typesafe feature that incomplete struct can't be serialized and inserted into db.

patrickfreed:
Hm, that's an interesting idea; I filed RUST-852 to discuss this with the team. Thanks for the suggestion!

PR #262, opened 2020-10-20
RUST-549 Increase ulimit in evergreen tests

patrickfreed:
RUST-549

This PR increases the ulimit to 64000 before starting orchestration or any tests to prevent the "too many open files" test failures we see on async-std with macOS. They seem to be caused by EVG-13244, so this change will not be necessary at some point in the future. Until then, I think it's worth manually setting the ulimit to avoid test failures.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -81,6 +81,10 @@ functions:
               export SSL=${SSL}
               export TOPOLOGY=${TOPOLOGY}
               export MONGODB_VERSION=${MONGODB_VERSION}
+
+              if [ "Windows_NT" != "$OS" ]; then
+                  ulimit -n 64000
+              fi
            EOT
            # See what we've done
            cat expansion.yml


.evergreen/config.yml
```diff
@@ -81,6 +81,10 @@ functions:
               export SSL=${SSL}
               export TOPOLOGY=${TOPOLOGY}
               export MONGODB_VERSION=${MONGODB_VERSION}
+
+              if [ "Windows_NT" != "$OS" ]; then
```
patrickfreed, line 85: Windows machines complain if we modify this, so I restricted it to macOS and Linux. macOS machines are the main culprit for our too many files errors anyways.

patrickfreed:
Looks like the existing failures will be fixed in #258 , so after these both go in, we should have a fully green waterfall :tada:

.evergreen/config.yml
```diff
@@ -81,6 +81,10 @@ functions:
...
+                  ulimit -n 64000
```
isabelatkinson, line 86: Is there any reason for 64000 specifically?

patrickfreed: This is the value evergreen hosts are supposed to have for max open files (and most do).

isabelatkinson (approved):

saghm (approved):

PR #263, opened 2020-10-21
RUST-555 Implement exhaust cursor support

saghm:
This pull request augments the driver with two new features: the ability to send the exhaust flag to the server through the internal cursor API, and the ability to read exhaust responses from the server (indicated by the `moreToCome` flag). These are decoupled from a behavioral standpoint; if the driver sends the `exhaust` flag, it can still read non-exhaust responses fine if that's what the server returns, and the driver can properly handle responses with `moreToCome` regardless of whether it sent `exhaust` on the operation corresponding to the response. The changes to the driver fall into three buckets:

* operation execution API changes

The operation execution API (i.e. `src/client/executor.rs`) was updated so that top-level execution methods all return an `Option<Connection>` along with the actual response from the operation. A connection will be present if and only if the server response contained the `moreToCome` flag

* operation layer changes

The operation trait was updated to allow operations to specify whether or not they support sending the exhaust flag to the server. This behavior is only implemented for the `getMore` command, and it is not exposed through any public API to users.

Additionally, the `Command` struct was updated to support setting the `exhaust` flag when it's serialized to the wire protocol.

* cursor implementation changes

The cursor API was updated so that a connection can optionally be passed in when constructing a cursor, with an invariant that a connection being passed in means that the initial operation response contained the `moreToCome` flag. Additionally, `GetMoreProvider` was augmented to support reading an exhaust response, and `GetMoreProviderResult` was augmented to support retrieving the connection in the case that the server response contains the `moreToCome` flag. The corresponding cursor implementations were updated to implement these new features through the new API methods on those traits.

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -216,11 +216,11 @@ impl Client {
             }) => {
                 let connection = if more_to_come { Some(conn) } else { None };
 
-                return Ok(CursorResponse {
+                Ok(CursorResponse {
                     response,
                     connection,
                     session: None,
-                });
+                })
             }
 
             Err(err) => {
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -1,6 +1,6 @@
 use super::{Client, ClientSession};
 
-use std::{collections::HashSet, sync::Arc};
+use std::{collections::HashSet, fmt::Debug, sync::Arc};
 
 use lazy_static::lazy_static;
 use time::PreciseTime;
@@ -31,13 +31,37 @@ lazy_static! {
     };
 }
 
+/// The result of executing an operation.
+#[derive(Debug)]
+pub(crate) struct OperationResult<T: Debug> {
+    /// The response from the server.
+    pub(crate) response: T,
+
+    /// The connection used for the operation. This will _only_ be present if the server's response
+    /// contained the `moreToCome` flag.
+    pub(crate) connection: Option<Connection>,
+}
+
+/// A wrapped response to an operation.
+#[derive(Debug)]
+pub(crate) struct OperationResponse<T: Debug> {
+    /// The response from the server.
+    pub(crate) response: T,
+
+    /// Whether or not the `moreToCome` flag was present in the server's response.
+    pub(crate) more_to_come: bool,
+}
+
 impl Client {
     /// Execute the given operation.
     ///
     /// Server selection will performed using the criteria specified on the operation, if any, and
     /// an implicit session will be created if the operation and write concern are compatible with
     /// sessions.
-    pub(crate) async fn execute_operation<T: Operation>(&self, op: T) -> Result<T::O> {
+    pub(crate) async fn execute_operation<T: Operation>(
+        &self,
+        op: T,
+    ) -> Result<OperationResult<T::O>> {
         // TODO RUST-9: allow unacknowledged write concerns
         if !op.is_acknowledged() {
             return Err(ErrorKind::ArgumentError {
@@ -56,7 +80,7 @@ impl Client {
     pub(crate) async fn execute_cursor_operation<T: Operation>(
         &self,
         op: T,
-    ) -> Result<(T::O, Option<ClientSession>)> {
+    ) -> Result<(OperationResult<T::O>, Option<ClientSession>)> {
         let mut implicit_session = self.start_implicit_session(&op).await?;
         self.execute_operation_with_retry(op, implicit_session.as_mut())
             .await
@@ -69,7 +93,7 @@ impl Client {
         &self,
         op: T,
         session: &mut ClientSession,
-    ) -> Result<T::O> {
+    ) -> Result<OperationResult<T::O>> {
         self.execute_operation_with_retry(op, Some(session)).await
     }
 
@@ -79,7 +103,7 @@ impl Client {
         &self,
         op: T,
         mut session: Option<&mut ClientSession>,
-    ) -> Result<T::O> {
+    ) -> Result<OperationResult<T::O>> {
         let server = self.select_server(op.selection_criteria()).await?;
 
         let mut conn = match server.checkout_connection().await {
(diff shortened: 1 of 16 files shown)


src/cursor/common.rs
```diff
@@ -200,3 +223,23 @@ pub(crate) struct CursorInformation {
     pub(crate) batch_size: Option<u32>,
     pub(crate) max_time: Option<Duration>,
 }
+
+/// Reads a `getMore` response form the server without sending a `getMore` command.
+pub(super) async fn read_exhaust_get_more(
```
patrickfreed, line 228: I worry that implementing this in such a direct manner outside of the executor could miss out on some required functionality. For instance, should we be updating the client and sessions' cluster times based off of get more responses? Also, in the event of network failure, we need to update the topology and clear the connection pool, which I don't think we can do here. Lastly, there seem to be tests verifying this behavior, but shouldn't we be emitting command monitoring events for these (perhaps with mocked command started events)? We probably need command failed events at least to ensure errors are properly logged.

src/coll/mod.rs
```diff
@@ -319,7 +328,10 @@ where
         resolve_options!(self, options, [write_concern]);
 
         let delete = Delete::new(self.namespace(), query, Some(1), options);
-        self.client().execute_operation(delete).await
+        self.client()
+            .execute_operation(delete)
```
patrickfreed, line 332: It could reduce on boilerplate a little bit to have a separate `execute_cursor_operation` so that all these other methods which could never make use of the returned connection don't have to deal with it.

src/client/executor.rs
```diff
@@ -106,8 +130,23 @@ impl Client {
             .execute_operation_on_connection(&op, &mut conn, &mut session, txn_number)
             .await
         {
-            Ok(result) => {
-                return Ok(result);
+            Ok(OperationResponse {
```
patrickfreed, line 133: I personally prefer this match with the condition checking in the branch like this:
```rust
match blah {
   Ok(response) => {
        let connection = if response.more_to_come { Some(conn) } else { None };
        return Ok(OperationResult { response: response.response, connection })
    },
    Err(err) => { ... }
}
```
This way the duplicated `OperationResponse { ... }` matching can be omitted. Up to you if you think it's worth changing.

patrickfreed (commented):
Implementation looks pretty clean, though I do have some concerns about not sharing code with the executor when sending exhausted getmores.

src/coll/mod.rs
saghm: There's already an `execute_cursor_operation` and it does something different

src/cursor/common.rs
saghm: All of this fair. The difficulty is that in the absence of the specification actually being written, we don't know what the answers to any of these questions are. The cluster times might not make sense to keep track of, as the time that we're reading the responses could be much later than when the server sent them. Ditto with network failure; if we can't reach the server, that might be because the server _did_ shut down, and that's why it sent the responses in exhaust mode rather than waiting for getMore commands. Command monitoring is specifically an open question that the scope says the design will deal with, but the design is not being prioritized, so we don't know what the solution will be.

Despite all this, I still think this PR is worth merging because large parts of it will remain the same whenever the design does get written, and it will be less work to keep the code up to date by having it compiled as part of CI rather than trying to keep an external branch up to date through rebasing or letting it get stale and doing one giant rebase later. Maybe the best solution is to pre-emptively file a ticket for implementing the design that will eventually be written and then add a TODO comment here explaining that all these issues will need to be addressed once the spec work is done.

saghm:
I've rebased to fix conflicts/fix the lint and refactored the operation execution changes a bit

src/cursor/common.rs
patrickfreed: That sounds good to me.

src/coll/mod.rs
patrickfreed: Oh right, I forgot we already added something similar before. While it does currently do something different, it actually needs to be expanded and used for exhaust cursors too. Otherwise, the implicit session used by an exhaust cursor will get returned to the session pool as soon as the initial command is executed, and it could then be used by other operations while the cursor is still alive (i.e. concurrently).

saghm: In the latest commit I pushed, I think this is handled properly. The session used for the initial command is passed into the cursor constructor and kept around for as long as the cursor is alive, with a reference being passed in to each getMore operation. I moved it into the `CursorResponse` type (which was called `OperationResult` in the first commit of this PR) rather than having a tuple of `(OperationResult, Option<Session>)` being returned and then having to pass both elements in separately to the constructor.

patrickfreed:
The evg run is pretty red, looks like it might need to be skipped on older versions?

src/cmap/conn/mod.rs
```diff
@@ -246,11 +246,26 @@ impl Connection {
...
+    pub(crate) async fn read_response(&mut self) -> Result<CommandResponse> {
         let response_message_result = Message::read_from(&mut self.stream).await;
         self.command_executing = false;
         self.error = response_message_result.is_err();
 
-        CommandResponse::new(self.address.clone(), response_message_result?)
+        let response_message = response_message_result?;
+
+        if !response_message.more_to_come() {
+            self.command_executing = false;
```
isabelatkinson, line 265: Is this assignment necessary? It looks like `command_executing` is already set to false in all cases on line 259 above.

src/cursor/common.rs
```diff
@@ -200,3 +223,24 @@ pub(crate) struct CursorInformation {
     pub(crate) batch_size: Option<u32>,
     pub(crate) max_time: Option<Duration>,
 }
+
+/// Reads a `getMore` response form the server without sending a `getMore` command.
```
isabelatkinson, line 227: "form" instead of "from" here

src/cmap/conn/mod.rs
saghm: Oops, good catch! That first one should be setting it to `true`.

src/cursor/common.rs
saghm: Good catch! Fixed

src/cmap/conn/mod.rs
saghm: Ah wait, no, it needs to just delay setting false until the if statement. Since this method is called by `send_command`, we want to only mark the connection as not being in use once we're sure that there aren't any more messages in the buffer. Still a bug as written though, so good catch!

saghm:
@patrickfreed test failures turned out to be due to exhaust cursors not being supported on mongos, so I skipped the test in the case that the cluster is sharded

saghm:
Based on offline discussion, we'll leave this in a branch in case we want to reference it later and close this PR

PR #265, opened 2020-10-23
RUST-576 Fix count_documents fails when writeConcern set on Collection

patrickfreed:
RUST-576

This PR fixes a bug where `count_documents` would return an error if called on a client that had a write concern set. This only applies to 3.6 servers it seems (I only tested with 4.4 and 3.6, did not apply to 4.4).

### DIFF
--- a/src/operation/aggregate/mod.rs
+++ b/src/operation/aggregate/mod.rs
@@ -54,8 +54,6 @@ impl Operation for Aggregate {
             if let Ok(cursor_doc) = body.get_document_mut("cursor") {
                 cursor_doc.remove("batchSize");
             }
-        } else {
-            body.remove("writeConcern");
         }
 
         Ok(Command::new(
--- a/src/operation/aggregate/mod.rs
+++ b/src/operation/aggregate/mod.rs
@@ -54,6 +54,8 @@ impl Operation for Aggregate {
             if let Ok(cursor_doc) = body.get_document_mut("cursor") {
                 cursor_doc.remove("batchSize");
             }
+        } else {
+            body.remove("writeConcern");
         }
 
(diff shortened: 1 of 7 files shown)


src/operation/aggregate/mod.rs
```diff
@@ -54,6 +54,8 @@ impl Operation for Aggregate {
             if let Ok(cursor_doc) = body.get_document_mut("cursor") {
                 cursor_doc.remove("batchSize");
             }
+        } else {
+            body.remove("writeConcern");
```
patrickfreed, line 58: I decided just removing any specified writeConcern was the easiest thing to do here. This will prevent the driver from erroring when a user passes a write concern into a read aggregation via `Collection::aggregate`, but I didn't think that was much of a concern.

saghm: What do you think of instead changing `count_documents` to manually create and execute an aggregation operation rather than going through `Collection::aggregate`? I think that should fix the issue without causing any changes to the behavior of other aggregate operations.

patrickfreed: Good point, that does seem to be a better solution, updated.

In fact, now that we're considering doing wrapping operations (e.g. for `watch`), this probably makes sense as its own operation wrapping `Aggregate`. Should I go ahead and do that as well?

saghm: Yep, sounds good!

patrickfreed: Done. I also noticed that `CountOptions` didn't have read concern or selection_criteria, so I added those options as well.

isabelatkinson (approved):

saghm (approved):

PR #266, opened 2020-10-28
RUST-577 Versioned API implementation

alcaeus:
This PR drafts necessary changes to accommodate the versioned API. Note that a significant portion of the diff is the unified tests runner, which I squashed in from #249. I plan to keep this PR up to date as that PR progresses.

Todo list
- [x] Allow declaring an API version in a language idiomatic matter
- [x] Submit declared API version fields with every command
- [x] Figure out where to append Versioned API fields to command
- [x] Format code according to style guide
- [x] Verify correct behaviour using spec tests
- [x] Provide easier way to create `ApiVersion`

See below for notes about open issues in the list above.

---

### Figure out where to append versioned API fields

I initially planned to append the Versioned API fields in `Client::execute_operation_on_connection`, but realised that this is only used when running commands that are part of an operation. Other commands, such as `isMaster` or all authentication commands are not run through this path, which makes this unfeasible.
The common code point would be `Message::with_command`, which will have to receive the declared API version from the connection that the message is being sent on. This has two downsides: instead of only keeping the API version declared on the client, we need to pass it down through connection options. Furthermore, since command monitoring is hooked into `Client::execute_operation_on_connection` instead of `Connection::send_command`, we need to manually append Versioned API fields to command documents for command monitoring. I'm not particularly happy about this as I'd like to keep the logic as contained as possible, but with `Connection::send_command` being used in multiple places I feel that passing the declared API version as part of the command is not feasible. I'm happy about any ideas how to best implement this.

### Format code

Will be done at the end - please forgive me for visually butchering the language.

### Spec tests

While the unified test runner POC is being worked on, not all tests may pass. Currently it fails to compare command objects, but I haven't investigated any further aside from making sure that the Versioned API fields appear as expected.
Due to this driver not implementing the transaction spec, the transaction tests will be verified using the second POC driver.

### Easier API for ApiVersion

The API version is currently declared as follows:
```rust
let mut options = ClientOptions::parse_uri(uri, None).await.unwrap();
options.server_api = ServerApi {
    version: ServerApiVersion::Version1,
    strict: None,
    deprecation_errors: None,
};
```

This feels cumbersome, as users are required to pass values for `strict` and `deprecation_errors` explicitly, even when they want to rely on the default. A factory would be useful, but I'm not sure what an idiomatic factory in Rust would look like. Following options come to mind:
```rust
// Variant 1: Factory to create ServerApi object from string, set other fields directly:
let mut server_api = ServerApi::from_string("1");
server_api.strict = true;
server_api.deprecation_errors = true;

// Variant 2: Have methods to modify initially created object
let server_api = ServerApi::from_string("1");
// Example: enable strict mode and deprecation errors directly
let server_api = ServerApi::from_string("1").strict(true).deprecation_errors(true);
```


src/client/options/mod.rs
```diff
@@ -192,6 +192,63 @@ impl fmt::Display for StreamAddress {
...
+
+    /// Whether the server should return errors for features that are not part of the API version
+    pub strict: Option<bool>,
+
+    /// Whether the server should return errors for deprecated features
+    pub deprecation_errors: Option<bool>,
+}
+
+impl ServerApi {
+    pub(crate) fn append_to_command(&self, command: &mut Document) {
```
alcaeus, line 231: Technically we need this kind of API:
```rust
    pub(crate) fn append_to_command(&self, command: &mut Command) {
        self.append_to_command_body(command.name.as_str(), &mut command.body);
    }

    pub(crate) fn append_to_command_body(&self, command_name: &str, body: &mut Document) {
        if !self.applies_to_command(command_name) {
            return;
        }

        body.insert("apiVersion", self.version.as_str());

        if let Some(strict) = self.strict {
            body.insert("apiStrict", strict);
        }

        if let Some(deprecation_errors) = self.deprecation_errors {
            body.insert("apiDeprecationErrors", deprecation_errors);
        }
    }
```

`append_to_command()` would be used in `Message::with_command` where we have a `Command` instance, while `append_to_command_body` would be used in `Executor::execute_operation_on_connection`, where we operate on a `Document` instance. The advantage would be that the calling code doesn't need to know that they have to check `applies_to_command` first (as is currently the case).

NB: `apply_to_command` may be a more feasible name, but I feel like it's too early for bikeshedding.

src/cmap/conn/wire/message.rs
```diff
@@ -36,6 +37,12 @@ impl Message {
                 .insert("$readPreference", read_pref.into_document());
         };
 
+        if let Some(server_api) = server_api {
+            if server_api.applies_to_command(command.name.as_str()) {
+                server_api.append_to_command(&mut command.body);
+            }
```
alcaeus, line 43: With the alternative API in `ServerApi`:
```suggestion
            server_api.append_to_command(&mut command);
```

src/client/executor.rs
```diff
@@ -252,7 +252,15 @@ impl Client {
             let command_body = if should_redact {
                 Document::new()
             } else {
-                cmd.body.clone()
+                // Append versioned API fields
+                let mut body = cmd.body.clone();
+                if let Some(ref server_api) = self.inner.options.server_api {
+                    if server_api.applies_to_command(cmd.name.as_str()) {
+                        server_api.append_to_command(&mut body);
+                    }
```
alcaeus, line 260: With the alternative API in ServerApi:
```suggestion
                    server_api.append_to_command_body(cmd.name.as_str(), &mut body);
```

src/client/options/mod.rs
saghm: I think the cleanest way to do this would likely be to add a method to `Command` similar to what we have for [sessions, cluster time, and transaction numbers](https://github.com/mongodb/mongo-rust-driver/blob/4eb171cd412a4d4227bd3c6690838a3a84b5158c/src/cmap/conn/command.rs#L49-L62). The general idea behind our `Command` type is to be a thin wrapper around `Document` that avoids the need to hardcode various field names that we have to access in multiple places. This seems like a good fit for setting the API version of a command.

src/client/options/mod.rs
```diff
@@ -192,6 +192,63 @@ impl fmt::Display for StreamAddress {
...
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[non_exhaustive]
+pub enum ServerApiVersion {
+    Version1,
+}
+
+impl ServerApiVersion {
+    pub fn from_string (version: &str) -> Self {
+        match version {
+            "1" => Self::Version1,
```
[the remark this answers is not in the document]
saghm: I think the most idiomatic way to do this would be to implement the [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) and [`Display`](https://doc.rust-lang.org/std/fmt/trait.Display.html) traits.

src/client/options/mod.rs
```diff
@@ -192,6 +192,63 @@ impl fmt::Display for StreamAddress {
...
+    }
+
+    pub(crate) fn as_str(&self) -> &str {
+        match self {
+            Self::Version1 => "1",
+        }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, TypedBuilder)]
```
saghm, line 216: `TypedBuilder` can be used to fix one of the issues you mentioned in your top-level comment. It generates a builder pattern for constructing the type and allows you to make fields implemented by default. As it's written right now, the following API would be valid:

```rust
ServerApi::builder()
    .version(some_version)
    .strict(true)
    .deprecation_errors(false)
    .build()
```

However, if you add `#[builder(default)]` above the declarations of the `strict` and `deprecation_errors` fields, then those can be omitted, so the following uses would also be valid:

```rust
ServerApi::builder()
    .version(some_version)
    .strict(true)
    .build()
```
```rust
ServerApi::builder()
    .version(some_version)
    .deprecation_errors(false)
    .build()
```
```rust
ServerApi::builder()
    .version(some_version)
    .build()
```

src/cmap/conn/mod.rs
```diff
@@ -70,6 +71,8 @@ pub(crate) struct Connection {
 
     #[derivative(Debug = "ignore")]
     handler: Option<Arc<dyn CmapEventHandler>>,
+
+    server_api: Option<ServerApi>,
```
saghm, line 75: Out of curiosity, is putting the server API version on a connection a requirement of the spec? It's not immediately obvious to me why this would be necessary, so I'd like to understand better why this is being done

src/cmap/conn/mod.rs
```diff
@@ -93,6 +96,7 @@ impl Connection {
             ready_and_available_time: None,
             stream: AsyncStream::connect(stream_options).await?,
             address,
+            server_api: options.as_ref().and_then(|options| options.server_api.clone()),
```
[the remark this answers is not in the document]
saghm: Cloning here is extremely cheap due to `ServerApiVersion` being quite small, so I think this is fine to leave as it is. Using a reference would actually be a lot more work, as that would require us to add a lifetime parameter to `Connection` to ensure that it doesn't outlive the `ServerApiVersion` that it references.

src/cmap/conn/mod.rs
```diff
@@ -230,7 +235,7 @@ impl Connection {
         command: Command,
         request_id: impl Into<Option<i32>>,
     ) -> Result<CommandResponse> {
-        let message = Message::with_command(command, request_id.into());
+        let message = Message::with_command(command, request_id.into(), &self.server_api);
```
saghm, line 238: Related to my comment above, I think the cleanest way to do this would be to have `Command` have a `set_api_version` method and then call that before passing the command to `Message::with_command`.

src/cmap/conn/mod.rs
alcaeus: Authentication commands and the handshake bypass the operation executor and create their own command objects. This means that the "lowest" common place to inject the server API version would be `Connection::send_command`, requiring us to store the `ServerApi` object on the connection.

src/client/options/mod.rs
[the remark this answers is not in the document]
alcaeus: Changed the enum to implement both `FromStr` and `Display`.

src/client/options/mod.rs
alcaeus: Ah, great to know. Made the change 👍

src/cmap/conn/mod.rs
alcaeus: Changed this. This allows me to simplify the logic in `Client::execute_operation_on_connection` and inject the client server API information in the command before emitting the CommandStarted event. On the other hand, this means that the connection will set the server API again, which wasn't in the original plan. I don't think this has any direct side effects, but wanted to point this out.

src/cmap/conn/mod.rs
saghm: Right; I'm wondering if it would make more sense to just having the calling context set the API version on a command before calling `send_command` like we do with some other properties. This seems like more of a Rust driver-specific architecture question, so I'm going to tag in @patrickfreed and @isabelatkinson to see if they have any opinions on the options here.

src/cmap/conn/mod.rs
saghm: I don't think the redundancy is a huge deal, but it might be mitigated if we do the change I mention above (i.e requiring that the API version is set before `send_command` rather than inside it).

saghm:
It looks like there are some git conflicts, likely due to the CMAP stuff that Patrick recently merged. Once everything in this branch is all settled, we can handle the rebase (which one of us can do if you want, since it's always hard to rebase code that you didn't write or review).

alcaeus:
@saghm the conflicts come from the unified test runner changes, which I've squashed into this PR. I'll update this as new changes are pushed to #249, which should then also fix the conflicts.

saghm:
Ah, that makes sense. Sounds good!

alcaeus:
@saghm I'll have to take a deeper look at the conflicts that are shown here. Locally, a merge into master resolves cleanly, so this may be due to GitHub getting a little confused.

Either way, this PR is now up-to-date with the latest changes from the spec itself. I've dropped the transaction tests as they don't apply here due to the driver not supporting transactions. As for the build failures, there are two distinct failures:
* Some auth builds (e.g. AWS, X509) fail due to the `getParameter` command being called in `TestClient` (necessary to check the new `serverParameters` option in `runOnRequirements`). The user in question is not allowed to run this command and I haven't visited this issue yet as I'm not quite sure where we need to make changes.
* Builds on latest fail due to extra commands in the `db.aggregate` test. However, the same test passes locally on the same version, so I'm not quite sure what's going on there.

Either way, this PR is ready for another review while I try to figure out where the build problems originate.

src/client/options/mod.rs
```diff
@@ -327,6 +390,12 @@ pub struct ClientOptions {
     #[builder(default)]
     pub selection_criteria: Option<SelectionCriteria>,
 
+    /// The declared API version
+    ///
+    /// The default value is to have no declared API version
+    #[builder(default)]
+    pub server_api: Option<ServerApi>,
```
patrickfreed, line 397: In addition to passing this via the options, do we want a specific `with_uri_str` method that also takes a version since it can't be specified via string?

e.g.
```rust
let server_api = ServerApi::builder().version(ServerApiVersion::v1).build();
let client = Client::with_uri_str_and_server_api("mongodb://...", server_api).await?;
```
It's a bit of a mouthful but it's still a little easier than parsing options first.

alcaeus: I wasn't sure what kind of factories you'd want to have. I'll defer to @saghm and @isabelatkinson on a decision on this and can implement it if everyone agrees 👍

src/cmap/options.rs
```diff
@@ -70,6 +70,12 @@ pub struct ConnectionPoolOptions {
     #[builder(default)]
     pub min_pool_size: Option<u32>,
 
+    /// The declared API version
+    ///
+    /// The default value is to have no declared API version
+    #[builder(default)]
+    pub server_api: Option<ServerApi>,
```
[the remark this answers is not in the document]
saghm: ConnectionPoolOptions isn't a user-facing type, so non-exhaustivity doesn't really make a difference here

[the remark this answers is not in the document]
saghm: Oh, I see, it's exposed in the PoolCreatedEvent. I guess a better design would have been not to use this type itself in the event but made a separate one to map to that only contained a subset of fields.

src/client/options/mod.rs
saghm: Given that there's already a way to parse a string into options and then pass that, I think I'd prefer not to do that until we have a strong sense that this is something that users want. If we ever have another option that can't be specified in the string, the combinations start to get unwieldy. There's a decent risk that we might wish we had done something more extensible to multiple options later.

patrickfreed: SGTM

alcaeus: > If we ever have another option that can't be specified in the string, the combinations start to get unwieldy.

Just pointing out that FLE will bring `autoEncryptionOpts` which will be the same kind of client option as `serverApi`, so not including option-specific constructors sounds sensible 👍

src/cmap/conn/mod.rs
alcaeus: From what I can see, moving this from `send_command` to the calling context would require us to pass server api info into the auth mechanism implementations and append it to the commands there.

On the other hand, if (and this is a big if at the moment) we were to allow users to override the declared API version on a more granular level (e.g. on a per-operation basis), automatically setting the api version on the connection might not be the right call, but as is I think it's the most logical place to keep this. I'll defer to your judgment on this since, but I might require some help if you decide that this shouldn't be in `send_command`.

saghm: I think Patrick's suggestion is probably the most consistent with how we're using commands and connections in the driver. The way things work right now, whatever is passed to `send_command` is exactly what's sent to the server; I'm not sure I want to go down the route where the connection modifies commands before sending them to the server, as then it gets trickier to reason about what's sent to the server from outside of the `send_command` function.

patrickfreed: So to support this approach, we'd need to do the store the `ServerApi` value on the `ConnectionEstablisher` and pass it through through to `handshaker.handshake` and `credential.authenticate_stream` as necessary (https://github.com/mongodb/mongo-rust-driver/blob/master/src/cmap/establish/mod.rs#L12). I think that in conjunction with the existing storage of `ServerApi` on `ClientInner` and setting in `execute_operation`, this should cover the cases we need without having to store anything on the `Connection` object itself.

src/cmap/options.rs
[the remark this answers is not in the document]
patrickfreed: I filed RUST-591 to fix these issues. For now I think this PR can just proceed as-is, and we just need to make sure to do RUST-591 before these changes are released.

src/cmap/conn/mod.rs
alcaeus: I've made the changes in a3c6a9b347452bb97b35ecf4305b5d4fa9b28bc3. One notable difference is that I didn't pass `server_api` through `handshaker.handshake`. Since the `isMaster` command is generated when the Handshaker instance is created, we can also apply server_api parameters there and don't have to worry about them later. I think that takes care of everything, but please let me know if I missed anything.

src/client/auth/plain.rs
```diff
@@ -23,12 +27,15 @@ pub(crate) async fn authenticate_stream(
         .as_ref()
         .ok_or_else(|| Error::authentication_error("PLAIN", "no password supplied"))?;
 
-    let sasl_start = SaslStart::new(
+    let mut sasl_start = SaslStart::new(
```
saghm, line 30: It might make sense to add `server_api` as a parameter to `SaslStart` and `SaslContinue` to reduce the repeated code on each call.

src/client/auth/scram.rs
```diff
@@ -131,10 +134,14 @@ impl ScramVersion {
         &self,
         conn: &mut Connection,
         credential: &Credential,
+        server_api: Option<&ServerApi>,
     ) -> Result<FirstRound> {
         let client_first = self.build_client_first(credential, false)?;
 
-        let command = client_first.to_command(self);
+        let mut command = client_first.to_command(self);
```
saghm, line 141: Similar to my suggestion above, I think it makes sense to modify the methods like `build_client_first` to take in server API as a parameter. The idea here is that all of the logic for building the commands is separate from the authentication flow, which makes each of them easier to understand independent of each other.

src/cmap/establish/test.rs
```diff
@@ -70,6 +71,7 @@ async fn speculative_auth_test(
         doc! { "find": "foo", "limit": 1  },
     );
 
+    // TODO: will have to add versioned API options before running command
```
saghm, line 74: We generally put ticket numbers on all of our TODOs that we commit to the driver. Am I correct in thinking that this is blocked on us having a new enough server version in our evergreen matrix to support this feature? If so, I can file a ticket for adding that to the matrix and then you can put the ticket number here.

alcaeus: This is currently blocked on not being able to declare a server API version for the test runner. Drivers will eventually have to run their tests against a server with `requireApiVersion` enabled, which will require being able to pass in an API version through environment variables (as it can't be declared using the connection string). However, that is not possible yet as mongo-orchestration doesn't support starting a server with `requireApiVersion` enabled (due to its internal `isMaster` check not declaring an API version, which leads it to believe that it wasn't able to start the cluster).

I can either file an issue pre-emptively (which may become obsolete if we were to create a DRIVERS ticket for this) and add that number here, or I can drop the TODO comments, as I only added them to highlight any commands that will not work on a server with `requireApiVersion` enabled.

alcaeus:
@saghm @patrickfreed with Unified Test Format done, this PR is ready for review and I even got Evergreen to play along! 🎉

src/cmap/establish/test.rs
saghm: Assuming the tests will fail when run against a server with requireApiVersion before we make any code changes, a TODO comment probably isn't needed. If they will silently pass though, we definitely want a comment with a ticket number so that we can keep track of the fact that need to do extra work whenever it becomes unblocked.

src/sdam/monitor.rs
```diff
@@ -160,6 +160,7 @@ async fn is_master(connection: &mut Connection) -> Result<IsMasterReply> {
     );
 
     let start_time = PreciseTime::now();
+    // TODO: will have to add versioned API options before running command
```
saghm, line 163: Ditto above comment about TODO's without ticket numbers

src/test/spec/unified_runner/test_file.rs
```diff
@@ -33,6 +37,9 @@ pub struct TestFile {
     pub create_entities: Option<Vec<TestFileEntity>>,
     pub initial_data: Option<Vec<CollectionData>>,
     pub tests: Vec<TestCase>,
+    // TODO: don't read this field
```
saghm, line 40: Is this TODO copied from the unified spec test runner, or did you add this?

src/test/spec/unified_runner/test_file.rs
```diff
@@ -0,0 +1,369 @@
...
+        uri.push_str(&format!("{}={}&", &key, value));
+    }
+
+    // remove the trailing '&' from the URI (or '?' if no options are present)
+    uri.pop();
+
+    Ok(uri)
+}
+
+fn deserialize_server_api<'de, D>(
```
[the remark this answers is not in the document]
saghm: If possible, it would be preferable to just implement Deserialize directly on `ServerApi` to avoid needing to write custom code. From looking at the code here, it looks like we could do it with just an attribute on the `version` field like this (omitting the existing attributes on the struct for brevity):

```rust
#[derive(Deserialize)]
struct ServerApi {
    #[serde(deserialize_with="FromStr::from_str")]
    version: Option<ServerApiVersion>,

    strict: Option<bool>,

    deprecationErrors: Option<bool>, 
}
```

src/test/spec/unified_runner/test_file.rs
alcaeus: I added this. I didn't find a way to ignore a field in serde, so I'll remove this comment and leave the field as-is.

src/test/spec/unified_runner/test_file.rs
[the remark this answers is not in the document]
alcaeus: This yields:
```
error[E0308]: mismatched types
   --> src/client/options/mod.rs:223:24
    |
223 | #[derive(Clone, Debug, Deserialize, PartialEq, TypedBuilder)]
    |                        ^^^^^^^^^^^ expected `&str`, found type parameter `__D`
    |
    = note:   expected reference `&str`
            found type parameter `__D`
    = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0308]: mismatched types
   --> src/client/options/mod.rs:223:24
    |
223 | #[derive(Clone, Debug, Deserialize, PartialEq, TypedBuilder)]
    |                        ^^^^^^^^^^^ expected associated type, found struct `error::Error`
    |
    = note: expected associated type `<__D as selection_criteria::_::_serde::Deserializer<'de>>::Error`
                        found struct `error::Error`
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider constraining the associated type `<__D as selection_criteria::_::_serde::Deserializer<'de>>::Error` to `error::Error`
    |
223 | #[derive(Clone, Debug, Deserialize<Error = error::Error>, PartialEq, TypedBuilder)]
    |                                   ^^^^^^^^^^^^^^^^^^^^^^

error[E0308]: mismatched types
   --> src/client/options/mod.rs:223:24
    |
223 | #[derive(Clone, Debug, Deserialize, PartialEq, TypedBuilder)]
    |                        ^^^^^^^^^^^ expected associated type, found struct `error::Error`
    |
    = note: expected associated type `<__D as selection_criteria::_::_serde::Deserializer<'de>>::Error`
                        found struct `error::Error`
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider constraining the associated type `<__D as selection_criteria::_::_serde::Deserializer<'de>>::Error` to `error::Error`
    |
223 | #[derive(Clone, Debug, Deserialize<Error = error::Error>, PartialEq, TypedBuilder)]
    |                                   ^^^^^^^^^^^^^^^^^^^^^^
```

src/test/util/mod.rs
```diff
@@ -83,28 +85,46 @@ impl TestClient {
...
-        let server_version = info.version.split('-').next().unwrap();
-        let server_version = Version::parse(server_version).unwrap();
+            let info: BuildInfo = bson::from_bson(Bson::Document(response)).unwrap();
+            let server_version_str = info.version.split('-').next().unwrap();
+            server_version = Some(Version::parse(server_version_str).unwrap());
+
+            let get_parameters =
+                RunCommand::new("admin".into(), doc! { "getParameter": "*" }, None).unwrap();
+
+            if let Ok(response) = client
```
isabelatkinson, line 107: Is it okay to ignore an error here? Might be better to unwrap the response to catch some kind of failure.

src/test/spec/unified_runner/mod.rs
```diff
@@ -236,3 +239,10 @@ async fn test_examples() {
     let _guard = LOCK.run_exclusively().await;
     run_spec_test(&["unified-runner-examples"], run_unified_format_test).await;
 }
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn test_versioned_api() {
+    let _guard = LOCK.run_exclusively().await;
```
isabelatkinson, line 246: We usually annotate the type on these (i.e. `let _guard: RwLockWriteGuard<_> = LOCK.run_exclusively().await`) because the compiler doesn't generate any warnings/errors if you forget to `await` without the annotation. It's not an issue here but it would be good to add it in anyway for the sake of consistency with other tests.

src/test/util/mod.rs
alcaeus: I specifically chose to ignore a failure for the time being (and probably should've added a comment explaining so) because the current AWS and X509 auth tests don't allow for running `getParameter` commands. Since those tests don't need to check parameters (yet), I decided to ignore errors. The result of this is that any unified spec test that has a `runOnRequirement` that includes `serverParameters` will be skipped since the parameters will always be unequal.

src/test/spec/unified_runner/mod.rs
alcaeus: That was a shameless copy/paste from the test above this one. I'll update both methods.

src/test/util/mod.rs
isabelatkinson: Ah gotcha, thanks for explaining that!

src/test/mod.rs
```diff
@@ -43,10 +47,20 @@ lazy_static! {
...
         let mut options = ClientOptions::parse_without_srv_resolution(&uri).unwrap();
         options.max_pool_size = Some(MAX_POOL_SIZE);
+        options.server_api = SERVER_API.clone();
 
         options
     };
     pub(crate) static ref LOCK: TestLock = TestLock::new();
     pub(crate) static ref DEFAULT_URI: String =
         std::env::var("MONGODB_URI").unwrap_or_else(|_| "mongodb://localhost:27017".to_string());
+    pub(crate) static ref SERVER_API: Option<ServerApi> = match std::env::var("MONGODB_API_VERSION")
```
saghm, line 57: From looking below, this is just needed for the server API spec tests, right?

src/test/spec/unified_runner/test_file.rs
saghm: You can have serde ignore a field with [`#[serde(skip)]`](https://serde.rs/field-attrs.html#skip).

src/test/spec/unified_runner/test_file.rs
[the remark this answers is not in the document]
saghm: Ah, sorry, my mistake; I misread what you're doing here. The most idiomatic way to do this would be to implement Deserialize directly on `ServerApiVersion`, which then would allow `Deserialize` to be derived on `ServerApi` without needing any custom code. Here's the code for that (and I promise I made sure it compiles this time!):

```rust
use std::str::FromStr;

use serde::{Deserialize, de::{self, Deserializer, Unexpected}};

pub enum ServerApiVersion {
    Version1,
}

impl<'de> Deserialize<'de> for ServerApiVersion {
    fn deserialize<D>(
        deserializer: D
    ) -> Result<Self, D::Error> where D: Deserializer<'de> {
        let s = String::deserialize(deserializer)?;
        
        ServerApiVersion::from_str(&s)
            .map_err(|_| {
                de::Error::invalid_value(Unexpected::Str(&s), &"a valid version number")
            })
    }
}

#[derive(Deserialize)]
struct ServerApi {
    version: Option<ServerApiVersion>,

    strict: Option<bool>,

    deprecationErrors: Option<bool>, 
}
```

src/test/spec/unified_runner/test_file.rs
alcaeus: Gave this a shot, but I believe that the `deny_unknown_fields` option prevents this from working. Using `#[serde(skip)]`, even in combination with `rename`, causes a panic about `_yamlAnchors` being an unknown field:
> unknown field `_yamlAnchors`, expected one of `description`, `schemaVersion`, `runOnRequirements`, `allowMultipleMongoses`, `createEntities`, `initialData`, `tests`

I've removed the todo comment, but it looks like we'll have to consume the field from the test files and then ignore it.

src/test/mod.rs
alcaeus: This is needed to be able to run the tests against a server with `requireApiVersion=true`. It is used when creating test clients when there's no explicit API version set, but only if the `MONGODB_API_VERSION` environment variable was set.
Without this, the test suite fails all over the place due to the server rejecting commands without `apiVersion` set. For those servers, we run `MONGODB_API_VERSION=1 cargo test --verbose` and our test clients will have an API version declared.

The versioned API spec tests don't need this, as they always declare an API version in the client config.

src/test/spec/unified_runner/test_file.rs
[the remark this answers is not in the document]
alcaeus: Ah, this works now. There was prior art for this (and lots of it at that), but I wasn't sure how to go about this. Thanks for helping out!

src/client/options/mod.rs
patrickfreed: Rather than having separate `Command::set_server_api` and this `ServerApi::append_to_command`, could we consolidate on a single `Command::set_server_api`?

patrickfreed (commented):
The actual driver changes look pretty good, my main concern right now has to do with the changes to `TestClient`. Also, I just merged some stuff that looks to be in conflict with this PR: don't worry about those for now, I can do the rebase once this branch looks ready.

src/test/util/mod.rs
```diff
@@ -83,28 +85,49 @@ impl TestClient {
...
 
-        let build_info = RunCommand::new("test".into(), doc! { "buildInfo":  1 }, None).unwrap();
+        let mut server_version = None;
+        let mut server_parameters = None;
 
-        let response = client
-            .execute_operation_with_session(build_info, &mut session)
-            .await
-            .unwrap();
+        if collect_server_info {
```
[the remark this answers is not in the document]
saghm: I'm a little worried about the potential proliferation of environmental variables needed to run our tests. Right now, the only env var needed to invoke the tests (besides ones that need secret info like the Atlas connectivity tests and AWS auth) is MONGODB_URI, and if possible I'd like to keep it that way since it makes it much easier to run the tests locally. (I was going to touch on this in my comment above about the MONGODB_API_VERSION variable above as well.)

I don't have a good sense of which cases in the future that we'll actually need to have the API version set for and which ones we won't. Will it be a 50/50 split of our variants on whichever server version first implements it, where we run all our tests both with and without it? Will we just have one variant running a few tests with the API version set and then run the rest without, or vice-versa? Or will we run all of our tests on newer server versions with the API version set? If the only tests we'll miss running without an env var set are specific to the API version functionality, or if it's like TLS where we just run the entire test suite with or without it, I wouldn't be opposed to a new env var, but if this is going to be a requirement for most or all of our tests going forward, I'd really prefer not to require that we need to have another env var set when invoking the tests locally.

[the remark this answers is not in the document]
patrickfreed: Yeah I think these questions are worth sorting out. The [spec](https://github.com/alcaeus/specifications/blob/drivers-996-versioned-api/source/versioned-api/tests/README.rst#id2) is currently a little vague on this and only says the following: 

>Drivers MUST test against a server with the requireApiVersion parameter enabled that also requires authentication. Since API versioning options can't be specified using the connection string, drivers MUST rely on environment variables to receive the API version to use in tests.

Assuming we treat API version as a tls-like option and run the entire suite with different values for it (I think this is the plan since the server needs startup parameters), then I think we'd also need to introduce other environment variables for server version, etc. as suggested in my previous comment since we can't use `buildInfo` in those cases. Note, this means that we'd only introduce new environment variables when doing API version testing and normal testing can just the the URI as we do today so I don't think this should impact running tests locally too much.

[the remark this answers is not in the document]
alcaeus: Trying to answer the various questions in the comments above here.

> I'm a little worried about the potential proliferation of environmental variables needed to run our tests.

The spec calls for running tests against a server with `requireApiVersion=true` to have a functional test suite and ensure that everything works as expected (for example, this uncovered a server bug with `getMore` commands simultaneously requiring and forbidding `apiVersion` options). Since the API version can only be declared as a client option and is not exposed as a URI option (see rationale in the spec), we have to use an environment variable or similar mechanism to declare an API version for tests. However, I don't expect this variant to be tested locally very often, unless someone is explicitly working on versioned API features.

> I don't have a good sense of which cases in the future that we'll actually need to have the API version set for and which ones we won't.

Just like we run tests on a server with and without authentication enabled, we'll also run tests on servers with `requireApiVersion=true`. For the time being this is only a single server version, but I expect us to test this with all server versions that support the feature. However, this always runs the entire test. The spec tests for versioned API test the driver logic for commands explicitly to ensure that API inheritance is implemented properly. These tests will change if we decide to change inheritance rules for the declared API version (e.g. allowing users to declare a different API version on a `MongoDatabase` or `MongoCollection` object). I hope that clears up the confusion as to what variants should be tested. Note that the evergreen pipeline currently does not test against a server with `requireApiVersion=true` for multiple reason: tests would be failing until the `getMore` issue explained above is fixed, and we currently can't use mongo-orchestration to start such a cluster as the `isMaster` command run by mongo-orchestration fails as it doesn't declare an API version.

> I worry about this approach where we sometimes don't get the server information, since in those cases all of `TestClient`'s methods panic when called.

Typically when creating test clients, they will not be running with `apiStrict=true`, only with `apiVersion=1`. The use-case for `collect_server_info` are client entities in the unified test runner. These are created with custom options, including a strict API version. After discussing this with @isabelatkinson I decided that the sensible thing would be to disable collection of this data if disabled. I'm open to alternatives, but most of these would go beyond the scope of this PR. One option would be to use a different implementation for client entities, another would be to make fetching buildinfo lazy and only do it when the info is needed. I'm not entirely happy with the added constructor argument, but it was the most logical choice to work around the limitation of not being able to test any strict API variants.

src/client/options/mod.rs
alcaeus: This was previously extracted because of where and how API version was injected into commands. I can remove the `impl` for `ServerApi` and move the logic to `Command` if that's more appropriate.

alcaeus:
> Also, I just merged some stuff that looks to be in conflict with this PR: don't worry about those for now, I can do the rebase once this branch looks ready.

No worries, I rebased to make sure this implementation doesn't break due to previous changes. I've also pushed a commit that inlines appending server parameters from a `ServerAPI` implementation into the `Command` implementation.

src/test/util/mod.rs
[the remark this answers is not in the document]
patrickfreed: To summarize the testing requirements based on our slack discussion:
- Drivers will be required to run their entire test suites with certain API versions, configured via environment variables (similar to auth)
- As part of normal test runs, the versioned API spec tests will be run and configure an API version they read from the JSON formats, no environment variable needed.

For the first point, I think the versioned API test readme needs to be updated to explicitly require drivers to do this and also to lay out specifically which API versions / options combos need to be tested in such a manner (as we discussed in slack). 

With regards to the second point, I'm okay with temporarily merging the `TestClient` changes for the sake of moving this project along, but the Rust driver team needs to follow up and figure out a more robust solution that doesn't introduce so much panic risk. 

Does this sound good to everyone?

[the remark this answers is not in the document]
saghm: :+1:

src/client/options/mod.rs
```diff
@@ -202,6 +206,62 @@ impl fmt::Display for StreamAddress {
     }
 }
 
+#[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub enum ServerApiVersion {
```
patrickfreed, line 211: Can we add a docstring to this enum and its case?

src/client/options/mod.rs
```diff
@@ -202,6 +206,62 @@ impl fmt::Display for StreamAddress {
...
+
+        ServerApiVersion::from_str(&s)
+            .map_err(|_| Error::invalid_value(Unexpected::Str(&s), &"a valid version number"))
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, TypedBuilder)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct ServerApi {
```
patrickfreed, line 252: ditto docstring for this struct

patrickfreed (commented):
Code all looks good to me, just missing a few docstrings.

re: mongo-orchestration, do we know what the status is on getting that ready to strart servers that require an API version? Will there be a follow on DRIVERS ticket that ensures we update our evergreen config accordingly?

alcaeus:
@patrickfreed Added both docstrings, let me know if you prefer wording improvements.

We'll spawn a drivers ticket when mongo-orchestration is capable of handling versioned API. This is blocked on PyMongo supporting versioned API.

patrickfreed (approved):
LGTM!

saghm (approved):

isabelatkinson (approved):

PR #268, opened 2020-11-17
RUST-575 Consider connection pool health during server selection

patrickfreed:
RUST-575

This PR updates server selection to consider the number of ongoing operations a each server has instead of randomly selecting from among those in the latency window. This done as part of DRIVERS-781 Avoiding Connection Storms. See https://github.com/mongodb/specifications/commit/07790d1978ebe6b8ff7f574c9ac5e1365dd16e26 for the spec changes.

This also fixes RUST-592.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ lazy_static = "1.4.0"
 md-5 = "0.8.0"
 os_info = { version = "2.0.6", default-features = false }
 percent-encoding = "2.0.0"
-rand = "0.7.2"
+rand = { version = "0.7.2", features = ["small_rng"] }
 serde_with = "1.3.1"
 sha-1 = "0.8.1"
 sha2 = "0.8.0"
--- a/src/cmap/worker.rs
+++ b/src/cmap/worker.rs
@@ -2,6 +2,7 @@ use derivative::Derivative;
 
 use super::{
     conn::PendingConnection,
+    connection_requester,
     connection_requester::{
         ConnectionRequest,
         ConnectionRequestReceiver,
@@ -9,6 +10,7 @@ use super::{
         RequestedConnection,
     },
     establish::ConnectionEstablisher,
+    manager,
     manager::{ManagementRequestReceiver, PoolManagementRequest, PoolManager},
     options::{ConnectionOptions, ConnectionPoolOptions},
     Connection,
@@ -132,10 +134,10 @@ impl ConnectionPoolWorker {
             .as_ref()
             .map(|pool_options| ConnectionOptions::from(pool_options.clone()));
 
-        let (handle_listener, handle) = HandleListener::new();
+        let (handle, handle_listener) = handle_channel();
         let (connection_requester, request_receiver) =
-            ConnectionRequester::new(address.clone(), handle);
-        let (manager, management_receiver) = PoolManager::new();
+            connection_requester::channel(address.clone(), handle);
+        let (manager, management_receiver) = manager::channel();
 
         let worker = ConnectionPoolWorker {
             address,
@@ -498,11 +500,26 @@ impl From<PoolManagementRequest> for PoolTask {
     }
 }
 
+/// Constructs a new channel for for monitoring whether this pool still has references
+/// to it.
+fn handle_channel() -> (PoolWorkerHandle, HandleListener) {
+    let (sender, receiver) = mpsc::channel(1);
+    (PoolWorkerHandle { sender }, HandleListener { receiver })
+}
+
 /// Handle to the worker. Once all handles have been dropped, the worker
 /// will stop waiting for new requests and drop the pool itself.
 #[derive(Debug, Clone)]
 pub(super) struct PoolWorkerHandle {
     sender: mpsc::Sender<()>,
+}
+
+impl PoolWorkerHandle {
+    #[cfg(test)]
+    pub(super) fn new_mocked() -> Self {
+        let (s, _) = handle_channel();
+        s
+    }
 }
 
 /// Listener used to determine when all handles have been dropped.
@@ -512,11 +529,6 @@ struct HandleListener {
 }
 
 impl HandleListener {
-    fn new() -> (Self, PoolWorkerHandle) {
-        let (sender, receiver) = mpsc::channel(1);
-        (Self { receiver }, PoolWorkerHandle { sender })
-    }
-
     /// Listen until all handles are dropped.
     /// This will not return until all handles are dropped, so make sure to only poll this via
     /// select or with a timeout.
--- a/src/sdam/description/topology/server_selection/mod.rs
+++ b/src/sdam/description/topology/server_selection/mod.rs
@@ -1,24 +1,84 @@
 #[cfg(test)]
 mod test;
 
-use std::{fmt, time::Duration};
+use std::{collections::HashMap, fmt, ops::Deref, sync::Arc, time::Duration};
 
-use rand::seq::IteratorRandom;
+use rand::{rngs::SmallRng, seq::SliceRandom, SeedableRng};
 
 use super::TopologyDescription;
 use crate::{
     error::{ErrorKind, Result},
+    options::StreamAddress,
     sdam::{
         description::{
             server::{ServerDescription, ServerType},
             topology::TopologyType,
         },
         public::ServerInfo,
+        Server,
     },
     selection_criteria::{ReadPreference, SelectionCriteria, TagSet},
 };
 
 const DEFAULT_LOCAL_THRESHOLD: Duration = Duration::from_millis(15);
+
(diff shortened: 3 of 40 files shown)


src/sdam/description/topology/server_selection/test/logic.rs
```diff
@@ -26,62 +20,6 @@ struct TestFile {
     error: Option<bool>,
 }
 
-#[derive(Debug, Deserialize)]
```
patrickfreed, line 29: these were moved into `server_selection/test/mod.rs` so they could be shared between the logic and in-window tests.

src/cmap/worker.rs
```diff
@@ -505,6 +505,14 @@ pub(super) struct PoolWorkerHandle {
     sender: mpsc::Sender<()>,
 }
 
+impl PoolWorkerHandle {
+    #[cfg(test)]
+    pub(super) fn new_mocked() -> Self {
+        let (_, s) = HandleListener::new();
```
saghm, line 511: Seeing this usage of `HandleListener::new()` makes me realize that it's a bit confusing to have a `new` method return a tuple rather than just `Self`. Could we rename that method to be a bit more descriptive?

src/sdam/description/topology/server_selection/mod.rs
```diff
@@ -1,25 +1,91 @@
...
+    type Target = Server;
+
+    fn deref(&self) -> &Server {
+        self.server.deref()
+    }
+}
+
+impl Drop for SelectedServer {
+    fn drop(&mut self) {
+        self.server.decrement_operation_count();
```
saghm, line 52: I'm not sure this is correct; we generally stop using the server after we've checked out a connection, and the compiler is free to drop the server at any point after the last usage, which could be before the current scope ends.

src/sdam/description/topology/server_selection/mod.rs
```diff
@@ -1,25 +1,91 @@
...
+/// Choose a server from several suitable choices within the latency window according to
+/// the algorithm laid out in the server selection specification.
+fn select_server_in_latency_window(in_window: Vec<&Arc<Server>>) -> Option<Arc<Server>> {
+    if in_window.is_empty() {
+        return None;
+    } else if in_window.len() == 1 {
+        return Some(in_window[0].clone());
+    }
+
+    let mut rng = SmallRng::from_entropy();
```
saghm, line 80: I'm not familiar with the spec changes yet, but if I'm understanding correctly, the algorithm is to choose two random servers within the latency window and then pick the one with fewer operations on it?

src/sdam/description/topology/server_selection/mod.rs
```diff
@@ -1,25 +1,91 @@
...
+    let mut choices = in_window.into_iter().choose_multiple(&mut rng, 2);
+    choices.shuffle(&mut rng);
```
saghm, line 82: I think the shuffling can be avoided by using calling `choose_multiple` [directly on the slice](https://docs.rs/rand/0.7.3/rand/seq/trait.SliceRandom.html#tymethod.choose_multiple) rather than on the iterator.

src/sdam/description/topology/server_selection/mod.rs
```diff
@@ -39,29 +105,17 @@ impl TopologyDescription {
...
-    ) -> Result<Option<&'a ServerDescription>> {
+    ) -> Result<Vec<&'a ServerDescription>> {
         if let Some(message) = self.compatibility_error() {
             return Err(ErrorKind::ServerSelectionError {
                 message: message.to_string(),
             }
             .into());
         }
 
-        // Although `suitable_servers` has to handle the Unknown case for testing, we skip calling
```
saghm, line 53: Given that the new functions above are not actually called by this method anywhere, why do we need to remove these optimizations? Returning an empty or singleton vector will still be correct in these cases, since we'll never have two servers to choose from.

src/sdam/description/topology/server_selection/test/in_window.rs
```diff
@@ -0,0 +1,194 @@
...
+        TestClient,
+        CLIENT_OPTIONS,
+        LOCK,
+    },
+    RUNTIME,
+};
+
+use super::TestTopologyDescription;
+
+#[derive(Debug, Deserialize)]
```
saghm, line 28: Just to make sure, these tests aren't covered by the unified runner, right?

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -1,166 +1,166 @@
...
-                })
-                .unwrap_or_default();
-
-            assert_eq!(parsed_options, actual_options);
-        }
-    }
-
-    run_spec_test(&["initial-dns-seedlist-discovery"], run_test).await;
-}
+// use serde::Deserialize;
```
saghm, line 1: I don't think we should commit the changes to comment out these tests

saghm:
Thanks for splitting out the race condition fix into a separate commit! Let's make sure to merge it as a separate commit as well instead of squashing it

src/sdam/description/topology/server_selection/mod.rs
patrickfreed: We continue using the checked out server until the operation we execute on it is done (see [here](https://github.com/patrickfreed/mongo-rust-driver/blob/RUST-575/health-rebased/src/client/executor.rs#L115) for example). Additionally, I don't believe the compiler is allowed to arbitrarily drop things when they're not being used. They have to be dropped at the end of a "drop scope" which is explcitly defined ([docs](https://doc.rust-lang.org/reference/destructors.html#drop-scopes)). If this wasn't the case, lock guards would be pretty cumbersome to use.

src/sdam/description/topology/server_selection/mod.rs
patrickfreed: Yep, thats it.

src/sdam/description/topology/server_selection/mod.rs
patrickfreed: Ah cool, fixed.

src/sdam/description/topology/server_selection/mod.rs
patrickfreed: I removed the first one since it would no longer be avoiding an allocation, but it looks like we want to keep that one in since we don't want to bother with the predicate for an unknown topology. 

I removed the second one since we were skipping the predicate in that case which is technically a minor bug (a user could decide to fail server selection if the single server doesn't meet their criteria). Also, the logic we're skipping is the same that case for read preferences anyways. 

The last optimization (previously on line 74) is also a bug, since we still need to filter the mongoses based on latency even if the user specifies a Primary read preference.

src/sdam/description/topology/server_selection/test/in_window.rs
patrickfreed: nope, this is a new format

src/test/spec/initial_dns_seedlist_discovery.rs
patrickfreed: Nope, just left them in here to get better evergreen results. It looks like the DNS issue may be fixed, so I'll revert this portion now.

src/sdam/description/topology/server_selection/mod.rs
saghm: Cool, thanks

Cargo.toml
```diff
@@ -41,7 +41,7 @@ lazy_static = "1.4.0"
 md-5 = "0.8.0"
 os_info = { version = "2.0.6", default-features = false }
 percent-encoding = "2.0.0"
-rand = "0.7.2"
+rand = { version = "0.7.2", features = ["small_rng"] }
```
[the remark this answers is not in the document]
saghm: Makes sense. Just to make sure, this is an additive feature, right? We're not making it impossible to use a more secure RNG in other places in the driver?

src/cmap/worker.rs
saghm: Awesome, thanks

src/sdam/description/topology/server_selection/mod.rs
saghm: Interesting; I had thought they dropped that guarantee when they switched to non-lexical lifetimes, but I must have been mistaken.

Cargo.toml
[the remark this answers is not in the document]
patrickfreed: Nope, it's just the SmallRng was gated by a feature flag.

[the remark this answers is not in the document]
saghm: Awesome

saghm (approved):

saghm:
I assume the evergreen failures were due to the now fixed DNS issue, so assuming it's green after a rebase to flatten all of the commits besides the race condition, this looks good to me!

isabelatkinson (approved):

PR #269, opened 2020-11-19
minor: fix lints for rust 1.48

saghm:
Rust 1.48 came out today, and as usual there are a host of new lints

### DIFF
--- a/src/operation/aggregate/test.rs
+++ b/src/operation/aggregate/test.rs
@@ -287,7 +287,7 @@ async fn handle_write_concern_error() {
     let response = CommandResponse::with_document(doc! {
         "cursor" : {
             "firstBatch" : [ ],
-            "id" : 0 as i64,
+            "id" : 0_i64,
             "ns" : "test.test"
         },
         "writeConcernError" : {
--- a/src/test/spec/unified_runner/operation.rs
+++ b/src/test/spec/unified_runner/operation.rs
@@ -34,12 +34,12 @@ use crate::{
 
 #[async_trait]
 pub trait TestOperation: Debug {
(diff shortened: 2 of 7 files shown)


src/operation/aggregate/test.rs
```diff
@@ -287,7 +287,7 @@ async fn handle_write_concern_error() {
     let response = CommandResponse::with_document(doc! {
         "cursor" : {
             "firstBatch" : [ ],
-            "id" : 0 as i64,
+            "id" : 0_i64,
```
saghm, line 290: I knew that you could suffix a literal with the integer type, but I never thought of using one of the underscores that gets ignored to make it more readable before!

src/test/spec/unified_runner/operation.rs
```diff
@@ -34,12 +34,12 @@ use crate::{
 
 #[async_trait]
 pub trait TestOperation: Debug {
-    async fn execute_test_runner_operation<'a>(&self, test_runner: &'a mut TestRunner);
```
isabelatkinson, line 37: Huh this is interesting; the compiler yelled at me for not having these when I originally implemented this but I guess some change along the way made them unnecessary. Good catch by the linter.

isabelatkinson (approved):

src/test/spec/unified_runner/operation.rs
saghm: In general, you won't ever need to annotate lifetimes on a function unless you're returning something that has a lifetime. I'm guessing at some point this method returned something that needed the lifetime to be tied to `test_runner`, but since it doesn't return anything now, it's no longer necessary.

patrickfreed (approved):

PR #271, opened 2020-11-24
RUST-598 Perform handshake when establishing monitoring connections

patrickfreed:
RUST-598

This fixes a bug where monitoring connections didn't perform the MongoDB handshake as part of establishing their connections. I decided to break this off from the other connection storms work since it seemed to be ready and the other stuff is still waiting on spec changes.

### DIFF
--- a/src/cmap/establish/handshake/mod.rs
+++ b/src/cmap/establish/handshake/mod.rs
@@ -147,50 +147,61 @@ impl Handshaker {
     /// Creates a new Handshaker.
     pub(crate) fn new(options: Option<HandshakerOptions>) -> Self {
         let mut metadata = BASE_CLIENT_METADATA.clone();
-
-        if let Some(app_name) = options.as_ref().and_then(|opts| opts.app_name.as_ref()) {
-            metadata.application = Some(AppMetadata {
-                name: app_name.to_string(),
-            });
-        }
-
-        if let Some(driver_info) = options.as_ref().and_then(|opts| opts.driver_info.as_ref()) {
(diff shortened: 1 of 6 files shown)


src/cmap/establish/handshake/mod.rs
```diff
@@ -136,92 +136,155 @@ lazy_static! {
...
-            metadata.driver.name.push('|');
-            metadata.driver.name.push_str(&driver_info.name);
+        let mut body = doc! {
+            "isMaster": 1,
+        };
 
-            if let Some(ref version) = driver_info.version {
-                metadata.driver.version.push('|');
-                metadata.driver.version.push_str(version);
+        if let Some(options) = options {
```
saghm, line 157: Wow, git did a terrible job with calculating the diff here...

src/cmap/establish/handshake/mod.rs
```diff
@@ -136,92 +136,155 @@ lazy_static! {
...
+            driver_info: options.driver_info,
+        }
+    }
+}
+
+/// Run the given isMaster command.
+///
+/// If the given command is not an isMaster, this function will return an error.
+pub(crate) async fn is_master(command: Command, conn: &mut Connection) -> Result<IsMasterReply> {
+    if !command.name.eq_ignore_ascii_case("ismaster") {
```
saghm, line 269: Interesting, I wasn't aware of that string method

saghm (approved):

isabelatkinson (approved):

PR #272, opened 2020-12-03
RUST-611 Port benchmarking suite from `unreviewed-impl` to current state of driver

saghm:
As part of the initial work to get benchmarks in evergreen, I updated the benchmarking suite from the `unreviewed-impl` branch to account for the API changes since it was written (which are mostly due to changing the driver to be async). I made a commit that dropped in the benchmarks as-is and then a subsequent one with my changes both to give @benjirewis credit for his work and to make the review process for the changes easier.


benchmarks/Cargo.toml
```diff
@@ -4,11 +4,19 @@ version = "0.1.0"
...
+
 [dependencies]
 mongodb = { path = ".." }
-bson = "0.13.0"
-serde_json = "1.0.39"
-lazy_static = "1.3.0"
-num_cpus = "1.10.1"
-clap = "2.33.0"
-indicatif = "0.11.0"
+serde_json = "1.0.59"
```
saghm, line 14: I upgraded all of the existing dependencies to their latest versions without running into any issues

benchmarks/src/bench/find_many.rs
```diff
@@ -1,13 +1,11 @@
-use std::{fs::File, path::PathBuf};
+use std::{convert::TryInto, fs::File, path::PathBuf};
 
-use bson::Bson;
-use mongodb::{Client, Collection, Database};
+use anyhow::{bail, Result};
```
saghm, line 3: Since our error types for BSON have changed a bit since last summer and I didn't think it was worth manually updating the boilerplate, I switched this to use [`anyhow`](https://crates.io/crates/anyhow), which is basically a wrapper around `Box<dyn Error>`. While this might be marginally less performant due to allocating for errors, the benchmarks shouldn't ever actually run into errors, and if they do, we don't really care about the performance results, so I think it's fine.

benchmarks/src/bench/find_many.rs
```diff
@@ -21,40 +19,43 @@ pub struct Options {
...
-    fn setup(options: Self::Options) -> Result<Self> {
-        let client = Client::with_uri_str(&options.uri)?;
+    async fn setup(options: Self::Options) -> Result<Self> {
+        let client = Client::with_uri_str(&options.uri).await?;
         let db = client.database(&DATABASE_NAME);
-        db.drop()?;
+        db.drop(None).await?;
 
-        let mut file = File::open(options.path)?;
+        let num_iter = options.num_iter;
```
saghm, line 31: Since the `spawn_blocking_and_await` macro I wrote implicitly wraps the contents in a `move` closure, we can't reuse `options` below. Since `num_iter` is just an int, it's simple to just copy beforehand.

benchmarks/src/bench/find_many.rs
```diff
@@ -21,40 +19,43 @@ pub struct Options {
...
 
-        let json: Value = serde_json::from_reader(&mut file)?;
-        let doc = match json.into() {
+        let mut file = spawn_blocking_and_await!(File::open(options.path))?;
```
saghm, line 33: One interesting part of the Rust async ecosystem is that only network I/O is truly async; the filesystem APIs offered by tokio and async-std are actually just wrappers around their APIs for spawning blocking tasks. Since it would require more work to abstract over each libraries filesystem APIs, I opted to just write a macro that called the `spawn_blocking` function from whichever runtime is enabled and use around the standard library filesystem API. With one exception in the `json_multi_export` tests (see the comment below), this didn't require any functional changes to the benchmarks. We can always come back later and do the work to properly use the runtime-specific filesystem APIs, but I didn't think it was a high enough priority to be worth taking time away from the other parts of this project to do that right now.

benchmarks/src/bench/find_one.rs
```diff
@@ -22,49 +24,46 @@ pub struct Options {
...
-    fn do_task(&self) -> Result<()> {
-        let find_options = FindOptions::builder().limit(Some(1)).build();
+    async fn do_task(&self) -> Result<()> {
         for i in 0..self.num_iter {
-            let mut cursor = self
-                .coll
-                .find(Some(doc! { "_id": i as i32 }), Some(find_options.clone()))?;
-            let _doc = cursor.next();
+            self.coll
+                .find_one(Some(doc! { "_id": i as i32 }), None)
```
saghm, line 58: I had forgotten that we didn't originally have `find_one` in our API!

benchmarks/src/bench/json_multi_export.rs
```diff
@@ -4,106 +4,117 @@ use std::{
...
+        db.drop(None).await?;
 
         let coll = db.collection(&COLL_NAME);
 
+        let mut tasks = FuturesUnordered::new();
+
         for i in 0..TOTAL_FILES {
-            let json_file_name = options.path.join(format!("ldjson{:03}.txt", i));
-            let file = File::open(&json_file_name)?;
+            let path = options.path.clone();
```
saghm, line 40: Importing all this data takes an _extremely_ long time, so I expedited it by using the powers of async to parallelize it. It's somewhat ironic that the setup for the benchmark was so slow that we actually had to do the optimization that we do in the benchmark itself to save time!

benchmarks/src/bench/json_multi_export.rs
```diff
@@ -4,106 +4,117 @@ use std::{
...
     }
 
-    fn do_task(&self) -> Result<()> {
-        let mut num_each = TOTAL_FILES / self.num_threads;
-        let remainder = TOTAL_FILES % self.num_threads;
-        if remainder != 0 {
-            num_each += 1;
-        }
+    async fn do_task(&self) -> Result<()> {
+        let mut tasks = FuturesUnordered::new();
```
saghm, line 68: Previously, this benchmark spawned a number of threads manually and divided up the work among them equally. Now that the driver is async, it's more performant to just spawn a bunch of tasks and then let the runtime do the work of dividing them in a performant manner. Running this one manually a spawnhost showed promising performance gains from this switch, although I'll wait until we can actually verify the results in our automatic runs before doing any victory laps.

benchmarks/src/bench/json_multi_export.rs
```diff
@@ -4,106 +4,117 @@ use std::{
...
+                    .write(true)
+                    .open(&file_name)
+                    .unwrap();
+
+                let mut cursor = coll_ref
+                    .find(Some(doc! { "file": i as i32 }), None)
+                    .await
+                    .unwrap();
+
+                let (sender, mut receiver) = tokio::sync::mpsc::unbounded_channel();
```
saghm, line 91: This is the one place in the benchmarks where the conversion to async required a bit of refactoring. The naive implementation of just using `spawn_blocking` like elsewhere doesn't pass the borrow checker due to the fact that the `File` needs to be moved into the `spawn_blocking` closure in order to be passed to `writeln!`. However, this isn't possible due to needing to invoke `writeln!` multiple times in a loop. This could be mitigated by using the runtime-specific async filesystem APIs, but this would require a decent amount of work due to the fact that tokio 0.2.x (the version the driver still supports) does not use the same async read/write traits that async-std uses, so we'd have to abstract over that. The workaround here is to do all of the async driver work in a separate task, then send the results to a blocking task through a channel to be written out. While this might be less performant than using the runtime-specific filesystem APIs, that optimization would only be in the benchmarks themselves and wouldn't actually change anything for end users, so I didn't think it was worth spending time on right now. If later we want to try to squeeze every ounce of performance out for these benchmarks, we can come back and switch to the runtime APIs.

patrickfreed: Cross driver comparisons of benchmarks are part of the value of running a unified set though, so I think its pretty important we try to cut down on any losses here. The other stuff is fine to leave because it's all in the setup phase. What would be involved in making the requisite abstractions for this test?

benchmarks/src/bench/json_multi_import.rs
```diff
@@ -1,99 +1,85 @@
...
 
-                for i in uploaded_files..uploaded_files + num_files {
-                    let json_file_name = path.join(format!("ldjson{:03}.txt", i));
-                    let file = File::open(&json_file_name).unwrap();
+                let json_file_name = path.join(format!("ldjson{:03}.txt", i));
+                let file: File = spawn_blocking_and_await!(File::open(&json_file_name)).unwrap();
 
-                    let mut new_docs = parse_json_file_to_documents(file).unwrap();
+                let mut new_docs =
+                    spawn_blocking_and_await!(parse_json_file_to_documents(file)).unwrap();
```
patrickfreed, line 65: ditto what would the work be to abstract over the filesystem APIs to implement this?

benchmarks/src/bench/json_multi_export.rs
saghm: Nope; the old version was incredibly slow too from testing it, but we didn't have such low hanging fruit to speed it up like we do now

benchmarks/src/bench/json_multi_export.rs
```diff
@@ -4,106 +4,117 @@ use std::{
...
+                ok
+            });
+        }
 
-            for mut doc in docs {
-                doc.insert("file", i as i32);
-                coll.insert_one(doc, None)?;
-            }
+        while let Some(result) = tasks.next().await {
+            println!("done!");
```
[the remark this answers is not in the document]
saghm: Oops, this was for my own testing locally. I'll remove it.

benchmarks/src/bench/json_multi_export.rs
saghm: Basically the same type of stuff we do in the driver for abstracting over tokio/async-std API differences, e.g. with the stream API. I feel somewhat strongly that working to improve the benchmarks themselves but not the driver isn't a priority of this project; right now, we just want to make as big an impact on the actual driver's performance, and putting these benchmarks in evergreen is just a way to measure that. Put another way, time we spend optimizing this benchmark is time we're not spending on stuff like the raw document type, which would be a win for actual users and not just the benchmarks.

benchmarks/src/bench/json_multi_import.rs
saghm: Answered above

benchmarks/src/bench/json_multi_export.rs
patrickfreed: I agree that optimizing the driver is the highest priority here, but my concern is that if the benchmark suite is artificially slow in a way that affects the results, it makes comparisons against a baseline harder since those baselines come from other drivers which do not have any such overhead. These baseline comparisons are really important, since they guide our expectations and measures of success for our implementations. For instance, we'd like our BSON benchmarks to be roughly on par or better than Java, say, but if they're much slower, we won't be able to tell if its coming from the BSON implementation or the benchmark without further investigation, which also costs time. And when we do fix the runner in the future, we would no longer use our old benchmark data as a baseline either. Another issue is correctness--if our benchmarks aren't using async in the way we expect users to, then it's harder to say if the benchmarks are reflective of real-world performance.

The reason I was wondering how much work would be involved is because if it's a few hours or a day, then it's probably worth it to ensure we have accurate benchmark results, since we likely won't have an opportunity to update these again for a long time. I haven't looked too much into the filesystem APIs, but I can't imagine it's that much different than our socket abstractions which are quite small. Also, we wouldn't have to abstract over the entirety of the filesystem API, but just the pieces that we're actively measuring.

saghm:
Per offline discussion, this PR has been updated to wrap around the runtime-specific filesystem APIs where possible. Please review again once you have a chance!

patrickfreed (approved):
LGTM!

isabelatkinson (commented):
looks good, just a few questions! (also "unreviewed" is misspelled in the PR name)

benchmarks/src/bench/json_multi_export.rs
saghm: Based on offline discussion, all places that are measured by the benchmarking have been updated to use the runtime-specific APIs for filesystem access.

benchmarks/README.md
```diff
@@ -0,0 +1,40 @@
...
+| Find many and empty the cursor | 5  |
+| Small doc bulk insert          | 6  |
+| Large doc bulk insert          | 7  |
+| LDJSON multi-file import       | 8  |
+| LDJSON multi-file export       | 9  |
+| All benchmarks                 | all|
+
+Note that in order to compare against the other drivers, an inMemory mongod instance should be used.
+
+At this point, BSON and GridFS benchmarks are not implemented because we do not own the Rust BSON library, and GridFS has not been implemented
```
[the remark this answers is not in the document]
saghm: I'll update this in my PR for the BSON benchmarks

benchmarks/src/error.rs
```diff
@@ -0,0 +1 @@
+
```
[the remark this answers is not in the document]
saghm: Oops, nope! I thought I had deleted it from git history, but I guess I just deleted the contents but not the file.

isabelatkinson (approved):
lgtm mod the typo in the commit message

saghm:
I think the commit message is right, but the PR title is wrong. I'll fix it though!

PR #273, opened 2020-12-04
RUST-610 Test that monitors wait at least minHeartbeatFrequencyMS between failed checks

patrickfreed:
RUST-610

This PR implements a new SDAM spec prose test that ensures the monitors wait at least 500ms between failed checks. This is part of the Avoiding Connection Storms project (DRIVERS-781).

relevant spec change: https://github.com/mongodb/specifications/commit/6c5b2acbf5a30b2c79e904532d51ba9ed8783247

### DIFF
--- a/src/sdam/test.rs
+++ b/src/sdam/test.rs
@@ -1,4 +1,7 @@
-use std::{sync::Arc, time::Duration};
+use std::{
+    sync::Arc,
+    time::{Duration, Instant},
+};
 
 use bson::doc;
 use semver::VersionReq;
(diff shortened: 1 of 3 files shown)


patrickfreed:
It looks like only one of the errors in the CI run is related to this test, and it appears that the driver hit the 5 second server selection timeout, which is a little concerning. I'm not sure how to "fix" the test though without reducing its efficacy.

failure: https://evergreen.mongodb.com/task_log_raw/mongo_rust_driver_tests__async_runtime~async_std_auth_and_tls~auth_and_tls_os~ubuntu_16.04_test_latest_standalone_patch_32d67deb1aa970a385e16a1142ec072a1d6f0be7_5fcac77357e85a0f8b1ad7c5_20_12_04_23_34_12/0?type=T#L1403

src/sdam/test.rs
```diff
@@ -0,0 +1,66 @@
...
+        return;
+    }
+
+    if setup_client.server_version_lt(4, 9) {
+        println!("skipping min_heartbeat_frequency test due to server version being less than 4.9");
+        return;
+    }
+
+    let fp_options = FailCommandOptions::builder()
+        .app_name("SDAMMinHeartbeatFrequencyTest".to_string())
```
saghm, line 33: Interesting, I've never seen this before! I assume that this makes it possible to only fail commands for a certain client (assuming we use unique app names). Would we be able to use this to remove the need for a global test lock, or is this only supported on newer server versions?

patrickfreed: It's a relatively new option (introduced in 4.2.9, started working for initial handshakes in 4.9). Due to those version restrictions, I don't think we can use it to replace the locks unfortunately.

isabelatkinson (approved):

saghm (approved):

patrickfreed:
This will likely have merge conflicts with #278 , and I think it'll be easier to rebase this on top of that rather than the other way around, so I'm going to hold off on merging this until that goes in.

PR #277, opened 2020-12-10
RUST-617 Add BSON benchmarks

saghm:
It looks like in my previous rebase and merge, I accidentally left off the changes to using the runtime-specific I/O APIs, so I included it here as a separate commit from the BSON work (and tagged separately with the ticket number for the previous PR as well).


benchmarks/src/main.rs
```diff
@@ -270,6 +272,95 @@ async fn parallel_benchmarks(uri: &str, more_info: bool, ids: &[bool]) -> Result
...
+    if ids[14] {
+        let bson_full_encode_options = bench::bson_encode::Options {
+            num_iter: 10_000,
+            path: DATA_PATH.join("extended_bson").join("full_bson.json"),
+        };
+        println!("Running BSON full encode...");
+        let bson_full_encode =
+            bench::run_benchmark::<BsonEncodeBenchmark>(bson_full_encode_options).await?;
+
+        comp_score += score_test(bson_full_encode, "BSON full encode", 57.34, more_info);
```
patrickfreed, line 357: From looking at the implementation of `score_test` it appears to be a little broken in that it relies on `MAX_ITERATIONS`  when it doesn't need to. The only iteration count it should consider is the number of iterations that actually occurred. This is something that I should've caught last time, but fortunately there's a commit in this PR for last ticket anyways :).

benchmarks/src/main.rs
```diff
@@ -394,6 +497,11 @@ Run benchmarks by id number (comma-separated):
         comp_score += parallel_benchmarks(uri, verbose, &ids).await.unwrap();
     }
 
+    // Bson
+    if ids[9] || ids[10] || ids[11] || ids[12] || ids[13] {
```
patrickfreed, line 501: i think this is missing `ids[14]` too.

benchmarks/README.md
```diff
@@ -2,7 +2,8 @@
 
 This suite implements the benchmarks described in this (spec)[https://github.com/mongodb/specifications/blob/master/source/benchmarking/benchmarking.rst].
 
-In order to run the microbenchmarks, first run `../etc/microbenchmark-test-data.sh` to download the data.
+In order to run the microbenchmarks, first run `./download-data.sh`. (NOTE: the data for the deeply nested BSON encoding and decoding is
+currently broken, so these benchmarks will not be runnable until that's fixed).
```
patrickfreed, line 6: in the meantime, we can just use the old version (pre-corruption) since it doesn't use any extJSON v2. We just checked it into the Swift benchmarking repo (it's really small: https://github.com/mongodb/mongo-swift-driver-performance/blob/main/data/deep_bson.json), so I think we can do the same here temporarily. I think it's probably important that we do these deep_bson benchmarks if we're planning on BSON being the biggest bottleneck in the driver right now.

kmahar: @saghm if you want you can also just do SPEC-1752 yourself. it should be really quick to replace the file and re-upload the fixed tarball, I just haven't been able to prioritize it. if not I can do it myself sometime this week

saghm: Yep, I can do that. The only reason I didn't go ahead and do it before is that I couldn't actually find the correct file anywhere, but now that Patrick linked it, I'll make the PR to the specs repo.

benchmarks/src/main.rs
saghm: Updated to use `len() / 2` instead of `MAX_ITERATIONS / 2`.

benchmarks/src/main.rs
```diff
@@ -270,6 +272,95 @@ async fn parallel_benchmarks(uri: &str, more_info: bool, ids: &[bool]) -> Result
...
+        };
+        println!("Running BSON full encode...");
+        let bson_full_encode =
+            bench::run_benchmark::<BsonEncodeBenchmark>(bson_full_encode_options).await?;
+
+        comp_score += score_test(bson_full_encode, "BSON full encode", 57.34, more_info);
+    }
+
+    println!("\nBSON benchmark composite score: {}\n", comp_score);
+    Ok(comp_score)
```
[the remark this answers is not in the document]
saghm: Good catch! This will improve the composite results quite a bit as well

benchmarks/src/main.rs
saghm: Ah, yep. Fixed.

benchmarks/src/main.rs
```diff
@@ -394,6 +497,11 @@ Run benchmarks by id number (comma-separated):
         comp_score += parallel_benchmarks(uri, verbose, &ids).await.unwrap();
     }
 
+    // Bson
+    if ids[9] || ids[10] || ids[11] || ids[12] || ids[13] {
+        comp_score += bson_benchmarks(verbose, &ids).await.unwrap();
```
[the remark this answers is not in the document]
saghm: Good catch; fixed

saghm (commented):
Just pushed the requested changes. I kept the changes for the existing code separate from the BSON  changes so that the rebase will be easier once this is reviewed, but I'll hold off on reordering the commits so that the non-BSON changes are all first until the review is complete.

benchmarks/src/main.rs
```diff
@@ -61,8 +49,8 @@ fn get_nth_percentile(durations: &[Duration], n: f64) -> Duration {
     durations[std::cmp::max(index, 1) - 1]
 }
 
-fn score_test(durations: Vec<Duration>, name: &str, task_size: f64, more_info: bool) -> f64 {
-    let median = get_nth_percentile(&durations, (*MAX_ITERATIONS / 2) as f64);
+fn score_test(mut durations: Vec<Duration>, name: &str, task_size: f64, more_info: bool) -> f64 {
+    let median = get_nth_percentile(&durations, durations.len() / 2 as f64);
```
patrickfreed, line 53: given that this function accepts a percentile, don't actually need to use len at all and can just say `50.0`.

benchmarks/src/main.rs
```diff
@@ -61,8 +49,8 @@ fn get_nth_percentile(durations: &[Duration], n: f64) -> Duration {
     durations[std::cmp::max(index, 1) - 1]
```
patrickfreed, line 49: this function needs to update its usage of `MAX_ITERATIONS` to `durations.len()`. Once this is fixed we can remove the restriction in `score_test` that the various other percentiles are only used when `MAX_ITERATIONS == 100`

benchmarks/src/main.rs
```diff
@@ -61,8 +49,8 @@ fn get_nth_percentile(durations: &[Duration], n: f64) -> Duration {
...
 }
 
-fn score_test(durations: Vec<Duration>, name: &str, task_size: f64, more_info: bool) -> f64 {
-    let median = get_nth_percentile(&durations, (*MAX_ITERATIONS / 2) as f64);
+fn score_test(mut durations: Vec<Duration>, name: &str, task_size: f64, more_info: bool) -> f64 {
```
patrickfreed, line 52: Is this `mut` necessary? It seemed to pass the linter but I don't see anywhere durations is mutated.

benchmarks/src/main.rs
patrickfreed: That wasn't quite what I was suggesting. I left a few more specific comments up there on the actual code now that there are diffs up there (:shakefist: at GitHub for not letting me comment outside of diffs).

benchmarks/src/main.rs
saghm: Ah, true! I think it might actually be incorrect like this when the length is not 100

benchmarks/src/main.rs
saghm: Oops, not, it's not. I added it when sorting to the vector inside the method, but then later realized it was sorted elsewhere, but I forgot to remove the `mut` here. I'll do that now.

benchmarks/src/main.rs
saghm: I'm not sure you're correct about this one; the specification says the following about the formula:

```
For each micro-benchmark, the 10th, 25th, 50th, 75th, 90th, 95th, 98th and 99th percentiles will be recorded using the following algorithm:

* Given a 0-indexed array A of N iteration wall clock times
* Sort the array into ascending order (i.e. shortest time first)
* Let the index i for percentile p in the range [1,100] be defined as: i = int(N * p / 100) - 1
```

In this case, the algorithm's `n` is `durations.len()` in Rust, the algorithm's `p` is (confusingly) `n` in Rust, and `100` is `MAX_ITERATIONS` in Rust (because we allow the user to configure it). If we just used `durations.len()` again, it would be `int(N * p / N) - 1`, which is just `p - 1`. That would work if we always completed 100 measurements, but if we did, we wouldn't need this formula in the first place! The idea here is that if we _don't_ complete 100 measurements, we estimate calculate which index in the sorted list of durations is closest to each percentile; for example, if we only finished 20 runs, then the 50th percentile would be index `(20 * 50 / 100) - 1`, or index `9` in the vector, or the 10th element out of 20.

patrickfreed: The 100 here is just converting `p` which is in [1,100] to an actual percentage in [0.01, 1], it's just a coincidence that `MAX_ITERATIONS` is also 100. You're right that the `durations.len()` in the numerator needs to be removed too, though. The algorithm here is as simple as `index = array length * fraction - 1`, where fraction is `p/100` and `array length` is `N` or the number of iterations. So if we're looking for the median, fraction is 0.5 and we get the middle element.

As a side note, `MAX_ITERATIONS` is somewhat poorly named, since it isn't necessarily a maximum. It's also not a minimum though. I'm not exactly sure what to call it.

saghm: Okay, I'll update it. As for a new name, how about `TARGET_ITERATION_COUNT`?

patrickfreed: Yeah that sounds good :+1:

saghm:
Ready for review again

benchmarks/src/main.rs
patrickfreed: I think the `as usize` needs to come after the division. As it is now, `50.0 as usize / 100` will be 0.

benchmarks/README.md
patrickfreed: Once https://github.com/mongodb/specifications/pull/889 goes in can we update this accordingly?

patrickfreed (commented):
Looks good, just a few minor comments

benchmarks/README.md
saghm: Definitely! I think it should probably be possible to merge that before this so we don't have to do a follow-up commit

benchmarks/src/main.rs
saghm: Fixed

patrickfreed: Sorry for the extra round trip, but i think we also need an `as f64` for the multiplication portion, and then the end result is the only thing cast to `usize`.

benchmarks/src/main.rs
```diff
@@ -42,31 +28,33 @@ use clap::{App, Arg, ArgMatches};
...
 }
 
 fn score_test(durations: Vec<Duration>, name: &str, task_size: f64, more_info: bool) -> f64 {
-    let median = get_nth_percentile(&durations, (*MAX_ITERATIONS / 2) as f64);
+    let median = get_nth_percentile(&durations, 50.0);
     let score = task_size / (median.as_millis() as f64 / 1000.0);
     println!("TEST: {} -- Score: {}\n", name, score);
 
-    if more_info && *MAX_ITERATIONS == 100 {
+    if more_info && *TARGET_ITERATION_COUNT == 100 {
```
patrickfreed, line 57: Now that `get_nth_percentile` is fixed, we can drp the TARGET_ITERATION_COUNT requirement of this condition.

benchmarks/src/main.rs
```diff
@@ -278,15 +283,118 @@ async fn parallel_benchmarks(uri: &str, more_info: bool, ids: &[bool]) -> Result
...
+        };
+        println!("Running BSON flat encode...");
+        let bson_flat_encode =
+            bench::run_benchmark::<BsonEncodeBenchmark>(bson_flat_encode_options).await?;
+
+        comp_score += score_test(bson_flat_encode, "BSON flat encode", 75.31, more_info);
+        benchmark_count += 1;
+    }
+
+    // BSON deep document encode
```
patrickfreed, line 332: encode -> decode

benchmarks/src/main.rs
```diff
@@ -278,15 +283,118 @@ async fn parallel_benchmarks(uri: &str, more_info: bool, ids: &[bool]) -> Result
...
+        };
+        println!("Running BSON deep encode...");
+        let bson_deep_encode =
+            bench::run_benchmark::<BsonEncodeBenchmark>(bson_deep_encode_options).await?;
+
+        comp_score += score_test(bson_deep_encode, "BSON deep encode", 19.64, more_info);
+        benchmark_count += 1;
+    }
+
+    // BSON full document encode
```
patrickfreed, line 360: encode -> decode

benchmarks/src/main.rs
saghm: Done

benchmarks/src/main.rs
saghm: Fixed

benchmarks/src/main.rs
saghm: Fixed

patrickfreed (approved):
LGTM!

isabelatkinson (approved):

PR #278, opened 2020-12-11
RUST-590 Allow pool to be paused

patrickfreed:
RUST-590

This PR updates the connection pool to support the "pausing" feature introduced in https://github.com/mongodb/specifications/pull/878 (SPEC-1530 / SPEC-1242). This allows the pool to halt operations (i.e. return errors from `check_out`, stop populating minPoolSize) when the endpoint it is associated with goes down and also before it is discovered to be in a known state. This also allows the pool to fully clear its WaitQueue when the pool is cleared. 

This is the last PR as part of Avoiding Connection Storms (DRIVERS-781)


src/client/executor.rs
```diff
@@ -82,16 +82,7 @@ impl Client {
...
-            Ok(conn) => conn,
-            Err(err) => {
-                self.inner
-                    .topology
-                    .handle_pre_handshake_error(err.clone(), server.address.clone())
-                    .await;
-                return Err(err);
-            }
-        };
+        let mut conn = server.pool.check_out().await?;
```
patrickfreed, line 85: In order to unify establishment error handling, I removed any special handling from in the executor and had the monitor do it instead by listening to a channel broadcasted by the pool itself. This allows us to use the same error handling regardless of whether a connection was established as part of operation execution or via the background thread.

src/cmap/connection_requester.rs
```diff
@@ -31,112 +26,87 @@ pub(super) fn channel(
 #[derive(Clone, Debug)]
 pub(super) struct ConnectionRequester {
     address: StreamAddress,
-    sender: mpsc::UnboundedSender<oneshot::Sender<RequestedConnection>>,
+    sender: mpsc::UnboundedSender<oneshot::Sender<ConnectionRequestResult>>,
     handle: PoolWorkerHandle,
 }
 
 impl ConnectionRequester {
```
patrickfreed, line 33: The error handling was split/duplicated between the `ConnectionRequester` and `ConnectionPool::check_out`, and it was pretty confusing. To simplify this, I moved most of the logic to `check_out` and only kept the parts in here that involved communicating with the worker thread. Once a response was received, it is returned and then processed for errors and what not by `check_out`.

src/cmap/manager.rs
```diff
@@ -24,6 +24,22 @@ impl PoolManager {
         let _ = self.sender.send(PoolManagementRequest::Clear);
     }
 
+    /// Mark the pool as "ready" as per the CMAP specification.
+    pub(super) async fn mark_as_ready(&self) {
+        let (sender, receiver) = tokio::sync::oneshot::channel();
+        if self
+            .sender
+            .send(PoolManagementRequest::MarkAsReady {
+                completion_handler: sender,
```
patrickfreed, line 33: We use a completion handler here because we don't want to mark the server as known before we know the pool is ready, otherwise we could select a server whose pool can't process  `check_out` requests.

src/cmap/status.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn default() -> Self {
+        PoolStatus {
+            generation: 0,
+            establishment_error: None,
+        }
+    }
+}
+
+/// Create a channel for publishing and receiving updates to the pool's generation.
+pub(super) fn channel() -> (PoolGenerationPublisher, PoolGenerationSubscriber) {
```
patrickfreed, line 24: This channel is currently used by the monitor to detect connection establishment errors, but it will also be used in the future to get the current generation as part of implementing the rest of RUST-355.

Happy to discuss design / naming options for the types in this file.

src/cmap/worker.rs
```diff
@@ -173,7 +205,7 @@ impl ConnectionPoolWorker {
 
         loop {
             let task = tokio::select! {
-                Some(request) = self.request_receiver.recv(), if self.can_service_connection_request() => {
```
patrickfreed, line 176: Because we need to be able to clear the WaitQueue now, we need to buffer the requests into a separate data structure instead of leaving them in the channel.

src/cmap/worker.rs
```diff
@@ -191,13 +223,43 @@ impl ConnectionPoolWorker {
...
+                PoolTask::HandleManagementRequest(
+                    PoolManagementRequest::HandleConnectionFailed {
+                        error,
+                        error_generation,
+                    },
+                ) => self.handle_connection_failed(error, error_generation),
                 PoolTask::Maintenance => self.perform_maintenance(),
             }
+
+            if self.can_service_connection_request() {
```
patrickfreed, line 268: We still only need to process them after a task is performed, since this worker task has exclusive access to the inner state. If we can't process a connection request at a certain point, we still won't be able to process any more unless another task has been performed that might change that.

src/cmap/worker.rs
```diff
@@ -308,15 +377,16 @@ impl ConnectionPoolWorker {
...
-            self.clear();
-        }
+    fn handle_connection_failed(&mut self, error: Error, error_generation: u32) {
         // Establishing a pending connection failed, so that must be reflected in to total
         // connection count.
         self.total_connection_count -= 1;
         self.pending_connection_count -= 1;
+
+        // if this error has not been reported yet, clear the pool.
+        if self.generation == error_generation {
```
patrickfreed, line 397: this is a portion of RUST-355 that is required by the DRIVERS-781 changes.

src/cmap/worker.rs
```diff
@@ -462,41 +563,35 @@ async fn establish_connection(
...
+    /// Connections may not be checked out nor created in the background to satisfy minPoolSize.
+    Paused,
+
+    /// Pool is operational.
+    Ready,
+}
+
 /// Task to process by the worker.
 #[derive(Debug)]
 enum PoolTask {
```
patrickfreed, line 591: It was getting pretty annoying to maintain essentially two copies of this enum, so I just moved all the management request ones into a single case.

src/test/spec/runner/test_event.rs
```diff
@@ -48,46 +36,16 @@ impl Matchable for TestEvent {
                 }
                 actual_command.matches(expected_command)
             }
-            (
```
patrickfreed, line 51: The v2 runner only actually asserts on command started events, so I removed the others to simplify things.

src/test/util/event.rs
```diff
@@ -68,46 +75,110 @@ impl CommandEvent {
     }
 }
 
-#[derive(Default)]
+#[derive(Clone, Debug)]
```
patrickfreed, line 78: This is part of the work involved in RUST-572, but I needed it now for the two prose tests introduced in `src/sdam/test.rs`. There is some copy pasted code here from the CMAP runner's implementation of this. We can remove the duplicate in RUST-572 and just use the single `EventHandler` then.

src/cmap/test/mod.rs
```diff
@@ -250,17 +283,19 @@ impl Operation {
...
-                    // wait for event to be emitted to ensure drop has completed.
+
                     subscriber
-                        .wait_for_event(EVENT_TIMEOUT, |e| {
-                            matches!(e, Event::ConnectionPoolCleared(_))
-                        })
+                        .wait_for_generation_change(EVENT_TIMEOUT)
                         .await
-                        .expect("did not receive ConnectionPoolCleared event after clearing pool");
+                        .expect("generation did not change after clearing pool");
```
patrickfreed, line 293: Because clearing a cleared pool doesn't necessarily emit an event but always bumps the generation, we use the generation subscriber here to determine if the call occurred or not.

src/sdam/monitor.rs
```diff
@@ -1,82 +1,132 @@
...
-
-        RUNTIME.execute(async move {
-            monitor.execute().await;
-        });
+            connection: None,
+        }
     }
 
-    async fn execute(&mut self) {
+    async fn execute(&mut self, weak_topology: WeakTopology) {
```
saghm, line 95: Out of curiosity, what's the rationale for having the topology be a parameter to execute instead of a field on the struct? (Likewise for UpdateMonitor below)

src/sdam/state/mod.rs
```diff
@@ -254,33 +270,36 @@ impl Topology {
...
+        &self,
+        error: Error,
+        server: &Server,
+        state_lock: RwLockWriteGuard<'_, TopologyState>,
+    ) -> bool {
+        let description = ServerDescription::new(server.address.clone(), Some(Err(error)));
+        self.update_and_notify(description, state_lock).await
     }
 
-    /// Updates the topology using the given `ServerDescription`. Monitors for new servers will
```
saghm, line 268: Is there a reason for deleting the doc comment here? If it no longer applies, I think a new one should be written to replace it

src/sdam/state/mod.rs
```diff
@@ -254,33 +270,36 @@ impl Topology {
...
-    /// be started as necessary.
-    ///
-    /// Returns true if the topology changed as a result of the update and false otherwise.
-    pub(crate) async fn update(&self, server_description: ServerDescription) -> bool {
+    async fn update_and_notify(
```
saghm, line 291: Can this method get a doc comment too? I think the SDAM machinery can be tricky to understand for someone not already used to it, so having all these methods documented helps make that easier.

src/sdam/state/mod.rs
```diff
@@ -372,17 +412,19 @@ impl TopologyState {
     /// Adds a new server to the cluster.
     ///
     /// A reference to the containing Topology is needed in order to start the monitoring task.
-    fn add_new_server(&mut self, address: StreamAddress, options: ClientOptions) {
+    fn add_new_server(
```
saghm, line 422: Can you update the doc comment to indicate that a monitor is started as part of this method call to help prevent us from accidentally calling it in tests?

src/sdam/state/server.rs
```diff
@@ -29,31 +33,24 @@ impl Server {
         }
     }
 
-    pub(crate) fn new(
+    pub(crate) fn create(
```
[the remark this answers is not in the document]
saghm: This seems fine to me, although I think it would be useful to add a doc comment indicating that the monitoring tasks are started as part of this method.

More generally, since we're starting to get in the habit of starting larger numbers of tasks, I think we should also make it a habit to note that in the doc comments of a method when doing that.

src/cmap/options.rs
```diff
@@ -45,6 +45,12 @@ pub struct ConnectionPoolOptions {
     #[serde(skip)]
     pub event_handler: Option<Arc<dyn CmapEventHandler>>,
 
+    /// How often the background thread performas its maintenance (e.g. ensure minPoolSize).
```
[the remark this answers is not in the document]
patrickfreed: fixed

src/cmap/test/event.rs
```diff
@@ -110,6 +114,20 @@ impl EventSubscriber<'_> {
             .ok()
             .flatten()
     }
+
+    /// Returns the received events without waiting for any more.
+    pub fn all<F>(&mut self, filter: F) -> Vec<Event>
+    where
+        F: Fn(&Event) -> bool,
+    {
+        let mut events = Vec::new();
```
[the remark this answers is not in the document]
patrickfreed: So this invocation would wait for the entire stream to be consumed and then collect all the results I think (`collect` is `async`), whereas we just want to aggregate any events that are buffered in the stream already without waiting for any more.

src/cmap/test/file.rs
```diff
@@ -59,14 +60,31 @@ impl ThreadedOperation {
...
+    Wait {
+        ms: u64,
+    },
+    WaitForThread {
+        target: String,
+    },
+    WaitForEvent {
+        event: String,
+        count: usize,
+        #[serde(deserialize_with = "bson_util::deserialize_duration_from_u64_millis")]
```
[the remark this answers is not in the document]
patrickfreed: Yeah I think so. It would be useful for us in the driver at least. cc @isabelatkinson

src/sdam/monitor.rs
patrickfreed: I moved the creation of the `Monitor` struct to happen at the same time as the creation of the `Server`, but when a `Topology` is being created itself it makes a bunch of `Server`s first to avoid having to access a lock and make `Topology::new` async. Now that I think about it though, we can use `futures::executor::block_on` like we do for `watch` channels since that lock would be acquired immediately, and we can avoid this situation altogether. Should I go ahead and make that change?

src/sdam/state/mod.rs
patrickfreed: So the git diff is a little weird here. This method and doc comment were actually just moved down the file. That being said, this method does need a docstring so I added one.

This also made me realize `update_and_notify` should be handling marking the pool as ready, since it is the common entry point to updating the topology. the `pub(crate) update` is now just the internal API for doing so (as used by monitors).

src/sdam/state/mod.rs
patrickfreed: Yep done.

src/sdam/state/mod.rs
patrickfreed: I think the existing one conveys that, no?

>/// A reference to the containing Topology is needed in order to start the monitoring task.

What's funny is I did not add this comment, so it seems like this method used to start monitors but then it stopped doing so at one point.

src/sdam/state/server.rs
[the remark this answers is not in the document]
patrickfreed: So this method (currently) doesn't start any monitoring tasks, though it probably should. The only reason it doesn't is that weird lock shenanigans I mentioned in my other response to one of your comments. If we go ahead with that change I'll be sure to update this docstring to indicate the monitor gets started.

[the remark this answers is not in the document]
saghm: Oh, I missed the fact that monitors need to be explicitly started now rather than starting when creating. Never mind!

src/sdam/state/mod.rs
saghm: That's weird, I totally missed that during review. Maybe the github diff was hiding it.

src/sdam/monitor.rs
saghm: Yep, that sounds good to mean, provided that there's a comment indicating why `block_on` is fine similar to the one for the channels.

src/sdam/state/mod.rs
saghm: Ah yeah, sorry about that. Adding a new comment is fine.

src/sdam/monitor.rs
patrickfreed: So it looks like the `Topology::new_from_hosts` also needs the ability to create a server without starting a monitor, so I'm thinking it's probably best if we just leave it as it is.

saghm: `new_from_hosts` is just used for tests IIRC, so we should feel free to work around if (e.g. write a new method or something) if needed.

src/sdam/state/server.rs
```diff
@@ -33,6 +33,8 @@ impl Server {
         }
     }
 
+    /// Create a new reference counted `Server` instance and a `Monitor` for that server.
+    /// The monitor is not started as part ofthis; call `Monitor::execute` to start it.
```
saghm, line 37: Missing a space between "of" and "this"

src/cmap/test/event.rs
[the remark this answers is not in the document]
saghm: Ah, looks like you're right. That's a little disappointing; this feels like the exact type of thing we should be able to do with the Stream API instead of having to do it manually.

src/sdam/monitor.rs
patrickfreed: Updated to store the `WeakTopology` on the monitor. I kept the two phase `execute` (which now takes no parameters) to allow for `new_from_hosts` to create servers without starting the monitors.

src/sdam/state/mod.rs
patrickfreed: All these methods have docstrings now.

src/sdam/state/server.rs
patrickfreed: fixed

src/sdam/state/mod.rs
```diff
@@ -87,26 +87,36 @@ impl Topology {
...
             state: Arc::new(RwLock::new(state)),
             common,
-        }
+        };
+
+        // we can block in place here because we're the only ones with access to the lock, so it
+        // should be acquired immediately.
+        let mut topology_state = RUNTIME.block_in_place(topology.state.write());
+
+        hosts.into_iter().for_each(|address| {
```
saghm, line 105: Out of curiosity, why not use `for address in hosts` instead?

saghm:
Looks like there are a couple of clippy failures; other than that (and my one clarifying question), this looks good to me!

src/sdam/state/mod.rs
patrickfreed: This used to `collect` into something so it was the fewest character change to just make it a `for_each` I think. Anyways rewritten to `for in`.

isabelatkinson (approved):

saghm (approved):

PR #281, opened 2021-01-06
minor: fix clippy errors on 1.49

saghm:


src/cmap/test/integration.rs
```diff
@@ -122,20 +122,22 @@ async fn concurrent_connections() {
...
-    let mut consecutive_creations = 0;
-    for event in events.iter() {
-        match event {
-            Event::ConnectionCreated(_) => {
-                consecutive_creations += 1;
-            }
-            Event::ConnectionReady(_) => {
-                if consecutive_creations < 2 {
-                    panic!("connections not created concurrently");
+    {
```
saghm, line 125: I tried using `drop(events)` on line 141 instead of using a block, but clippy strangely still gave the error about `events` being held across an `await` point

patrickfreed: Yeah this appears to be a bug in clippy: https://github.com/rust-lang/rust-clippy/issues/6446

patrickfreed (approved):

isabelatkinson (approved):

src/cmap/test/integration.rs
saghm: Ah, nice catch

PR #284, opened 2021-01-29
minor: hide versioned API from public API

patrickfreed:
In preparation of releasing the alpha, hide the not-yet-complete versioned API implementation from the public API.

### DIFF
--- a/src/cmap/options.rs
+++ b/src/cmap/options.rs
@@ -85,8 +85,8 @@ pub struct ConnectionPoolOptions {
     /// The declared API version
     ///
     /// The default value is to have no declared API version
-    #[builder(default)]
-    pub server_api: Option<ServerApi>,
+    #[builder(skip, default)]
+    pub(crate) server_api: Option<ServerApi>,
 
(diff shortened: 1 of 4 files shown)


src/cmap/options.rs
```diff
@@ -85,8 +85,8 @@ pub struct ConnectionPoolOptions {
     /// The declared API version
     ///
     /// The default value is to have no declared API version
-    #[builder(default)]
-    pub server_api: Option<ServerApi>,
+    #[builder(skip, default)]
```
saghm, line 88: The tiniest of nits: can we be consistent about the order when doing both `default` and `skip`, or does it actually matter?

patrickfreed: I don't believe the order is significant. For the sake of not aborting the evergreen patch in progress, is it alright if I leave it as-is? The `skip` tag will be removed in the future anyways.

saghm: Sure, I guess so :/

isabelatkinson (approved):

saghm (approved):

PR #285, opened 2021-02-01
RUST-603 Lint documentation in CI

isabelatkinson:


saghm (approved):

patrickfreed (commented):
Can we enable the [`missing_docs`](https://doc.rust-lang.org/rustdoc/lints.html#missing_docs) and `missing_crate_level_docs` lints? They're disabled by default so we currently aren't checking for them.

src/error.rs
```diff
@@ -205,6 +205,7 @@ impl std::ops::Deref for Error {
 }
 
 /// The types of errors that can occur.
+#[allow(missing_docs)]
```
isabelatkinson, line 208: This suppresses warnings for missing comments on all of the `message` fields -- doesn't seems super necessary/helpful to users to have an individual comment for each of these fields given they're fairly self-explanatory.

src/selection_criteria.rs
```diff
@@ -93,12 +93,15 @@ pub type Predicate = Arc<dyn Send + Sync + Fn(&ServerInfo) -> bool>;
 /// option and will be sent to the server as an integer number of seconds.
 ///
 /// See the [MongoDB docs](https://docs.mongodb.com/manual/core/read-preference) for more details.
+#[allow(missing_docs)]
```
isabelatkinson, line 96: Ditto above for all of the `options` fields.

patrickfreed (approved):
LGTM!

PR #290, opened 2021-02-12
release 1.2.0

isabelatkinson:


saghm (commented):
The code LGTM! The only request I have is to swap the order of the clippy lint fix commit and the "release 1.2.0" commit so that the latter is the latest commit of the actual 1.2.0 release.

patrickfreed:
Before we can go ahead with this release, we have to do a 1.2.0 release of the BSON library, then we can update our `Cargo.toml` to use that new release instead of 1.1.0.

patrickfreed (approved):
LGTM :tada:

saghm (approved):

PR #291, opened 2021-02-12
minor: fix clippy

isabelatkinson:


src/runtime/mod.rs
```diff
@@ -46,6 +46,7 @@ impl AsyncRuntime {
     /// If the runtime is still running, this will return a handle to the background task.
     /// Otherwise, it will return `None`. As a result, this must be called from an async block
     /// or function running on a runtime.
+    #[allow(clippy::unnecessary_wraps)]
```
isabelatkinson, line 49: clippy is producing an `unnecessary_wraps` error here because the return value is always wrapped in `Some` for async-std; however, this isn't true for tokio so rather than changing the method signature based on the feature I just suppressed the warning.

saghm (approved):

patrickfreed (approved):

PR #297, opened 2021-02-22
RUST-682 Update encryption dependencies

bugadani:
This PR leaves PBKDF2 at 0.6 instead of 0.7 to avoid bumping MSRV to 1.47.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1135,9 +1135,9 @@ axes:
   - id: "extra-rust-versions"
     values:
       - id: "min"
-        display_name: "1.45 (minimum supported version)"
+        display_name: "1.47 (minimum supported version)"
         variables:
-          RUST_VERSION: "1.45.2"
+          RUST_VERSION: "1.47.0"
       - id: "nightly"
         display_name: "nightly"
         variables:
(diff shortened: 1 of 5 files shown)


agolin95:
Tracked in https://jira.mongodb.org/browse/RUST-682

jszwedko:
Just noting that this is required to resolve https://rustsec.org/advisories/RUSTSEC-2020-0146 due to the transient dependency through `sha2` -> `block-buffer` -> `generic-array`.

tobin-crypto (approved):
In regards to clearing the `generic-array` RUSTSEC advisory. I applied this to a local fork of `mongodb`. It builds and it removes the `cargo deny` error for a crate that depends on `mongodb`.

isabelatkinson (approved):
LGTM, tagging in the rest of the team for review!

saghm (commented):
The code all looks good to me!

> This PR leaves PBKDF2 at 0.6 instead of 0.7 to avoid bumping MSRV to 1.47

@isabelatkinson @patrickfreed since we're doing this for our 2.0 release, do we think we should just bump this all the way and bump our minimum Rust version accordingly?

patrickfreed:
Bumping pdbkf2 to 0.7 and the MSRV to 1.47 for driver 2.0 sounds good to me.

isabelatkinson:
SGTM as well, @bugadani this will require some updates in our CI so I'll take it from here if that's alright with you!

saghm (approved):

patrickfreed (approved):

PR #298, opened 2021-02-22
RUST-684 Use async Rwlock for cache

bugadani:


agolin95:
Tracked in https://jira.mongodb.org/browse/RUST-684

Cargo.toml
```diff
@@ -31,6 +31,7 @@ sync = ["async-std-runtime"]
 bson-u2i = ["bson/u2i"]
 
 [dependencies]
+async-lock = "2.3.0"
```
patrickfreed, line 34: To avoid bringing in a new dependency, can we update this PR to use tokio's [`RwLock`](https://docs.rs/tokio/1.2.0/tokio/sync/struct.RwLock.html) here instead? We use it in a few places in the driver already.

bugadani: Thanks, that's a fair point!

patrickfreed (approved):
Thanks for the PR! Once again running CI and tagging in other reviewers.

isabelatkinson (approved):

saghm (approved):

PR #299, opened 2021-02-22
RUST-683 Use hashmap's entry api

bugadani:
Hash cache key only once.

### DIFF
--- a/src/client/auth/scram.rs
+++ b/src/client/auth/scram.rs
@@ -1,6 +1,6 @@
 use std::{
     borrow::Cow,
-    collections::HashMap,
+    collections::{hash_map::Entry, HashMap},
     fmt::{self, Display, Formatter},
     ops::{BitXor, Range},
     str,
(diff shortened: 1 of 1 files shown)


agolin95:
Tracked in https://jira.mongodb.org/browse/RUST-683

src/client/auth/scram.rs
```diff
@@ -258,8 +258,13 @@ impl ScramVersion {
 
         if should_update_cache {
             if let Ok(ref mut cache) = CREDENTIAL_CACHE.write() {
-                if cache.get(&cache_entry_key).is_none() {
-                    cache.insert(cache_entry_key, salted_password);
+                match cache.entry(cache_entry_key) {
```
saghm, line 261: I think this can be simplified by using `if let` to avoid defining an empty branch:

```rust
if let Entry::Vacant(entry) = cache.entry(cache_entry_key) {
    entry.insert(salted_password);
}
```

bugadani: I'm sorry this fell under the rag. I've pushed an update.

saghm: No worries! Thanks for making the change

saghm (approved):

isabelatkinson (approved):

patrickfreed (approved):

PR #301, opened 2021-03-09
RUST-679 Simplify error API to ease pattern matching

saghm:
There were several different places in the driver that we assumed we could clone errors, so I handled each of those places (along with cascading changes due to other places related to that code) in a separate commit to make it easier to follow along with each one. The final commit removes the Arc and the Clone derive from the Error type. Because of this, I think the easier way to review this PR is to review each commit separately so that all of the different unrelated changes don't make things confusing. I'll push a separate commit for each change that's suggested during review, and then once everything looks okay, I'll rebase to move each change into the commit associated with the part that changed and request one final review to make sure that all of the changes look okay after that rebase.

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -103,7 +103,7 @@ impl Client {
             Err(err) => {
                 self.inner
                     .topology
-                    .handle_post_handshake_error(err.clone(), &conn, server)
+                    .handle_post_handshake_error(&err, &conn, server)
                     .await;
 
                 // Retryable writes are only supported by storage engines with document-level
@@ -162,7 +162,7 @@ impl Client {
             Err(err) => {
                 self.inner
                     .topology
-                    .handle_post_handshake_error(err.clone(), &conn, server)
+                    .handle_post_handshake_error(&err, &conn, server)
                     .await;
 
                 let err = match retryability {
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -108,7 +108,7 @@ impl Client {
 
                 // Retryable writes are only supported by storage engines with document-level
                 // locking, so users need to disable retryable writes if using mmapv1.
-                if let ErrorKind::CommandError(ref err) = err.kind.as_ref() {
+                if let ErrorKind::CommandError(ref err) = err.kind {
                     if err.code == 20 && err.message.starts_with("Transaction numbers") {
                         let mut err = err.clone();
                         err.message = "This MongoDB deployment does not support retryable writes. \
--- a/src/cursor/common.rs
+++ b/src/cursor/common.rs
@@ -10,7 +10,7 @@ use futures::{Future, Stream};
 
 use crate::{
     bson::Document,
-    error::{ErrorKind, Result},
+    error::{Error, ErrorKind, Result},
     options::StreamAddress,
     results::GetMoreResult,
     Client,
@@ -73,12 +73,13 @@ impl<T: GetMoreProvider> Stream for GenericCursor<T> {
             if let Some(future) = self.provider.executing_future() {
                 match Pin::new(future).poll(cx) {
                     // If a result is ready, retrieve the buffer and update the exhausted status.
-                    Poll::Ready(mut get_more_result) => {
-                        let buffer_result = get_more_result.take_buffer();
-                        self.exhausted = get_more_result.exhausted();
+                    Poll::Ready(get_more_result) => {
+                        let exhausted = get_more_result.exhausted();
+                        let (result, session) = get_more_result.into_parts();
 
-                        self.provider.clear_execution(get_more_result);
-                        self.buffer = buffer_result?;
+                        self.exhausted = exhausted;
+                        self.provider.clear_execution(session, exhausted);
+                        self.buffer = result?.batch;
                     }
                     Poll::Pending => return Poll::Pending,
                 }
@@ -113,7 +114,11 @@ pub(super) trait GetMoreProvider: Unpin {
     fn executing_future(&mut self) -> Option<&mut Self::GetMoreFuture>;
 
     /// Clear out any state remaining from previous getMore executions.
-    fn clear_execution(&mut self, result: Self::GetMoreResult);
+    fn clear_execution(
+        &mut self,
+        session: <Self::GetMoreResult as GetMoreProviderResult>::Session,
+        exhausted: bool,
+    );
 
     /// Start executing a new getMore if one isn't already in flight.
     fn start_execution(&mut self, spec: CursorInformation, client: Client);
@@ -121,16 +126,11 @@ pub(super) trait GetMoreProvider: Unpin {
 
 /// Trait describing results returned from a `GetMoreProvider`.
 pub(super) trait GetMoreProviderResult {
-    /// A result containing a mutable reference to the raw getMore result.
-    fn as_mut(&mut self) -> Result<&mut GetMoreResult>;
+    type Session;
 
-    /// A result containing a reference to the raw getMore result.
-    fn as_ref(&self) -> Result<&GetMoreResult>;
+    fn as_ref(&self) -> std::result::Result<&GetMoreResult, &Error>;
 
-    /// Take the buffer from the getMore result.
-    fn take_buffer(&mut self) -> Result<VecDeque<Document>> {
-        self.as_mut().map(|res| std::mem::take(&mut res.batch))
-    }
+    fn into_parts(self) -> (Result<GetMoreResult>, Self::Session);
 
     /// Whether the response from the server indicated the cursor was exhausted or not.
     fn exhausted(&self) -> bool {
--- a/src/cursor/common.rs
+++ b/src/cursor/common.rs
@@ -137,7 +137,7 @@ pub(super) trait GetMoreProviderResult {
         match self.as_ref() {
             Ok(res) => res.exhausted,
             Err(e) => {
-                matches!(e.kind.as_ref(), ErrorKind::CommandError(e) if e.code == 43 || e.code == 237)
+                matches!(e.kind, ErrorKind::CommandError(ref e) if e.code == 43 || e.code == 237)
             }
         }
     }
--- a/src/error.rs
+++ b/src/error.rs
@@ -27,7 +27,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 #[non_exhaustive]
 pub struct Error {
     /// The type of error that occurred.
-    pub kind: Arc<ErrorKind>,
+    pub kind: ErrorKind,
     labels: Vec<String>,
 }
 
@@ -63,7 +63,7 @@ impl Error {
 
     /// Whether this error is an "ns not found" error or not.
     pub(crate) fn is_ns_not_found(&self) -> bool {
-        matches!(self.kind.as_ref(), ErrorKind::CommandError(err) if err.code == 26)
+        matches!(self.kind, ErrorKind::CommandError(ref err) if err.code == 26)
     }
 
     /// Whether a read operation should be retried if this error occurs.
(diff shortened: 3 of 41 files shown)


src/sdam/description/server.rs
```diff
@@ -74,7 +73,7 @@ pub(crate) struct ServerDescription {
     // allows us to ensure that only valid states are possible (e.g. preventing that both an error
     // and a reply are present) while still making it easy to define helper methods on
     // ServerDescription for information we need from the isMaster reply by propagating with `?`.
-    pub(crate) reply: Result<Option<IsMasterReply>>,
+    pub(crate) reply: Result<Option<IsMasterReply>, String>,
```
saghm, line 76: Because the only place we use the errors on server descriptions is in the error messages to users when server selection fails, the simplest way to avoid needing to copy errors around here is to just store the description of the error instead of the error itself, since that's what will be shown to users anyhow. The rest of the changes in this specific commit are due to this one change.

src/sdam/description/topology/mod.rs
```diff
@@ -661,8 +666,13 @@ impl TopologyDescription {
     }
 
     /// Create a new ServerDescription for each address and add it to the topology.
-    fn add_new_servers<'a>(&mut self, servers: impl Iterator<Item = &'a String>) -> Result<()> {
-        let servers: Result<Vec<_>> = servers.map(|server| StreamAddress::parse(server)).collect();
+    fn add_new_servers<'a>(
```
saghm, line 669: Rustfmt changed the spacing a bit here, so it might not be super clear what the change is: the return value of the function was changed from `Result<()>` to `Result<(), String>`, and then a call to `to_string()` on the error was added to the iterator expression.

src/sdam/description/topology/mod.rs
```diff
@@ -731,15 +741,17 @@ pub(crate) struct TopologyDescriptionDiff {
     pub(crate) new_addresses: HashSet<StreamAddress>,
 }
 
-fn verify_max_staleness(max_staleness: Option<Duration>) -> Result<()> {
+fn verify_max_staleness(max_staleness: Option<Duration>) -> crate::error::Result<()> {
```
saghm, line 744: There were two existing uses of this function before this commit, one of which does return the error to users, so we still want to be able to return an actual Error rather than a string in this case. I split this function into two so that we don't have to pattern match the error and assert that it's an ArgumentError (when we know it always will be) when it's called in a place where we don't return the error to users.

src/cursor/common.rs
```diff
@@ -73,12 +73,13 @@ impl<T: GetMoreProvider> Stream for GenericCursor<T> {
             if let Some(future) = self.provider.executing_future() {
                 match Pin::new(future).poll(cx) {
                     // If a result is ready, retrieve the buffer and update the exhausted status.
-                    Poll::Ready(mut get_more_result) => {
-                        let buffer_result = get_more_result.take_buffer();
-                        self.exhausted = get_more_result.exhausted();
+                    Poll::Ready(get_more_result) => {
+                        let exhausted = get_more_result.exhausted();
```
saghm, line 77: Previously, the GetMoreResult trait allowed for getting the exhausted state and the buffer separately, with the latter cloning the error if it existed. This clone turned out not to be necessary, as it was only done to preserve the ownership of the GetMoreResult so it could be passed wholesale to the `clear_execution` method on the provider. This method needs to know whether the cursor is exhausted and potentially have access to the session used for the cursor, so the simplest fix was just to add a method on GetMoreResult that splits it into the session (with the existing associated type for the GetMoreResult being returned) and the buffer (or owned error if it exists). GetMoreResult is internal to the cursor module, so these changes were possible without changing anything in the rest of the driver or any cursor semantics.

src/sdam/state/server.rs
```diff
@@ -74,7 +74,10 @@ impl Server {
 /// TODO: add success cases from application handshakes.
 #[derive(Debug)]
 pub(crate) enum ServerUpdate {
-    Error { error: Error, error_generation: u32 },
+    Error {
+        error: String,
```
saghm, line 78: Passed an owned error over the channel is no longer possible since the error also needs to be returned to the user in some places (e.g. when a new connection is created but fails to establish during a checkout). Luckily, the other side of the channel only uses the errors to mark the server as unknown in the topology, and an earlier commit in this PR already updated ServerDescription to use String instead of Error, so we can just pass the string across the channel eagerly instead of lazily getting the message on the other end.

src/error.rs
```diff
@@ -23,12 +23,12 @@ pub type Result<T> = std::result::Result<T, Error>;
...
 /// [`ErrorKind`](enum.ErrorKind.html) is wrapped in an `Arc` to allow the errors to be
 /// cloned.
-#[derive(Clone, Debug, Error)]
+#[derive(Debug, Error)]
 #[error(display = "{}", kind)]
 #[non_exhaustive]
 pub struct Error {
     /// The type of error that occurred.
-    pub kind: Arc<ErrorKind>,
+    pub kind: ErrorKind,
```
saghm, line 31: This is the only substantive change in this commit; the rest are just fixes for the Arc being removed

src/client/executor.rs
```diff
@@ -108,7 +108,7 @@ impl Client {
 
                 // Retryable writes are only supported by storage engines with document-level
                 // locking, so users need to disable retryable writes if using mmapv1.
-                if let ErrorKind::CommandError(ref err) = err.kind.as_ref() {
+                if let ErrorKind::CommandError(ref err) = err.kind {
```
patrickfreed, line 111: is there any reason we would want `kind` to be a method instead of accessing the field directly? This is what [`std::io::Error` does](https://doc.rust-lang.org/std/io/struct.Error.html#method.kind), but I don't really have any further justification than that.

src/runtime/acknowledged_message.rs
```diff
@@ -21,8 +21,8 @@ impl<M, R> AcknowledgedMessage<M, R> {
     }
 
     /// Get the message.
-    pub(crate) fn message(&self) -> &M {
-        &self.message
+    pub(crate) fn message(self) -> M {
```
patrickfreed, line 24: Now that this takes ownership, can we name it `into_message`?

src/error.rs
```diff
@@ -23,12 +23,12 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// An error that can occur in the `mongodb` crate. The inner
 /// [`ErrorKind`](enum.ErrorKind.html) is wrapped in an `Arc` to allow the errors to be
 /// cloned.
-#[derive(Clone, Debug, Error)]
```
patrickfreed, line 26: Before we go through with removing the `Clone` implementation altogether, have we revisited just wrapping the `std::io::Error` branch in an `Arc` to preserve `Clone` on our type? I went back to see [why we didn't opt for this originally](https://docs.google.com/document/d/1SexV0QyBbb8BlhzDD9Wfib7eL3b8Hc_ltoU98wXGqtk/edit#heading=h.domcvrirm8h0), and it seems to be to avoid having users handle the indirection of `Arc` non-uniformally. For `std::io::Error` specifically, I'm not sure if `Arc` actually introduces any additional work for users though, since they have to invoke a method on the error (e.g. `kind()`) to get any useful result anyways.

I've looked over the other branches of our `ErrorKind`, and the only ones that also don't implement `Clone` are our BSON errors (which we own and can make `Clone`) and `TokioTimeoutElapsed` error, which we don't actually even use since `RUNTIME.timeout` converts everything to `std::io::Error` anyways.

I think if we manage to keep `Clone` alongside the changes made here to ease in matching, we'll have the best of both worlds.

patrickfreed (commented):
Overall looks good! I have one question about possibly preserving `Clone` and a few other minor questions. Also, it looks like this broke the async-std tests + lint.

src/client/executor.rs
saghm: The main benefit would be that we would be able to change the way that we store the field in the struct without breaking API. As it is currently, removing or changing the type of the field would break user code that accesses the field directly.

src/error.rs
saghm: I considered that a bit too when working on this PR. I didn't end up making the change (although I hadn't looked back into why we didn't choose that originally), although a bit of that did make it into one of the proposals I made in the google doc in the context of making `CommandFailedEvent` use an alternate error struct.

I guess my main question is what benefit we think users will get from having our error type implement `Clone`. The only thing I can think of is the reason that we wanted it in the first place (to make their own wrapping error type cloneable), but I don't think this would be quite as common an issue for our users due to the fact that most of them are using the driver in the context of an application rather than a library, and the current consensus in the Rust community is that applications should just use `Box<dyn Error>` (or something like [`anyhow`](https://crates.io/crates/anyhow) to abstract that). While there are some libraries out there that wrap the driver (and obviously more could be written in the future), my general instinct is to make any performance costs (e.g. heap allocations and reference counts) opt-in for users who need them rather than on by default with no way to opt out of, especially since the number of users who will need it in this case are far fewer than the ones who won't. My opposition to doing this isn't particularly strong, though.

src/runtime/acknowledged_message.rs
saghm: Updated

src/client/executor.rs
patrickfreed: I don't really have any strong feelings on this either way, if you think that protection is worthwhile then I'd say go ahead and make the change and if not, seems fine to me to leave it as is.

src/error.rs
patrickfreed: I can't really think of a specific example of what a user might need the cloning for, but I imagine it could be something similar to what we needed clone for as you mentioned. Even if they don't _require_ the cloning (as we didn't), having to update their existing code could be non-trivial, as is seen in this PR. Also, if they really do need cloning, they'll need to wrap our entire error in an `Arc`, which is pretty annoying to deal with (hence this PR :slightly_smiling_face:), so even if it is somewhat uncommon, the frustration felt could be high and worth avoiding. Lastly, if we don't go for `Clone` now, I don't think we'll be able to do so until 3.0, so if it does turn out to be a burden for users, we can't really give them any good fix.

Regarding performance costs, I definitely agree, but given that this Arc would only be allocated in single specific error case (i.e. not the "hot" path), I don't think it's much of a concern.

So overall, for reasons of reducing user friction + safeguarding us against any future need, I think we should try to preserve `Clone` here.

This also has made me wonder, if we preserve `Clone`, could we avoid having to make the `String` changes in this PR? I realize this is an frustrating question to ask _after_ all the work has been done and I'm really sorry about that; it's just that up to this point I had assumed that preserving `Clone` wasn't going to be possible or easy but now have realized there could be another option.

saghm: If we kept Error being Clone, then yes, I think we would want to undo the changes to CommandFailedEvent. All of the other changes are probably fine to keep though, since they don't put any burden on the user and are marginal perf improvements.

src/error.rs
```diff
@@ -188,14 +188,44 @@ where
...
+    }
+}
+
+impl From<std::io::ErrorKind> for ErrorKind {
+    fn from(err: std::io::ErrorKind) -> Self {
+        Self::Io(Arc::new(err.into()))
+    }
+}
+
+impl From<tokio::time::error::Elapsed> for ErrorKind {
```
patrickfreed, line 221: I think this error and the async-std equivalent we can just drop, since we convert them both to `std::io::Error`s anyways. Alternatively, we could just implement a custom clone for this (possibly via `Derivative` or something) since [`tokio::time::error::Elapsed` is a unit struct](https://docs.rs/tokio/0.3.2/src/tokio/time/error.rs.html#39).

saghm: Sounds good; I've updated it to remove those variants

patrickfreed (approved):
LGTM! I think this will make our error API a lot nicer to use.

patrickfreed:
Looks like clippy needs to be satisfied but otherwise LGTM (I don't need to re-review the clippy fix)

isabelatkinson (approved):

saghm:
Looks like the clippy issue is just a slight memory inefficiency in the CMAP tests, which isn't really anything to worry about.

PR #303, opened 2021-03-10
RUST-705 replace err-derive with thiserror

saghm:


patrickfreed (approved):
LGTM. Very cool to see how similar the two libraries ended up being.

saghm:
I think this also is technically not a breaking change (even though we'll be doing it as part of 2.0), since we're still implementing `Error` with the exact same error messages, as neither err-derive or thiserror end up exposing anything beyond a `std::error::Error` implementation to the user.

isabelatkinson (approved):

PR #304, opened 2021-03-15
RUST-52 Implement Sessions API

isabelatkinson:


src/test/spec/sessions/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+                .run_command(doc, None)
+                .await
+                .unwrap();
+        }
+
+        let options = match test.session_options {
+            Some(ref options) => options.get("session0").cloned(),
+            None => None,
+        };
+        let mut session0 = Some(client.start_session(options).await.unwrap());
```
isabelatkinson, line 124: These sessions need to be wrapped in options because one of the possible test operations is `endSession`, which in our case requires calling `drop` (i.e. taking ownership of the session) -- to get around the borrow checker, the session is taken out of the option when it's necessary to `drop` and unwrapped everywhere else. The unwraps are safe since the tests never use a session after it's been ended.

src/test/spec/sessions/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+                    } else {
+                        operation
+                            .execute_on_session(session1.as_ref().unwrap())
+                            .await;
+                    }
+                    continue;
+                }
+                OperationObject::TestRunner => {
+                    match operation.name.as_str() {
+                        "assertDifferentLsidOnLastTwoCommands" => {
```
isabelatkinson, line 180: All of these operations take in different arguments so it didn't make much sense to integrate them into the usual operation parsing -- I'll abstract this out a little more if it gets too unwieldy when the additional transactions operations are introduced.

src/test/spec/sessions/operation.rs
```diff
@@ -0,0 +1,1178 @@
+use std::{collections::HashMap, fmt::Debug, ops::Deref};
```
isabelatkinson, line 1: This is all quite similar to the operations stuff in the other runners, just some minor adjustments to fit in with the specific test schema for sessions/transactions.

src/client/executor.rs
```diff
@@ -70,6 +79,13 @@ impl Client {
         op: T,
         session: &mut ClientSession,
     ) -> Result<T::O> {
+        // TODO RUST-9: allow unacknowledged write concerns
```
saghm, line 82: Can we remove the check above for this if we're going to start checking it here, or do we need to do it in both places?

src/client/options/mod.rs
```diff
@@ -2057,3 +2057,8 @@ mod tests {
         );
     }
 }
+
+/// Contains the options that can be used to create a new
+/// [`ClientSession`](../struct.ClientSession.html).
+#[derive(Clone, Debug, Deserialize)]
+pub struct SessionOptions {}
```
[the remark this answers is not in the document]
saghm: It's a little strange to define a public type as part of our API that users can't actually create on their own. Should we at least derive `TypedBuilder` on it?

src/coll/mod.rs
```diff
@@ -178,13 +180,31 @@ where
...
+        &self,
+        options: impl Into<Option<DropCollectionOptions>>,
+        session: impl Into<Option<&mut ClientSession>>,
+    ) -> Result<()> {
         let mut options = options.into();
         resolve_options!(self, options, [write_concern]);
 
         let drop = DropCollection::new(self.namespace(), options);
-        self.client().execute_operation(drop).await
+        self.client().execute_operation(drop, session.into()).await
```
saghm, line 192: My understanding was that we were planning on changing the return type of each of these methods so that that a user could either `await` and get back the result without a session (so that the API worked the same as before in that case) or could call a method like `with_session(&mut session)` and then `await` that to execute the operation with a session. Is there a reason that we didn't go that route with this implementation?

src/cursor/session.rs
```diff
@@ -1,44 +1,78 @@
...
+/// # use mongodb::{Client, error::Result, ClientSession, SessionCursor};
+/// #
+/// # async fn do_stuff() -> Result<()> {
+/// # let client = Client::with_uri_str("mongodb://example.com").await?;
+/// # let mut session = client.start_session(None).await?;
+/// # let coll = client.database("foo").collection("bar");
+/// # let mut cursor = coll.find_with_session(None, None, &mut session).await?;
+/// #
+/// while let Some(doc) = cursor.with_session(&mut session).next().await {
+/// println!("{}", doc?)
```
saghm, line 36: Slight nit: could we indent the `println!` here with four extra spaces (beyond the original space after the `///`) so that it renders cleanly in our documentation?

src/selection_criteria.rs
```diff
@@ -131,11 +133,22 @@ impl<'de> Deserialize<'de> for ReadPreference {
 
         match preference.mode.as_str() {
             "Primary" => Ok(ReadPreference::Primary),
-            "Secondary" => Ok(ReadPreference::Secondary { options: preference.options }),
-            "PrimaryPreferred" => Ok(ReadPreference::PrimaryPreferred { options: preference.options }),
-            "SecondaryPreferred" => Ok(ReadPreference::SecondaryPreferred { options: preference.options }),
-            "Nearest" => Ok(ReadPreference::Nearest { options: preference.options }),
-            other => Err(D::Error::custom(format!("Unknown read preference mode: {}", other))),
+            "Secondary" => Ok(ReadPreference::Secondary {
```
saghm, line 136: I assume rustfmt did this? I've been running into a few cases lately where it has reformatted code that I didn't touch but was in the same file I was working in. I guess that's the cost of sticking with the nightly version.

src/test/spec/sessions/mod.rs
saghm: Sounds good! We had to do a similar thing with the connection pool spec tests.

src/coll/mod.rs
patrickfreed: The API here is in line with the proposed [design](https://docs.google.com/document/d/1xdIk3RVTUh_Wrna52nqChq1ZoqOBqbaak5ZvfwuhzEw/edit#heading=h.hjltafp465xm). Back when we were first looking at this, we didn't consider the builder approach too much since it would technically be a breaking change, and I don't think we really ever addressed it again. Now that we're doing 2.0, we could re-evaluate it, though all of this code has already been written.

Note that cursors do have a `with_session` in the design, since they can't retain ownership of the session or a reference to it without preventing other operations from using that session.

patrickfreed: Looking back at the [original-original design](https://gist.github.com/patrickfreed/e37d084500c0bdfe676556db1dc7cc1b#cons-2), it looks like a couple options were laid out without one explicitly being picked:
>This could entail duplicating the entire API or having users specify None whenever they don't need a session. Alternatively, we could use a builder pattern on the returned future to specify whether a session is required or not. e.g.
>```rust
>coll.find(doc! { "x": 1 }).with_session(&mut session).await; // Cursor<'a>
>coll.find(doc! { "x": 1 }).await; // Cursor<'static>
>coll.insert_one(doc! { "x": 1 }).with_session(&mut session).await;
>coll.insert_one(doc! { "x": 1 }).await;
>```
>This works nicely for async because we're always returning futures, but it doesn't apply directly to sync which has no intermediate return values. For sync, we'd either have to introduce them or just duplicate the API:
>```rust
>coll.find_with_session(&mut session, doc! { "x": 1 }); // Cursor<'a>
>coll.find(doc! { "x": 1 }); // Cursor<'static>
>coll.insert_one_with_session(&mut session, doc! { "x": 1 });
>coll.insert_one(doc! { "x": 1 }).await;
>```
It looks like consistency with the sync API was another concern, but again I think it mostly came down to not wanting to make a breaking change.

isabelatkinson: Hm I wasn't aware of the alternative builder design; I used the first doc Patrick linked as a reference when implementing this. I personally find the current API a bit more straightforward, but do see the benefit in reducing duplication with the builder pattern. I'd be fine with re-implementing the other way if either of you feels strongly about it (although worth noting that a significant overhaul will probably delay us on being able to get transactions out a little longer).

patrickfreed: I don't feel strongly either way. Builder is slightly "prettier" maybe? but also slightly easier to forget to do and likely non-trivial to implement. I guess my opinion then would be to stick with the current implementation because it's already done and I personally don't see a huge difference between the two at the moment.

src/client/executor.rs
isabelatkinson: removing this method and sending everything through `execute_operation` instead

src/coll/mod.rs
saghm: I'm mostly not super thrilled with having to duplicate all the methods in our API, but if you both are fine with it, I'll resolve this thread.

src/coll/mod.rs
```diff
@@ -318,6 +469,39 @@ where
...
+        &self,
+        filter: impl Into<Option<Document>>,
+        options: impl Into<Option<FindOneOptions>>,
+        session: &mut ClientSession,
+    ) -> Result<Option<T>> {
+        let mut options: FindOptions = options
+            .into()
+            .map(Into::into)
+            .unwrap_or_else(Default::default);
+        options.limit = Some(-1);
```
patrickfreed, line 484: I think this limit line actually ought to be in the `From<FindOneOptions>` implementation for `FindOptions`

src/db/mod.rs
```diff
@@ -232,14 +275,41 @@ impl Database {
...
-    pub async fn create_collection(
+    /// Gets the names of the collections in the database using the provided `ClientSession`.
+    pub async fn list_collection_names_with_session(
+        &self,
+        filter: impl Into<Option<Document>>,
+        session: &mut ClientSession,
+    ) -> Result<Vec<String>> {
+        let list_collections =
+            ListCollections::new(self.name().to_string(), filter.into(), true, None);
+        let mut cursor: SessionCursor<Document> = self
```
patrickfreed, line 286: Given that `SessionCursorHandle` implements `Stream`, we should be able to write a generic version of this to avoid duplicating the code.

src/sdam/description/topology/mod.rs
```diff
@@ -289,16 +289,6 @@ impl TopologyDescription {
             return;
         }
 
-        if server_description.server_type == ServerType::Standalone {
```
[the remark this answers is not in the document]
patrickfreed: Yep, this is old behavior that got clarified (extremely shortly) after we implemented it here. Should be fine to remove.

src/test/spec/sessions/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+                Some("session0") => session0.as_mut(),
+                Some("session1") => session1.as_mut(),
+                Some(other) => panic!("unknown session name: {}", other),
+                None => None,
+            };
+
+            let result = match operation.object {
+                OperationObject::Collection => {
+                    let result = operation.execute_on_collection(&coll, session).await;
+                    if test.description == "Server supports implicit sessions" {
```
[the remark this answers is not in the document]
patrickfreed: Sounds good, could you add a comment into the code here explaining that?

src/test/spec/sessions/operation.rs
patrickfreed: The README for the runner claims it has the same format as the transactions runner, which should mean it has the v2 format (transactions / retryable reads / crud v2). Is there any chance we'll be able to reuse code there instead of introducing an entirely new runner?

patrickfreed (commented):
Looks really good so far! I haven't gotten a chance to review the whole thing yet, but I do have a few minor comments / suggestions on the bits I have seen.

src/client/session/mod.rs
```diff
@@ -31,21 +32,28 @@ lazy_static! {
...
-pub(crate) struct ClientSession {
+pub struct ClientSession {
     cluster_time: Option<ClusterTime>,
     server_session: ServerSession,
     client: Client,
     is_implicit: bool,
+    options: Option<SessionOptions>,
 }
 
 impl ClientSession {
```
patrickfreed, line 43: From the specification, it looks like we need to make a few of these methods public, namely:
- an accessor for the client
- `id()`
- `cluster_time()`
- `advance_cluster_time()`

src/coll/mod.rs
patrickfreed: Ah yep, I was probably to blame for the original implementation then :slightly_smiling_face: 

BTW, this assignment can be removed now that the `From` implementation takes care of it.

patrickfreed (commented):
Looks great! I just have a few minor comments about the `ClientSession` type and the tests.

src/test/util/event.rs
```diff
@@ -406,6 +414,24 @@ impl EventClient {
         self.handler.command_events.write().unwrap().clear();
         self.handler.pool_cleared_events.write().unwrap().clear();
     }
+
+    /// Gets the last two command started events.
+    pub fn get_last_two_command_started_events(
```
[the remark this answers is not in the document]
isabelatkinson: this is actually not super necessary with the introduction of `get_all_command_started_events` so I'm going to remove

src/client/session/mod.rs
```diff
@@ -89,6 +97,19 @@ impl ClientSession {
...
+    /// Whether this session is dirty.
+    #[cfg(test)]
+    pub(crate) fn is_dirty(&self) -> bool {
+        self.server_session.dirty
+    }
+
+    /// Returns a mutable reference to this session wrapped in an option. This method is necessary
+    /// to bypass the construction of an option taking ownership of a mutable reference.
+    #[allow(clippy::unnecessary_wraps)]
+    pub(crate) fn as_mut_ref_option(&mut self) -> Option<&mut Self> {
```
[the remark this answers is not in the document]
isabelatkinson: Huh that's super interesting; I'm surprised there isn't better documentation of this in the Rust book. Updated to use an explicit reborrow.

patrickfreed:
Code changes LGTM, it looks like this just needs to be rebased to fix the merge conflict. I think the compilation issue on 1.45 should go away after that too.

patrickfreed (approved):
LGTM! Great job with this!

Side note: there may be some tricky merge conflicts with the error changes in #301, particularly with cursors. It may help to coordinate somehow with @saghm on merging them in.

isabelatkinson:
> Side note: there may be some tricky merge conflicts with the error changes in #301, particularly with cursors. It may help to coordinate somehow with @saghm on merging them in.

@saghm I think it would be easiest for you to merge the error changes first; once that's in I can rebase and you can review whatever changes are needed to apply that work to this PR.

src/client/session/cluster_time.rs
```diff
@@ -10,11 +10,11 @@ use crate::bson::{Document, Timestamp};
 #[derive(Debug, Deserialize, Clone, Serialize, Derivative)]
 #[derivative(PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
-pub(crate) struct ClusterTime {
-    cluster_time: Timestamp,
+pub struct ClusterTime {
+    pub cluster_time: Timestamp,
```
saghm, line 14: I don't think these should be public; this allows a user who owns a ClusterTime to modify the timestamp or the signature, which is not useful since it would presumably break it. If users do actually need to inspect these values, we should make them private and add getters, but I think that this is supposed to be a mostly opaque value to users (that they receive from the driver and then can pass back in other places without needing to care about the internals), in which case we wouldn't need getters. Since the struct derives Debug, users will still be able to log cluster times even if we don't add getters, so that wouldn't be a reason to add getters either.

src/test/spec/v2_runner/mod.rs
```diff
@@ -0,0 +1,304 @@
+pub mod operation;
```
saghm, line 1: This (and the rest of the v2 test runner) is just moved from the previous location in the CRUD spec tests, correct?

saghm (commented):
One question/suggestion about making something public and one clarification question about the v2 runner, but other than those, this looks good to me. Great job on this!

src/client/session/cluster_time.rs
isabelatkinson: That makes sense to me, the reasoning for exposing these was mainly for inspecting, but I think deriving `Debug` should take care of that.

src/test/spec/v2_runner/mod.rs
isabelatkinson: yep, I just did some renaming for the sake of clarity

saghm (approved):

PR #307, opened 2021-03-22
RUST-711 Update rustfmt configuration to use imports_granularity

isabelatkinson:


src/lib.rs
```diff
@@ -72,7 +72,6 @@
 
 #![warn(missing_docs)]
 #![warn(missing_crate_level_docs)]
-
```
isabelatkinson, line 75: Looks like some more formatting updates were pulled into this commit

saghm (approved):

saghm:
It looks like this also hit the issue with the min compiler version, so make sure to rebase this to get the update

patrickfreed (approved):

PR #309, opened 2021-03-25
minor: fix clippy

isabelatkinson:


src/sdam/description/server.rs
```diff
@@ -19,11 +19,11 @@ const DRIVER_MAX_WIRE_VERSION: i32 = 7;
 pub enum ServerType {
     Standalone,
     Mongos,
-    RSPrimary,
-    RSSecondary,
-    RSArbiter,
-    RSOther,
-    RSGhost,
+    RsPrimary,
```
saghm, line 22: I think these are technically breaking changes since this type is exposed through the selection criteria API, but since we're doing a major release, that's not an issue

patrickfreed: It might be worth filing a ticket then for this change so that it'll be represented in the release notes.

patrickfreed (approved):

src/sdam/description/server.rs
isabelatkinson: Filed [RUST-719](https://jira.mongodb.org/browse/RUST-719), going to split this PR into two commits so the ticket number will show up in our log.

saghm (approved):
Oops, thought I had approved this. Sorry about that!

PR #310, opened 2021-03-25
Sync CRUD spec tests and fix failures

saghm:
When I sync'd the CRUD spec tests as part of RUST-658, there were a few failures in the CRUD v2 tests due to various bugs. Since they were all small and easy to fix, I filed tickets where needed and compiled the fixes into a single PR (with a separate commit per fix to make reviewing easier). If any changes need to be made as part of review, I'll commit them as separate commits, and then once everything is approved I'll rebase locally to squash each commit from review with the associated commit of the fix it's related to, and then I'll use the "rebase and merge" option to ensure that each fix gets merged as a separate commit.


src/test/util/matchable.rs
```diff
@@ -67,7 +67,7 @@ impl Matchable for Document {
                     return false;
                 }
             } else {
-                return false;
+                return v == &Bson::Null;
```
saghm, line 70: The spec test for checking that `allowDiskUse` is not sent as part of a find when it's not specified as an option [explicitly includes an `allowDiskUse` field in the CommandStartedEvent expectation with a null value](https://github.com/mongodb/specifications/blob/7077aa39aafe1abd6e5d5529b13d66ab9ded90bf/source/crud/tests/v2/find-allowdiskuse.json#L20-L30). Given that we prefer to sync the spec tests as-is instead of modifying them, the most straightforward fix in code is to have our matching functionality treat a null value in the expected document as a match for a missing value in our actual document.

isabelatkinson: Should this only return when `v != &Bson::Null`? Otherwise I think this is terminating the loop too early.

saghm: Oops, yes, you're right. Fixed!

isabelatkinson (approved):

src/test/util/matchable.rs
patrickfreed: I think we're still returning too early here actually, since we still need to inspect the other keys. The `if` should have the `v == &Bson::Null` in it and the body still just have `return false` I think. I believe we have the same bug in the Swift driver test runner actually: https://github.com/mongodb/mongo-swift-driver/blob/main/Tests/MongoSwiftSyncTests/SpecTestRunner/Match.swift#L75

saghm: The latest version of the code has the following code:

```rust
if let Some(actual_v) = self.get(k) {
    if !actual_v.matches(v) {
        return false;
    }
} else if v != &Bson::Null {
    return false;
}            
```

This will return false if there is a value for the key in the actual document (i.e. in `self`) and it doesn't match the expected key, and it will return false if the key doesn't in the actual document and the value in the actual document isn't null (meaning the value that the test expects is not present in the actual document). Otherwise, it doesn't return and continues looping to check the other keys. This seems correct to me; can you clarify what types of keys and values in expected/actual will cause it to terminate too early?

src/coll/mod.rs
```diff
@@ -513,6 +513,16 @@ where
         options: impl Into<Option<ReplaceOptions>>,
     ) -> Result<UpdateResult> {
         let replacement = to_document(&replacement)?;
+
+        if let Some(first_key) = replacement.keys().next() {
```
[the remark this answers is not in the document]
saghm: Oops, you're right! There's also an `update_document_check` already in BSON-util, and used in update_many, but _not_ used in `update_one`. I'm a lot less confused as to how we got by without having done anything for this before without anyone noticing; we did all the functionality, but just forgot to call it in `update_one`.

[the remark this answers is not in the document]
saghm: Updated to remove the redundant checks and re-use the existing helper in `update_one`

src/coll/mod.rs
```diff
@@ -541,6 +551,18 @@ where
         options: impl Into<Option<UpdateOptions>>,
     ) -> Result<UpdateResult> {
         let update = update.into();
+
+        if let UpdateModifications::Document(ref update) = update {
```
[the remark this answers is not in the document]
saghm: See response in the previous comment

src/test/util/matchable.rs
patrickfreed: Ah I think I was looking at an outdated diff, it looks correct to me.

src/coll/mod.rs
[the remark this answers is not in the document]
patrickfreed: This check here still needs to be deleted.

patrickfreed (commented):
Everything looks good bar removing that one last redundant check. The lint failures seem like they'll be resolved by #309.

src/coll/mod.rs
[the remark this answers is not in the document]
saghm: Updated

patrickfreed (approved):

PR #311, opened 2021-03-25
 RUST-658 Change estimatedDocumentCount() to use the $collStats agg stage 

saghm:
To implement this feature, I augmented the `Count` operation so that it sends an `aggregation` command instead of a `count` command on server versions new enough to support $collStats. However, this required being able to deserialize either type of command response in `Operation::handle_response`. The cleanest way to do this seemed to be to pass in the StreamDescription to `handle_response` in the same way it's done in `build` (which is trivial to do from the execution layer), so I updated the Operation trait to include that in the signature. This required a large number of changes elsewhere in the driver, but it won't require any additional changes moving forward, so I think it's worth it.


src/coll/mod.rs
```diff
@@ -265,7 +265,11 @@ where
         resolve_options!(self, options, [read_concern, selection_criteria]);
 
         let op = Count::new(self.namespace(), options);
-        self.client().execute_operation(op, None).await
+
+        match self.client().execute_operation(op, None).await {
```
saghm, line 269: The section of the CRUD spec on this functionality indicated that we should return `0` for the count if this method is invoked on a non-existent collection.

src/operation/count/mod.rs
```diff
@@ -52,8 +75,29 @@ impl Operation for Count {
...
+        let response_body: ResponseBody = match description.max_wire_version {
+            Some(v) if v >= SERVER_4_9_0_WIRE_VERSION => {
+                let CursorBody { mut cursor } = response.body()?;
+
+                cursor
+                    .first_batch
+                    .pop_front()
+                    .and_then(|doc| bson::from_document(doc).ok())
+                    .ok_or_else(|| {
+                        Error::from(ErrorKind::ArgumentError {
```
patrickfreed, line 92: I think for this case we typically use `ResponseError`, though honestly we could just replace any usages of `ResponseError` with `InternalError` if we wanted to.

src/coll/mod.rs
patrickfreed: We can actually use `Operation::handle_error` to do this at the operation layer. We currently do something [similar for `drop_collection`](https://github.com/mongodb/mongo-rust-driver/blob/master/src/operation/drop_collection/mod.rs#L58).

patrickfreed (commented):
Looks pretty good, I just have a few minor comments. And I agree, passing the stream description to `handle_response` seems like the right move, hopefully the refactor was relatively painless.

src/coll/mod.rs
saghm: Oh, right! I had forgotten about that

src/operation/count/mod.rs
saghm: Oops, good catch

patrickfreed (approved):
LGTM!

isabelatkinson (approved):

PR #312, opened 2021-03-31
RUST-655 / RUST-691 Simplification of SDAM error handling

patrickfreed:
RUST-655 / RUST-691

This PR updates and simplifies how errors are handled by the SDAM machinery based on the specification updates in DRIVERS-1152 and DRIVERS-1476. These two tickets were closely related so I opted to include them in the same PR. I was sort of working on both of them at the same time, so it was hard to go back at the end and separate them into logical commits; sorry if this makes reviewing more difficult.

Notable changes:
- consolidation of all application error handling to single `handle_application_error` method (RUST-691)
- updated the SDAM spec test runner to support the new errors tests (both)
- only error codes are inspected for SDAM purposes (RUST-655)


src/client/executor.rs
```diff
@@ -100,30 +100,36 @@ impl Client {
                 return Ok(result);
             }
             Err(err) => {
```
patrickfreed, line 102: The changes here are mostly just shuffling the order so that the error mutations are reflected in the error stored in the server description.

src/cmap/manager.rs
```diff
@@ -20,8 +20,15 @@ pub(super) struct PoolManager {
 
 impl PoolManager {
     /// Lazily clear the pool.
-    pub(super) fn clear(&self) {
-        let _ = self.sender.send(PoolManagementRequest::Clear);
+    pub(super) async fn clear(&self) {
```
patrickfreed, line 23: Because clearing the pool needs to occur only while the lock on the topology is held, this was turned into a blocking operation.

src/cmap/mod.rs
```diff
@@ -176,20 +176,16 @@ impl ConnectionPool {
...
         self.manager.mark_as_ready().await;
     }
 
-    /// Subscribe to updates to the pool's generation.
-    ///
-    /// This can be used to listen for errors that occur during connection
-    /// establishment or to get the current generation of the pool.
-    pub(crate) fn subscribe_to_generation_updates(&self) -> PoolGenerationSubscriber {
-        self.generation_subscriber.clone()
+    pub(crate) fn generation(&self) -> u32 {
```
patrickfreed, line 188: Since `clear` is async now, we no longer need to subscribe to updates and can just `await` on the clear. We still need to be able to get the generation at a point in time though.

src/error.rs
```diff
@@ -508,6 +531,7 @@ pub struct WriteError {
 #[non_exhaustive]
 pub struct BulkWriteError {
     /// Index into the list of operations that this error corresponds to.
+    #[serde(default = "default_index")]
```
patrickfreed, line 532: this is required since one of the tests omits this field, filed DRIVERS-1637 for updating it.

src/sdam/description/topology/test/sdam.rs
```diff
@@ -76,7 +145,15 @@ async fn run_test(test_file: TestFile) {
...
     let test_description = &test_file.description;
-    let mut topology_description = TopologyDescription::new(options).expect(test_description);
+
+    // TODO: RUST-360 unskip tests that rely on topology version
+    if test_description.contains("topologyVersion") {
+        println!("Skipping {} (RUST-360)", test_description);
+        return;
+    }
+
+    let topology = Topology::new_mocked(options.clone());
```
patrickfreed, line 155: Because we need to handle errors as part of this runner now, we need to go through the `Topology` instead of the `TopologyDescription`, which required some changes to the `Topology` to support "mocked" scenarios (i.e. don't start any monitors for new servers).

src/sdam/monitor.rs
```diff
@@ -159,7 +158,7 @@ impl HeartbeatMonitor {
                         .map(|sd| sd.is_available())
                         .unwrap_or(false)
                 {
-                    self.handle_error(e.to_string(), topology, server).await;
+                    self.handle_error(e, topology, server).await;
```
patrickfreed, line 161: I needed to go back to regular `Error` in a few places since we need the ability to handle the error differently based on its characteristics, which would be hard/impossible with just the error message.

src/sdam/state/mod.rs
```diff
@@ -225,54 +240,55 @@ impl Topology {
         self.common.message_manager.notify_topology_changed();
     }
 
-    /// Updates the topology based on an error that occurs before the handshake has completed during
-    /// an operation.
-    pub(crate) async fn handle_pre_handshake_error(&self, error: String, server: &Server) -> bool {
+    pub(crate) async fn handle_application_error(
```
patrickfreed, line 243: This derived from the new pseudocode in the SDAM spec: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#error-handling-pseudocode

src/error.rs
saghm: `Default` is already implemented for `usize` to be zero, so I think this can just be `#[serde(default)]` (without needing a custom function).

src/sdam/monitor.rs
saghm: Since we ended up deciding to keep Error cloneable, this seems fine.

src/error.rs
patrickfreed: Oh nice, I didn't know that, fixed.

saghm (approved):

src/sdam/test.rs
```diff
@@ -233,3 +233,97 @@ async fn sdam_min_pool_size_error() {
...
+
+    subscriber
+        .wait_for_event(Duration::from_millis(2000), |event| {
+            matches!(event, Event::CmapEvent(CmapEvent::ConnectionPoolCleared(_)))
+        })
+        .await
+        .expect("should see pool cleared event");
+
+    // Wait a little while for the server to be marked as Unknown.
+    // Once we have SDAM monitoring, this wait can be removed and be replaced
```
isabelatkinson, line 304: Can you add a todo with the ticket number here? (RUST-232)

patrickfreed: Done

isabelatkinson (approved):

patrickfreed:
The test I added ended up being pretty flaky on sharded clusters since the failpoint could be enabled on a different mongos than was used in subsequent inserts. I updated the test to use only a single host to fix this.

PR #313, opened 2021-04-02
RUST-728 Inherit collection options in find and count_documents

isabelatkinson:


src/test/spec/retryable_writes/mod.rs
```diff
@@ -57,15 +56,7 @@ async fn run_spec_tests() {
             let db_name = get_default_name(&test_case.description);
             let coll_name = "coll";
 
-            let write_concern = WriteConcern::builder().w(Acknowledgment::Majority).build();
```
isabelatkinson, line 60: `find` properly inheriting these options from the collection was messing with some of the retryable writes tests. I'm not sure why I included these in the first place (it's not specified to use a majority read/write concern anywhere in the retryable writes runner spec) so I removed them.

patrickfreed (approved):

PR #314, opened 2021-04-02
RUST-604 Use mongodb::Client for unified runner entities

patrickfreed:
RUST-604

This PR updates the unified runner to use regular `mongodb::Client` for the client entities instead of `EventClient` / `TestClient`, which allows us to unconditionally use `buildInfo` and drop the `collect_server_info` option in `TestClient`.

### DIFF
--- a/src/test/spec/unified_runner/entity.rs
+++ b/src/test/spec/unified_runner/entity.rs
@@ -1,6 +1,12 @@
-use std::ops::Deref;
+use std::{ops::Deref, sync::Arc};
 
-use crate::{bson::Bson, test::util::EventClient, Collection, Database};
+use crate::{
+    bson::Bson,
+    test::{CommandEvent, EventHandler},
+    Client,
+    Collection,
+    Database,
+};
 
 #[derive(Clone, Debug)]
 pub enum Entity {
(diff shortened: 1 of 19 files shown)


src/test/spec/unified_runner/entity.rs
```diff
@@ -13,11 +19,63 @@ pub enum Entity {
...
+            observer,
+            observe_events,
+            ignore_command_names,
+        }
+    }
+
+    /// Gets a list of all of the events of the requested event types that occurred on this client.
+    /// Ignores any event with a name in the ignore list. Also ignores all configureFailPoint
+    /// events.
+    pub fn get_filtered_events(
```
patrickfreed, line 46: This was previously on `EventClient`, since it's only used by the unified runner I just moved it here. I added `get_filtered_command_events` to `EventHandler` in case we want to use it elsewhere.

src/test/util/mod.rs
```diff
@@ -87,32 +87,23 @@ impl TestClient {
...
 
-            let get_parameters =
-                RunCommand::new("admin".into(), doc! { "getParameter": "*" }, None).unwrap();
+        let info: BuildInfo = bson::from_bson(Bson::Document(response)).unwrap();
+        let server_version_str = info.version.split('-').next().unwrap();
+        let server_version = Version::parse(server_version_str).unwrap();
 
-            // The command above may fail due to insufficient permissions. In that case, the unified
-            // test runner will skip any tests with a serverParameters runOnRequirement as the check
-            // will fail.
```
patrickfreed, line 111: It wasn't clear when this actually failed and we were unwrapping it anyways in the run on requirement, so I just removed this handling and unwrapped here.

isabelatkinson (approved):

PR #317, opened 2021-04-05
RUST-714 Fix possible deadlock during retry (1.x driver)

patrickfreed:
RUST-714

This backports a fix for a potential deadlock that can occur during a retry when all connections are checked out from the pool.

### DIFF
--- a/src/test/spec/retryable_reads.rs
+++ b/src/test/spec/retryable_reads.rs
@@ -36,9 +36,9 @@ async fn retry_releases_connection() {
     client_options.retry_reads = Some(true);
     client_options.max_pool_size = Some(1);
 
-    let client = TestClient::with_options(Some(client_options.clone()), true).await;
+    let client = TestClient::with_options(Some(client_options), true).await;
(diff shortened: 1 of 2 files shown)


src/test/spec/retryable_reads.rs
```diff
@@ -1,10 +1,59 @@
...
+    let collection = client
+        .database("retry_releases_connection")
+        .collection("retry_releases_connection");
+    collection.insert_one(doc! { "x": 1 }, None).await.unwrap();
+
+    let options = FailCommandOptions::builder().error_code(91).build();
+    let failpoint = FailPoint::fail_command(&["find"], FailPointMode::Times(1), Some(options));
+    let _fp_guard = client.enable_failpoint(failpoint, None).await.unwrap();
+
+    RUNTIME
```
patrickfreed, line 54: Prior to this change, this timeout would get hit every time since the driver was stuck. Post-change it should get through every time.

isabelatkinson (approved):

PR #318, opened 2021-04-05
RUST-714 Fix possible deadlock during retry

patrickfreed:
This PR adds a test to ensure deadlock doesn't occur during a retry when all connections are checked out from the pool. This was possible in the released 2.0.0-alpha, but was incidentally fixed in a subsequent PR.

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -125,10 +125,13 @@ impl Client {
                     .topology
                     .handle_application_error(
                         err.clone(),
-                        HandshakePhase::after_completion(conn),
+                        HandshakePhase::after_completion(&conn),
                         &server,
                     )
                     .await;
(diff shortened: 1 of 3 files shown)


src/client/executor.rs
```diff
@@ -125,10 +125,13 @@ impl Client {
                     .topology
                     .handle_application_error(
                         err.clone(),
-                        HandshakePhase::after_completion(conn),
+                        HandshakePhase::after_completion(&conn),
```
patrickfreed, line 128: This was technically fixed already since `after_completion` dropped the connection, but I updated this anyways to make the drop explicit in case `after_completion` is updated in the future to somehow hold on to the connection for longer than it ought to.

isabelatkinson (approved):
good catch!

PR #322, opened 2021-04-06
RUST-724 Include versioned API options in the public API

patrickfreed:
RUST-724

This updates the `versioned-api` branch to include the versioned API options in the public API for private beta users to use. We'll need to do this on the master branch in the future, so I filed RUST-732 to track that.


isabelatkinson (approved):
code changes lgtm, did you mean to link a different ticket for the second one you mentioned for unhiding this stuff on master? they both link to the release tasks ticket

patrickfreed:
Oh whoops, yeah that was a typo. Updated to RUST-732.

PR #323, opened 2021-04-08
RUST-735 Remove Document as default generic type on Collection and Cursor

patrickfreed:
RUST-735

This PR removes `Document` as the default generic type for `Collection` and `Cursor` and updates `Database::collection` and `Database::collection_with_options` to return `Collection<T>`. As part of this, the existing typed helpers became redundant and were removed. These changes were made to ease and promote the use of serde models with the driver instead of regular Documents.

Note: I didn't update the README examples to avoid confusing users of the 1.2.x or 2.0.0-alpha versions. I filed RUST-736 to cover the work for updating that, which we can do right before the release.

### DIFF
--- a/src/cursor/mod.rs
+++ b/src/cursor/mod.rs
@@ -43,11 +43,11 @@ use common::{GenericCursor, GetMoreProvider, GetMoreProviderResult};
 ///
 /// ```rust
 /// # use futures::stream::StreamExt;
-/// # use mongodb::{Client, error::Result};
+/// # use mongodb::{bson::Document, Client, error::Result};
 /// #
 /// # async fn do_stuff() -> Result<()> {
 /// # let client = Client::with_uri_str("mongodb://example.com").await?;
-/// # let coll = client.database("foo").collection("bar");
+/// # let coll = client.database("foo").collection::<Document>("bar");
 /// # let mut cursor = coll.find(None, None).await?;
 /// #
 /// while let Some(doc) = cursor.next().await {
--- a/src/cursor/mod.rs
+++ b/src/cursor/mod.rs
@@ -81,7 +81,7 @@ use common::{GenericCursor, GetMoreProvider, GetMoreProviderResult};
 /// # }
 /// ```
 #[derive(Debug)]
-pub struct Cursor<T = Document>
+pub struct Cursor<T>
 where
     T: DeserializeOwned + Unpin,
 {
@@ -139,7 +139,7 @@ where
         let coll = self
             .client
             .database(ns.db.as_str())
-            .collection(ns.coll.as_str());
+            .collection::<Document>(ns.coll.as_str());
         let cursor_id = self.wrapped_cursor.id();
         RUNTIME.execute(async move { coll.kill_cursor(cursor_id).await });
     }
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -39,7 +39,7 @@ use crate::{
 /// ```rust
 /// 
 /// # #[cfg(not(feature = "sync"))]
-/// # use mongodb::{Client, error::Result};
+/// # use mongodb::{bson::Document, Client, error::Result};
 /// # #[cfg(feature = "async-std-runtime")]
 /// # use async_std::task;
 /// # #[cfg(feature = "tokio-runtime")]
@@ -55,7 +55,7 @@ use crate::{
 ///     let db_ref = db.clone();
 ///
 ///     task::spawn(async move {
-///         let collection = db_ref.collection(&format!("coll{}", i));
+///         let collection = db_ref.collection::<Document>(&format!("coll{}", i));
 ///
 ///         // Do something with the collection
 ///     });
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -129,36 +129,17 @@ impl Database {
         self.inner.write_concern.as_ref()
     }
 
-    /// Gets a handle to a collection specified by `name` of the database. The `Collection` options
-    /// (e.g. read preference and write concern) will default to those of the `Database`.
-    ///
-    /// This method does not send or receive anything across the wire to the database, so it can be
-    /// used repeatedly without incurring any costs from I/O.
-    pub fn collection(&self, name: &str) -> Collection {
-        Collection::new(self.clone(), name, None)
-    }
-
     /// Gets a handle to a collection with type `T` specified by `name` of the database. The
     /// `Collection` options (e.g. read preference and write concern) will default to those of the
     /// `Database`.
     ///
     /// This method does not send or receive anything across the wire to the database, so it can be
     /// used repeatedly without incurring any costs from I/O.
-    pub fn collection_with_type<T>(&self, name: &str) -> Collection<T>
+    pub fn collection<T>(&self, name: &str) -> Collection<T>
     where
         T: Serialize + DeserializeOwned + Unpin + Debug,
     {
         Collection::new(self.clone(), name, None)
-    }
-
-    /// Gets a handle to a collection specified by `name` in the cluster the `Client` is connected
-    /// to. Operations done with this `Collection` will use the options specified by `options` by
-    /// default and will otherwise default to those of the `Database`.
-    ///
-    /// This method does not send or receive anything across the wire to the database, so it can be
-    /// used repeatedly without incurring any costs from I/O.
-    pub fn collection_with_options(&self, name: &str, options: CollectionOptions) -> Collection {
-        Collection::new(self.clone(), name, Some(options))
     }
 
     /// Gets a handle to a collection with type `T` specified by `name` in the cluster the `Client`
@@ -167,7 +148,7 @@ impl Database {
     ///
     /// This method does not send or receive anything across the wire to the database, so it can be
     /// used repeatedly without incurring any costs from I/O.
-    pub fn collection_with_type_and_options<T>(
+    pub fn collection_with_options<T>(
         &self,
         name: &str,
         options: CollectionOptions,
@@ -213,7 +194,7 @@ impl Database {
         &self,
         filter: impl Into<Option<Document>>,
         options: impl Into<Option<ListCollectionsOptions>>,
-    ) -> Result<Cursor> {
+    ) -> Result<Cursor<Document>> {
         let list_collections = ListCollections::new(
             self.name().to_string(),
             filter.into(),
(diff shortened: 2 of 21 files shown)


src/db/mod.rs
```diff
@@ -129,45 +129,26 @@ impl Database {
...
-    }
-
     /// Gets a handle to a collection with type `T` specified by `name` of the database. The
     /// `Collection` options (e.g. read preference and write concern) will default to those of the
     /// `Database`.
     ///
     /// This method does not send or receive anything across the wire to the database, so it can be
     /// used repeatedly without incurring any costs from I/O.
-    pub fn collection_with_type<T>(&self, name: &str) -> Collection<T>
+    pub fn collection<T>(&self, name: &str) -> Collection<T>
```
patrickfreed, line 138: I know we discussed the possibility of introducing a `collection_with_document` helper (or something like that) as an ergonomic way to retrieve a `Collection<Document>`, but after writing a few of these conversions, it seemed simpler / easier to just use the turbofish (`::<>`) operator on the existing helpers.

e.g.
```rust
let collection: Collection<Document> = db.collection("foo"); // explicit type way
let collection = db.collection::<Document>("foo"); // turbofish
let collection = db.collection_with_document("foo"); // helper
```

To me, the helper didn't really make it any easier to type nor did it actually make it clearer what was happening, so it seemed fine to me to just not have it. Thoughts?

src/db/mod.rs
```diff
@@ -213,7 +194,7 @@ impl Database {
         &self,
         filter: impl Into<Option<Document>>,
         options: impl Into<Option<ListCollectionsOptions>>,
-    ) -> Result<Cursor> {
+    ) -> Result<Cursor<Document>> {
```
patrickfreed, line 197: In Swift, we have this return a `MongoCursor<CollectionSpecification>`, and I think it would be a nice improvement here as well for 2.0. The versioned API will ensure that we won't have to worry about deserialization errors for the foreseeable future.

src/sync/coll.rs
```diff
@@ -153,7 +153,7 @@ where
         &self,
         pipeline: impl IntoIterator<Item = Document>,
         options: impl Into<Option<AggregateOptions>>,
-    ) -> Result<Cursor> {
+    ) -> Result<Cursor<Document>> {
```
patrickfreed, line 156: We could consider returning a `Cursor<O>` here (where `O` is a generic parameter to the `aggregate` method, not the `T` from the `Collection`). Or should we have a separate `aggregate_with_output_type` or something? I think with aggregation it's more likely that a user might just want to retrieve a document than with a `find`, since they may mutate the output along the way, so it's a little riskier to have this one be pure generic. Thoughts?

patrickfreed (commented):
I have a few non-as-yet-discussed comments/questions about the public API, so tagging in @kmahar as well.

src/cursor/mod.rs
```diff
@@ -139,7 +139,7 @@ where
         let coll = self
             .client
             .database(ns.db.as_str())
-            .collection(ns.coll.as_str());
+            .collection::<Document>(ns.coll.as_str());
```
kmahar, line 142: this is honestly not really relevant to the PR, but I'm just curious for my own knowledge and future reference for Swift- aren't some cursors not actually tied to particular collections e.g. one resulting from db-level aggregate? how does this work in that case?

src/db/mod.rs
kmahar: sgtm (though that's probably not surprising since I already agreed to it once...) 

if so, would be nice to use a `DatabaseSpecification` type for `list_databases` too (assuming `listDatabases` is also in API version 1?)

src/db/mod.rs
kmahar: from my outsider perspective it doesn't seem to make a significant difference, but I'd defer to you two on whether `collection_with_document` is in any way more idiomatic

src/sync/coll.rs
kmahar: per slack conversation: seems we're all on the same page the `T` from the collection should not be used since people often do transformations here.

as for whether to require a type be specified in this method (which I was not thinking about the option of in the slack discussion), or just having a separate `aggregate_with_type` I don't have particularly strong feelings. I agree with your comment that it does seem a bit at odds with the choice to make requiring a type the default for `db.collection`. OTOH maybe it's less likely that people will bother writing custom `Deserialize` types for the output types of single methods?

src/db/mod.rs
patrickfreed: Good call, I filed RUST-740 to cover the both of them.

src/sync/coll.rs
patrickfreed: Yeah, my original inclination was to just leave this as `Document` by default and introduce the new separate method, since I imagine it'll be pretty common to use `Document` at least while prototyping the pipeline. I'm starting to flip back to wanting to maintain consistency though, given how easy it is to just append a `::<Document>` to the end of the `aggregate` call. Especially in the face of complex pipelines, handling the generic type would be a piece of cake.

_However_

This got me trying it out and I noticed an interesting / frustrating technical challenge to doing it the `Cursor<O>` way: turbofish cannot be used in conjunction with `impl Trait` in the argument position (yet, see https://github.com/rust-lang/rust/issues/83701).

e.g.
```rust
fn foo<T>(bar: impl ToString) { ... }
foo::<usize>("blah"); // compile error!
```
yields the following compiler error:
```
error[E0632]: cannot provide explicit generic arguments when `impl Trait` is used in argument position
  --> src/main.rs:21:11
   |
21 |     foo::<usize>("hello");
   |           ^^^^^ explicit generic argument not allowed

```
So you'd have to do it the explicit type parameter way:
```rust
let x: Cursor<Document> = coll.aggregate(vec![], None).await?
```
Which could be pretty frustrating.

This will make it tough to implement the one that takes in a generic argument, since we use impl Trait in all our CRUD methods for our options types (e.g. to be able to say `options` instead of having to say `Some(options)`). This makes me think we may want to do the builder approach here and have the method be on `Cursor` actually, something like:

```rust
let cursor = coll.aggregate(vec![ ... ], None).await?.with_type::<MyDeserializeType>();
```
This is pretty ugly admittedly, but I think it's preferable to breaking consistency with how options are provided to the other CRUD methods or _requiring_ users to use explicit type annotations instead of turbofish.

kmahar: I agree changing the way we pass options either here or everywhere doesn't seem like a great path forward.

is turbofish common/used rather than type annotations that people would be running into this a lot?

adding the method to cursor seems reasonable though it does look clunky.... it also seems consistent with the `clone_with_type` behavior that we currently have on `Collection` to change the generic type. and maybe it would be useful in other use cases as well - e.g. if you provide a `projection` via `FindOptions` that changes the shape of your documents as well, and you could call this on the resulting cursor.

isabelatkinson: Yeah none of these options is very ideal to me; `with_type` seems like it would be more ergonomic for chaining method calls together, since providing a type annotation would require breaking out the call to `aggregate` into a separate line. Updating our CRUD API to use the builder pattern would get rid of this issue, right?

In any case since we're opting to go the `aggregate_with_type` route, I think we can defer coming to a conclusion here til we implement that (and make a decision regarding the builder pattern).

isabelatkinson (commented):
Code looks good, just a few questions. Also wanted to double check that the examples in the README are still good, I believe they should be unaffected but would be worth taking a glance to sanity check if you haven't yet. (It would also probably be a good idea to put those in a test somewhere to make sure they don't become outdated.)

src/sync/coll.rs
patrickfreed: I think even with the builder pattern for options/sessions we'd still have the same issue unfortunately, since we use `impl IntoIterator` for the pipeline.

Also, to clarify, I don't think `aggregate_with_type` method will really be possible, since we'd have to ditch the `impl Trait` arguments, so I think we kind of have to do the `Cursor::with_type` approach. That being said, we don't need to implement that functionality now.

patrickfreed (commented):
I was originally planning on holding off on updating the README until close to the release (RUST-736), but now that we're planning on doing a second alpha I went ahead and updated them. We can still use RUST-736 to track the work for updating the examples to be use serde more prominently. Also, good point on having tests for them--I added a file to our `tests` folder since the examples only use the public API to ensure the examples compile.

src/sync/coll.rs
isabelatkinson: Ah right, I forgot we use `impl` for the pipeline.

isabelatkinson (approved):

isabelatkinson:
whoops, just noticed there's a clippy failure in the readme file (lgtm otherwise though, no need for another round of review on taht)

kmahar (approved):

seanpianka:
After updating from alpha to alpha.1, I'm now seeing this run-time error with the same code:
```
failed to update collection: Error { kind: ArgumentError { message: \"update document must have first key starting with \\\'$\" }, labels: [] }" }
```

patrickfreed:
The `update_` family of methods are meant to be used with [update operator expressions](https://docs.mongodb.com/manual/reference/command/update/#std-label-update-command-upsert) rather than replacement documents, which the `replace` family of methods are mean to be used with, but we only started validating this in alpha.1 (RUST-718). To fix your issue, update any usages of `update_one` that use a replacement document to use `replace_one` instead.

e.g.
```rust
coll.update_one(doc! { "x": 2 }, doc! { "x": 3 }, None).await?; // error
coll.replace_one(doc! { "x": 2 }, doc! { "$inc": { "x": 1 } }, None).await?; // error

coll.update_one(doc! { "x": 2 }, doc! { "$inc": { "x": 1 } }, None).await?; // correct
coll.replace_one(doc! { "x": 2 }, doc! { "x": 3 }, None).await?; // correct
```

If you have any more questions or concerns, please [open a ticket on our Jira project](https://github.com/mongodb/mongo-rust-driver#bug-reporting--feature-requests) rather than continuing the discussion here or on any other PR, since it can be difficult for the team or other users to find the discussion in the future otherwise.

seanpianka:
Gotcha, I appreciate your detailed reply, thank you @patrickfreed! In the future, I'll open a ticket in the Jira project. 

I managed to fix the `update_one` call via `doc! { "$set": replacement_doc }`, but `replace_one` more closely captures my intent. I'll out that, thanks again!

patrickfreed:
No problem, glad you were able to get it working!

PR #325, opened 2021-04-21
RUST-615 Upgrade typed-builder to 0.9.0

patrickfreed:
RUST-615

This PR updates the dependency on `typed-builder` to 0.9.0. This is a breaking change, since we now use `strip_option` instead of `into` for the various builder methods, meaning those method take in `T` instead of `impl Into<Option<T>>`. In practice, this results in the following changes:
- type inference can now be used on the underlying type 
  - e.g. `.thing(Some(other_thing.into()))` can now just be written as `.thing(other_thing.into())`
  - `.number(5 as u32)` can now be written as `.number(5)`
 - `Option<T>` instances cannot be passed into the builder methods
   - The options type now needs to be built (either via the builder or `MyOptoinsType::default()` and then mutated


src/cmap/options.rs
```diff
@@ -109,18 +99,18 @@ pub struct ConnectionPoolOptions {
 
 impl ConnectionPoolOptions {
     pub(crate) fn from_client_options(options: &ClientOptions) -> Self {
-        Self::builder()
```
patrickfreed, line 112: This situation here is one of the more painful examples not being able to pass in an `Option<T>` anymore, though I doubt this will be common among our users. The [standard examples for the builder pattern](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html) all don't take in `Option`, and it seems neither does the actual [`std::process::Command`](https://doc.rust-lang.org/std/process/struct.Command.html) that it's based off of, so I think we should be okay here.

kmahar: I agree, it doesn't seem like much of an issue. I can't think of a lot of use cases for storing some value you're going to pass in as an option in an optional variable

kmahar (approved):
a small question mainly for my own understanding, but LGTM!

src/client/options/mod.rs
```diff
@@ -289,54 +289,54 @@ pub struct ClientOptions {
     /// The application name that the Client will send to the server as part of the handshake. This
     /// can be used in combination with the server logs to determine which Client is connected to a
     /// server.
-    #[builder(default)]
+    #[builder(default, setter(strip_option))]
     pub app_name: Option<String>,
 
-    #[builder(default)]
+    #[builder(default, setter(skip))]
```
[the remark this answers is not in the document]
patrickfreed: Since `heartbeat_freq_test` has the `#[cfg(test)]` attribute, it only gets compiled when the tests are being run. The builder appears to inherit this as well from testing it out. For the non-`#[cfg(test)]` ones, they actually do get included in the public builder, regardless of the field's ACL. In fact, the 1.x driver actually leaks these :grimacing: (https://docs.rs/mongodb/1.2.1/mongodb/options/struct.ClientOptions.html#method.builder)

PR #326, opened 2021-04-27
RUST-591 Use separate ConnectionPoolOptions types for monitoring and internal usage

patrickfreed:
RUST-591

This PR introduces a public `event::cmap::ConnectionPoolOptions` type separate from the (now internal) `ConnectionPoolOptions` used to configure the connection pool. As part of this, the credential was removed from `event::cmap::ConnectionPoolOptions` and the struct was also marked as `non_exhaustive`.


src/cmap/options.rs
```diff
@@ -94,23 +93,42 @@ pub struct ConnectionPoolOptions {
...
-        pool_options.credential = options.credential.clone();
-        pool_options.driver_info = options.driver_info.clone();
-        pool_options.event_handler = options.cmap_event_handler.clone();
-        pool_options.max_idle_time = options.max_idle_time;
-        pool_options.min_pool_size = options.min_pool_size;
-        pool_options.tls_options = options.tls_options();
-        pool_options.wait_queue_timeout = options.wait_queue_timeout;
-
-        pool_options
+        Self {
```
patrickfreed, line 101: Using the struct literal here ensures that no fields can accidentally be forgotten (e.g. like `server_api` was)

src/event/cmap.rs
```diff
@@ -31,6 +35,70 @@ pub struct PoolCreatedEvent {
...
+/// client-level, `ConnectionPoolOptions` is exposed for the purpose of CMAP event handling.
+#[derive(Clone, Default, Deserialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct ConnectionPoolOptions {
+    /// The application name specified by the user. This is sent to the server as part of the
+    /// handshake that each connection makes when it's created.
+    pub app_name: Option<String>,
+
+    /// The connect timeout passed to each underlying TcpStream when attemtping to connect to the
```
isabelatkinson, line 48: nit: typo on "attempting" (both here and in the other `ConnectionPoolOptions` comment)

isabelatkinson (approved):
looks good, just one typo (don't need to do another round of review for that fix)

src/event/cmap.rs
patrickfreed: fixed

PR #328, opened 2021-04-27
RUST-740 Return Deserialize structs from list_databases and list_collections

patrickfreed:
RUST-740

This PR updates `Client::list_databases` and `Database::list_collections` to return structs that implement `Deserialize` rather than just `Document`.

### DIFF
--- a/src/db/options.rs
+++ b/src/db/options.rs
@@ -85,7 +85,7 @@ pub struct CreateCollectionOptions {
 
 /// Specifies how strictly the database should apply validation rules to existing documents during
 /// an update.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub enum ValidationLevel {
@@ -100,7 +100,7 @@ pub enum ValidationLevel {
 
 /// Specifies whether the database should return an error or simply raise a warning if inserted
 /// documents do not pass the validation.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub enum ValidationAction {
@@ -111,7 +111,7 @@ pub enum ValidationAction {
 }
 
 /// Specifies default configuration for indexes created on a collection, including the _id index.
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, TypedBuilder, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct IndexOptionDefaults {
--- a/src/db/options.rs
+++ b/src/db/options.rs
@@ -29,7 +29,7 @@ pub struct DatabaseOptions {
 /// These are the valid options for creating a collection with
 /// [`Database::create_collection`](../struct.Database.html#method.create_collection).
 #[skip_serializing_none]
-#[derive(Debug, Default, TypedBuilder, Serialize)]
+#[derive(Clone, Debug, Default, TypedBuilder, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 #[builder(field_defaults(default, setter(strip_option)))]
 #[non_exhaustive]
@@ -55,8 +55,7 @@ pub struct CreateCollectionOptions {
     /// Specifies a validator to restrict the schema of documents which can exist in the
     /// collection. Expressions can be specified using any query operators except `$near`,
     /// `$nearSphere`, `$text`, and `$where`.
-    #[serde(rename = "validator")]
-    pub validation: Option<Document>,
+    pub validator: Option<Document>,
 
     /// Specifies how strictly the database should apply the validation rules to existing documents
     /// during an update.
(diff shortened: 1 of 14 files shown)


src/db/options.rs
```diff
@@ -55,8 +55,7 @@ pub struct CreateCollectionOptions {
     /// Specifies a validator to restrict the schema of documents which can exist in the
     /// collection. Expressions can be specified using any query operators except `$near`,
     /// `$nearSphere`, `$text`, and `$where`.
-    #[serde(rename = "validator")]
-    pub validation: Option<Document>,
+    pub validator: Option<Document>,
```
patrickfreed, line 58: This was originally a typo I believe, all other drivers have this as `validator`. Took the opportunity to fix it up while I'm modifying these files.

src/db/options.rs
```diff
@@ -112,8 +111,9 @@ pub enum ValidationAction {
 }
 
 /// Specifies default configuration for indexes created on a collection, including the _id index.
-#[derive(Clone, Debug, PartialEq, Serialize)]
+#[derive(Clone, Debug, TypedBuilder, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
+#[non_exhaustive]
 pub struct IndexOptionDefaults {
```
patrickfreed, line 117: This struct was missing a few attributes so I added them. I don't think any of our users will use this though.

src/operation/list_databases/mod.rs
```diff
@@ -85,5 +85,4 @@ impl Operation for ListDatabases {
 #[derive(Debug, Deserialize)]
 struct ResponseBody {
     databases: Vec<Document>,
-    total_size: Option<i64>,
```
patrickfreed, line 88: This wasn't being used and in fact wasn't even being deserialized, since the server returned this value as a double until 5.0.

src/results.rs
```diff
@@ -71,3 +72,94 @@ pub(crate) struct GetMoreResult {
...
+#[non_exhaustive]
+/// A struct modeling the information about an individual database returned from
+/// [`Client::list_databases`](../struct.Client.html#method.list_databases).
+pub struct DatabaseSpecification {
+    /// The name of the database.
+    pub name: String,
+
+    /// The amount of disk space in bytes that is consumed by the database.
+    #[serde(deserialize_with = "crate::bson_util::deserialize_i64_from_bson_number")]
+    pub size_on_disk: i64,
```
patrickfreed, line 157: This value is returned as a bunch of different types depending on the server version, so we need a helper to deserialize it from whatever type to an integer. See [here](https://jira.mongodb.org/browse/SERVER-52544?focusedCommentId=3505206&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3505206) for an overview if interested.

src/results.rs
```diff
@@ -71,3 +72,94 @@ pub(crate) struct GetMoreResult {
...
+pub struct CollectionSpecification {
+    /// The name of the collection.
+    pub name: String,
+
+    /// Type of the data store.
+    #[serde(rename="type")]
+    pub collection_type: CollectionType,
+
+    /// The options used to create the collection.
+    pub options: CreateCollectionOptions,
```
patrickfreed, line 136: I thought about creating a separate type for this in case they ever diverge (and also because it contains fields like `write_concern` that don't make sense in this context), but the [docs](https://docs.mongodb.com/manual/reference/command/listCollections/#output) explicitly state that this field will contain the exact options provided to `createCollection`, and looking through the server code confirms also this, so I figured this was pretty low risk. Let me know what you think.

patrickfreed (commented):
Tagging in @kmahar for the public API changes.

src/db/options.rs
kmahar: not sure what you've got in mind in terms of a change log/migration guide for 2.0, but would it make sense to file a separate ticket for this just so we remember to mention it?

src/db/options.rs
kmahar: TIL `create` accepts `indexOptionDefaults`

src/results.rs
```diff
@@ -71,3 +72,94 @@ pub(crate) struct GetMoreResult {
...
+#[non_exhaustive]
+pub enum CollectionType {
+    /// Indicates that the data store is a view.
+    View,
+
+    /// Indicates that the data store is a collection.
+    Collection,
+
+    /// An unknown collection type. This is included for forwards compatibility.
+    Other(String),
```
kmahar, line 89: given this command is part of the versioned API (which was part of our reasoning for why introducing such structs is pretty safe), is this something we actually need to worry about? 

the versioned API scope lists under "breaking changes":
> Add a new value to a StableCommand reply field's enum-like fixed set of values, e.g. a new index type (unless there's an opt-in mechanism besides API version).

a new collection type I would think falls into the same category as a new index type, BUT I will say I don't totally understand what an "opt-in mechanism" would be and if that negates our ability to ignore the "other" case here.

thoughts?

src/results.rs
kmahar: I agree it seems low risk

src/results.rs
kmahar: that makes sense; sgtm

src/results.rs
kmahar: I agree u64 does seem to make sense for a size field

kmahar (approved):
LGTM mod Isabel's comment, I don't need to re-review

src/results.rs
patrickfreed: I thought so too originally, but this is a value that's deserialized from server responses, meaning it can _possibly_ have negative values. I checked elsewhere in the driver for precedent and saw that we use `i64` for similar values, e.g. [`matched_count`](https://github.com/mongodb/mongo-rust-driver/blob/master/src/results.rs#L52) in `UpdateResult`. That being said, it seems like we probably would be totally fine using an unsigned value here if we wanted to, so I've updated it.

I also updated `get_int` to be a little more lenient when reading integers from doubles in case of a floating point rounding issue.

patrickfreed: I filed RUST-764 to discuss updating other such fields.

isabelatkinson (approved):

PR #329, opened 2021-04-28
RUST-738 Remove Arc wrapper around BSON errors

patrickfreed:
RUST-738

Now that the BSON errors are `Clone`, we can remove the `Arc` wrapper around them in `ErrorKind`. To bring in these changes, this PR also updates the `Cargo.toml` to track the master branch of `bson-rust`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ bson-u2i = ["bson/u2i"]
 async-trait = "0.1.42"
 base64 = "0.13.0"
 bitflags = "1.1.0"
-bson = "1.2.0"
+bson = { git = "https://github.com/mongodb/bson-rust" }
 chrono = "0.4.7"
 derivative = "2.1.1"
 futures = "0.3.5"
--- a/src/cursor/mod.rs
+++ b/src/cursor/mod.rs
@@ -151,11 +151,11 @@ type ImplicitSessionCursor = GenericCursor<ImplicitSessionGetMoreProvider>;
 
 struct ImplicitSessionGetMoreResult {
     get_more_result: Result<GetMoreResult>,
-    session: Option<ClientSession>,
+    session: Option<Box<ClientSession>>,
 }
 
 impl GetMoreProviderResult for ImplicitSessionGetMoreResult {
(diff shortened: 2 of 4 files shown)


src/cursor/mod.rs
```diff
@@ -170,7 +170,7 @@ impl GetMoreProviderResult for ImplicitSessionGetMoreResult {
 /// This is to be used with cursors associated with implicit sessions.
 enum ImplicitSessionGetMoreProvider {
```
patrickfreed, line 171: Clippy started complaining that the `Executing` variant was much bigger than the `Idle` one and suggested we `Box` the `Idle` one to fix this. I believe this is because, after the switch to `indexmap` in `bson`, the size of `Document` got a little bigger and was enough to trip the lint here.

src/sdam/description/server.rs
```diff
@@ -257,7 +257,7 @@ impl ServerDescription {
             .as_ref()
             .map_err(Clone::clone)?
             .as_ref()
-            .and_then(|reply| reply.command_response.election_id.clone());
+            .and_then(|reply| reply.command_response.election_id);
```
patrickfreed, line 260: `ObjectId` is now `Copy`, so clippy wanted us to fix this too.

Cargo.toml
```diff
@@ -34,7 +34,7 @@ bson-u2i = ["bson/u2i"]
 async-trait = "0.1.42"
 base64 = "0.13.0"
 bitflags = "1.1.0"
-bson = "1.2.0"
+bson = { git = "https://github.com/mongodb/bson-rust" }
```
patrickfreed, line 37: Before we release the next alpha or beta on crates.io, we'll need to release and alpha/beta of `bson` to depend on here. In the meantime, just tracking the git repo should be fine.

src/cursor/mod.rs
isabelatkinson: I'm actually getting this locally (without the BSON update), I think it's just a new complaint from clippy. Using `Box` sgtm.

isabelatkinson (approved):

PR #330, opened 2021-04-29
RUST-742 / RUST-739 Improve Error API and eliminate public dependencies on unstable crates

patrickfreed:
RUST-742 / RUST-739

This PR makes a number of breaking updates to the `Error` struct and `ErrorKind` enum, as well as some changes to other areas of the API where we re-export types from unstable crates. Each commit has been made separate to include changes specific to each ticket.

Notable changes
- Wrapped `Error::kind` in a `Box` to reduce the size of `Error`
    - This is the change that a user suggested and I did that writeup about, and they're included here with the unstable dependency stuff since they both modified and reduced the size of `ErrorKind`. I know @kmahar hasn't had a chance to read through the writeup on this yet, but I wanted to open this PR to get the reviews going on the later changes (which are the majority of this PR). Since they're separate commits anyways I can always go back and modify the first one as needed, but I didn't really anticipate having to do so.
- Consolidated several redundant error cases
- Rewrote error cases that simply wrapped errors from other crates to instead just use their messages
- Defined a wrapper `ResolverConfig` type within the driver rather than accepting `trust-dns-resolver`'s `ResolverConfig`, which is unstable
- Hid `TlsOptions::into_rustls_config` from the public API

### DIFF
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -1,6 +1,8 @@
 #[cfg(all(test, not(feature = "sync")))]
 mod test;
 
+mod resolver_config;
+
 use std::{
     collections::HashSet,
     fmt::{self, Display, Formatter},
@@ -28,7 +30,6 @@ use serde::{
     Deserializer,
 };
 use strsim::jaro_winkler;
-pub use trust_dns_resolver::config::ResolverConfig;
 use typed_builder::TypedBuilder;
 use webpki_roots::TLS_SERVER_ROOTS;
 
@@ -43,6 +44,8 @@ use crate::{
     selection_criteria::{ReadPreference, SelectionCriteria, TagSet},
     srv::SrvResolver,
 };
+
+pub use resolver_config::ResolverConfig;
 
 const DEFAULT_PORT: u16 = 27017;
 
@@ -820,7 +823,7 @@ impl ClientOptions {
         options.resolver_config = resolver_config.clone();
 
         if srv {
-            let mut resolver = SrvResolver::new(resolver_config).await?;
+            let mut resolver = SrvResolver::new(resolver_config.map(|config| config.inner)).await?;
             let mut config = resolver
                 .resolve_client_options(&options.hosts[0].hostname)
                 .await?;
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -149,8 +149,8 @@ impl StreamAddress {
         let hostname = match parts.next() {
             Some(part) => part,
             None => {
-                return Err(ErrorKind::InvalidHostname {
-                    hostname: address.to_string(),
+                return Err(ErrorKind::ArgumentError {
+                    message: format!("invalid server address: \"{}\"", address),
                 }
                 .into())
             }
@@ -158,13 +158,19 @@ impl StreamAddress {
 
         let port = match parts.next() {
             Some(part) => {
-                let port = u16::from_str(part).map_err(|_| ErrorKind::InvalidHostname {
-                    hostname: address.to_string(),
+                let port = u16::from_str(part).map_err(|_| ErrorKind::ArgumentError {
+                    message: format!(
+                        "port must be valid 16-bit unsigned integer, instead got: {}",
+                        part
+                    ),
                 })?;
 
                 if parts.next().is_some() {
-                    return Err(ErrorKind::InvalidHostname {
-                        hostname: address.to_string(),
+                    return Err(ErrorKind::ArgumentError {
+                        message: format!(
+                            "address \"{}\" contains more than one unescaped ':'",
+                            address
+                        ),
                     }
                     .into());
                 }
@@ -592,9 +598,8 @@ impl TlsOptions {
         if let Some(path) = self.ca_file_path {
             store
                 .add_pem_file(&mut BufReader::new(File::open(&path)?))
-                .map_err(|_| ErrorKind::ParseError {
-                    data_type: "PEM-encoded root certificate".to_string(),
-                    file_path: path,
+                .map_err(|_| ErrorKind::TlsConfigError {
+                    message: format!("Unable to parse PEM-encoded root certificate from {}", path),
                 })?;
         } else {
             store.add_server_trust_anchors(&TLS_SERVER_ROOTS);
@@ -607,9 +612,11 @@ impl TlsOptions {
             let certs = match pemfile::certs(&mut file) {
                 Ok(certs) => certs,
                 Err(()) => {
-                    return Err(ErrorKind::ParseError {
-                        data_type: "PEM-encoded client certificate".to_string(),
-                        file_path: path,
+                    return Err(ErrorKind::TlsConfigError {
+                        message: format!(
+                            "Unable to parse PEM-encoded client certificate from {}",
+                            path
+                        ),
                     }
                     .into())
                 }
@@ -619,16 +626,19 @@ impl TlsOptions {
             let key = match pemfile::rsa_private_keys(&mut file) {
                 Ok(key) => key,
                 Err(()) => {
-                    return Err(ErrorKind::ParseError {
-                        data_type: "PEM-encoded RSA key".to_string(),
-                        file_path: path,
+                    return Err(ErrorKind::TlsConfigError {
+                        message: format!("Unable to parse PEM-encoded RSA key from {}", path),
                     }
                     .into())
                 }
             };
 
             // TODO: Get rid of unwrap.
-            config.set_single_client_cert(certs, key.into_iter().next().unwrap())?;
+            config
+                .set_single_client_cert(certs, key.into_iter().next().unwrap())
+                .map_err(|e| ErrorKind::TlsConfigError {
+                    message: e.to_string(),
+                })?;
         }
 
         Ok(config)
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -588,7 +588,7 @@ impl ServerCertVerifier for NoCertVerifier {
 
 impl TlsOptions {
     /// Converts `TlsOptions` into a rustls::ClientConfig.
-    pub fn into_rustls_config(self) -> Result<rustls::ClientConfig> {
+    pub(crate) fn into_rustls_config(self) -> Result<rustls::ClientConfig> {
         let mut config = rustls::ClientConfig::new();
 
         if let Some(true) = self.allow_invalid_certificates {
--- a/src/client/options/resolver_config.rs
+++ b/src/client/options/resolver_config.rs
@@ -1,0 +1,43 @@
+use trust_dns_resolver::config::ResolverConfig as TrustDnsResolverConfig;
+
+/// Configuration for the upstream nameservers to use for resolution.
+///
+/// This is a thin wrapper around a `trust_dns_resolver::config::ResolverConfig` provided to ensure
+/// API stability.
+#[derive(Clone, Debug, PartialEq)]
+pub struct ResolverConfig {
+    pub(crate) inner: TrustDnsResolverConfig,
+}
+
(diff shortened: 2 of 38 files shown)


src/sync/test.rs
```diff
@@ -33,22 +33,10 @@ where
     T: Serialize + DeserializeOwned + Unpin + Debug + Send + Sync,
 {
     let coll = client.database(db_name).collection(coll_name);
-    drop_collection(&coll);
+    coll.drop(None).unwrap();
     coll
 }
 
-pub fn drop_collection<T>(coll: &Collection<T>)
```
patrickfreed, line 40: This is no longer necessary since our `drop` helper ignores ns not found errors already.

src/client/options/mod.rs
```diff
@@ -149,22 +149,28 @@ impl StreamAddress {
         let hostname = match parts.next() {
             Some(part) => part,
             None => {
-                return Err(ErrorKind::InvalidHostname {
```
patrickfreed, line 152: It didn't seem super helpful to have a separate error case for this function only, so I lumped it in with the already existing `ArgumentError` case.

src/client/options/mod.rs
```diff
@@ -592,9 +598,8 @@ impl TlsOptions {
         if let Some(path) = self.ca_file_path {
             store
                 .add_pem_file(&mut BufReader::new(File::open(&path)?))
-                .map_err(|_| ErrorKind::ParseError {
```
patrickfreed, line 595: These `ParseError`'s only ever occurred when parsing TLS certificates, so I updated them to use the `TlsConfigError` variant.

src/error.rs
```diff
@@ -240,11 +260,7 @@ impl std::ops::Deref for Error {
 #[derive(Clone, Debug, Error)]
 #[non_exhaustive]
 pub enum ErrorKind {
-    /// Wrapper around [`std::net::AddrParseError`](https://doc.rust-lang.org/std/net/struct.AddrParseError.html).
-    #[error("{0}")]
-    AddrParse(#[from] std::net::AddrParseError),
```
patrickfreed, line 245: This was never actually emitted from the driver.

src/error.rs
```diff
@@ -271,45 +287,19 @@ pub enum ErrorKind {
...
+    #[non_exhaustive]
+    DnsResolveError { message: String },
 
     #[error("Internal error: {message}")]
     #[non_exhaustive]
     InternalError { message: String },
 
-    /// Wrapper around `webpki::InvalidDNSNameError`.
-    #[error("{0}")]
-    InvalidDnsName(#[from] webpki::InvalidDNSNameError),
```
patrickfreed, line 286: This is lumped into `DnsResolveError`

src/error.rs
```diff
@@ -271,45 +287,19 @@ pub enum ErrorKind {
...
-    #[error("Unable to parse hostname: {hostname}")]
-    #[non_exhaustive]
-    InvalidHostname { hostname: String },
-
     /// Wrapper around [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html).
     #[error("{0}")]
     Io(Arc<std::io::Error>),
 
-    #[error("No DNS results for domain {0}")]
-    NoDnsResults(StreamAddress),
```
patrickfreed, line 298: Also lumped into `DnsResolveError`.

src/error.rs
```diff
@@ -271,45 +287,19 @@ pub enum ErrorKind {
...
-
-    /// A database operation failed to send or receive a reply.
-    #[error("A database operation failed to send or receive a reply: {message}")]
-    #[non_exhaustive]
-    OperationError { message: String },
```
patrickfreed, line 303: All uses of this were rewritten as either `ResponseError` or `InternalError`, so this wasn't needed anymore.

src/error.rs
```diff
@@ -326,22 +316,13 @@ pub enum ErrorKind {
     #[non_exhaustive]
     ServerSelectionError { message: String },
 
-    /// An error occurred during SRV record lookup.
-    #[error("An error occurred during SRV record lookup: {message}")]
-    #[non_exhaustive]
-    SrvLookupError { message: String },
```
patrickfreed, line 332: Both this and `TxtLookupError` were rolled into `DnsResolveError`.

src/runtime/resolver.rs
```diff
@@ -41,8 +42,14 @@ impl AsyncResolver {
         Ok(lookup)
     }
 
-    pub async fn txt_lookup<N: IntoName>(&self, query: N) -> Result<TxtLookup> {
-        let lookup = self.resolver.txt_lookup(query).await?;
-        Ok(lookup)
+    pub async fn txt_lookup<N: IntoName>(&self, query: N) -> Result<Option<TxtLookup>> {
```
patrickfreed, line 45: The only caller of this ignored the NoRecordsFound error, so I updated this to include that instead and just return an `Option`.

src/client/options/resolver_config.rs
```diff
@@ -0,0 +1,43 @@
...
+/// Configuration for the upstream nameservers to use for resolution.
+///
+/// This is a thin wrapper around a `trust_dns_resolver::config::ResolverConfig` provided to ensure
+/// API stability.
+#[derive(Clone, Debug, PartialEq)]
+pub struct ResolverConfig {
+    pub(crate) inner: TrustDnsResolverConfig,
+}
+
+impl ResolverConfig {
```
patrickfreed, line 12: I opted to provide just the predefined configs rather than exposing further functionality since that is likely the only things our users need and it could get hard to maintain the additional API surface. (see [`trust-dns-resolver` docs](https://docs.rs/trust-dns-resolver/0.20.2/trust_dns_resolver/config/struct.ResolverConfig.html) for the full type).

src/client/options/mod.rs
```diff
@@ -588,7 +588,7 @@ impl ServerCertVerifier for NoCertVerifier {
 
 impl TlsOptions {
     /// Converts `TlsOptions` into a rustls::ClientConfig.
-    pub fn into_rustls_config(self) -> Result<rustls::ClientConfig> {
+    pub(crate) fn into_rustls_config(self) -> Result<rustls::ClientConfig> {
```
patrickfreed, line 591: I think this is mostly useful for us rather than users, so it seems safe to just internalize.

src/error.rs
```diff
@@ -227,26 +231,6 @@ impl From<std::io::ErrorKind> for ErrorKind {
     }
 }
 
-impl From<trust_dns_resolver::error::ResolveError> for ErrorKind {
```
patrickfreed: Just pushed a commit to remove these which I missed in the first pass. Thankfully the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/necessities.html#public-dependencies-of-a-stable-crate-are-stable-c-stable) called this case out specifically since it was so easy to miss.

kmahar:
just some high-level comments on a couple categories of changes, I'll leave it to @isabelatkinson to ensure the code changes are reasonable:

* Wrapped Error::kind in a Box to reduce the size of Error
    * If this is really an issue, handling it in the same way as trust-dns seems like a reasonable choice. But I guess I still wonder if it actually is? Your writeup sounded ambivalent ("it could potentially be an issue worth tackling") so I guess I'm just wondering if there's anything else we want to be thinking about here. Are there any possible downsides to this?
* Rewrote error cases that simply wrapped errors from other crates to instead just use their messages
    * seems ok if there isn't any other interesting information contained in error properties besides the message; if there is, should we consider stringifying and adding to messages?

patrickfreed:
>If this is indeed an issue, handling it in the same way as trust-dns seems like a reasonable choice. But I guess I still wonder if it actually is? Your writeup sounded ambivalent ("it could potentially be an issue worth tackling") so I guess I'm just wondering if there's anything else we want to be thinking about here. Are there any possible downsides to this?

I phrased it that way since I couldn't be certain / prove via profiling that it is in fact an issue, especially given the prevailing wisdom that in Rust you don't need to worry too much about the sizes of things being returned due to optimizations in the compiler. However, given that a user reported it so soon after we made `ErrorKind` larger and that trust-dns did see some major improvements by wrapping their large `ErrorKind` in a `Box` suggests that the prevailing wisdom could be wrong in this case, and that it is probably safer to reduce the size in the absence of any proof that it's not an issue. As another interesting note, [clippy has a lint](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/large_enum_variant.rs#L13) for enums with variants that are too different in size which actually [had us `Box` a large internal variant](https://github.com/mongodb/mongo-rust-driver/blob/master/src/cursor/mod.rs#L173) for performance reasons. This is relevant since our `Result` type is just an enum where one case is a generic type `T` (possibly small, e.g. in the count helpers) and the other case is our `Error`, which prior to this change is pretty big.

 >seems fine to me if there isn't any other interesting information contained in error properties besides the message; if there is, should we consider stringifying and adding to messages?

I think we can defer to the underlying error's `Display` implementations for this, which (if implemented properly) should stringify all the relevant information. (note: `to_string` calls through to the error's `Display` implementation).

isabelatkinson (commented):
sorry for the delay! looks good, just one small comment.

src/error.rs
```diff
@@ -240,11 +244,7 @@ impl std::ops::Deref for Error {
 #[derive(Clone, Debug, Error)]
 #[non_exhaustive]
 pub enum ErrorKind {
-    /// Wrapper around [`std::net::AddrParseError`](https://doc.rust-lang.org/std/net/struct.AddrParseError.html).
-    #[error("{0}")]
-    AddrParse(#[from] std::net::AddrParseError),
-
-    /// An invalid argument was provided to a database operation.
+    /// An invalid argument was provided.
```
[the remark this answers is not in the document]
patrickfreed: good catch, fixed.

kmahar:
> I phrased it that way since I couldn't be certain / prove via profiling that it is in fact an issue, especially given the prevailing wisdom that in Rust you don't need to worry too much about the sizes of things being returned due to optimizations in the compiler. However, given that a user reported it so soon after we made `ErrorKind` larger and that trust-dns did see some major improvements by wrapping their large `ErrorKind` in a `Box` suggests that the prevailing wisdom could be wrong in this case, and that it is probably safer to reduce the size in the absence of any proof that it's not an issue. As another interesting note, [clippy has a lint](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/large_enum_variant.rs#L13) for enums with variants that are too different in size which actually [had us `Box` a large internal variant](https://github.com/mongodb/mongo-rust-driver/blob/master/src/cursor/mod.rs#L173) for performance reasons. This is relevant since our `Result` type is just an enum where one case is a generic type `T` (possibly small, e.g. in the count helpers) and the other case is our `Error`, which prior to this change is pretty big.

That makes sense. And given that we expect most of the time to return success values and not errors, it does make sense to try to keep the enum closer to the size of the value in the success/common case.  I was mainly wondering whether there are any negative tradeoffs we're making by taking this approach, but it sounds like not really? 

> I think we can defer to the underlying error's `Display` implementations for this, which (if implemented properly) should stringify all the relevant information. (note: `to_string` calls through to the error's `Display` implementation).

👍

patrickfreed:
The main tradeoff is that it's slightly less ergonomic to work with since it requires a dereference (`*`) to get an owned value or an `.as_ref()` invocation to get a reference, though these aren't particularly huge drawbacks IMO. It's still a big improvement from `Arc`, which wouldn't let you get an owned value at all. One benefit from `Arc` that we lose in either the `Box` or no `Box` case is that cloning the errors is now somewhat expensive though (since it actually makes a copy rather than just making a new reference), though I don't think that's really a big deal either.

kmahar (approved):

isabelatkinson (approved):

PR #331, opened 2021-05-04
RUST-754 Accept impl Borrow<T> in insert and replace methods

patrickfreed:
RUST-754

This PR updates the various insert and replace methods to accept `impl Borrow<T>`, which means that both owned values and values that can be converted into references can be accepted in these methods. 

In a related change, the trait requirements on `T` in the sync driver were relaxed now that tokio and async-std's blocking functions no longer require it.

### DIFF
--- a/src/test/coll.rs
+++ b/src/test/coll.rs
@@ -541,7 +541,7 @@ async fn empty_insert() {
         .database(function_name!())
         .collection::<Document>(function_name!());
     match coll
-        .insert_many(Vec::new(), None)
+        .insert_many(Vec::<Document>::new(), None)
         .await
         .expect_err("should get error")
         .kind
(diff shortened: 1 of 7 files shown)


src/test/coll.rs
```diff
@@ -541,7 +541,7 @@ async fn empty_insert() {
         .database(function_name!())
         .collection::<Document>(function_name!());
     match coll
-        .insert_many(Vec::new(), None)
+        .insert_many(Vec::<Document>::new(), None)
```
patrickfreed, line 544: In certain cases, this will make type inference more difficult and require the usage of type arguments, such as in this case. I think this is worth it over taking in just `&T` though, which rules out owned values and other smart-pointer types from being passed in (e.g. `Arc`, `Box`). I think in most cases users won't be relying on the concrete type in these methods and the annotations won't be necessary.

isabelatkinson (approved):

PR #332, opened 2021-05-04
RUST-90 Implement Transactions API

isabelatkinson:
Spec: https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst


src/client/session/mod.rs
```diff
@@ -128,6 +360,12 @@ impl Drop for ClientSession {
             txn_number: self.server_session.txn_number,
         };
 
+        if self.transaction.state == TransactionState::InProgress {
+            RUNTIME.block_on(async {
```
isabelatkinson, line 364: This was a bit tricky to implement, as the underlying functions called in `RUNTIME::execute` require the data within the `async move` block to have a `'static` lifetime. This isn't possible for `self` in this context because of the constraints of the signature of `drop`, so it seems like the only solution is to use `RUNTIME::block_on`. This doesn't seem great to me (since this blocks on the operation rather than executing it in the background) but I'm not sure there's a way to get around it -- do you have any better ideas?

isabelatkinson (commented):
Moving this into review as I continue to investigate the stack overflow failures on Windows async-std.

src/operation/mod.rs
```diff
@@ -103,6 +108,13 @@ pub(crate) trait Operation {
     fn retryability(&self) -> Retryability {
         Retryability::None
     }
+
+    // Updates this operation's write concern.
+    fn update_write_concern(&mut self) {}
```
patrickfreed, line 113: I think this is probably a bit too niche to include on the `Operation` trait, since I think it's unlikely other operations would need to use it. What about if we introduced a `fn build_retry` instead that `CommitTransaction` could implement which uses the appropriate write concern, and as a default implementation it just calls through to the regular `build`? This is a little more general and potentially could be useful in the future.

patrickfreed (commented):
Looks really good so far! This is only a partial review that I figured I should release to get the PR going. Will take a look at the rest on Monday.

src/client/session/mod.rs
```diff
@@ -116,6 +186,168 @@ impl ClientSession {
...
+                    .await
+            }
+            TransactionState::Committed {
+                data_committed: true,
+            } => {
+                let mut commit_transaction =
+                    CommitTransaction::new(self.transaction.options.clone());
+                commit_transaction.update_write_concern();
+                self.client
+                    .clone()
```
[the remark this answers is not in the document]
isabelatkinson: Yeah, `self.client` (without the clone) borrows `self`, so we can't re-borrow `self` as the second argument to `execute_operation` if the client isn't cloned.

src/client/executor.rs
```diff
@@ -373,25 +447,59 @@ impl Client {
...
+/// A TransientTransactionError label should be added if a transaction is in progress and the error
+/// is a network or server selection error.
+///
+/// On a pre-4.4 connection, a RetryableWriteError label should be added to any write-retryable
+/// error. On a 4.4+ connection, a label should only be added to network errors. Regardless of
+/// server version, a label should only be added if the `retry_writes` client option is not set to
+/// `false`, the operation during which the error occured is write-retryable, and a
+/// TransientTransactionError label has not already been added.
+async fn get_error_with_labels(
+    err: Error,
```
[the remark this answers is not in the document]
isabelatkinson: Refactored as you suggested. I also updated `with_label` to be `add_label` as there was some unnecessary cloning going on in there (not sure why I implemented it like that in the first place).

src/operation/mod.rs
isabelatkinson: Hm, that would require some refactoring since `build` is called further down in `execute_operation_on_connection` (which doesn't have any indication of whether we're on the first try or retry). Maybe we just rename this `update_for_retry` or something along those lines? I agree it's a little niche as-is.

isabelatkinson (commented):
Replies to the first round of review. I also realized I forgot to add the sync API; I'll include that in a later commit.

src/coll/mod.rs
```diff
@@ -185,11 +187,17 @@ where
...
+
         let mut options = options.into();
         resolve_options!(self, options, [write_concern]);
 
-        let drop = DropCollection::new(self.namespace(), options);
-        self.client().execute_operation(drop, session.into()).await
+        let drop = DropCollection::new(
+            self.namespace(),
+            options,
+            transaction_read_concern(&session),
```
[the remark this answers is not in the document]
patrickfreed: In order to avoid having to do this for each new write operation that we introduce, we can include this in `execute_operation`. To do so, we'd need a helper on `Command` that we could check in the event that the transaction state is starting, and if the command contains no read concern but the transaction does, set it there.

src/db/mod.rs
```diff
@@ -353,15 +355,61 @@ impl Database {
...
+    /// default read concern or write concern. If specific read concern or write concern is
+    /// desired, it must be specified manually.
     pub async fn run_command_with_session(
         &self,
-        command: Document,
+        mut command: Document,
         selection_criteria: impl Into<Option<SelectionCriteria>>,
         session: &mut ClientSession,
     ) -> Result<Document> {
+        let mut selection_criteria = selection_criteria.into();
```
patrickfreed, line 371: I think this logic is more suited to `RunCommand::build`. You'll need to pass the transaction state into `RunCommand::new` but I think that's okay.

src/sdam/description/topology/mod.rs
```diff
@@ -701,6 +740,31 @@ impl SessionSupportStatus {
...
+    /// Transactions are not supported by this topology.
+    Unsupported,
+
+    /// Transactions are supported by this topology. A topology supports transactions if it
+    /// supports sessions and its maxWireVersion >= 7, or its maxWireVersion >= 8 and the topology
+    /// type is sharded.
+    ///
+    /// Note that meeting these conditions does not guarantee that a deployment
+    /// supports transactions; any other missing qualification will be reported by the server.
+    Supported,
```
patrickfreed, line 759: Can we update this docstring to indicate we don't support sharded transactions yet + TODO?

src/client/executor.rs
```diff
@@ -363,36 +443,112 @@ impl Client {
...
+            .unwrap_or(false)
+        {
+            match op.retryability() {
+                Retryability::Read if self.inner.options.retry_reads != Some(false) => {
+                    return Ok(Retryability::Read);
+                }
+                Retryability::Write if conn.stream_description()?.supports_retryable_writes() => {
+                    // commitTransaction and abortTransaction should be retried regardless of the
+                    // value for retry_writes set on the Client
+                    if op.name() == "commitTransaction"
```
patrickfreed, line 485: We can use `CommitTransaction::NAME` and `AbortTransaction::NAME` here (ditto for above).

src/client/executor.rs
```diff
@@ -363,36 +443,112 @@ impl Client {
...
+                            self.add_label(RETRYABLE_WRITE_ERROR);
+                        }
+                    }
+                }
+                if self.should_add_unknown_transaction_commit_result_label() {
+                    self.add_label(UNKNOWN_TRANSACTION_COMMIT_RESULT);
+                }
+            }
+            TransactionState::Aborted => {
+                self.add_label(RETRYABLE_WRITE_ERROR);
```
patrickfreed, line 538: For abort, we don't need to check if the client supports retryability, but we do need to ensure the error does.
Per the spec:
>abortTransaction is a retryable write command. Drivers MUST retry once
after abortTransaction fails with a retryable error according to the
`Retryable Writes Specification`_., regardless of whether retryWrites is set
on the MongoClient or not.

src/client/executor.rs
```diff
@@ -76,38 +96,69 @@ impl Client {
...
+        let mut conn = match server.pool.check_out().await {
+            Ok(conn) => conn,
+            Err(mut err) => {
+                err.add_labels(None, &session, None)?;
+                return Err(err);
+            }
+        };
 
-        let mut conn = server.pool.check_out().await?;
+        // If the current transaction has been committed/aborted and it is not being
```
patrickfreed, line 118: This needs to be moved above server selection and connection check out to ensure the state gets updated properly and that the appropriate labels are attached.

src/client/executor.rs
```diff
@@ -363,36 +443,112 @@ impl Client {
...
+            &session.transaction.state
+        });
+        match transaction_state {
+            TransactionState::Starting | TransactionState::InProgress => {
+                if self.is_network_error() || self.is_server_selection_error() {
+                    self.add_label(TRANSIENT_TRANSACTION_ERROR);
+                }
+            }
+            TransactionState::Committed { .. } => {
+                if let Some(conn) = conn {
```
patrickfreed, line 526: given that this is computed in a few branches, could we bring it out to a variable that's computed outside the `match`? It's pretty cheap to compute so it shouldn't be a perf issue.

src/test/spec/v2_runner/mod.rs
```diff
@@ -140,6 +143,9 @@ pub async fn run_v2_test(test_file: TestFile) {
         let mut session1 = Some(client.start_session(options).await.unwrap());
         let session1_lsid = session1.as_ref().unwrap().id().clone();
 
+        // initialize with abortTransaction to account for implicit aborts in command monitoring
+        let mut command_names = vec!["abortTransaction".into()];
```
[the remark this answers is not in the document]
isabelatkinson: It's because of [this test](https://github.com/mongodb/specifications/blob/33916af7c70397732c203b4c3f55c7bbf8e2aebc/source/transactions/tests/legacy/create-collection.json#L16) (and other similar tests); the `assertCollectionExists` and `assertCollectionNotExists` both use `list_collection_names` which emit command events that aren't included in the expected events. We can't filter these out indiscriminately because other tests using the v2 runner (e.g. retryable reads) do look for these in expected events.

src/test/spec/v2_runner/test_file.rs
```diff
@@ -79,6 +83,15 @@ pub struct Test {
     pub outcome: Option<Outcome>,
 }
 
+fn deserialize_uri_options_to_uri_string_option<'de, D>(
+    deserializer: D,
+) -> std::result::Result<Option<String>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    Ok(Some(deserialize_uri_options_to_uri_string(deserializer)?))
```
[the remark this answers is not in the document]
isabelatkinson: Not every test includes the `clientOptions` field. For those that don't, `client_uri` will be given a `None` value because the field is tagged with `#[serde(default)]`; for those that do, they'll go through this function. It needs to return wrapped in an option to be consistent with the type of `client_uri`.

src/lib.rs
```diff
@@ -134,22 +133,20 @@ define_if_single_runtime_enabled! {
...
     #[cfg(feature = "sync")]
     pub(crate) use crate::{
-        client::Client,
+        client::{Client, session::ClientSession},
         coll::Collection,
         cursor::{Cursor, session::{SessionCursor, SessionCursorHandle}},
         db::Database,
     };
 
-    pub use client::session::ClientSession;
```
isabelatkinson, line 151: The sync API originally used the same `ClientSession` as async because `ClientSession` didn't have any async components; however, the transactions methods are async so I needed to introduce a new sync `ClientSession`. This is a breaking change but luckily this shouldn't be a problem given the 2.0 (I doubt this will affect many users either) -- we should still call it out in the release notes when this goes out, though.

src/db/mod.rs
patrickfreed: We also added the logic for adding the transaction's read concern to the first command in a transaction to `execute_operation` (see `executor.rs` line 277). This was done for commands that don't accept a read concern via their options, but it also applies here. Now that I think about it, it also applies to the commands which _do_ take a read concern, meaning we can also remove the logic for adding it to the options structs in the `resolve_read_concern_with_session!` macro too. Then the read concern and write concern macros will look pretty much the same. 

Getting back to `run_command`, the validation that the command document doesn't contain a read concern does need to remain here though, just like we still need to verify the options structs don't contain read concerns / write concerns.

src/test/spec/v2_runner/mod.rs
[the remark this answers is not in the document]
patrickfreed: Ah I see. So I don't think the v2 runner mentions this explicitly, but the `testRunner` operations are supposed to use the "internal client" (or at least, _an_ "internal client" that's separate from the client used as part of the test execution). This is the same client used to create the collections at the beginning and to seed the initial data. The unified test format makes this requirement explicit: 

>The assertCollectionExists operation instructs the test runner to assert that the given collection exists in the database. The test runner MUST use the internal MongoClient for this operation.

Once this is done, we can be sure that the test runner assertions aren't interfering with any of the expected behavior or event assertions included in the actual test itself.

src/db/mod.rs
isabelatkinson: Oh gotcha, yeah that was just for the write commands but seems reasonable to move all of the logic there instead.

patrickfreed (commented):
Looks really good! I only have a few docs suggestions but otherwise LGTM!

patrickfreed:
Oh and it looks like there are some merge conflicts that need to be fixed.

isabelatkinson:
Updated the documentation as suggested. Once this is approved I'll squash all my commits and rebase onto master to fix the merge conflicts.

patrickfreed (approved):
LGTM! Fantastic job with this.

One note: it looks like a doc test needs to be fixed but otherwise seems good.

PR #333, opened 2021-05-04
RUST-764 Use unsigned integers for values that should always be non-negative

patrickfreed:
RUST-764

This updates a number of results and options in the driver that current use signed integers where unsigned integers could be used instead, since the values stored there will always be positive.

### DIFF
--- a/src/coll/options.rs
+++ b/src/coll/options.rs
@@ -10,7 +10,8 @@ use crate::{
         deserialize_duration_from_u64_millis,
         serialize_batch_size,
         serialize_duration_as_int_millis,
-        serialize_u32_as_i32,
+        serialize_u32_option_as_i32,
+        serialize_u64_option_as_i64,
     },
     concern::{ReadConcern, WriteConcern},
     options::Collation,
@@ -498,7 +499,8 @@ pub struct CountOptions {
     pub hint: Option<Hint>,
 
     /// The maximum number of documents to count.
-    pub limit: Option<i64>,
+    #[serde(serialize_with = "serialize_u64_option_as_i64")]
+    pub limit: Option<u64>,
 
     /// The maximum amount of time to allow the query to run.
     ///
@@ -508,7 +510,8 @@ pub struct CountOptions {
     pub max_time: Option<Duration>,
 
     /// The number of documents to skip before counting.
-    pub skip: Option<i64>,
+    #[serde(serialize_with = "serialize_u64_option_as_i64")]
+    pub skip: Option<u64>,
 
     /// The collation to use for the operation.
     ///
@@ -618,7 +621,7 @@ pub struct FindOptions {
     /// only the number of documents kept in memory at a given time (and by extension, the
     /// number of round trips needed to return the entire set of documents returned by the
     /// query.
-    #[serde(serialize_with = "serialize_u32_as_i32")]
+    #[serde(serialize_with = "serialize_u32_option_as_i32")]
     pub batch_size: Option<u32>,
 
     /// Tags the query with an arbitrary string to help trace the operation through the database
@@ -650,7 +653,8 @@ pub struct FindOptions {
     ///
     /// Note: this option is deprecated starting in MongoDB version 4.0 and removed in MongoDB 4.2.
     /// Use the maxTimeMS option instead.
-    pub max_scan: Option<i64>,
+    #[serde(serialize_with = "serialize_u64_option_as_i64")]
+    pub max_scan: Option<u64>,
 
     /// The maximum amount of time to allow the query to run.
     ///
@@ -689,7 +693,8 @@ pub struct FindOptions {
     pub show_record_id: Option<bool>,
 
     /// The number of documents to skip before counting.
-    pub skip: Option<i64>,
+    #[serde(serialize_with = "serialize_u64_option_as_i64")]
+    pub skip: Option<u64>,
 
     /// The order of the documents for the purposes of the operation.
     pub sort: Option<Document>,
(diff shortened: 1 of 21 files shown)


src/coll/options.rs
```diff
@@ -519,7 +520,6 @@ pub struct CountOptions {
     /// The criteria used to select a server for this operation.
     ///
     /// If none specified, the default set on the collection will be used.
-    #[serde(skip_serializing)]
```
patrickfreed, line 522: `CountOptions` isn't actually `Serialize`, so this wasn't needed.

src/coll/options.rs
```diff
@@ -618,7 +618,7 @@ pub struct FindOptions {
     /// only the number of documents kept in memory at a given time (and by extension, the
     /// number of round trips needed to return the entire set of documents returned by the
     /// query.
-    #[serde(serialize_with = "serialize_u32_as_i32")]
+    #[serde(serialize_with = "serialize_u32_option_as_i32")]
     pub batch_size: Option<u32>,
 
     /// Tags the query with an arbitrary string to help trace the operation through the database
```
patrickfreed, line 624: negative limit on `FindOptions` has semantic meaning, so it is preserved here.

src/coll/options.rs
```diff
@@ -806,7 +809,8 @@ pub struct FindOneOptions {
     pub show_record_id: Option<bool>,
 
     /// The number of documents to skip before counting.
-    pub skip: Option<i64>,
+    #[serde(serialize_with = "crate::bson_util::serialize_u64_option_as_i64")]
```
kmahar, line 812: little nit/question: is there a reason the helper function is fully qualified here with the `crate::bson_util` prefix but that's omitted in usages above? (is it doing anything different?)

kmahar (commented):
this seems totally fine to me if the unsigned int types are idiomatic to use here in Rust (unlike Swift where we just adopted the "`Int` for everything" policy)

just quickly scanning through the rest of the code, I do see a couple more cases that could be worth updating:

- `w` values for a `WriteConcern`... I think w has to be >= 0?
- zlib compression level (looks to be internal at the moment, but maybe worth updating so it's consistent if we do ever add zlib support?)
- collation "strength" value 
 
on the other hand we don't control the range of valid values for the latter two since they're defined by zlib and some external spec so maybe those are riskier to restrict

(relatedly, looking at the `Acknowledgement` type I noticed the docstring seems to imply we support unacknowledged writes, but iirc we don't actually, right?)

LGTM otherwise

patrickfreed:
>w values for a WriteConcern... I think w has to be >= 0?

Good catch, updated.

>zlib compression level (looks to be internal at the moment, but maybe worth updating so it's consistent if we do ever add zlib support?)

It looks like -1 is actually a valid value for the level (search for `Z_DEFAULT_COMPRESSION` on https://www.zlib.net/manual.html)

>collation "strength" value

Also a good catch, though I think for this one (and other collation options) we actually should use an enum with descriptive types, as per the [C-CUSTOM-TYPE](https://rust-lang.github.io/api-guidelines/type-safety.html#c-custom-type) Rust API guideline. The [Java driver](http://mongodb.github.io/mongo-java-driver/3.6/javadoc/com/mongodb/client/model/Collation.html) also takes this approach. If you agree, I'll file a separate ticket to include that change.

src/coll/options.rs
patrickfreed: No reason, probably just because I didn't have autocomplete here. Removed.

kmahar:
> It looks like -1 is actually a valid value for the level (search for `Z_DEFAULT_COMPRESSION` on https://www.zlib.net/manual.html)

ah good catch!



> Also a good catch, though I think for this one (and other collation options) we actually should use an enum with descriptive types, as per the [C-CUSTOM-TYPE](https://rust-lang.github.io/api-guidelines/type-safety.html#c-custom-type) Rust API guideline. The [Java driver](http://mongodb.github.io/mongo-java-driver/3.6/javadoc/com/mongodb/client/model/Collation.html) also takes this approach. If you agree, I'll file a separate ticket to include that change.

I like that idea, sgtm 👍

kmahar (approved):

src/concern/mod.rs
```diff
@@ -244,18 +246,9 @@ impl WriteConcern {
...
-        if let Some(Acknowledgment::Nodes(i)) = self.w {
-            if i < 0 {
-                return Err(ErrorKind::ArgumentError {
-                    message: "write concern `w` field cannot be negative integer".to_string(),
-                }
-                .into());
-            }
-        }
-
+    pub(crate) fn validate(&self) -> Result<()> {
```
isabelatkinson, line 251: Why was this removed from the public API?

isabelatkinson (commented):
looks good, just one question

src/concern/mod.rs
patrickfreed: Oh good catch, I meant to call this out in a comment but forgot to. This method seemed really niche and like something that was more useful for us internally than for users, especially since it requires an unacknowledged write concern which we don't currently support.

The method originally stems from this requirement in the [spec](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#inconsistent-writeconcern):
>Drivers MUST raise an error when an inconsistent WriteConcern is specified.

From my reading, this doesn't require that users need to be able to eagerly determine if a WC is valid or even that an error needs to be thrown when an inconsistent WC is created, just when one is "specified". So as long as we throw an error operation execution, we should still be abiding by the spec in my eyes. I doubt users were using this method anyways. Also, we throw errors on all `w: 0` write concerns, so its a bit of a moot point anyways.

I filed RUST-784 to cover this change so that we remember to include it in the release notes as a breaking change.

isabelatkinson: Gotcha, yeah I agree with your reading of the spec. Even if we were to allow unacknowledged writes I would guess most users would rely upon us to validate their write concerns rather than doing it themselves, so this seems fine to me to remove.

PR #337, opened 2021-05-07
RUST-765 Conform API to the Rust API guidelines

patrickfreed:
RUST-765 / RUST-791 / RUST-788 / RUST-786

This PR makes numerous small changes to the API to conform to the [Rust API Guidelines Checklist](https://rust-lang.github.io/api-guidelines/checklist.html). Each commit is tagged with the checklist item that its addressing. Most of the commits are just associated with the parent RUST-765 ticket, but for some of the more meaningful changes I filed separate tickets for the sake of the release notes.


Cargo.toml
```diff
@@ -37,7 +37,10 @@ bitflags = "1.1.0"
 bson = { git = "https://github.com/mongodb/bson-rust" }
 chrono = "0.4.7"
 derivative = "2.1.1"
-futures = "0.3.5"
+futures-core = "0.3.14"
+futures-io = "0.3.14"
+futures-util = { version = "0.3.14", features = ["io"] }
+futures-executor = "0.3.14"
```
patrickfreed, line 43: The `futures` crate contains a number of async-related APIs that are widely used in the community, the most important one (for us) being the [`Stream`](https://docs.rs/futures-core/0.3.14/futures_core/stream/trait.Stream.html) trait, which our various cursor types conform to (it's basically the async iterator trait). Because we publicly conform to it, we'll need to do so for each major version of `futures` to ensure full compatibility. The `futures` crate is really just a collection of rexports from "sub" crates though (many of which are listed here now), each of which has differing stability exceptions. The `futures-core` one, which contains `Stream`, is purposefully much more stable than the others, and it will likely go straight from `0.3` to `1.0` without any more breaking versions, whereas `futures-util` and `futures` itself will have potentially many major versions. For that reason, I updated this to depend directly on `futures-core` instead of `futures` to reduce the number of separate `Stream` implementations we'll need to provide for the lifetime of 2.0. This [page](https://smallcultfollowing.com/babysteps/blog/2019/12/09/async-interview-2-cramertj/#the-futures-crate-is-really-a-facade) gives a nice overview of the situation and why we need to do this.

src/concern/mod.rs
```diff
@@ -165,10 +165,8 @@ pub struct WriteConcern {
 pub enum Acknowledgment {
     /// Requires acknowledgement that the write has reached the specified number of nodes.
     ///
-    /// Note: specifying 0 here indicates that the write is unacknowledged. Doing so means
-    /// the driver will not receive a response indicating whether an operation succeeded or failed.
-    /// It also means that the operation cannot be associated with a session. It is reccommended to
-    /// avoid using unacknowledged write concerns.
+    /// Note: specifying 0 here indicates that the write concern is unacknowledged, which is
```
patrickfreed, line 168: I meant to include this change in the last PR (I even commented that I did but forgot to push the commit), so I included it here.

src/error.rs
```diff
@@ -193,6 +193,98 @@ impl Error {
     pub(crate) fn from_resolve_error(error: trust_dns_resolver::error::ResolveError) -> Self {
         ErrorKind::DnsResolveError { message: error.to_string() }.into()
     }
+
```
patrickfreed, line 196: These methods were moved from being `impl ErrorKind` to just being on `impl Error` since the `Deref` implementation is gone now. They're all internal anyways so there isn't any impact on the public API.

src/error.rs
```diff
@@ -257,11 +341,11 @@ pub enum ErrorKind {
 
     /// Wrapper around `bson::de::Error`.
     #[error("{0}")]
-    BsonDecode(crate::bson::de::Error),
+    BsonDeserializationError(crate::bson::de::Error),
```
patrickfreed, line 344: I updated the terminology here and appended the "Error" suffix to the end to match the other cases. I wonder if it would actually be better if we dropped all the "Error" suffixes though, since they're kind of redundant with `ErrorKind`. I'm not really sure which way would be better, but I just opted for this since it was the least amount of changes.
```rust
match *e.kind {
    ErrorKind::IoError(e) => { ... },
    ErrorKind::CommandError(e) => { ... },
    ErrorKind::ArgumentError { message } => { ... },
    // rest of cases...
}

// vs
match *e.kind {
    ErrorKind::Io(e) => { ... },
    ErrorKind::Command(e) => { ... },
    ErrorKind::Argument { message } => { ... },
    // rest of cases...
}
```

src/error.rs
```diff
@@ -320,99 +404,6 @@ pub enum ErrorKind {
     WriteError(WriteFailure),
```
patrickfreed, line 404: I left `ErrorKind::SessionsNotSupported` without an "error" prefix since it doesn't have any error data associated with it. Let me know if that seems weird or if it makes sense.

src/lib.rs
```diff
@@ -83,6 +83,8 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![cfg_attr(test, type_length_limit = "80000000")]
 
+#![doc(html_root_url = "https://docs.rs/mongodb/2.0.0-alpha.1")]
```
kmahar, line 86: seems like something we'd want to make sure we update programmatically with each release?

src/collation.rs
```diff
@@ -16,15 +20,15 @@ pub struct Collation {
 
     /// The level of comparison to perform. Corresponds to [ICU Comparison Levels](http://userguide.icu-project.org/collation/concepts#TOC-Comparison-Levels).
     #[builder(default, setter(strip_option))]
-    pub strength: Option<i32>,
+    pub strength: Option<CollationStrength>,
 
-    /// Whether to include case comparison when `strength` is level 1 or 2.
+    /// Whether to include a separate level for case differenes. See [ICU Collation: CaseLevel](http://userguide.icu-project.org/collation/concepts#TOC-CaseLevel) for more information.
```
kmahar, line 25:
```suggestion
    /// Whether to include a separate level for case differences. See [ICU Collation: CaseLevel](http://userguide.icu-project.org/collation/concepts#TOC-CaseLevel) for more information.
```

src/collation.rs
```diff
@@ -48,3 +52,235 @@ pub struct Collation {
...
+impl std::fmt::Display for CollationStrength {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Display::fmt(&u32::from(*self), f)
+    }
+}
+
+/// Setting that determines sort order of case differences during case tertiary level comparisons.
+/// For more info, see http://userguide.icu-project.org/collation/customization.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
```
kmahar, line 153: well TIL kebab case is a thing... does it actually do anything here though given these enum cases are all one word?

src/error.rs
kmahar: I would say feels a little more consistent to me to add on the "Error" at the end, because it seems like the enum instance itself is an error, whether or not the case has associated values (or whatever the rust term is for those)

src/client/mod.rs
```diff
@@ -98,8 +98,8 @@ impl Client {
     ///
     /// See the documentation on
     /// [`ClientOptions::parse`](options/struct.ClientOptions.html#method.parse) for more details.
-    pub async fn with_uri_str(uri: &str) -> Result<Self> {
-        let options = ClientOptions::parse_uri(uri, None).await?;
+    pub async fn with_uri_str(uri: impl AsRef<str>) -> Result<Self> {
```
isabelatkinson, line 101: I think we should also update the `collection`/`collection_with_options` and `database`/`database_with_options` methods to take `impl AsRef<str>` for the `name` params.

src/lib.rs
isabelatkinson: this would probably be worth adding to the release checklist doc

src/error.rs
isabelatkinson: There's actually a [relevant clippy lint](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/enum_variants.rs) here that complains when all enum variants have the same suffix; I ran into this once before when I was making the `Retryability` enum and all of the variants were suffixed with `Retryable` (i.e. `ReadRetryable`, `WriteRetryable`, `NonRetryable`). I think we're avoiding it for now since `SessionsNotSupported` currently doesn't have the `Error` suffix as you point out below, but I do think it would be nicer to get rid of the `Error` in the names in the spirit of this lint. I don't feel super strongly about this though so I'll leave it up to you; I agree it's best to minimize changes when we don't have a strong opinion on something.

src/client/mod.rs
patrickfreed: Thank you for bringing this up, I meant to open a thread discussing this in my original review. 

Those two methods are a little tricky because they have generic parameters, so if we include `impl AsRef` as an argument, the turbofish operator will no longer work. If we introduce it as a regular generic argument instead, then users will have to specify both of them if they use the turbofish. When weighing all of these, I figured the inconsistency of requiring `&str` here was the best option of the three. What do you think? cc @kmahar

src/error.rs
kmahar: given that, dropping `Error` SGTM

patrickfreed: Updated to drop suffix. I added the `Invalid` prefix to the `Argument`, `Response`, and `TlsConfig` variants to convey that something malformed was produced somewhere.

src/error.rs
patrickfreed: Resolving per discussion above.

src/client/mod.rs
kmahar: sgtm

PR #338, opened 2021-05-11
RUST-757 Remove `wait_queue_timeout` option

patrickfreed:
RUST-757

This PR removes the `wait_queue_timeout` option from `ClientOptions`, since it will be superseded by the general `timeoutMS` option from the Client Side Operations Timeout specification.

In the meantime, users can just use the timeout functionality from their respective runtimes as a workaround.

e.g. for tokio:
```rust
tokio::time::timeout(Duration::from_secs(5), coll.insert_one(doc! { "x": 1 }, None)).await?;
```
Note this may lead to some connection churn, but it seems to be an acceptable alternative for now.

### DIFF
--- a/src/cmap/connection_requester.rs
+++ b/src/cmap/connection_requester.rs
@@ -31,7 +31,6 @@ pub(super) struct ConnectionRequester {
 
 impl ConnectionRequester {
     /// Request a connection from the pool that owns the receiver end of this requester.
-    /// Returns None if it takes longer than wait_queue_timeout before the pool returns a result.
     pub(super) async fn request(&self) -> ConnectionRequestResult {
         let (sender, receiver) = oneshot::channel();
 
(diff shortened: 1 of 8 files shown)


src/cmap/connection_requester.rs
```diff
@@ -33,27 +32,16 @@ pub(super) struct ConnectionRequester {
 impl ConnectionRequester {
     /// Request a connection from the pool that owns the receiver end of this requester.
     /// Returns None if it takes longer than wait_queue_timeout before the pool returns a result.
```
isabelatkinson, line 34: Can you remove the second line from this comment?

isabelatkinson (commented):
looks good, just one doc change

src/cmap/connection_requester.rs
patrickfreed: Good catch, done.

isabelatkinson (approved):

PR #341, opened 2021-05-14
RUST-796 Provide easy way to reborrow session in between cursor iterations

patrickfreed:
RUST-796

This PR adds a `SessionCursor::next(&mut ClientSession)` method to allow for iteration of the cursor that allows for easy reborrowing of the session. As part of this, `SessionCursor::with_session` was renamed to `SessionCursor::stream` to more accurately reflect its intent now that `next` is around. Similar changes were made for the sync driver.

I also went ahead and updated the documentation for cursors in general to be a little more straightforward.

### DIFF
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -488,7 +488,7 @@ where
         let mut cursor = self
             .find_with_session(filter, Some(options), session)
             .await?;
-        let mut cursor = cursor.with_session(session);
+        let mut cursor = cursor.stream(session);
         cursor.next().await.transpose()
     }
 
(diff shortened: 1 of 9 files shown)


isabelatkinson (approved):

isabelatkinson:
Just noticed there's a `rustdoc` failure in CI but LGTM otherwise; don't need to do another round for that fix

patrickfreed:
Ah it looks like `SessionCursorIterator` (now `SessionCursorIter`) wasn't being exported publicly, updated it to do so now. On a related note, I updated the sync version of `stream()` / `SessionCursorStream` to match as `iter()` / `SessionCursorIter`.

kmahar (approved):

PR #342, opened 2021-05-19
RUST-768 Test that versioned API options are passed to transaction-continuing commands

isabelatkinson:
RUST-768

This PR introduces a test to ensure that versioned API options are passed to all commands within a transaction. The `getMore`-related changes will be implemented in future work.

### DIFF
--- a/src/sync/test.rs
+++ b/src/sync/test.rs
@@ -187,12 +187,10 @@ fn typed_collection() {
 #[test]
 #[function_name::named]
 fn transactions() {
-    let (should_skip, should_create_collection) = RUNTIME.block_on(async {
+    let should_skip = RUNTIME.block_on(async {
         let test_client = AsyncTestClient::new().await;
         // TODO RUST-122: Unskip this test on sharded clusters
-        let should_skip = !test_client.is_replica_set() || test_client.server_version_lt(4, 0);
-        let should_create_collection = test_client.server_version_lt(4, 4);
-        (should_skip, should_create_collection)
+        !test_client.is_replica_set() || test_client.server_version_lt(4, 0)
     });
     if should_skip {
         return;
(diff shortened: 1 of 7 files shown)


src/sync/test.rs
```diff
@@ -205,12 +203,10 @@ fn transactions() {
         .expect("session creation should succeed");
     let coll = init_db_and_typed_coll(&client, function_name!(), function_name!());
 
-    if should_create_collection {
-        client
-            .database(function_name!())
-            .create_collection(function_name!(), None)
-            .expect("create collection should succeed");
-    }
+    client
```
isabelatkinson, line 206: A few errors were showing up on 4.4 replica sets for this test with the following message: "Unable to read from a snapshot due to pending collection catalog changes; please retry the operation." It seems like this has caused some [drivers issues](https://jira.mongodb.org/browse/SERVER-41532) in the past related to inserts in transactions. Creating the collection before starting the transaction fixes the issue.

src/test/spec/versioned_api.rs
```diff
@@ -0,0 +1,67 @@
...
+    if TestClient::new().await.server_version_gte(5, 0) {
+        return;
+    }
+    run_spec_test(&["versioned-api"], run_unified_format_test).await;
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[function_name::named]
+async fn transaction_handling() {
```
patrickfreed, line 25: This is the prose version of a spec test right? If so, can we add a comment indicating that along with a TODO with the ticket to convert transactions to the unified runner (RUST-817, or maybe RUST-734)?

patrickfreed (commented):
looks good, just one suggestion for a TODO comment

src/test/spec/versioned_api.rs
isabelatkinson: oh good catch I meant to do this but must've forgotten, fixed

patrickfreed (approved):

PR #345, opened 2021-05-25
RUST-824 Add Snapshot variant to ReadConcernLevel

isabelatkinson:


patrickfreed (commented):
It looks like a bunch of the tests failed with an interrupted error, though that seems unrelated to these changes? I wonder if evergreen did a deploy while they were running or something. Maybe just restart the patch and see if it is green afterwards.

isabelatkinson:
Just restarted, the tests are passing locally so hopefully it's just an evg issue.

src/test/spec/read_write_concern/operation.rs
```diff
@@ -0,0 +1,12 @@
+use tokio::sync::RwLockReadGuard;
+
+use crate::test::{run_spec_test, LOCK};
+
+use super::super::run_v2_test;
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn run() {
+    let _lock: RwLockReadGuard<_> = LOCK.run_concurrently().await;
```
patrickfreed, line 10: Ah I think I know what the problem is. The v2 runner executes `killAllSessions` between test executions, and because this is running concurrently with other tests that are using implicit sessions, their operations get terminated and result in that error. So if we update this to be an exclusive lock, we should be good.

isabelatkinson: ahh okay good catch, fixed

patrickfreed (approved):
LGTM!

PR #348, opened 2021-05-27
Update to semver 1.0.0

dtolnay:


patrickfreed:
Thanks for the PR! It looks like our (non-standard) rustfmt invocation needs to be run though (see `.evergreen/check-rustfmt.sh`), so if a commit for that could be pushed up that'd be great. I filed RUST-831 to look into why it's that way and to see if we can switch to a more vanilla config in the future.

patrickfreed (approved):
LGTM, thanks for the contribution!

PR #349, opened 2021-05-27
RUST-813 Add documentation examples for transactions

isabelatkinson:


tests/transaction_examples.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+async fn update_employee_info(session: &mut ClientSession) -> Result<()> {
+    let transaction_options = TransactionOptions::builder()
+        .read_concern(ReadConcern::snapshot())
+        .write_concern(WriteConcern::builder().w(Acknowledgment::Majority).build())
+        .build();
+    session.start_transaction(transaction_options).await?;
+
+    execute_transaction_with_retry(execute_employee_info_transaction, session).await
```
isabelatkinson, line 80: Running into some lifetime issues here -- the compiler error indicates that the `&mut ClientSession` argument in the function pointer needs a lifetime, but adding a lifetime parameter causes a multiple mutable references issue in the loop on line 21. This issue doesn't appear in a similar loop in the docs examples in `src/client/session/mod.rs`, so it seems like it has something to do with the lifetime that I can't quite figure out.

src/client/session/mod.rs
```diff
@@ -37,6 +37,57 @@ lazy_static! {
...
+/// #     bson::doc,
+/// #     error::{Result, TRANSIENT_TRANSACTION_ERROR, UNKNOWN_TRANSACTION_COMMIT_RESULT},
+/// #     Client,
+/// #     ClientSession,
+/// #     Collection,
+/// # };
+/// # async fn do_stuff() -> Result<()> {
+/// # let client = Client::with_uri_str("mongodb://example.com").await?;
+/// # let coll = client.database("foo").collection("bar");
+/// # let mut session = client.start_session(None).await?;
```
patrickfreed, line 58: the line creating the session and the imports used in the visible part of the example can be revealed since they add to the example imo.

src/client/session/mod.rs
```diff
@@ -37,6 +37,57 @@ lazy_static! {
...
+/// session.start_transaction(None).await?;
```
patrickfreed, line 59: I think it would be good to demonstrate best practices here and specify a majority read / write concern in the transaction options.

src/client/session/mod.rs
```diff
@@ -37,6 +37,57 @@ lazy_static! {
...
+///         break;
+///     }
+/// }
+/// # Ok(())
+/// # }
+///
+/// async fn execute_transaction(coll: &Collection, session: &mut ClientSession) -> Result<()> {
+///     coll.insert_one_with_session(doc! { "x": 1 }, None, session).await?;
+///     coll.delete_one_with_session(doc! { "y": 2 }, None, session).await?;
+///     // An "UnknownTransactionCommitResult" label indicates that it is unknown whether the
```
patrickfreed, line 73: can we update `commit_transaction` to include this documentation?

src/client/session/mod.rs
```diff
@@ -37,6 +37,57 @@ lazy_static! {
...
+/// #     Client,
+/// #     ClientSession,
+/// #     Collection,
+/// # };
+/// # async fn do_stuff() -> Result<()> {
+/// # let client = Client::with_uri_str("mongodb://example.com").await?;
+/// # let coll = client.database("foo").collection("bar");
+/// # let mut session = client.start_session(None).await?;
+/// session.start_transaction(None).await?;
+/// // A "TransientTransactionError" label indicates that the entire transaction can be retried
```
patrickfreed, line 60: can we update `start_transaction` to include this info?

tests/transaction_examples.rs
isabelatkinson: Got it. That macro isn't too daunting but IMO isn't ideal for basic transactions starter code to provide for users -- I think it might be best to deviate a bit from the other examples and eliminate the closure (e.g. by calling `execute_employee_info_transaction`) directly. Thoughts?

patrickfreed: Yeah I think you're right. Maybe we ought to just include the retry part directly in `update_employee_info` then instead of invoking the macro.

patrickfreed (approved):
LGTM!

PR #350, opened 2021-05-27
RUST-830 Consolidate error label storage to `Error::labels`

patrickfreed:
RUST-830

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,9 +1,6 @@
 //! Contains the `Error` and `Result` types that `mongodb` uses.
 
-use std::{
-    fmt::{self, Debug},
-    sync::Arc,
-};
+use std::{collections::HashSet, fmt::{self, Debug}, sync::Arc};
 
 use serde::Deserialize;
 use thiserror::Error;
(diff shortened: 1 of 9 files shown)


src/test/spec/retryable_writes/mod.rs
```diff
@@ -63,7 +64,15 @@ async fn run_spec_tests() {
                     .expect(&test_case.description);
             }
 
-            if let Some(ref fail_point) = test_case.fail_point {
+            if let Some(ref mut fail_point) = test_case.fail_point {
+                // TODO: DRIVERS-1385 remove this logic for moving errorLabels to the top level.
```
patrickfreed, line 68: I ended up just changing the failpoints to be accurate. This had the benefit of not requiring any driver changes, and it's likely how the tests will be fixed in the future.

src/error.rs
```diff
@@ -159,19 +163,8 @@ impl Error {
...
-                WriteFailure::WriteError(_) => &self.labels,
-                WriteFailure::WriteConcernError(ref err) => &err.labels,
-            },
-            ErrorKind::BulkWrite(ref err) => match err.write_concern_error {
-                Some(ref err) => &err.labels,
-                None => &self.labels,
-            },
-            _ => &self.labels,
-        }
+    pub fn labels(&self) -> &HashSet<String> {
```
patrickfreed, line 166: I changed this to a `HashSet` since ordering doesn't matter, we don't want duplicates, and generally only care about whether a label is present or not. Arguably we could just not expose this at all given that we have `contains_label` already but I figured we already had it so might as well keep it.

PR #352, opened 2021-06-04
RUST-571 Fix race between SDAM discovery and SRV polling

patrickfreed:
RUST-571

The initial dns seedlist tests were pretty flaky, and I believe for (at least) these 3 reasons:
- the test lock was not acquired before running it
- we don't have SDAM monitoring yet, so our checking of the discovered hosts was racy
- there was a race between SRV polling and discovery of an Unknown topology
  - We were not waiting before doing the first SRV poll after client creation, so the SRV pol could update the topology before the monitors had a chance to mark it as a replica set, causing the polling and SDAM to get into a loop of updating the topology back and forth

This PR fixes these three issues, though sadly it appears that we [still have failures in this test](https://evergreen.mongodb.com/task/mongo_rust_driver_tests__async_runtime~tokio_auth_and_tls~auth_and_tls_os~macos_10.14_test_4.2_replica_set_patch_3569cc2279c8a357f0bc332df4b8f990acf61939_60a8394dd1fe07741eec0f7e_21_05_21_22_51_00), though much less frequently. As the macOS queue is heavily stalled right now (my patch is 1200th in the queue after over an hour), I'm just going to put this PR up now and monitor the failures going forward.

### DIFF
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -47,7 +47,7 @@ use crate::{
     options::ReadConcernLevel,
     sdam::MIN_HEARTBEAT_FREQUENCY,
     selection_criteria::{ReadPreference, SelectionCriteria, TagSet},
-    srv::SrvResolver,
+    srv::{OriginalSrvInfo, SrvResolver},
 };
 
 pub use resolver_config::ResolverConfig;
@@ -523,8 +523,10 @@ pub struct ClientOptions {
     #[builder(default, setter(skip))]
     pub(crate) zlib_compression: Option<i32>,
 
+    /// Information from the SRV URI that generated these client options, if applicable.
     #[builder(default, setter(skip))]
-    pub(crate) original_srv_hostname: Option<String>,
+    #[serde(skip)]
+    pub(crate) original_srv_info: Option<OriginalSrvInfo>,
 
     #[builder(default, setter(skip))]
     pub(crate) original_uri: Option<String>,
@@ -762,7 +764,7 @@ impl From<ClientOptionsParser> for ClientOptions {
             credential: parser.credential,
             cmap_event_handler: None,
             command_event_handler: None,
-            original_srv_hostname: None,
+            original_srv_info: None,
             original_uri: Some(parser.original_uri),
             resolver_config: None,
             server_api: None,
@@ -778,7 +780,10 @@ impl ClientOptions {
     #[cfg(test)]
     pub(crate) fn new_srv() -> Self {
         Self {
-            original_srv_hostname: Some("localhost.test.test.build.10gen.cc".into()),
+            original_srv_info: Some(OriginalSrvInfo {
+                hostname: "localhost.test.test.build.10gen.cc".into(),
+                min_ttl: Duration::from_secs(60),
+            }),
             ..Default::default()
         }
     }
@@ -897,8 +902,12 @@ impl ClientOptions {
                 .resolve_client_options(&options.hosts[0].host())
                 .await?;
 
-            // Save the original SRV hostname to allow mongos polling.
-            options.original_srv_hostname = Some(options.hosts[0].host().to_string());
+            // Save the original SRV info to allow mongos polling.
+            options.original_srv_info = OriginalSrvInfo {
+                hostname: options.hosts[0].host().to_string(),
+                min_ttl: config.min_ttl,
+            }
+            .into();
 
             // Set the ClientOptions hosts to those found during the SRV lookup.
             options.hosts = config.hosts;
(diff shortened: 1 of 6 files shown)


src/client/options/mod.rs
```diff
@@ -523,8 +523,10 @@ pub struct ClientOptions {
     #[builder(default, setter(skip))]
     pub(crate) zlib_compression: Option<i32>,
 
+    /// Information from the SRV URI that generated these client options, if applicable.
     #[builder(default, setter(skip))]
-    pub(crate) original_srv_hostname: Option<String>,
+    #[serde(skip)]
+    pub(crate) original_srv_info: Option<OriginalSrvInfo>,
```
patrickfreed, line 529: originally we were just storing the SRV hostname, but we also need the [min TTL](https://github.com/mongodb/specifications/blob/master/source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst#should-dns-polling-use-heartbeatfrequencyms-or-dns-cache-ttls) to know how long to sleep for before doing another poll.

src/sdam/srv_polling/mod.rs
```diff
@@ -58,23 +54,34 @@ impl SrvPollingMonitor {
...
+        std::cmp::max(self.rescan_interval, MIN_RESCAN_SRV_INTERVAL)
+    }
+
     async fn execute(&mut self) {
+        fn should_poll(tt: TopologyType) -> bool {
+            matches!(tt, TopologyType::Sharded | TopologyType::Unknown)
+        }
+
         while self.topology.is_alive() {
+            RUNTIME.delay_for(self.rescan_interval()).await;
```
patrickfreed, line 67: The issue here was that we were not waiting at all after the client was created to start SRV polling. This is an issue because the very first SRV lookup happens before client creation in `ClientOptions::parse`. So the order of events could end up like this:
1. `ClientOptions::parse("mongodb+srv://...").await`
2. `Client::with_options(options)`
3. SDAM monitor for `<address resolved from SRV records>` starts
4. SRV polling starts
5. Monitor check succeeds, updates topology with `localhost:27017`, `localhost:27018`, and `localhost:27019` and topology type ReplicaSetWithPrimary, removes `<address resolved from SRV records>` from topology
6. SRV poll finishes, replaces all hosts with `<address resolved from SRV records>` again
7. Repeat

By adding this wait to the beginning, we allow the SDAM monitors to discover that the topology is a replica set, and if so the SRV polling monitor will not attempt to update the topology.

src/sdam/srv_polling/mod.rs
```diff
@@ -95,36 +102,23 @@ impl SrvPollingMonitor {
...
-            #[allow(clippy::single_match)]
-            match host {
-                Ok(host) => new_hosts.push(host),
-                Err(_) => {
-                    // TODO RUST-230: Log error with host that was returned.
-                }
-            }
-        }
-
-        Ok(LookupHosts {
```
patrickfreed, line 118: this min ttl calculation was moved to the SRV resolver

src/srv.rs
```diff
@@ -65,33 +75,19 @@ impl SrvResolver {
         let lookup_hostname = format!("_mongodb._tcp.{}", original_hostname);
 
         let srv_lookup = self.resolver.srv_lookup(lookup_hostname.as_str()).await?;
+        let mut srv_addresses: Vec<Result<ServerAddress>> = Vec::new();
+        let mut min_ttl = u32::MAX;
```
patrickfreed, line 79: the changes here are kind of hard to read, but basically it combines the `min_ttl` map and the `srv_addresses` map into a single for loop over the same information.

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -113,10 +117,7 @@ async fn run() {
 
             let mut options_with_tls = options.clone();
             if requires_tls {
-                let tls_options = TlsOptions::builder()
-                    .allow_invalid_certificates(true)
-                    .build();
-                options_with_tls.tls = Some(Tls::Enabled(tls_options));
+                options_with_tls.tls = CLIENT_OPTIONS.tls.clone();
```
isabelatkinson, line 120: What's the reasoning behind this change? From looking over DRIVERS-1324 it seems like we still need to be running these tests with TLS enabled.

isabelatkinson (commented):
looks good, just one question. (it's so nice to see a green patch!)

src/test/spec/initial_dns_seedlist_discovery.rs
patrickfreed: This ensures the test respects the TLS settings specified in the `MONGODB_URI` environment variable. The previously hard-coded TLS options here happened to work with our evergreen config, but it didn't work when I was testing locally, for example, but the new config works in both places. Also, on line 94 we verify that TLS is enabled in `CLIENT_OPTIONS` so copying it here also enables it.

isabelatkinson (approved):

PR #353, opened 2021-06-04
RUST-162 / RUST-814 Update evergreen test variants

patrickfreed:
RUST-162 / RUST-814

This PR updates our evergreen configuration to run on Ubuntu 18.04 ARM64, Ubuntu 18.04, and Ubuntu 20.04, and it removes testing on Ubuntu 16.04.

I also introduced the concept of "bang builds" (i.e. required builds) by adding an exclamation point in front of the more important variants. This will automatically sort them to the top of the waterfall / a patch view and can help for easy differentiation from the other, less important tasks. I borrowed this idea from the server's [evergreen config](https://evergreen.mongodb.com/waterfall/mongodb-mongo-master). I propose that we limit the tasks scheduled as part of a PR to just the bang builds and have the waterfall take care of running the other tasks, especially now that we have more variants. This should help cut back on needless evergreen cycles and speed up the PR process. Let me know if you all like this approach.

The bang builds are as follows:
- regular tests
- lint
- compile

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1330,9 +1330,9 @@ buildvariants:
 - matrix_name: "aws-auth"
   matrix_spec:
     os:
-      - ubuntu-18.04
+      - ubuntu-20.04
     async-runtime: "tokio"
-  display_name: "AWS Auth ${os} with ${async-runtime}"
+  display_name: "${os} AWS Auth with ${async-runtime}"
   tasks:
     - ".aws-auth"
 # TODO: RUST-361 enable these tests once OCSP support is implemented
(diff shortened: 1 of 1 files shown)


patrickfreed:
The test failures are all from known issues. The purple failures are due to the boxes not being able to reach github it seems--not sure what that's about. They don't seem to be related to the new variants so I just restarted those tasks.

isabelatkinson (approved):
lgtm! bang build idea also sgtm.

kmahar:
> The purple failures are due to the boxes not being able to reach github it seems--not sure what that's about.

I was getting some of those last week too

PR #357, opened 2021-06-08
RUST-847 Redact `Credential` `Debug` output (1.2.x)

patrickfreed:
RUST-847

This PR redacts the debug output for `Credential` to just `Credential("REDACTED")` in the 1.2.x driver. I opted to include a redacted implementation rather than omit one so that the `Credential` struct is more easily usable in structs that are `Debug`. This also does allow some amount of debugging of it via print statements in that you can tell one is present in an options struct.

This fixes the security issue reported as part of RUST-591 for the 1.2.x driver. The issue was already resolved in the 2.0.0-beta release of the driver.

### DIFF
--- a/src/client/auth/mod.rs
+++ b/src/client/auth/mod.rs
@@ -10,7 +10,7 @@ mod scram;
 mod test;
 mod x509;
 
-use std::{borrow::Cow, str::FromStr};
+use std::{borrow::Cow, fmt::Debug, str::FromStr};
 
 use hmac::Mac;
 use rand::Rng;
@@ -324,7 +324,7 @@ impl FromStr for AuthMechanism {
 ///
 /// Some fields (mechanism and source) may be omitted and will either be negotiated or assigned a
 /// default value, depending on the values of other fields in the credential.
-#[derive(Clone, Debug, Default, Deserialize, TypedBuilder, PartialEq)]
+#[derive(Clone, Default, Deserialize, TypedBuilder, PartialEq)]
 #[non_exhaustive]
 pub struct Credential {
     /// The username to authenticate with. This applies to all mechanisms but may be omitted when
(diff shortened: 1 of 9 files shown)


src/client/auth/mod.rs
```diff
@@ -437,6 +437,14 @@ impl Credential {
     }
 }
 
+impl Debug for Credential {
```
patrickfreed, line 440: This and the added test are the meat of the changes. There were also some unrelated clippy updates that were made on the master branch but not this one.

src/test/spec/unified_runner/mod.rs
```diff
@@ -245,5 +245,10 @@ async fn test_examples() {
 #[cfg_attr(feature = "async-std-runtime", async_std::test)]
 async fn test_versioned_api() {
     let _guard: RwLockWriteGuard<_> = LOCK.run_exclusively().await;
+
+    // TODO RUST-725 Unskip these tests on 5.0
+    if TestClient::new().await.server_version_gte(5, 0) {
```
patrickfreed, line 250: This skip is on master but not 1.2.x, adding here to prevent the latest tests from failing.

src/cmap/test/mod.rs
```diff
@@ -435,3 +438,44 @@ async fn cmap_spec_tests() {
...
+
+    let handler = Arc::new(crate::test::EventHandler::new());
+    let mut subscriber = handler.subscribe();
+    options.cmap_event_handler = Some(handler.clone());
+
+    let _client = Client::with_options(options).unwrap();
+
+    subscriber
+        .wait_for_event(EVENT_TIMEOUT, |e| {
+            if let crate::test::Event::CmapEvent(CmapEvent::ConnectionPoolCreated(event)) = e {
```
kmahar, line 467: is it worth also explicitly checking that REDACTED is in the debug output?

kmahar (approved):
one suggestion but otherwise LGTM

src/cmap/test/mod.rs
patrickfreed: Updated to check.

isabelatkinson (approved):

PR #363, opened 2021-06-11
RUST-652 Add script to run all linter tests

NBSquare:
Pretty straightforward! If we want to add any other tests to this we should probably add them manually here.

I also think it could be helpful to add this script into a pre-commit hook, but that's probably another discussion.

### DIFF
--- a/.evergreen/check-all.sh
+++ b/.evergreen/check-all.sh
@@ -2,7 +2,8 @@
 
 set -o errexit
 
-EVERGREEN_DIR=$(dirname "$0")
+# Note: This script must run from root, or it won't check src/ files correctly.
+EVERGREEN_DIR=".evergreen"
 
 bash "${EVERGREEN_DIR}/check-clippy.sh"
 bash "${EVERGREEN_DIR}/check-rustdoc.sh"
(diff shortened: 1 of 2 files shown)


.evergreen/check-all.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -o errexit
+
+EVERGREEN_DIR=$(dirname "$0")
```
isabelatkinson, line 5: Why'd you opt for using this rather than the `.evergreen` directory name directly?

isabelatkinson (commented):
looks good! just a question for my own understanding. also, feel free to file a ticket for adding a commit hook so we can discuss as a team during our next triage

.evergreen/check-all.sh
NBSquare: I added it just in case the user was running the script within the `.evergreen` directory.

But now that you mention it, it looks like `check-rustfmt.sh` breaks if you run it from `.evergreen` instead of from the root, so this might actually be redundant.

Now that you mention it, we should probably replace it with this:
```suggestion
EVERGREEN_DIR=".evergreen"
```

That way, if the user runs the script from the wrong folder, it will immediately fail.

isabelatkinson (approved):
lgtm!

abr-egn (approved):
LGTM!

PR #364, opened 2021-06-16
RUST-35 add integration tests for write concerns on some methods

abr-egn:


src/concern/test.rs
```diff
@@ -210,3 +217,251 @@ async fn assert_event_contains_read_concern(client: &EventClient) {
...
+    let mut options = CLIENT_OPTIONS.clone();
+    options.command_event_handler = Some(buffer.clone());
+    let client = TestClient::with_options(Some(options)).await;
+    let db = client.database("test");
+    let coll: Collection = db.collection(name);
+    coll.drop(None).await?;
+    Ok((buffer, coll))
+}
+
+struct CommandBuffer {
```
patrickfreed, line 436: We actually already have some helper types to facilitate this kind of thing, namely `EventClient` (which is a `Client` wrapper which automatically records all events) and `EventHandler` (which is just a handler that can be used when more granularity over the recorded events is required). In this case, I think `EventClient` should suffice. Check out [`index_options_defaults_test`](https://github.com/mongodb/mongo-rust-driver/blob/master/src/test/db.rs#L368) for an example of how to use it.

abr-egn: Handy, thank you!  Updated.

patrickfreed (approved):
LGTM! The doc lint failure looks unrelated and can be fixed separately.

PR #366, opened 2021-06-17
RUST-847 Redact `Credential` `Debug` output

patrickfreed:
RUST-847

This PR redacts the `Debug` output of `Credential` to prevent the accidental leaking of credential via printing. This is basically just a cherry pick of #357 without the test since we no longer include the credential in CMAP events in 2.0.


src/cmap/test/mod.rs
```diff
@@ -21,12 +21,15 @@ use crate::{
     test::{
         assert_matches,
         run_spec_test,
+        CmapEvent,
```
isabelatkinson, line 24: This looks like a holdover from the test that was removed

patrickfreed: Oh whoops, fixed.

isabelatkinson (approved):

PR #367, opened 2021-06-17
RUST-849 Remove extra fields from `ConnectionPoolOptions` included in CMAP events

patrickfreed:
RUST-849

This PR removes the fields from the `ConnectionPoolOptions` that are included in CMAP events that are not mandated by the CMAP specification. This will protect us from accidentally leaking any sensitive information from these events in the future (e.g. from `TlsOptions`).

### DIFF
--- a/src/cmap/options.rs
+++ b/src/cmap/options.rs
@@ -109,14 +109,9 @@ impl ConnectionPoolOptions {
 
     pub(crate) fn to_event_options(&self) -> EventOptions {
         EventOptions {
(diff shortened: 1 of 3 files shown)


isabelatkinson (commented):
Changes LGTM, this is just exposed for CMAP event handling right? i.e. users don't actually construct `ConnectionPoolOptions`?

patrickfreed:
Yep, this struct is purely for `ConnectionPoolCreated` events.

isabelatkinson (approved):

PR #368, opened 2021-06-17
RUST-858 Use `into` field setter for `typed-builder`

isabelatkinson:


patrickfreed (commented):
Changes here look good. I think we should also update the now unnecessary workarounds we had in the driver for times we wanted to pass options in (e.g. in `Collection::clone_with_type`). These can be found by grepping for lines with `mut` and `builder().build()` in them.

patrickfreed (approved):

PR #371, opened 2021-06-21
RUST-766 Use `hyper` in place of `reqwest`

AsianIntel:
The PR replaces `reqwest` in favor of `hyper::Client`. This will also allow upstream libraries and end-users to have a choice in whether to use `reqwest` or not.

- Replaces `reqwest::Client` with `hyper::Client`
- Adds `serde_json` as direct dependencies (They were already dependencies of `reqwest`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ exclude = [
 
 [features]
 default = ["tokio-runtime"]
-tokio-runtime = ["tokio/macros", "tokio/net", "tokio/rt", "tokio/time", "hyper", "hyper-rustls", "serde_bytes", "serde_json"]
+tokio-runtime = ["tokio/macros", "tokio/net", "tokio/rt", "tokio/time", "hyper", "serde_bytes", "serde_json"]
 async-std-runtime = ["async-std", "async-std/attributes", "async-std-resolver", "tokio-util/compat"]
 sync = ["async-std-runtime"]
 # The bson/u2i feature enables automatic conversion from unsigned to signed types during
@@ -47,8 +47,7 @@ futures-util = { version = "0.3.14", features = ["io"] }
 futures-executor = "0.3.14"
 hex = "0.4.0"
 hmac = "0.10.1"
-hyper = { version = "0.14", default-features = false, features = ["client", "http1", "http2", "runtime"], optional = true }
-hyper-rustls = { version = "0.22", default-features = false, features = ["webpki-tokio"], optional = true }
+hyper = { version = "0.14", default-features = false, features = ["client", "http1", "runtime"], optional = true }
 lazy_static = "1.4.0"
 md-5 = "0.9.1"
 os_info = { version = "3.0.1", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ futures-util = { version = "0.3.14", features = ["io"] }
 futures-executor = "0.3.14"
 hex = "0.4.0"
 hmac = "0.10.1"
-hyper = { version = "0.14", default-features = false, features = ["client", "http1", "runtime"], optional = true }
+hyper = { version = "0.14", default-features = false, features = ["client", "http1", "tcp"], optional = true }
 lazy_static = "1.4.0"
 md-5 = "0.9.1"
 os_info = { version = "3.0.1", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ exclude = [
 
 [features]
 default = ["tokio-runtime"]
-tokio-runtime = ["tokio/macros", "tokio/net", "tokio/rt", "tokio/time", "hyper", "hyper-rustls", "serde_bytes"]
+tokio-runtime = ["tokio/macros", "tokio/net", "tokio/rt", "tokio/time", "hyper", "hyper-rustls", "serde_bytes", "serde_json"]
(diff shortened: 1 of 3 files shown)


patrickfreed:
Hi @AsianIntel, thanks for the PR! I'm running the CI patch now and will take a look at the code once that's complete.

AsianIntel:
I'm not sure where the linter is failing, it's telling me it's fine locally

NBSquare:
Try rebasing on master (`git pull upstream master --rebase`) and force pushing once the rebase is resolved.
I fixed this particular lint issue around the same time you made this PR and it probably wasn't merged. 🙂

AsianIntel:
Done lmao

Cargo.toml
```diff
@@ -47,11 +47,13 @@ futures-util = { version = "0.3.14", features = ["io"] }
 futures-executor = "0.3.14"
 hex = "0.4.0"
 hmac = "0.10.1"
+hyper = { version = "0.14", default-features = false, features = ["client", "http1", "tcp"], optional = true }
```
isabelatkinson, line 50: Would you mind updating this to be a block (i.e. `[dependencies.hyper]` rather than the single line here) for readability?

isabelatkinson (commented):
Thanks for the PR, @AsianIntel! I have one small request but otherwise this looks great. I'm also going to tag in the rest of the team for review.

abr-egn (approved):

isabelatkinson (approved):

src/runtime/http.rs
```diff
@@ -54,28 +69,50 @@ impl HttpClient {
...
-        let response = headers
-            .into_iter()
-            .fold(self.inner.request(method, uri), |request, (k, v)| {
-                request.header(*k, *v)
-            })
-            .send()
-            .await?;
+        uri: Uri,
+        headers: &'a [(&'a str, &'a str)],
+    ) -> Pin<Box<dyn Future<Output = Result<Response<Body>, Box<dyn Error>>> + 'a + Send>> {
```
patrickfreed, line 91: Can we write this type as `BoxFuture<'a, Result<Response<Body>, Box<dyn Error>>>` instead? It's a little shorter / easier to read. This may allow us to drop the clippy suppression too.

src/runtime/http.rs
```diff
@@ -15,26 +27,29 @@ impl HttpClient {
...
-            .await?
-            .json()
+        let res = self
+            .request(Method::GET, Uri::from_str(uri)?, headers)
             .await?;
 
-        Ok(value)
+        let mut buf = body::aggregate(res.into_body()).await?;
+        let mut bytes = vec![0; buf.remaining()];
+        buf.copy_to_slice(&mut bytes);
```
patrickfreed, line 41: Since we're just copying to a slice anyways, we can use `body::to_bytes` directly. Ditto for the other use of `body::aggregate` in `request_and_read_string`.

src/runtime/http.rs
```diff
@@ -54,28 +69,47 @@ impl HttpClient {
...
-    ) -> reqwest::Result<Response> {
-        let response = headers
-            .into_iter()
-            .fold(self.inner.request(method, uri), |request, (k, v)| {
-                request.header(*k, *v)
-            })
-            .send()
-            .await?;
+        uri: Uri,
+        headers: &'a [(&'a str, &'a str)],
```
NBSquare, line 87: Is there a reason you replaced the `IntoIterator` logic? The `Iterator::fold` pattern used previously would probably work here too.

src/runtime/http.rs
```diff
@@ -54,28 +69,47 @@ impl HttpClient {
...
+
+            if response.status().is_redirection() {
+                if let Some(Ok(location)) = response.headers().get(LOCATION).map(|u| u.to_str()) {
+                    if let (Some(scheme), Some(authority)) = (uri.scheme_str(), uri.authority()) {
+                        let uri = Uri::builder()
+                            .scheme(scheme)
+                            .authority(authority.as_str())
+                            .path_and_query(location)
+                            .build()?;
+                        return self.request(method, uri, headers).await;
```
NBSquare, line 107: We should probably remove Authorization and other sensitive headers before following the redirect.

src/runtime/http.rs
AsianIntel: Yeah, https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold takes `self` as an argument causing the iterator to get consumed. I think this prevents it from being used further in the redirect.

src/runtime/http.rs
AsianIntel: I'm not sure what you mean here. This module is only used internally here (for AWS). As far as I see, none of the uses have any sensitive headers.

isabelatkinson:
Hi @AsianIntel, upon further discussion with the team we've decided that we don't want to commit to maintaining a lower-level HTTP implementation given some of the concerns that have arisen in this PR. We do agree, though, that pulling in `reqwest` is unnecessary for most users. To that end we think the best course of action is to move our AWS authentication implementation behind a feature flag. Feel free to make those changes in this pull request if you'd like; otherwise we can take on this work. Apologies for the change of course here! Hopefully the introduction of a feature flag will address the issues you're facing.

patrickfreed:
Going to close this out in favor of a new PR that simply moves AWS auth behind a feature flag. Thanks again for your efforts @AsianIntel!

PR #374, opened 2021-06-23
RUST-853 Only mark connection pools ready for data bearing servers or single topologies

abr-egn:


patrickfreed (approved):
LGTM! On a more meta note, I think the commit description would be a little more accurately phrased as "Only mark connection pools ready for data bearing servers or single topologies" or something along those lines, as we were already marking pools ready for direct connections.

isabelatkinson (approved):

PR #375, opened 2021-06-23
RUST-662 Expose errInfo as details in WriteError

abr-egn:


patrickfreed (approved):

isabelatkinson (commented):
Code changes look good, from the ticket description it looks like we also need to add a prose test for 5.0+ servers described here https://github.com/mongodb/specifications/commit/5141a8968a36cb4f29f203110c5d48aaaa281533

abr-egn:
> Code changes look good, from the ticket description it looks like we also need to add a prose test for 5.0+ servers described here [mongodb/specifications@5141a89](https://github.com/mongodb/specifications/commit/5141a8968a36cb4f29f203110c5d48aaaa281533)

I wasn't sure here how much to follow the letter vs the spirit - the previous version of the spec has a similar server-round-trip prose test for WriteConcernError.details that's covered in the rust driver by directly testing response parsing, so I followed that pattern.  Happy to add the roundtrip test if you think that's the way to go :)

isabelatkinson:
I think the roundtrip test is still useful here to ensure that the `errInfo` field from the server is properly deserialized by the driver -- the test in `src/operation/insert/test.rs` that you updated here uses a mock command response that we define ourselves, but we still want to verify that a real command response is properly handled.

isabelatkinson (approved):

PR #376, opened 2021-06-23
RUST-697 / RUST-701 Add sessions and transactions to unified test runner

NBSquare:
As above. Implemented a new unified test runner entity which wraps on a `ClientSession` and on its LSID, so that when the session is dropped, the LSID persists in the entity map to cross-reference with expected events.

Currently skipping one test in `poc-transactions` because it relies on index management, which isn't implemented AFAIK.

### DIFF
--- a/src/test/spec/unified_runner/entity.rs
+++ b/src/test/spec/unified_runner/entity.rs
@@ -118,7 +118,8 @@ impl Deref for ClientEntity {
 }
 
 impl SessionEntity {
-    pub fn new(client_session: ClientSession, lsid: Document) -> Self {
+    pub fn new(client_session: ClientSession) -> Self {
+        let lsid = client_session.id().clone();
         Self {
             client_session: Some(Box::new(client_session)),
             lsid,
--- a/src/test/spec/unified_runner/entity.rs
+++ b/src/test/spec/unified_runner/entity.rs
@@ -1,4 +1,7 @@
-use std::{ops::Deref, sync::Arc};
+use std::{
+    ops::{Deref, DerefMut},
+    sync::Arc,
+};
 
 use crate::{
     bson::{Bson, Document},
@@ -15,7 +18,7 @@ pub enum Entity {
     Client(ClientEntity),
     Database(Database),
     Collection(Collection<Document>),
-    Session(ClientSession),
+    Session(SessionEntity),
     Bson(Bson),
     None,
 }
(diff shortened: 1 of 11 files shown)


src/test/spec/unified_runner/test_runner.rs
```diff
@@ -115,8 +116,19 @@ impl TestRunner {
                     };
                     (id, collection.into())
                 }
-                TestFileEntity::Session(_) => {
-                    panic!("Explicit sessions not implemented");
+                TestFileEntity::Session(session) => {
+                    let id = session.id.clone();
+                    let client = self.get_client(&session.client);
+                    let options: Option<SessionOptions> = session
```
isabelatkinson, line 122: Rather than doing this conversion manually, we can update the `Session` struct [here](https://github.com/mongodb/mongo-rust-driver/blob/4bd9f8eb3a56bd8bdb8285438e55497fa1298c0d/src/test/spec/unified_runner/test_file.rs#L208) to use `Option<SessionOptions>` rather than `Option<Document>`. I put `Document` there originally as a placeholder because explicit sessions (and by extension the `SessionOptions` struct) hadn't been implemented yet.

isabelatkinson (commented):
This looks great! I just have a few small suggestions.

isabelatkinson (approved):
looks good! tagging in the rest of the team for review

src/test/spec/unified_runner/entity.rs
```diff
@@ -25,6 +31,12 @@ pub struct ClientEntity {
     pub ignore_command_names: Option<Vec<String>>,
 }
 
+#[derive(Clone, Debug)]
+pub struct SessionEntity {
+    pub lsid: Document,
+    pub client_session: Option<Box<ClientSession>>,
```
abr-egn, line 37: I'm not sure why the ClientSession is boxed here?

patrickfreed (commented):
Changes look great to me, just echoing @abr-egn's question about the boxed session.

src/test/spec/unified_runner/entity.rs
NBSquare: Clippy errors without the box, since the struct is too large otherwise.
Should I suppress the linter error instead?

patrickfreed: Oh in that case its probably just fine to keep the `Box` then imo.

patrickfreed (approved):

abr-egn (approved):

PR #377, opened 2021-06-23
RUST-536 Eliminate redundant clones

patrickfreed:
RUST-536

This PR removes the need to clone input documents or command responses in the driver, resulting in significant performance improvements (from 15-40% depending on the benchmark). It also removes unneeded clones that were used in the benchmark runner now that references can be accepted in the `insert` family of methods (#331). The benchmark runner was also updated to compile, as it had slipped out of date with the breaking changes in the driver / tokio.


src/error.rs
```diff
@@ -510,13 +508,16 @@ pub struct BulkWriteFailure {
 
     /// The error that occurred on account of write concern failure.
     pub write_concern_error: Option<WriteConcernError>,
+
+    pub(crate) inserted_ids: HashMap<usize, Bson>,
```
patrickfreed, line 512: This is necessary to track how many writes were attempted in the event of an error in an unordered bulk write, but it's also something that users might find useful (RUST-260).

src/operation/mod.rs
```diff
@@ -66,7 +66,8 @@ pub(crate) trait Operation {
     const NAME: &'static str;
 
     /// Returns the command that should be sent to the server as part of this operation.
-    fn build(&self, description: &StreamDescription) -> Result<Command>;
+    /// The operation may store some additional state that is required for handling the response.
+    fn build(&mut self, description: &StreamDescription) -> Result<Command>;
```
patrickfreed, line 70: This was updated to take `&mut self` so that `Insert` could store the `_id`'s it appended to the documents it inserts.

src/operation/insert/mod.rs
```diff
@@ -3,62 +3,103 @@ mod test;
...
+        let mut docs: Vec<Document> = vec![];
+        let mut size = 0;
+
+        for (i, d) in self
+            .documents
+            .iter()
+            .take(description.max_write_batch_size as usize)
+            .enumerate()
+        {
+            let mut doc = bson::to_document(d)?;
```
patrickfreed, line 64: By delaying the serialization until `build`, we can avoid cloning here. We will have to repeat the serialization in the retry making that case slower, but the common case gets much faster.

src/operation/insert/mod.rs
```diff
@@ -3,62 +3,103 @@ mod test;
...
+                        .get(i)
+                        .cloned()
+                        .unwrap_or_else(|| Bson::ObjectId(ObjectId::new()))
+                })
+                .clone();
+
+            let doc_size = bson_util::array_entry_size_bytes(i, &doc);
+
+            if (size + doc_size) <= description.max_bson_object_size as u64 {
+                if self.inserted_ids.len() <= i {
```
patrickfreed, line 78: only populated the `inserted_ids` vector if this is the first time `build` is being called.

src/error.rs
```diff
@@ -510,13 +508,17 @@ pub struct BulkWriteFailure {
...
+    #[serde(skip)]
+    pub(crate) inserted_ids: HashMap<usize, Bson>,
 }
 
 impl BulkWriteFailure {
     pub(crate) fn new() -> Self {
         BulkWriteFailure {
             write_errors: None,
             write_concern_error: None,
+            inserted_ids: Default::default(),
```
abr-egn, line 521: Out of curiosity, why not `HashMap::new()` ?

src/operation/mod.rs
abr-egn: It's a pity that so many sites had to be changed to accommodate the single `Insert` case; I wonder if it would be better to keep this as `&self` and cheat with a `RefCell` in `Index`?

abr-egn (approved):

isabelatkinson (approved):

PR #379, opened 2021-06-25
RUST-852 Move `Collection` generic trait bounds to implementation

isabelatkinson:


univerz:
my fault, looks like i didn't look at the proposed solution with enough understanding. is `T` in `Collection<T>` needed at all? my original thought was to move `T` completely to relevant methods. for example there are some where `T` doesn't even make sense:

```db.collection::<Document>("col").count_documents(None, None).await```

File: src/test/coll.rs [hunk 1]
```diff
@@ -1020,3 +1020,28 @@ async fn drop_skip_serializing_none() {
     let options = DropCollectionOptions::builder().build();
     assert!(coll.drop(options).await.is_ok());
 }
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[function_name::named]
+async fn collection_generic_bounds() {
```
patrickfreed, line 1027: this test and the other one need to acquire the test lock

patrickfreed (commented):
driver changes LGTM, just one comment about the tests

isabelatkinson:
Hi @univerz, we considered removing the generic bound from `Collection` entirely (as you proposed), but decided not to because eliminating this bound leads to some verbose type annotations when calling `find` methods:

```
// does not compile because find_one_and_delete needs a type annotation
match coll.find_one_and_delete(doc! { “x”: 1 }, None).await? {
    Some(doc) => ...,
    None => ...,
}

// also does not compile because the turbofish operator (i.e. ::<Document>)
// cannot be used when a method parameter uses impl (i.e. the second options
// parameter)
match coll.find_one_and_delete::<Document>(doc! { “x”: 1 }, None).await? {
    Some(doc) => ...,
    None => ...,
}

// need to use a type annotation and store the data in a variable
let option: Option<Document> = coll.find_one_and_delete(doc! { “x”: 1}, None).await?;
match option {
    Some(doc) => ...,
    None => ...,
}
```

This is a significant deviation from our current API and fairly burdensome for users who are calling `find` methods frequently. The changes here will still allow you to create and use a `Collection<T>` where `T` implements only `Deserialize` if you're only calling `find` methods (and vice versa with `Serialize` and `insert` methods).

src/test/coll.rs
isabelatkinson: added the lock to the async test (and to another async test that wasn't using it), we weren't using the lock at all in the sync tests so I updated all of those as well

abr-egn (approved):

patrickfreed (approved):

univerz:
thank you for the detailed explanation, i didn't anticipate the problem with generic+impl, it's really unfortunate.

PR #380, opened 2021-06-28
RUST-725 Use "hello" for handshake and heartbeat when an API version is declared

abr-egn:


patrickfreed (approved):
LGTM!

As far as testing goes, we'll get coverage for it when we implement the [required API version](https://github.com/mongodb/specifications/blob/master/source/versioned-api/tests/README.rst#id2) test variant and SDAM integration tests.

isabelatkinson (approved):

PR #381, opened 2021-06-29
RUST-666 Add options for timeseries collection creation

abr-egn:


src/db/options.rs
```diff
@@ -83,6 +85,17 @@ pub struct CreateCollectionOptions {
 
     /// The default configuration for indexes created on this collection, including the _id index.
     pub index_option_defaults: Option<IndexOptionDefaults>,
+
+    /// Specifies options for creating a timeseries collection.
+    pub timeseries: Option<TimeseriesOptions>,
+
+    /// Duration indicating after how long old time-series data should be deleted.
+    #[serde(
+        default,
```
abr-egn, line 94: I don't know why this field needs to be explicitly marked as using the default when it's not present, but without this annotation some tests fail in deserialization complaining about `missing field "expireAfterSeconds"`.

patrickfreed: I think it's because we provide special serialization / deserialization helpers for this field which override the default behavior.

patrickfreed (approved):
LGTM!

src/db/options.rs
```diff
@@ -83,6 +85,17 @@ pub struct CreateCollectionOptions {
 
     /// The default configuration for indexes created on this collection, including the _id index.
     pub index_option_defaults: Option<IndexOptionDefaults>,
+
+    /// Specifies options for creating a timeseries collection.
```
isabelatkinson, line 89: Can we document here that this feature is only available on 5.0+?

isabelatkinson (commented):
looks good! just one doc request

src/db/options.rs
abr-egn: Good call, done.

isabelatkinson (approved):

PR #383, opened 2021-07-01
RUST-122 Support `mongos` pinning for sharded transactions

NBSquare:
Implemented `mongos` pinning on sharded transactions.

Important notes:
- Because recovery token hasn't been implemented, there are a few places where the field is expected or a related error is raised. In order to test the desired functionality for this PR, these issues have been suppressed, along with TODO comments for RUST-97 (implement recovery tokens).
- `src/test/spec/json/transactions` has a number of V2 runner tests which are related to pinning, and these are being skipped through a simple check in `test::spec::transactions::run`. Removing this check would cause recovery token tests to fail. Rather than remove these, I created another method `test::spec::transactions::test_sharded_transactions` which runs a subset of these V2 tests, as well as the unpin test for the unified runner.


NBSquare:
Removing Patrick and Abraham until Isabel gets a chance to look at this.

NBSquare:
[Evergreen Patch](https://spruce.mongodb.com/version/60df30b757e85a7ced5367d5/tasks)

src/client/session/mod.rs
```diff
@@ -285,6 +298,9 @@ impl ClientSession {
                 }
                 .into());
             }
+            TransactionState::Committed { .. } => {
```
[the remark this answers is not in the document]
NBSquare: I think so. Per the spec, we want to unpin the session when:
> A new transaction is started on the ClientSession after the previous transaction has been committed. The session MUST be unpinned before performing server selection for the first operation of the new transaction.

src/sdam/description/topology/mod.rs
```diff
@@ -329,18 +329,12 @@ impl TopologyDescription {
...
-                    self.transaction_support_status = TransactionSupportStatus::Unsupported;
-                }
-                _ => {
-                    if max_wire_version < 7 {
-                        self.transaction_support_status = TransactionSupportStatus::Unsupported;
-                    } else {
-                        self.transaction_support_status = TransactionSupportStatus::Supported;
-                    }
-                }
+            // TODO RUST-734: Evaluate whether we should permanently support sharded transactions.
```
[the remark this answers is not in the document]
NBSquare: Makes sense 🙂

src/client/session/mod.rs
```diff
@@ -449,6 +465,7 @@ impl ClientSession {
             }
             .into()),
             TransactionState::Starting => {
+                self.pinned_session = None;
```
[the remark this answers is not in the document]
NBSquare: Unfortunately this isn't possible. However, what I had written here wasn't right, either. When a transaction is aborted, the session needs to unpin regardless of whether the abort command is executed, but the command needs to be executed on the pinned `mongos` or there's a risk the server won't be able to find the transaction.

What's more, if the abort operation fails with a retryable error, the session needs to unpin prior to server selection for the retry. So we can't put the unpin command at the end of this method, either.

durran:
I'm not familiar with the Rust driver's design but just wanted to chime in with a note that could make everyone's lives easier in the future. When implementing load balancer support, the driver will need to be able to pin connections as well, to both cursors and transactions. So may be something to think about now when the pinning related functionality for mongos goes in. Hope that information may help if no one had it yet.

src/test/spec/v2_runner/mod.rs
```diff
@@ -126,10 +133,22 @@ pub async fn run_v2_test(test_file: TestFile) {
             EventClient::with_additional_options(options, None, test.use_multiple_mongoses, None)
                 .await;
 
-        let _fp_guard = match test.fail_point {
-            Some(fail_point) => Some(fail_point.enable(client.deref(), None).await.unwrap()),
-            None => None,
-        };
+        // TODO RUST-900: Remove this extraneous call.
+        if test.operations.iter().any(|op| op.name == "distinct") {
```
isabelatkinson, line 137: As a slight optimization, we can also add conditions here only to run `distinct` on each host when we're working with a 4.2 sharded cluster, as that's the only topology where we run into this issue.

isabelatkinson (commented):
one small comment but otherwise this looks great! also it looks like changes from some newer commits on master are being shown in the diff for this PR, can you update the branch to include these commits as well to fix?

src/test/spec/v2_runner/mod.rs
NBSquare: That sounds good. As an aside, some of the other drivers have also run into issues related to `StaleDbVersion` on other versions. However, it sounds like these bugs are weird and driver-specific.

Since *we* only ran into the issue on 4.2 sharded, I think it makes sense to add this check, but I want to draw attention to the fact that the issue has impacted other versions on other drivers.

src/test/spec/v2_runner/mod.rs
```diff
@@ -134,7 +134,7 @@ pub async fn run_v2_test(test_file: TestFile) {
                 .await;
 
         // TODO RUST-900: Remove this extraneous call.
-        if test.operations.iter().any(|op| op.name == "distinct") {
+        if internal_client.is_replica_set() && internal_client.server_version_lte(4, 2) && test.operations.iter().any(|op| op.name == "distinct") {
```
isabelatkinson, line 137: Do we want this to be `is_sharded()` rather than `is_replica_set()`?

isabelatkinson (commented):
Just one question from me, going to tag in the rest of the team for review. Can you run a full evg patch (i.e. all the variants prefixed with `!`) and add a link to it on the ticket? (Note to reviewers: evergreen patches aren't running automatically because this PR is on a separate branch. Patches are linked in the comments on RUST-122.)

src/test/spec/v2_runner/mod.rs
NBSquare: Yeah 😢  I just noticed that myself, fixing it now.

src/sdam/description/topology/mod.rs
```diff
@@ -329,21 +329,15 @@ impl TopologyDescription {
...
+            //                If we leave the feature as unsupported, we should revert this code.
+            if max_wire_version < 7
```
abr-egn, line 334: Minor: I think this would be slightly more idiomatic with the assignment outside the if, i.e.
```
self.transaction_support_status = if ... { TransactionSupportStatus::Unsupported } else { TransactionSupportStatus::Supported };
```

abr-egn (approved):
LGTM!

src/sdam/description/topology/mod.rs
NBSquare: Ah yeah, that's much nicer. Don't know why we didn't have that before.

src/client/executor.rs
```diff
@@ -201,6 +226,18 @@ impl Client {
                 // release the selected server to decrement its operation count
                 drop(server);
 
+                if let Some(ref mut session) = session {
+                    if err.contains_label(TRANSIENT_TRANSACTION_ERROR) {
```
patrickfreed, line 230: I think this logic needs to live in `execute_operation_on_connection`, since otherwise retries that fail with these labels won't unpin on error.

patrickfreed (commented):
I have a few minor suggestions, but otherwise looks good!

src/test/spec/unified_runner/test_runner.rs
```diff
@@ -32,22 +33,33 @@ impl TestRunner {
 
     pub async fn insert_initial_data(&self, data: &CollectionData) {
         let write_concern = WriteConcern::builder().w(Acknowledgment::Majority).build();
```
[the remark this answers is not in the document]
NBSquare: Our original implementation of `insert_initial_data` didn't actually call `create` on the collection, and on 4.2 collections aren't created implicitly during transactions, so a transaction-related test was raising an error.

src/client/executor.rs
```diff
@@ -126,7 +133,17 @@ impl Client {
             }
         }
 
-        let server = match self.select_server(op.selection_criteria()).await {
+        let selection_criteria = if let Some(ref session) = session {
```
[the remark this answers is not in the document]
NBSquare: I knew there had to be a cooler way to write this. Thanks!

src/client/executor.rs
```diff
@@ -147,6 +164,14 @@ impl Client {
             }
         };
 
+        // If the transaction is being aborted, unpin session before retry and regardless of whether
+        // the command succeeds, fails, or is executed.
+        if let Some(ref mut session) = session {
```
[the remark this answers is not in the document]
NBSquare: Hm, that's a good point. The unpin has to happen after server selection, but it should be agnostic of whether the operation is executed, if I'm understanding the behavior correctly. How do you feel about leaving an unpin here (so server selection can occur as expected) and another at the end of `abort_transaction`?

I'm thinking we should write an internal test to verify the behavior in this case, since it's probably the strangest unpin.

src/client/executor.rs
NBSquare: They'll probably need to be in both places, in that case - If the error causes the operation to unpin, it has to unpin _before_ the retry. But yeah, there's also the situation where the operation doesn't unpin, but the retry does.

Again, a good situation for a custom test case.

NBSquare: Ah, right - because `execute_operation_on_connection` is called during the operation and on the retry... For some reason I was thinking that the method is only called on the first run.

That's my 5PM Thursday brain for ya...

src/client/executor.rs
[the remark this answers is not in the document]
patrickfreed: Looks good! I think we can remove this unpin here now.

src/client/executor.rs
```diff
@@ -460,13 +484,21 @@ impl Client {
...
 
-                if let Some(session) = session {
+                err.add_labels(Some(connection), session, Some(retryability))?;
+
+                if let Some(ref mut session) = session {
                     if err.is_network_error() {
                         session.mark_dirty();
                     }
+
+                    if err.contains_label(TRANSIENT_TRANSACTION_ERROR)
```
patrickfreed, line 494: This probably won't matter in practice, but just in case: we probably should put this logic after the entire `match` in case `op.handle_response` below in the `Ok` branch happens to return an error with one of these labels.

src/client/executor.rs
```diff
@@ -637,6 +660,7 @@ impl Error {
             TransactionState::Starting | TransactionState::InProgress => {
                 if self.is_network_error() || self.is_server_selection_error() {
                     self.add_label(TRANSIENT_TRANSACTION_ERROR);
+                    session.as_mut().unwrap().unpin_mongos();
```
patrickfreed, line 663: the error may already have had a `TransientTransactionError` label appended to it via the server, so I think we actually need to move this unpin outside of the match on the transaction state. `UnknownTransactionCommitResult` is only appended by the driver so its technically not necessary to move that one too, but it could be nice to do so for symmetry with the other label.

patrickfreed (commented):
Looks really good, just have one minor suggestion

src/client/executor.rs
NBSquare: Also noticed this and made the fix, so I just saw this - I think it makes sense to move the other unpin out as well, so I'll go ahead and do that.

patrickfreed (approved):
LGTM! great job with this

isabelatkinson (approved):

abr-egn (approved):

PR #388, opened 2021-07-07
RUST-670 Expect unified test format operations to succeed

isabelatkinson:
Finished up this GBF ticket as I had a small amount of work left and will be out next Friday.

This PR adds validation that the `valid_fail` tests in the Unified Test Format specification fail upon error.


src/results.rs
```diff
@@ -147,7 +147,10 @@ pub struct DatabaseSpecification {
     pub name: String,
 
     /// The amount of disk space in bytes that is consumed by the database.
-    #[serde(deserialize_with = "crate::bson_util::deserialize_u64_from_bson_number")]
+    #[serde(
+        deserialize_with = "crate::bson_util::deserialize_u64_from_bson_number",
+        serialize_with = "crate::bson::serde_helpers::serialize_u64_as_i64"
```
isabelatkinson, line 152: This addition was necessary because the results from test operations are all serialized into `Bson`.

src/test/spec/mod.rs
```diff
@@ -62,22 +62,26 @@ where
...
+        run_single_test(test_file_full_path, &run_test_file).await;
+    }
+}
 
-        // Printing the name of the test file makes it easier to debug deserialization errors.
-        println!(
-            "Running tests from {}",
-            test_file_full_path.display().to_string()
-        );
+pub(crate) async fn run_single_test<T, F, G>(path: PathBuf, run_test_file: &F)
```
isabelatkinson, line 69: The `valid_fail` tests need to be able to call this method directly to ensure that each test file in the directory fails; just calling `run_spec_test` will panic on the first test file and return.

src/test/spec/unified_runner/mod.rs
```diff
@@ -165,40 +165,41 @@ pub async fn run_unified_format_test(test_file: TestFile) {
...
-                    if let Some(ref expect_result) = operation.expect_result {
-                        let result = result
-                            .unwrap_or_else(|e| {
-                                panic!(
-                                    "{} should succeed, but the following error: {}",
-                                    operation.name, e
-                                )
-                            })
-                            .unwrap_or_else(|| {
+                    if let Some(expect_error) = operation.expect_error {
```
isabelatkinson, line 168: The only change here is refactoring the conditionals to unwrap the result if `expect_error` is not present rather than just when `expect_result` is present.

src/test/spec/unified_runner/mod.rs
```diff
@@ -272,5 +273,38 @@ pub async fn run_unified_format_test(test_file: TestFile) {
...
+    ]
+    .iter()
+    .collect();
+
+    for entry in read_dir(&path).unwrap() {
+        let test_file_path = PathBuf::from(entry.unwrap().file_name());
+        let path = path.join(&test_file_path);
+        let path_display = path.display().to_string();
+
+        std::panic::AssertUnwindSafe(run_single_test(path, &run_unified_format_test))
```
isabelatkinson, line 305: My original plan for this work was to return errors directly from the unified test runner code, but given the amount of panics and unwraps in our tests this led to excessive verbosity. Instead I opted for [`catch_unwind`](https://docs.rs/futures/0.3.15/futures/future/trait.FutureExt.html#method.catch_unwind), which catches panics and converts them into a `Result` that can be examined. [`AssertUnwindSafe`](https://doc.rust-lang.org/nightly/std/panic/struct.AssertUnwindSafe.html) is essentially a wrapper that suppresses a compiler error that would otherwise appear indicating that the future returned from `run_single_test` cannot be unwound safely. I'd be hesitant to use this in driver code given that it bypasses some compiler safety, but because this is just test code this feels like the best solution that doesn't over-complicate the test runner itself.

src/test/spec/unified_runner/operation.rs
```diff
@@ -246,7 +246,7 @@ impl Deref for Operation {
 pub(super) struct DeleteMany {
     filter: Document,
     #[serde(flatten)]
-    options: Option<DeleteOptions>,
+    options: DeleteOptions,
```
isabelatkinson, line 249: The combination of `#[serde(flatten)]` and wrapping these fields in an `Option` was silencing certain deserialization errors.

src/coll/options.rs
```diff
@@ -509,7 +509,7 @@ pub struct CountOptions {
     ///
     /// This options maps to the `maxTimeMS` MongoDB query option, so the duration will be sent
     /// across the wire as an integer number of milliseconds.
-    #[serde(deserialize_with = "deserialize_duration_from_u64_millis")]
+    #[serde(default, deserialize_with = "deserialize_duration_from_u64_millis")]
```
NBSquare, line 512: Is there a reason these are the only fields on which you've applied the `default` attribute?

NBSquare (commented):
lgtm! just want a bit of clarification surrounding some of the changes to serialization logic, the restructures to testing all seem good though.

src/coll/options.rs
isabelatkinson: `default` is needed when `deserialize_with` is being used to default the field to `None` if it isn't specified. I added `default` to these fields specifically because they weren't properly deserializing from the test files.

NBSquare: Makes sense. Does it make more sense to include `default` as a container attribute instead now that the options structs aren't `Option`s? I guess every field is optional so it doesn't matter much either way.

isabelatkinson: Yeah it shouldn't matter, `default` is only really necessary when we're overriding the default deserialization behavior with `deserialize_with` because the method provided there doesn't handle `Option`s.

abr-egn (approved):

NBSquare (approved):

patrickfreed (approved):

PR #390, opened 2021-07-09
RUST-885 Support snapshot sessions

abr-egn:
RUST-885

This adds support for the `snapshot` session option, and the associated client-side machinery to track read operation session timestamps.  This also fixes RUST-878 and RUST-879.


src/cmap/conn/command.rs
```diff
@@ -120,11 +131,21 @@ impl CommandResponse {
         let cluster_time = raw_response
             .get("$clusterTime")
             .and_then(|subdoc| bson::from_bson(subdoc.clone()).ok());
+        let snapshot_time = raw_response
```
abr-egn, line 134: My initial approach was to add `atClusterTime` as an explicit field in the output of the relevant commands (find, aggregate, distinct), but that required:
* introducing a trait for the output type of an operation, which added a lot of churn, and
* additional special handling in operations implemented on top of one of those three (e.g. countDocuments).

The approach here avoids both of those issues at the cost of digging a bit into the raw document.

src/concern/mod.rs
```diff
@@ -21,10 +21,15 @@ use crate::{
...
 /// information about read concerns.
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct ReadConcern {
     /// The level of the read concern.
     pub level: ReadConcernLevel,
+
+    /// The snapshot read timestamp.
+    #[serde(skip_serializing_if = "Option::is_none")]
```
abr-egn, line 31: The server rejects commands with `atClusterTime: null`.

src/cursor/common.rs
```diff
@@ -152,22 +153,20 @@ pub(crate) struct CursorSpecification {
 
 impl CursorSpecification {
     pub(crate) fn new(
```
abr-egn, line 155: This refactor was in service of being able to pass an `at_cluster_time` field through easily, but even without that it seemed worthwhile to leave in.

src/error.rs
```diff
@@ -411,6 +412,11 @@ pub enum ErrorKind {
     #[error("{message}")]
     #[non_exhaustive]
     Transaction { message: String },
+
+    /// The server version does not support the operation.
+    #[error("The server version does not support a database operation: {message}")]
+    #[non_exhaustive]
+    ServerVersion { message: String },
```
patrickfreed, line 419: What do you think about naming this `IncompatibleServerVersion` instead for some added clarity? Or even `IncompatibleServer` in case there are more criteria in the future unrelated to server version that may cause an operation to fail?

src/cmap/conn/command.rs
patrickfreed: As part of #389, I actually did have to introduce a trait for the output type of an operation (there was definitely a lot of churn, but for that work it was required I think), so we could still go with your original plan once that PR is merged. For the purposes of avoiding merge conflicts, I think it'll probably be easier to merge this PR first though and then rebase #389 off of that and adjust for `atClusterTime` accordingly there.

src/error.rs
abr-egn: Good call, done.

src/cmap/conn/command.rs
abr-egn: Ah, with that trait existing it definitely makes sense to go back to extracting the snapshot time that way.  Your plan for merging SGTM.

patrickfreed (approved):
LGTM!

src/client/mod.rs
```diff
@@ -158,20 +158,40 @@ impl Client {
...
+    /// Gets information about each database present in the cluster the Client is connected to.
+    pub async fn list_databases(
+        &self,
+        filter: impl Into<Option<Document>>,
+        options: impl Into<Option<ListDatabasesOptions>>,
+    ) -> Result<Vec<DatabaseSpecification>> {
+        self.list_databases_common(filter, options, None).await
+    }
+
+    /// Gets information about each database present in the cluster the Client is connected to.
```
isabelatkinson, line 184: Can we add "using the provided `ClientSession`" to this doc (in the same vein as e.g. [this](https://docs.rs/mongodb/2.0.0-beta.2/mongodb/struct.Collection.html#method.aggregate_with_session))?

src/concern/mod.rs
isabelatkinson: Out of curiosity, would adding `#[serde(skip_serializing_none)]` to the `ReadConcern` struct do the trick here as well? We use that for a lot of our options structs but I'm wondering if that would break other behavior here.

abr-egn: Ah, I'd missed that, thank you!  It looks like that works as well, updated to be consistent with the rest of the codebase.

src/client/mod.rs
abr-egn: Good call, done.

isabelatkinson (approved):

PR #393, opened 2021-07-15
RUST-656 Allow "serverless" entry in runOnRequirement

abr-egn:
RUST-656

This allows the "serverless" entry, but does not set up the evergreen environment for running tests against a serverless deployment; the goal is to unblock other changes that happen to pick up spec test updates with `serverless: forbid`.

### DIFF
--- a/src/test/spec/unified_runner/test_file.rs
+++ b/src/test/spec/unified_runner/test_file.rs
@@ -20,7 +20,7 @@ use crate::{
         SelectionCriteria,
         WriteConcern,
     },
-    test::{spec::unified_runner::results_match, Serverless, TestClient, DEFAULT_URI},
+    test::{spec::unified_runner::results_match, Serverless, TestClient, DEFAULT_URI, SERVERLESS},
 };
 
 #[derive(Debug, Deserialize)]
@@ -101,6 +101,14 @@ impl RunOnRequirement {
                 None,
             ) {
                 return false;
+            }
+        }
+        if let Some(ref serverless) = self.serverless {
+            let is_serverless = SERVERLESS.as_ref().map_or(false, |s| s == "serverless");
+            match serverless {
+                Serverless::Forbid if is_serverless => return false,
+                Serverless::Require if !is_serverless => return false,
+                _ => (),
             }
         }
         true
--- a/src/test/spec/unified_runner/test_file.rs
+++ b/src/test/spec/unified_runner/test_file.rs
@@ -61,6 +61,7 @@ pub struct RunOnRequirement {
     max_server_version: Option<String>,
     topologies: Option<Vec<Topology>>,
     server_parameters: Option<Document>,
+    serverless: Option<Serverless>,
 }
 
 #[derive(Debug, Deserialize, PartialEq)]
@@ -71,6 +72,14 @@ pub enum Topology {
     Sharded,
     #[serde(rename = "sharded-replicaset")]
     ShardedReplicaSet,
+}
+
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase", deny_unknown_fields)]
+pub enum Serverless {
+    Require,
+    Forbid,
+    Allow,
 }
 
 impl RunOnRequirement {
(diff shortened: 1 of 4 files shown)


src/test/spec/unified_runner/test_file.rs
```diff
@@ -102,6 +103,14 @@ impl RunOnRequirement {
                 return false;
             }
         }
+        if let Some(ref serverless) = self.serverless {
+            let is_serverless = SERVERLESS.as_ref().map_or(false, |s| s == "serverless");
```
abr-egn, line 107: `Option<T>` only implements `PartialEq<Option<T>>`, so this little dance is needed to compare an `Option<String>` against an `Option<&str>` without a spurious temporary allocation.

src/test/spec/v2_runner/test_file.rs
```diff
@@ -27,10 +34,12 @@ pub struct TestFile {
 
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
+#[serde(deny_unknown_fields)]
 pub struct RunOn {
     pub min_server_version: Option<String>,
     pub max_server_version: Option<String>,
     pub topology: Option<Vec<String>>,
+    pub(crate) serverless: Option<Serverless>,
```
abr-egn, line 42: As far as I can tell, this is all unexported test code, so `pub` and `pub(crate)` are functionally equivalent; `pub(crate)` is needed here because the `Serverless` type is itself that, which in turn is because that's what the local file it's in uses for other things.  Might be worth doing a pass to make everything in src/test one or the other.

src/test/spec/v2_runner/test_file.rs
```diff
@@ -27,10 +34,12 @@ pub struct TestFile {
 
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
+#[serde(deny_unknown_fields)]
```
abr-egn, line 37: The v2 runner previously allowed unknown fields, which caused me some amount of confusion during testing when things I expected to fail didn't; I couldn't find a reason to be lax here so added the deny tag.

src/test/spec/unified_runner/test_file.rs
NBSquare: Personally I think it'd be more readable to do something along the lines of this, but I'm open to what others think.
```suggestion
            let is_serverless = if let Some(s) = SERVERLESS.as_ref() {
                s == "serverless"
            } else {
                false
            }
```

NBSquare (approved):
LGTM! One small suggestion but otherwise I'm glad to see this being done, I ran into this issue in my own work and this is a nicer solution than what I did.

src/test/spec/v2_runner/test_file.rs
patrickfreed: I personally try to only use `pub` for things that are truly public, even in test / private modules, though I think we didn't always do that.

src/test/spec/unified_runner/test_file.rs
patrickfreed: Can we move this logic to the `lazy_static` definition? That way it can be used here and elsewhere simply as a boolean.

patrickfreed (commented):
looks good, just have one tiny suggestion

src/test/spec/v2_runner/test_file.rs
isabelatkinson: Good call here, I usually include this tag on all of the test runner structs but must've missed this one. Tiny nit but can we move this into the previous `serde` tag (i.e. `#[serde(rename_all = "camelCase", deny_unknown_fields)]`)?

src/test/spec/v2_runner/test_file.rs
isabelatkinson: I don't have a strict policy on how I handle these things but it would be nice to have a convention -- I think `pub(crate)` on everything that's not actually user-facing is the way to go.

src/test/spec/v2_runner/test_file.rs
abr-egn: Good catch, done.

src/test/spec/unified_runner/test_file.rs
abr-egn: Done.

isabelatkinson (approved):

patrickfreed (approved):
LGTM!

PR #397, opened 2021-07-19
RUST-768 Pass versioned API parameters to `getMore` and transaction-continuing commands.

abr-egn:


patrickfreed (approved):

src/test/spec/versioned_api.rs
```diff
@@ -12,6 +12,10 @@ use super::run_unified_format_test;
 #[cfg_attr(feature = "async-std-runtime", async_std::test)]
 async fn run() {
     let _guard: RwLockWriteGuard<_> = LOCK.run_exclusively().await;
+    // TODO RUST-734 Unskip these tests on sharded deployments.
+    if TestClient::new().await.is_sharded() {
```
isabelatkinson, line 16: I think we ought to localize this skip to the tests in `transaction-handling.json`, as we'll lose test coverage for versioned API on sharded clusters otherwise. Unfortunately there's not a great way to do this; we could add a skip conditional on the test description in the unified runner for the affected tests, or alternatively `transaction-handling.json` could be moved temporarily to a separate directory and be given its own `run` function.

abr-egn: Good call; I went with a third option and added some variants to `run_spec_test` to allow filtering on filename.

isabelatkinson (approved):

PR #398, opened 2021-07-20
RUST-97 Support sharded transactions recovery token

NBSquare:
Implemented recovery tokens from [the spec](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#recoverytoken-field):

> Every successful (`ok:1`) command response in a sharded transaction includes a `recoveryToken` field. Drivers MUST track the most recently received `recoveryToken` field and MUST append this field to any subsequent commitTransaction or abortTransaction commands. Tracking the most recently returned `recoveryToken` allows the server to update the `recoveryToken` mid-transaction if needed.

This ticket retrieves the `recoveryToken` from responses and adds it next to the pinned `mongos` during sharded transactions, passing it back during the appropriate operations.


src/test/spec/v2_runner/mod.rs
```diff
@@ -271,7 +271,8 @@ pub async fn run_v2_test(test_file: TestFile) {
                                 .unwrap_or_else(|| panic!("ClientSession is not pinned"));
 
                             fail_point_guards.push(
-                                internal_client
+                                client
+                                    .deref()
```
NBSquare, line 275: As far as I can tell it doesn't strictly matter here whether the failpoint is enabled on `internal_client` or the `EventClient`, but for consistency with the other failpoint we enable I changed this to the `EventClient`.

src/test/util/failpoint.rs
```diff
@@ -48,10 +48,20 @@ impl FailPoint {
         client: &Client,
         criteria: impl Into<Option<SelectionCriteria>>,
     ) -> Result<FailPointGuard> {
+        // TODO: DRIVERS-1385 remove this logic for moving errorLabels to the top level.
```
NBSquare, line 51: This is necessary because the new spec test (`mongos-recovery-token`) sets up a failpoint with a `RetryableWriteError`, which requires `errorLabels` to be on the failpoint top level when it's initialized. This can also be observed [here](https://github.com/mongodb/mongo-rust-driver/blob/ab01ac04a08b40147a7fdbd5563c7d7cbcd7b159/src/test/spec/retryable_writes/mod.rs#L80)

src/test/util/matchable.rs
```diff
@@ -67,9 +67,12 @@ impl Matchable for Document {
             if k == "afterClusterTime" {
                 continue;
             }
-            // TODO RUST-97: Remove this logic to bypass recoveryToken
-            if k == "recoveryToken" {
-                continue;
+            if k == "recoveryToken" && v.is_placeholder() {
```
NBSquare, line 70: [`recoveryToken` is tested as follows.](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#recoverytoken)

src/client/session/mod.rs
```diff
@@ -118,12 +118,14 @@ pub(crate) struct Transaction {
...
     pub(crate) options: Option<TransactionOptions>,
     pub(crate) pinned_mongos: Option<SelectionCriteria>,
+    pub(crate) recovery_token: Option<Document>,
 }
 
 impl Transaction {
     pub(crate) fn start(&mut self, options: Option<TransactionOptions>) {
         self.state = TransactionState::Starting;
         self.options = options;
+        self.recovery_token = None;
```
isabelatkinson, line 128: Is this assignment necessary? Similarly to `pinned_mongos` I believe this field should be `None` by default upon the first transaction and reset to `None` after previous transactions have finished.

src/operation/run_command/mod.rs
```diff
@@ -109,7 +110,16 @@ impl super::Response for Response {
             .ok()
             .and_then(|doc| bson::from_document(doc.clone()).ok());
 
-        Ok(Self { doc, cluster_time })
+        let recovery_token = doc
+            .get_document("recoveryToken")
+            .ok()
+            .and_then(|doc| bson::from_document(doc.clone()).ok());
```
isabelatkinson, line 116: The conversion on this line shouldn't be necessary as the type for `recovery_token` is already `Option<Document>`.

src/test/spec/v2_runner/mod.rs
isabelatkinson: Hm I agree we should be consistent with which client we're using for failpoints but I think it would make more sense to do all of them with the `internal_client` rather than the `EventClient`. In general we just use the `EventClient` for operations that need to be tracked for command monitoring and the internal client is used for test setup and test runner operations.

src/test/util/matchable.rs
```diff
@@ -67,9 +67,12 @@ impl Matchable for Document {
...
+                if let Some(actual_v) = self.get(k) {
+                    if actual_v.as_document().is_some() {
```
isabelatkinson, line 72: We can make this a bit more concise with `if self.get_document(k).is_ok()`

isabelatkinson (commented):
looks good! just a few comments/suggestions

src/client/session/mod.rs
NBSquare: Yeah. From [the spec](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#recoverytoken-field):
> Drivers MUST clear a session's cached `recoveryToken` when transitioning to the "no transaction" or "starting transaction" state.

So, I'm clearing the `recoveryToken` on `Transaction::start` and `Transaction::reset`

isabelatkinson (approved):
lgtm! tagging in the rest of the team. can you run another evergreen patch and add it to the ticket?

abr-egn (approved):

src/test/util/matchable.rs
```diff
@@ -67,8 +67,7 @@ impl Matchable for Document {
             if k == "afterClusterTime" {
                 continue;
             }
-            // TODO RUST-97: Remove this logic to bypass recoveryToken
-            if k == "recoveryToken" {
+            if k == "recoveryToken" && v.is_placeholder() && self.get_document(k).is_ok() {
```
patrickfreed, line 70: I think this isn't needed since `.matches` accounts for placeholder already.

src/client/executor.rs
```diff
@@ -403,6 +410,13 @@ impl Client {
                     Ok(r) => {
                         self.update_cluster_time(&r, session).await;
                         if r.is_success() {
+                            // Retrieve recovery token from successful response.
+                            if let Some(ref mut session) = session {
```
patrickfreed, line 414: In the event of a deserialization error for an `ok: 1` response, we still want to update the recovery token (e.g. a user provides a `T` to a `find` that doesn't map correctly to their data). The branch for that is a little bit down on lines 463 and 464. We may want some sort of function for this now that it's being done in two places.

src/operation/mod.rs
```diff
@@ -153,6 +153,9 @@ pub(crate) trait Response: Sized {
     /// The `atClusterTime` field of the response.
     fn at_cluster_time(&self) -> Option<Timestamp>;
 
+    /// The `recoveryToken` field of the response.
+    fn recovery_token(&self) -> Option<Document>;
```
patrickfreed, line 157: For consistency with the other methods in this trait, can this be updated to return `Option<&Document>` instead? This will also help prevent any accidental over-cloning in the future.

patrickfreed (commented):
Looks pretty good! I just have a few minor comments / suggestions

src/test/util/matchable.rs
NBSquare: `.matches` checks that the placeholder exists, but not that the underlying value is a BSON Document, which the spec requires.

patrickfreed (approved):
LGTM!

NBSquare:
@abr-egn @isabelatkinson please take a look at the **last** commit when you get the chance - i added a prose test and changed some logic for retrieving the `recoveryToken` which is probably worth looking at.

isabelatkinson (approved):

PR #399, opened 2021-07-20
RUST-881 Run tests with `requireApiVersion: 1`

abr-egn:
RUST-881

This adds an evergreen test configuration to run tests against a server that requires the API version be set in every call.  A few tests needed to be updated because setting the API version on the client switches from `isMaster` to `hello`.


patrickfreed (approved):
Looks great! Hope your first foray into evergreen configuration wasn't too bad :slightly_smiling_face:

isabelatkinson (approved):

PR #401, opened 2021-07-22
 RUST-732 Mark the versioned API options public.

abr-egn:
RUST-732

This approximately reverts the [commit](https://github.com/mongodb/mongo-rust-driver/commit/ffae5c31c6ac368fc0c242c3bcf3b5d8811a7f09) hiding the options; it also updates the documentation for those options to link to the manual (RUST-605).

### DIFF
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -295,7 +295,8 @@ impl fmt::Display for ServerAddress {
 #[derive(Clone, Debug, PartialEq)]
 #[non_exhaustive]
 pub enum ServerApiVersion {
-    V1,
+    /// Use API version 1.
+    V1, 
 }
 
 impl FromStr for ServerApiVersion {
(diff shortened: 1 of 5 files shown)


src/client/options/mod.rs
```diff
@@ -500,8 +501,11 @@ pub struct ClientOptions {
     /// supported and is considered undefined behaviour. To run any command with a different API
     /// version or without declaring one, create a separate client that declares the
     /// appropriate API version.
-    #[builder(default, setter(skip))]
```
patrickfreed, line 503: Can we also update the builder for `ConnectionPoolOptions` to not skip this field? Alternatively we could just remove the builder API for `ConnectionPoolOptions` as it isn't a public-facing type and can be constructed from a `ClientOptions`, which does have a builder.

src/client/options/mod.rs
```diff
@@ -500,8 +501,11 @@ pub struct ClientOptions {
...
-    pub(crate) server_api: Option<ServerApi>,
+    ///
+    /// For more information, see the [Versioned API](
+    /// https://docs.mongodb.com/v5.0/reference/versioned-api/) manual page.
```
patrickfreed, line 506: Can we add this link to the `ServerApi` struct declaration too?

abr-egn: Done.

src/client/options/mod.rs
abr-egn: Removed the builder API.

patrickfreed (approved):
LGTM!

isabelatkinson (approved):
lgtm, sorry for the late review!

PR #402, opened 2021-07-23
RUST-882 Stop relying on `From<u64>` for `Bson`

patrickfreed:
RUST-882
The `From<u64>` impl for `Bson` was recently removed due to it being lossy for certain values. This functionality was relied upon by the driver in a few areas which needed to be updated.

### DIFF
--- a/src/selection_criteria.rs
+++ b/src/selection_criteria.rs
@@ -1,4 +1,5 @@
 use std::{collections::HashMap, sync::Arc, time::Duration};
+use std::convert::TryInto;
 
 use derivative::Derivative;
 use serde::{de::Error, Deserialize, Deserializer};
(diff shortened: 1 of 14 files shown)


abr-egn (approved):

src/selection_criteria.rs
```diff
@@ -304,7 +305,7 @@ impl ReadPreference {
         let mut doc = doc! { "mode": mode };
 
         if let Some(max_stale) = max_staleness {
-            doc.insert("maxStalenessSeconds", max_stale.as_secs());
+            doc.insert("maxStalenessSeconds", max_stale.as_secs().try_into().unwrap_or(i64::MAX));
```
isabelatkinson, line 308: Why use a default value here rather than return an error?

patrickfreed: My original thought was that for the purposes of `maxStalenessSeconds`, `i64::MAX` is effectively "forever", but you're probably right that we should be more precise and error here for invalid values. Updated.

isabelatkinson (approved):

PR #403, opened 2021-07-26
RUST-873 Test redaction of replies to security-sensitive commands

abr-egn:
RUST-873

This pulls in spec test changes, which required adding 1.5 to allowed spec versions and implementing support for `auth` and `observeSensitiveCommands`.  Said tests turned up a bug in the redaction - command response redaction hadn't picked up the `speculativeAuthenticate` clause - which is now fixed.

### DIFF
--- a/src/test/spec/unified_runner/entity.rs
+++ b/src/test/spec/unified_runner/entity.rs
@@ -76,7 +76,7 @@ impl ClientEntity {
             if let Some(ignore_command_names) = ignore_command_names {
                 if ignore_command_names
                     .iter()
-                    .any(|name| event.command_name() == name)
+                    .any(|name| event.command_name().eq_ignore_ascii_case(name))
                 {
                     return false;
                 }
--- a/src/test/spec/unified_runner/test_runner.rs
+++ b/src/test/spec/unified_runner/test_runner.rs
@@ -60,13 +60,10 @@ impl TestRunner {
                     let id = client.id.clone();
                     let observe_events = client.observe_events.clone();
                     let mut ignore_command_names = client.ignore_command_monitoring_events.clone();
-                    match client.observe_sensitive_commands {
-                        Some(true) => {}
-                        _ => {
-                            ignore_command_names
-                                .get_or_insert_with(Vec::new)
-                                .extend(REDACTED_COMMANDS.iter().map(|s| String::from(*s)));
-                        }
+                    if client.observe_sensitive_commands != Some(true) {
+                        ignore_command_names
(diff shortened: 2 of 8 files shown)


src/test/spec/unified_runner/entity.rs
```diff
@@ -76,7 +76,7 @@ impl ClientEntity {
             if let Some(ignore_command_names) = ignore_command_names {
                 if ignore_command_names
                     .iter()
-                    .any(|name| event.command_name() == name)
+                    .any(|name| event.command_name().eq_ignore_ascii_case(name))
```
abr-egn, line 79: The straightforward way to implement `observeSensitiveCommands: false` was to add the list of redacted commands to the ignore list; since that list is lowercase, this needed to change to be a case-insensitive comparison.

src/test/spec/unified_runner/test_runner.rs
```diff
@@ -59,7 +59,12 @@ impl TestRunner {
                 TestFileEntity::Client(client) => {
                     let id = client.id.clone();
                     let observe_events = client.observe_events.clone();
-                    let ignore_command_names = client.ignore_command_monitoring_events.clone();
+                    let mut ignore_command_names = client.ignore_command_monitoring_events.clone();
+                    if !client.observe_sensitive_commands.unwrap_or(false) {
```
NBSquare, line 63: I found the double negation here a little bit difficult to read. I'd be in favor of using `match` here, either leaving the `Some(true)` branch empty or assigning `ignore_command_names` as follows:
```suggestion
                    let ignore_command_names = match client.observe_sensitive_commands {
                        Some(true) => client.ignore_command_monitoring_events.clone(),
                        _ => // same as current logic
```
OR
```suggestion
                   match client.observe_sensitive_commands {
                        Some(true) => {}
                        _ => //same as above
```

NBSquare (approved):
LGTM! One small comment about readability but otherwise good.

src/test/spec/unified_runner/test_runner.rs
abr-egn: Done - went with the second.

isabelatkinson: Another option here would be to make the conditional `if client.observe_sensitive_events != Some(true)` -- I think that's a little more readable than the original and avoids the verbosity of a `match` with an empty case.

isabelatkinson (approved):
LGTM! one style suggestion but you can resolve however you choose

patrickfreed (approved):

PR #404, opened 2021-07-26
RUST-232 SDAM Monitoring

isabelatkinson:
RUST-232

https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst


src/sdam/description/server.rs
```diff
@@ -71,7 +91,7 @@ pub(crate) struct ServerDescription {
     // allows us to ensure that only valid states are possible (e.g. preventing that both an error
     // and a reply are present) while still making it easy to define helper methods on
     // ServerDescription for information we need from the isMaster reply by propagating with `?`.
-    pub(crate) reply: Result<Option<IsMasterReply>, String>,
+    pub(crate) reply: Result<Option<HelloReply>, String>,
```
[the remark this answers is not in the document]
patrickfreed: So in order to completely separate the implementation / public API completely, we introduced the `ServerInfo` struct (in `sdam::public`) a while back to be used for filtering in the `Predicate` selection criteria. I think we may want to also use that type for SDAM monitoring too to reduce duplication.

Given the naming requirements in the specification, we'd probably want to refactor `ServerInfo` to be named `ServerDescription` as part of that (the name won't clash internally since they're in different modules). One thing to consider with this is `ServerInfo` currently wraps a reference to our internal `ServerDescription`, and for the purposes of event emitting that will likely need to be updated to be an owned value. This may hurt the performance of `Predicate` based server selection though, so perhaps we would update `ServerInfo` to have a `Cow<'a, ServerDescription>` in it instead just a reference or just an owned value, and in the SDAM events that need it will just store a `ServerInfo<'static>`.

Similarly, instead of marking `TopologyDescription` public, we probably want to introduce a similar wrapper type into `sdam::public`.

src/is_master.rs
```diff
@@ -34,72 +43,166 @@ pub(crate) fn is_master_command(api: Option<&ServerApi>) -> Command {
...
 }
 
+/// The response to a `hello` command.
+///
+/// See the documentation [here](https://docs.mongodb.com/manual/reference/command/hello/) for more details.
 #[derive(Debug, Clone, Default, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub(crate) struct IsMasterCommandResponse {
+#[non_exhaustive]
+pub struct HelloCommandResponse {
```
patrickfreed, line 128: So this type is currently in a bit of a weird place in that some fields are optional where they shouldn't be (see RUST-726). Also, it's used quite a bit internally for handshakes and what not, so it has some risk of needing to be duplicated if we ever need to change it. Given those concerns and that SDAM monitoring is mostly a logging / debugging tool, I think we may want to keep this internal to the driver and just serve the reply as a `Document`. This will also allow the driver to log fields that it doesn't know about (e.g. from newer server versions). What do you think?

NBSquare (commented):
Looks pretty good! Here's a few initial questions/thoughts/impressions.

src/cmap/establish/handshake/mod.rs
```diff
@@ -150,6 +157,10 @@ impl Handshaker {
         let mut command =
             is_master_command(options.as_ref().and_then(|opts| opts.server_api.as_ref()));
 
+        let handler = options
+            .as_ref()
+            .and_then(|options| options.sdam_event_handler.clone());
```
abr-egn, line 162: Very minor style suggestion: I tend to explicitly write `Arc::clone(...)` to make it clear at the site that it's the pointer being cloned, not the value.

src/sdam/description/server.rs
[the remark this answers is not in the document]
abr-egn: Bikeshed: I'd really prefer to not have two types with the same name, even if the compiler sees them as distinct.

src/sdam/description/topology/mod.rs
```diff
@@ -20,63 +20,87 @@ use crate::{
...
+    /// A sharded topology.
     Sharded,
+
+    /// A topology whose type is not known.
     Unknown,
 }
 
+#[cfg(test)]
+impl TopologyType {
+    fn as_str(&self) -> &str {
```
abr-egn, line 45: I think this will infer the output string to have the same lifetime as `self`; it'll be more useful as `&'static str`.

src/sdam/description/server.rs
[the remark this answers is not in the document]
isabelatkinson: The wrappers would need to keep the internal field non-`pub` in order not to expose the current description types, right? My only concern here is that logging these events won't be super readable with the wrapper type + `Cow`.

(Also agree that duplicate naming can be somewhat confusing, although I'm not sure what we'd want to rename the internal `ServerDescription` type to.)

src/sdam/state/mod.rs
```diff
@@ -514,8 +583,46 @@ impl TopologyState {
         topology: WeakTopology,
     ) -> std::result::Result<Option<TopologyDescriptionDiff>, String> {
         let old_description = self.description.clone();
+
+        if let Some(ref handler) = options.sdam_event_handler {
```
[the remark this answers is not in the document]
patrickfreed: The server changed events need to be moved into the diff block as well I think, since other servers in the topology may get updated as part of the new description. An example is if we have a `ReplicaSetWithPrimary` topology and we get a server description from a server that says it's the primary, we need to mark the server we think is a primary as unknown. To implement this I think we'll need to update `TopologyDescription::diff` to check each server for equality and report the changed ones in another field similar to the other two.

src/is_master.rs
```diff
@@ -34,72 +43,166 @@ pub(crate) fn is_master_command(api: Option<&ServerApi>) -> Command {
...
+) -> Result<HelloReply> {
     if !command.name.eq_ignore_ascii_case("ismaster") && !command.name.eq_ignore_ascii_case("hello")
     {
         return Err(ErrorKind::Internal {
             message: format!("invalid ismaster command: {}", command.name),
         }
         .into());
     }
+
+    if let Some(ref handler) = handler {
```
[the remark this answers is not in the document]
patrickfreed: I did a brief scan and I couldn't find any places where it's constructed as `None`, so I think we can just make `round_trip_time` non-optional if we want. If that's the only blocker here then I think it's probably best for us to make that change and move this logic into the monitor.

src/sdam/description/topology/mod.rs
```diff
@@ -766,6 +842,7 @@ impl Default for TransactionSupportStatus {
 /// Returned from `TopologyDescription::diff`.
 #[derive(Debug)]
 pub(crate) struct TopologyDescriptionDiff {
+    pub(crate) old_addresses: HashSet<ServerAddress>,
```
patrickfreed, line 845: `old_addresses` sort of implies "pre existing" addresses rather than removed ones. could we update this to something like `removed_addresses`? Feel free to change the `new_addresses` to like `added_addresses` if you think that would make this more consistent.

also see my other comment about expanding this to include "changed" or "updated" servers.

src/sdam/public.rs
```diff
@@ -133,3 +173,125 @@ impl<'a> fmt::Display for ServerInfo<'a> {
...
+        match self.description.topology_type {
+            TopologyType::Unknown | TopologyType::ReplicaSetNoPrimary => false,
+            TopologyType::Single | TopologyType::Sharded => {
+                self.description.has_available_servers()
+            }
+            TopologyType::ReplicaSetWithPrimary => true,
+        }
+    }
+
+    /// Gets the type of the topology.
```
isabelatkinson, line 214: Added these getters for the same reasoning described [here](https://github.com/mongodb/mongo-rust-driver/pull/404#discussion_r677802805)

src/sdam/monitor.rs
```diff
@@ -184,7 +184,12 @@ impl HeartbeatMonitor {
         let result = match self.connection {
             Some(ref mut conn) => {
                 let command = is_master_command(self.client_options.server_api.as_ref());
-                run_is_master(command, conn).await
+                // SDAM events should only be emitted if the topology is still alive
```
patrickfreed, line 187: This will still be racy since the topology could become non alive while `run_is_master` is executing, so if we have tests relying on this behavior we may need to still have special handling.

src/event/sdam.rs
```diff
@@ -0,0 +1,206 @@
...
+use serde::Deserialize;
+
+use crate::{
+    bson::{oid::ObjectId, Document},
+    error::Error,
+    options::ServerAddress,
+};
+
+pub use crate::sdam::{
+    public::{ServerInfo as ServerDescription, TopologyInfo as TopologyDescription},
```
patrickfreed, line 14: If we use `type` here instead of `use`, we can elide the lifetime too:
e.g.
```rust
pub type ServerDescription = ServerInfo<'static>;
```
I think from the user's perspective `pub use` and `pub type` are otherwise pretty much the same thing.

src/sdam/public.rs
```diff
@@ -133,3 +173,125 @@ impl<'a> fmt::Display for ServerInfo<'a> {
         write!(f, " }}")
     }
 }
+
+/// A description of the most up-to-date information known about a topology. Further details can
+/// be found in the [Server Discovery and Monitoring specification](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst).
+#[derive(Clone)]
+pub struct TopologyInfo {
```
patrickfreed, line 180: (continuing the conversation here since GitHub won't let me comment on the old thread, cc @abr-egn)

It does seem a bit odd to name this struct `TopologyInfo` but have its only use be as an aliased type. Maybe a better middle ground would be to name this `EventTopologyDescription` and remove the alias on it and keep the alias on `ServerInfo` but update it to `EventServerDescription`. That way the number of names for each type corresponds to the number of use-cases for each. As part of that we'd need to move this definition to `src/event` to live with the alias I guess. What do you all think?

src/sdam/description/topology/test/sdam.rs
```diff
@@ -444,6 +510,87 @@ async fn errors() {
...
+        .database(function_name!())
+        .collection(function_name!())
+        .insert_one(doc! { "x": 1 }, None)
+        .await
+        .unwrap();
+
+    let handler = client.handler.clone();
+    drop(client);
+
+    let last_event = handler.get_all_sdam_events().into_iter().last().unwrap();
```
patrickfreed, line 536: I think it would be better to use the channel to receive the events here to avoid racing the `drop` implementation of `Topology`. To achieve the "last" assertion, you can just use a filter that matches any event and assert the timeout is reached.

src/is_master.rs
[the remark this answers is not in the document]
patrickfreed: Ah I see. So I think that line doesn't actually ever execute, since it only executes if the reply is `Some(Ok(_))` and the server type is unknown, but the server type is unknown only if the `IsMasterReply` is `None` or `Err`. Also, if the reply is `None` or `Err`, then the round trip time of the `ServerDescription` will be `None` as well, so I think we do abide by the spec anyways and we're still safe to remove the `Option` from `IsMasterReply`.

Only reason I push this is it helps make progress towards RUST-726 and also due to the potential races involved with having the logic here I mentioned in another comment.

patrickfreed (commented):
Looking good! Just have some follow-up comments on naming / location of the heartbeat event emitter.

src/sdam/public.rs
abr-egn: Sounds reasonable to me.

src/is_master.rs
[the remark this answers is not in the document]
isabelatkinson: Cool, made the change and it doesn't seem to impact anything. I did need to adjust the return type of `run_is_master` to return the `round_trip_time` in the error case.

src/is_master.rs
```diff
@@ -31,21 +31,35 @@ pub(crate) fn is_master_command(api: Option<&ServerApi>) -> Command {
 /// Run the given isMaster command.
 ///
 /// If the given command is not an isMaster, this function will return an error.
-pub(crate) async fn run_is_master(
-    command: Command,
-    conn: &mut Connection,
-) -> Result<IsMasterReply> {
+pub(crate) async fn run_is_master(command: Command, conn: &mut Connection) -> IsMasterResult {
     if !command.name.eq_ignore_ascii_case("ismaster") && !command.name.eq_ignore_ascii_case("hello")
```
patrickfreed, line 35: edit: also see my other comment about possibly ditching this function (at least in the monitor) altogether.

I think this check of the command name is proving to be a bit over-paranoid and we'd probably be better off just removing it. This function is only called in two places, both of which will always use a hello-type command, and having this check here makes us have to deal with the possibility of a `None` duration. If we do want to enforce this, it'd probably be better to do so as a compile time check via a newtype rather than runtime as well, though I'm not really sure that's necessary either.

src/sdam/monitor.rs
```diff
@@ -184,7 +189,45 @@ impl HeartbeatMonitor {
...
-                run_is_master(command, conn).await
+                if let Some(ref handler) = self.client_options.sdam_event_handler {
+                    if self.topology.is_alive() {
+                        let event = ServerHeartbeatStartedEvent {
+                            server_address: self.address.clone(),
+                        };
+                        handler.handle_server_heartbeat_started_event(event);
+                    }
+                }
+                let is_master_result = run_is_master(command, conn).await;
```
patrickfreed, line 200: I'm starting to think using the `run_is_master` function here is more trouble than its worth. If we were to inline the contents of it here, we could use the actual command response from the server in the `reply` field rather than round tripping through our struct, which currently may accidentally discard fields (the `ok` field being an example). It would also avoid the optional duration stuff I mentioned in my other comment.

I ran into something similar when implementing the raw BSON stuff in the executor and ended up storing the server response and the deserialized one separately, allowing the event handler to use the unmodified server resposne while the driver logic used the deserialized one.

src/sdam/monitor.rs
```diff
@@ -21,6 +27,16 @@ pub(super) const DEFAULT_HEARTBEAT_FREQUENCY: Duration = Duration::from_secs(10)
 
 pub(crate) const MIN_HEARTBEAT_FREQUENCY: Duration = Duration::from_millis(500);
 
+macro_rules! emit_event {
```
isabelatkinson, line 30: I needed to use a macro here instead of a method because the borrow checker does not allow you to pass an immutable reference to something as an argument while you hold an immutable reference to it. In this case, we're holding a mutable reference to `self` on in `perform_is_master` when matching on `self.connection` in the `Some` branch. Because macros are just expanded at the call site, we can bypass this rule by using a macro rather than a method on `HeartbeatMonitor` (or a method that takes references to `HeartbeatMonitor`'s fields) which would require an immutable reference.

src/sdam/monitor.rs
```diff
@@ -184,7 +200,46 @@ impl HeartbeatMonitor {
...
+                                        |err| doc! { "deserialization error": err.to_string() }
+                                    ),
+                                    server_address: self.address.clone(),
+                                },
+                                handle_server_heartbeat_succeeded_event
+                            );
+                            Ok(is_master_reply)
+                        }
+                        Err(err) => {
+                            self.emit_heartbeat_failed_event(err.clone(), round_trip_time);
```
isabelatkinson, line 234: We *can* borrow `self` here because we're finished with using the mutable borrow.

src/sdam/monitor.rs
```diff
@@ -184,7 +200,46 @@ impl HeartbeatMonitor {
...
+                    handle_server_heartbeat_started_event
+                );
+
+                let start_time = Instant::now();
+                let response_result = conn.send_command(command, None).await;
+                let end_time = Instant::now();
+
+                let round_trip_time = end_time.duration_since(start_time);
+
+                match response_result {
```
isabelatkinson, line 217: (cc @patrickfreed) Unfortunately matching on a tuple of `(response_result, response_result.and_then(|raw| raw.as_ref().as_is_master_response(round_trip_time)))` doesn't work here because both of the results in the tuple need ownership of the original error in `response_result`, which requires a clone.

src/sdam/monitor.rs
```diff
@@ -184,7 +200,46 @@ impl HeartbeatMonitor {
...
+                        Err(err) => {
+                            self.emit_heartbeat_failed_event(err.clone(), round_trip_time);
+                            Err(err)
+                        }
+                    },
+                    Err(err) => {
+                        self.emit_heartbeat_failed_event(err.clone(), round_trip_time);
+                        Err(err)
+                    }
+                }
```
patrickfreed, line 242: I believe we need to also emit events in the case that the heartbeat is performed as part of creating the monitoring connection (i.e. the `None` branch here), so we'll need to lift that logic out of the `match` entirely and possibly add a new one. This may alleviate that immutable / mutable borrow issues actually, since the mutable borrow of `conn` will no longer be in scope.

src/cmap/conn/command.rs
```diff
@@ -168,53 +171,33 @@ impl RawCommandResponse {
             .map_err(|_| Error::invalid_authentication_response(mechanism_name))
     }
 
-    /// Deserialize the raw bytes into a response backed by a `Document` for further processing.
-    pub(crate) fn into_document_response(self) -> Result<DocumentCommandResponse> {
-        let response: CommandResponse<Document> = self.body()?;
-        Ok(DocumentCommandResponse { response })
+    pub(crate) fn as_is_master_response(&self, round_trip_time: Duration) -> Result<IsMasterReply> {
```
patrickfreed, line 174: since this performs a somewhat costly conversion, this would be better named as `to_is_master_response`. Check out the [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) for detailed info on when to use `to` / `as` / `into`.

src/sdam/monitor.rs
isabelatkinson: So this ended up making things a bit more complicated because of how establishing a new connection and `Handshaker::handshake` work. A few issues that arose:

1. Creating a new connection can possibly return an error. In this case, we don't want to emit any SDAM events (including a `ServerHeartbeatStartedEvent`) because we won't be sending anything to the server. This means we can't emit the `ServerHeartbeatStartedEvent` prior to the `match` expression on `self.connection` because we're not sure we'll emit a corresponding success/failure event. Alternatively, we could emit this event within each branch of the `match` expression, but ideally we'd avoid duplicating these events and we would still encounter the ownership issue that requires the macro. Emitting the event after the `match` expression would technically emit the right events in the right order, but that would be out of compliance with the spec, which states that the `ServerHeartbeatStartedEvent` should be emitted prior to sending the command to the server.
2. `Handshaker::handshake` can possibly return an error when setting speculative auth info. This is a similar issue to the one presented by the incorrect command name error we removed from `run_is_master`, but this one cannot be removed. Similar to above, we need to ensure we make it past this error prior to emitting a `ServerHeartbeatStartedEvent`.
3. `Handshaker::handshake` uses the `IsMasterReply` parsed from the `RawCommandResponse` for updating the connection's stream description and moving the speculative authentication information from the `IsMasterCommandResponse` to a `FirstRound`. This means we'd need to move this logic out of the method to avoid returning both an `IsMasterReply` and `RawCommandResponse` or converting to an `IsMasterReply` twice.

Given these issues, we'd need to inline a pretty large chunk of `Handshaker::handshake` everywhere it's currently being called in order to catch the right errors and extract the necessary information. I don't think this is a great idea because this will require making duplicate changes in two places any time we need to update our handshake logic, which can be quite error-prone.

To that end, I decided to reintroduce `run_is_master` and contain all of the SDAM event logic in that function. This required passing in the `WeakTopology` to ensure that events are only emitted while the topology is still alive. The `Some` branch of the `match` on `self.connection` and `Handshaker::handshake` both call this method; the latter now takes optional arguments for the topology and handler which are propagated as necessary from `perform_is_master`. This played more nicely with the borrow checker so I was able to implement `emit_event` as a function rather than a macro.

patrickfreed: All good points. I think your solution of reintroducing `run_is_master` with SDAM monitoring included looks good. Sorry for leading you down all these rabbit holes! I think one big lesson learned from all this (which will come in handy for Swift) is that the SDAM implementation and SDAM monitoring implementation should happen at the same time, rather than one and then the other much later, since it can be hard to figure out where this stuff needs to live after the fact.

patrickfreed (commented):
Code changes all look good to me now! Just needs another rebase off of master

abr-egn (approved):
LGTM!

src/event/sdam/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// A [`Client`](../../struct.Client.html) will call this method on each registered handler when
+    /// a server is initialized.
+    fn handle_server_opening_event(&self, _event: ServerOpeningEvent) {}
+
+    /// A [`Client`](../../struct.Client.html) will call this method on each registered handler when
+    /// a server is closed.
+    fn handle_server_closed_event(&self, _event: ServerClosedEvent) {}
+
+    /// A [`Client`](../../struct.Client.html) will call this method on each registered handler when
+    /// its topology description changes. Note that this method will not be called until the
```
patrickfreed, line 187: Was this note about not being called until the `Client` is dropped meant to be included with the topolgy closed handler method?

patrickfreed (commented):
LGTM, just one question on a docstring

src/event/sdam/mod.rs
isabelatkinson: yep good catch, fixed

patrickfreed (approved):
LGTM! Great job with this!

PR #405, opened 2021-07-29
RUST-945 Check that explicit sessions were created on the correct client

isabelatkinson:


src/client/executor.rs
```diff
@@ -72,6 +72,15 @@ impl Client {
         }
         match session.into() {
             Some(session) => {
+                if !Arc::ptr_eq(&self.inner, &session.client().inner) {
```
isabelatkinson, line 75: We don't have `PartialEq` derived for `Client` (and it wouldn't make much sense to do so), so this verifies that the `inner` fields on both the clients point to the same data.

abr-egn (approved):

NBSquare (approved):
LGTM!

patrickfreed (approved):

PR #406, opened 2021-07-29
RUST-871 Serialize directly to BSON bytes in insert operations

patrickfreed:
RUST-871

This PR leverages the `bson::to_vec` function that was introduced in https://github.com/mongodb/bson-rust/pull/279 to greatly increase the performance of inserts. It also introduces the framework for updating other operations to skip the `Document` step during serialization, though that is left as future work for once the raw document API is introduced.


benchmarks/src/main.rs
```diff
@@ -444,7 +444,7 @@ fn parse_ids(matches: ArgMatches) -> Vec<bool> {
     ids
 }
 
-#[cfg_attr(feature = "tokio-runtime", tokio::main)]
+#[cfg_attr(feature = "tokio-runtime", tokio::main(flavor = "current_thread"))]
```
patrickfreed, line 447: Since the benchmarks are single threaded, this ends up giving a nice boost by avoiding the tokio overhead of managing a thread pool.

src/bson_util/mod.rs
```diff
@@ -164,128 +169,30 @@ where
         .ok_or_else(|| D::Error::custom(format!("could not deserialize u64 from {:?}", bson)))
 }
 
-pub fn doc_size_bytes(doc: &Document) -> u64 {
```
patrickfreed, line 167: These were no longer needed as we now just use the length of the serialized vec.

src/bson_util/mod.rs
```diff
@@ -164,128 +169,30 @@ where
...
-fn num_decimal_digits(n: usize) -> u64 {
-    let mut digits = 1;
-    let mut curr = 10;
-
-    while curr < n {
-        curr = match curr.checked_mul(10) {
-            Some(val) => val,
-            None => break,
-        };
+fn num_decimal_digits(mut n: usize) -> u64 {
```
patrickfreed, line 185: the old implementation of this method was incorrect actually so I rewrote it.

src/bson_util/mod.rs
```diff
@@ -300,63 +207,161 @@ pub(crate) fn read_document_bytes<R: Read>(mut reader: R) -> Result<Vec<u8>> {
...
-pub(crate) fn document_to_vec(doc: Document) -> Result<Vec<u8>> {
-    let mut v = Vec::new();
-    doc.to_writer(&mut v)?;
-    Ok(v)
+/// Get the value for the provided key from a buffer containing a BSON document.
+/// If the key is not present, None will be returned.
+/// If the BSON is not properly formatted, an internal error would be returned.
+///
+/// TODO: RUST-924 replace this with raw document API usage.
+pub(crate) fn raw_get(doc: &[u8], key: &str) -> Result<Option<Bson>> {
```
patrickfreed, line 215: Unfortunately this function needs to live in the driver for now as we need to be able to comb through the serialized command for `_id` values, but it can go away shortly once RUST-924 is complete.

src/bson_util/mod.rs
```diff
@@ -300,63 +207,161 @@ pub(crate) fn read_document_bytes<R: Read>(mut reader: R) -> Result<Vec<u8>> {
...
+
+    fn deserialize_error<T: std::error::Error>(_e: T) -> Error {
+        deserialize_error_no_arg()
+    }
+
+    fn deserialize_error_no_arg() -> Error {
+        Error::from(ErrorKind::Internal {
+            message: "failed to read from serialized document".to_string(),
+        })
+    }
```
patrickfreed, line 232: The documents queried here should always be valid as they're coming straight out of our serializer, so any error encountered here is mapped to an internal error.

src/operation/insert/mod.rs
```diff
@@ -1,36 +1,37 @@
...
+    runtime::SyncLittleEndianWrite,
     Namespace,
 };
 
-use super::CommandResponse;
+use super::{CommandBody, CommandResponse};
 
 #[derive(Debug)]
-pub(crate) struct Insert<T> {
+pub(crate) struct Insert<'a, T> {
```
patrickfreed, line 24: Previously this `T` was always actually a reference. This now makes it required to be a reference, which allows us to clone the input.

src/operation/insert/mod.rs
```diff
@@ -65,25 +67,37 @@ impl<T: Serialize> Operation for Insert<T> {
...
-                    self.inserted_ids
-                        .get(i)
-                        .cloned()
-                        .unwrap_or_else(|| Bson::ObjectId(ObjectId::new()))
-                })
-                .clone();
-
-            let doc_size = bson_util::array_entry_size_bytes(i, &doc);
+            let mut doc = bson::to_vec(d)?;
+            let id = match bson_util::raw_get(doc.as_slice(), "_id")? {
```
patrickfreed, line 71: Since we can't inspect the `T` for an `_id` without requiring it implement a trait, we scan through the raw document for it instead.

src/operation/insert/mod.rs
```diff
@@ -65,25 +67,37 @@ impl<T: Serialize> Operation for Insert<T> {
...
+                Some(b) => b,
+                None => {
+                    let oid = ObjectId::new();
+
```
patrickfreed, line 75: If it's not in there, we write the new element into the beginning of the document. Once we have the raw document API we could update this to use it.

src/operation/insert/mod.rs
```diff
@@ -96,20 +110,61 @@ impl<T: Serialize> Operation for Insert<T> {
...
+                length: size as i32,
+            },
+            options,
         };
 
-        append_options(&mut body, self.options.as_ref())?;
+        Ok(Command::new("insert".to_string(), self.ns.db.clone(), body))
+    }
+
+    fn serialize_command(&mut self, cmd: Command<Self::Command>) -> Result<Vec<u8>> {
```
patrickfreed, line 128: Since we already have raw BSON for the documents, we can't just use serde to serialize it, as it'll attempt to do so as a a `Binary`. Instead, we serialize the rest of the command and manually append the serialized documents. Again, this will be easier with the raw document API.

src/cmap/conn/mod.rs
```diff
@@ -240,17 +253,22 @@ impl Connection {
...
-        let response_message_result = Message::read_from(&mut self.stream).await;
-        self.command_executing = false;
-        self.error = response_message_result.is_err();
-
-        RawCommandResponse::new(self.address.clone(), response_message_result?)
+    /// Executes a `RawCommand` and returns a `CommandResponse` containing the result from the
+    /// server.
+    ///
+    /// An `Ok(...)` result simply means the server received the command and that the driver
+    /// driver received the response; it does not imply anything about the success of the command
```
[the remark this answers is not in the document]
patrickfreed: fixed

src/cmap/conn/command.rs
```diff
@@ -11,76 +11,105 @@ use crate::{
...
+            start_transaction: None,
+            autocommit: None,
+            read_concern: None,
+            recovery_token: None,
         }
     }
 
     pub(crate) fn set_session(&mut self, session: &ClientSession) {
-        self.body.insert("lsid", session.id());
+        self.lsid = Some(session.id().clone())
```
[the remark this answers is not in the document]
patrickfreed: some of them do have a smidge of logic (e.g. `set_autocommit` sets it to false which could be hard to remember), and they're nice to have if we ever have to do a refactor like this again, so I'd prefer if they were kept around personally.

src/client/auth/mod.rs
```diff
@@ -469,7 +469,7 @@ impl Debug for Credential {
 /// Contains the first client message sent as part of the authentication handshake.
 pub(crate) enum ClientFirst {
     Scram(ScramVersion, scram::ClientFirst),
-    X509(Command),
+    X509(Box<Command>),
```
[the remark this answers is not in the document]
patrickfreed: clippy complained that the `X509` variant was a lot bigger in size than the other one and recommended boxing to even it out a bit.

abr-egn (approved):

src/cmap/conn/command.rs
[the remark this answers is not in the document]
patrickfreed: This did make me realize that we no longer need to return `Result` from a bunch of these, so I went ahead and updated them to not return anything.

src/operation/abort_transaction/mod.rs
```diff
@@ -28,11 +30,12 @@ impl AbortTransaction {
 
 impl Operation for AbortTransaction {
     type O = ();
+    type Command = Document;
     type Response = CommandResponse<WriteConcernOnlyBody>;
 
     const NAME: &'static str = "abortTransaction";
 
-    fn build(&mut self, _description: &StreamDescription) -> Result<Command> {
+    fn build(&mut self, _description: &StreamDescription) -> Result<Command<Self::Command>> {
```
isabelatkinson, line 38: I don't think the addition of `Self::Command` here is necessary because the default `T` for `Command` is already `Document` -- can we revert this change or change the rest of the operations whose `build` signature wasn't updated (e.g. `ListCollections`) for consistency?

src/operation/insert/mod.rs
isabelatkinson: Can we add a TODO comment for this?

src/test/spec/unified_runner/test_file.rs
```diff
@@ -145,15 +145,38 @@ pub struct Client {
...
+    pub observe_sensitive_commands: Option<bool>,
+    #[serde(default, deserialize_with = "deserialize_server_api")]
     pub server_api: Option<ServerApi>,
 }
 
 fn default_uri() -> String {
     DEFAULT_URI.clone()
 }
 
+pub fn deserialize_server_api<'de, D>(
```
isabelatkinson, line 158: Why was this custom implementation needed?

patrickfreed: The server expects the fields to be named `apiVersion`, `apiStrict`, and `apiDeprecationErrors`, but the test format drops all of the `api-` prefixes. Now that we're using the `Serialize` implementation for `ServerApi` in the driver, we needed to introduce a custom one for deserializing from the test format.

src/operation/abort_transaction/mod.rs
patrickfreed: done.

src/operation/insert/mod.rs
patrickfreed: done

src/test/spec/unified_runner/test_file.rs
NBSquare: I know this is only used within this test file, but for clarity can we rename this serialization function? Maybe `deserialize_server_api_for_testing` or `deserialize_server_api_test_format`?

NBSquare (commented):
LGTM! Only a few small things from me.

src/test/spec/unified_runner/test_file.rs
isabelatkinson: Ah in that case I think you should be able to change the `rename` attributes on the fields in `ServerApi` to `#[serde(rename(serialize = "apiVersion"))]` so that the renames only apply to serialization, and then add `#[serde(rename_all = "camelCase", deny_unknown_fields)]` to the top of the struct for deserialization.

src/client/executor.rs
```diff
@@ -388,23 +388,29 @@ impl Client {
...
-                || HELLO_COMMAND_NAMES.contains(name.as_str())
-                    && cmd.body.contains_key("speculativeAuthenticate")
+        let should_redact = cmd.should_redact();
+
+        let cmd_name = cmd.name.clone();
+        let target_db = cmd.target_db.clone();
+
+        let serialized = op.serialize_command(cmd)?;
+        let raw_cmd = RawCommand {
+            name: cmd_name.clone(),
```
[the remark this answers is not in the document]
patrickfreed: I thought so too, but it's actually also used below when emitting the command succeeded / failed event.

src/test/spec/unified_runner/test_file.rs
patrickfreed: Oh cool, I didn't know that was possible. Unfortunately, it looks like we'll still need the separate implementation though, since `deny_unknown_fields` can't be used with `flatten` (see https://serde.rs/attr-flatten.html). `flatten` is needed with server API since they're all top level fields of commands rather than in a subdocument.

I did update the fuinction to be more precisely named (went with the `test_format` suggestion)

isabelatkinson (approved):

PR #408, opened 2021-07-30
RUST-734 Sharded Transactions

NBSquare:
This is the final PR to implement sharded transactions. The first two commits have already gone through the review process, and the third commit wraps up some loose ends surrounding documentation, testing, etc. So, that third commit is the most relevant one to look at IMO.


src/concern/test.rs
```diff
@@ -179,8 +179,9 @@ async fn snapshot_read_concern() {
         .database(function_name!())
         .collection::<Document>(function_name!());
 
-    // TODO RUST-734: Run this test on sharded clusters when transactions are complete.
-    if client.is_replica_set() && client.server_version_gte(4, 0) {
+    if client.is_replica_set() && client.server_version_gte(4, 0)
```
isabelatkinson, line 182: Can we add a `supports_transactions` method to `TestClient` to perform these checks and call that instead in the places where we have this logic?

isabelatkinson (commented):
looks good, just one small suggestion

isabelatkinson (approved):
lgtm, great work on this! tagging in the rest of the team

patrickfreed (commented):
Can we remove the Python file used to generate the transaction spec tests? Other than that LGTM!

abr-egn (approved):

patrickfreed (approved):
LGTM!

PR #412, opened 2021-07-30
RUST-580 Improve compile times of the test suite

patrickfreed:
RUST-580

This PR reimplements the various test runner `TestOperation` traits manually rather than relying on `async_trait`, resulting in much faster compile times of the tests. For context, see https://github.com/rust-lang/rust/issues/87012#issuecomment-881735995.

On my machine, an incremental change to `TestClient::new` (e.g. adding a single `println!`) used to take ~1m15s to run `cargo check --tests`, whereas the same change now takes ~10s. There's still a lot of room for improvement (such a change should ideally be near instant), but this is still a big difference.

### DIFF
--- a/src/test/spec/unified_runner/operation.rs
+++ b/src/test/spec/unified_runner/operation.rs
@@ -1,7 +1,6 @@
 use std::{collections::HashMap, convert::TryInto, fmt::Debug, ops::Deref, time::Duration};
 
-use async_trait::async_trait;
-use futures::stream::TryStreamExt;
+use futures::{future::BoxFuture, stream::TryStreamExt, FutureExt};
 use serde::{de::Deserializer, Deserialize};
 
 use super::{Entity, ExpectError, TestRunner};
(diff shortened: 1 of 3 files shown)


src/test/spec/unified_runner/operation.rs
```diff
@@ -37,15 +36,21 @@ use crate::{
     RUNTIME,
 };
 
-#[async_trait]
 pub trait TestOperation: Debug {
-    async fn execute_test_runner_operation(&self, test_runner: &mut TestRunner);
+    fn execute_test_runner_operation<'a>(
+        &'a self,
+        _test_runner: &'a mut TestRunner,
+    ) -> BoxFuture<'a, ()> {
```
abr-egn, line 43: Does this (and likewise below) need a default body?  It didn't have one before.

abr-egn (approved):

isabelatkinson (approved):
lgtm! this is awesome, these changes speed up compile time to ~10 seconds for me as well

src/test/spec/unified_runner/operation.rs
patrickfreed: They didn't have them previously because default implementations didn't place nicely with `async_trait`, see this comment for more info: https://github.com/mongodb/mongo-rust-driver/blob/master/src/test/spec/v2_runner/operation.rs#L43.

NBSquare (approved):
This looks great. Based on the issue you linked it does seem like it's still worth investigating why `async_trait` causes this issue. Avoiding use of the trait definitely hurts the readability of the test code IMO.

patrickfreed:
Yeah definitely, I updated the original issue and also filed one on `async_trait`: https://github.com/dtolnay/async-trait/issues/174

PR #415, opened 2021-08-03
RUST-950 Enable connection to a load balancer

abr-egn:
RUST-950

This implements the connection establishment portion of load balancer support, i.e.:
- Parsing the `loadBalanced` URI/TXT option
- Modifications to handshake and topology/server type tracking
- Per-`serviceId` connection pool tracking

The work here happened to also encompass RUST-955.

### DIFF
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -558,6 +558,7 @@ pub struct ClientOptions {
     pub(crate) heartbeat_freq_test: Option<Duration>,
 
     /// Allow use of the `load_balanced` option.
+    // TODO RUST-653 Remove this when load balancer work is ready for release.
     #[builder(default, setter(skip))]
     #[serde(skip)]
     pub(crate) allow_load_balanced: bool,
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -557,10 +557,14 @@ pub struct ClientOptions {
     #[cfg(test)]
     pub(crate) heartbeat_freq_test: Option<Duration>,
 
-    /// Allow connecting to a load balancer.
-    #[builder(default)]
+    /// Allow use of the `load_balanced` option.
+    #[builder(default, setter(skip))]
     #[serde(skip)]
-    pub(crate) allow_load_balancer: Option<bool>,
+    pub(crate) allow_load_balanced: bool,
+
+    /// Whether or not the client is connecting to a MongoDB cluster through a load balancer.
+    #[builder(default, setter(skip))]
+    pub(crate) load_balanced: Option<bool>,
 }
 
 fn default_hosts() -> Vec<ServerAddress> {
@@ -694,6 +698,7 @@ struct ClientOptionsParser {
     auth_mechanism_properties: Option<Document>,
     read_preference: Option<ReadPreference>,
     read_preference_tags: Option<Vec<TagSet>>,
+    load_balanced: Option<bool>,
     original_uri: String,
 }
 
@@ -926,7 +931,8 @@ impl From<ClientOptionsParser> for ClientOptions {
             server_api: None,
             #[cfg(test)]
             heartbeat_freq_test: None,
-            allow_load_balancer: None,
+            allow_load_balanced: false,
+            load_balanced: parser.load_balanced,
         }
     }
 }
@@ -1128,6 +1134,14 @@ impl ClientOptions {
         if let Some(ref write_concern) = self.write_concern {
             write_concern.validate()?;
         }
+
+        if !self.allow_load_balanced && self.load_balanced.is_some() {
+            return Err(ErrorKind::InvalidArgument {
+                message: "loadBalanced is not supported".to_string(),
+            }
+            .into());
+        }
+
         Ok(())
     }
 
@@ -1682,6 +1696,9 @@ impl ClientOptionsParser {
             k @ "journal" => {
                 let mut write_concern = self.write_concern.get_or_insert_with(Default::default);
                 write_concern.journal = Some(get_bool!(value, k));
+            }
+            k @ "loadBalanced" => {
+                self.load_balanced = Some(get_bool!(value, k));
             }
             k @ "localthresholdms" => {
                 self.local_threshold = Some(Duration::from_millis(get_duration!(value, k)))
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -1724,7 +1724,7 @@ impl ClientOptionsParser {
                 let mut write_concern = self.write_concern.get_or_insert_with(Default::default);
                 write_concern.journal = Some(get_bool!(value, k));
             }
-            k @ "loadBalanced" => {
+            k @ "loadbalanced" => {
                 self.load_balanced = Some(get_bool!(value, k));
             }
             k @ "localthresholdms" => {
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -1150,17 +1150,21 @@ impl ClientOptions {
             if self.hosts.len() > 1 {
                 return Err(ErrorKind::InvalidArgument {
                     message: "cannot specify multiple seeds with loadBalanced=true".to_string(),
-                }.into());
+                }
+                .into());
             }
             if self.repl_set_name.is_some() {
                 return Err(ErrorKind::InvalidArgument {
                     message: "cannot specify replicaSet with loadBalanced=true".to_string(),
-                }.into());
+                }
+                .into());
             }
             if self.direct_connection == Some(true) {
                 return Err(ErrorKind::InvalidArgument {
-                    message: "cannot specify directConnection=true with loadBalanced=true".to_string(),
-                }.into());
+                    message: "cannot specify directConnection=true with loadBalanced=true"
+                        .to_string(),
+                }
+                .into());
             }
         }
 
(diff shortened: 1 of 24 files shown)


src/client/options/mod.rs
```diff
@@ -552,6 +552,16 @@ pub struct ClientOptions {
     #[builder(default)]
     #[cfg(test)]
     pub(crate) heartbeat_freq_test: Option<Duration>,
+
+    /// Allow use of the `load_balanced` option.
+    // TODO RUST-653 Remove this when load balancer work is ready for release.
+    #[builder(default, setter(skip))]
+    #[serde(skip)]
+    pub(crate) allow_load_balanced: bool,
```
abr-egn, line 560: This awkward double-flag construction allows `load_balanced` to be populated by the URI/TXT parsing as it will be  on release, but prevents that from being used outside of tests that have access to set `allow_load_balanced` to be true.

src/cmap/worker.rs
```diff
@@ -447,10 +468,17 @@ impl ConnectionPoolWorker {
...
+            PoolState::Ready
+        } else {
+            self.generation += 1;
+            let prev = std::mem::replace(&mut self.state, PoolState::Paused(cause.clone()));
+            self.generation_publisher.publish(self.generation);
+            prev
+        };
 
         if matches!(previous_state, PoolState::Ready) {
             self.emit_event(|handler| {
```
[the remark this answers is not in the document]
abr-egn: Done.

src/sdam/state/mod.rs
```diff
@@ -245,19 +250,22 @@ impl Topology {
         error: Error,
         handshake: HandshakePhase,
         server: &Server,
+        service_id: Option<ObjectId>,
     ) -> bool {
         let state_lock = self.state.write().await;
         if handshake.generation() < server.pool.generation() {
```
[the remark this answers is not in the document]
abr-egn: Done; I ended up with `PoolGeneration` and `ConnectionGeneration` as distinct types.

The error cases turned out to be annoying; propagating the `Result`s wasn't an issue, but those cap out in the event loop of the worker thread, and I didn't see any way to usefully surface them to the user from there, so I put in a TODO to log them.  Very open to better options, I just didn't see any.

src/sdam/state/mod.rs
```diff
@@ -245,19 +250,37 @@ impl Topology {
...
-            return false;
+        match &handshake {
+            HandshakePhase::BeforeCompletion { generation } => {
+                match (generation, server.pool.generation()) {
+                    (PoolGeneration::Normal(hgen), PoolGeneration::Normal(sgen)) => {
+                        if *hgen < sgen {
+                            return false;
+                        }
+                    }
+                    (PoolGeneration::LoadBalanced(_), PoolGeneration::LoadBalanced(_)) => {}
```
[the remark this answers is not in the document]
abr-egn: Done.

src/cmap/worker.rs
```diff
@@ -568,7 +650,7 @@ async fn establish_connection(
     event_handler: Option<&Arc<dyn CmapEventHandler>>,
 ) -> Result<Connection> {
     let connection_id = pending_connection.id;
-    let generation = pending_connection.generation;
+    let generation = pending_connection.generation.clone();
     let address = pending_connection.address.clone();
 
     let mut establish_result = establisher.establish_connection(pending_connection).await;
```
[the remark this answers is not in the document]
abr-egn: Got it, thanks for breaking down the explanation.  I ended up with something that's mostly equivalent to your suggestion - a new `HandshakePhase`, and an `EstablishError` for the result of `establish_connection` to capture the needed state.

src/srv.rs
```diff
@@ -189,11 +191,28 @@ impl SrvResolver {
                 "replicaset" => {
                     config.replica_set = Some(parts[1].into());
                 }
+                "loadBalanced" => {
```
xforce, line 194: This needs to be lowercase. See Line 187.

abr-egn: Good catch, thank you!

abr-egn:
This has broken all tests with auth enabled; converted to draft while I track that down.

abr-egn:
> This has broken all tests with auth enabled; converted to draft while I track that down.

Ah, found it; I'd broken `HandshakePhase::is_before_completion`.  Good to review.

src/lib.rs
```diff
@@ -99,6 +99,8 @@ macro_rules! define_if_single_runtime_enabled {
 // In the case that neither tokio nor async-std is enabled, we want to disable all compiler errors
 // and warnings other than our custom ones.
 define_if_single_runtime_enabled! {
+    #[macro_use]
+    pub mod error;
```
patrickfreed, line 103: Rather than `#[macro_use]`, we can rely on the more "modern" `pub(crate) use <macro name>;` inside of the `error` module. See https://stackoverflow.com/a/67140319

src/sdam/state/mod.rs
```diff
@@ -477,15 +506,23 @@ impl TopologyState {
             topology.clone(),
             self.http_client.clone(),
         );
+        let is_load_balanced = matches!(topology.client_options().load_balanced, Some(true));
+        if is_load_balanced {
+            // Because load balancer servers don't have a monitoring connection, the associated
+            // connection pool needs to be directly marked as ready.
+            let server = Arc::clone(&server);
+            RUNTIME.execute(async move {
```
patrickfreed, line 514: Since this task is running in the background, there could be a potential race between the first operation and this mark as ready completing. I think we could avoid this if we the pool just marked itself as ready in `new` depending on if we're in load balanced mode or not.

src/sdam/state/mod.rs
```diff
@@ -557,23 +594,27 @@ impl TopologyState {
...
+#[derive(Debug, Clone)]
 pub(crate) enum HandshakePhase {
-    /// Describes an point that occurred before the handshake completed (e.g. when opening the
-    /// socket or while performing authentication)
-    BeforeCompletion { generation: u32 },
+    /// Describes a point that occurred before the initial hello completed (e.g. when opening the
+    /// socket).
+    PreHello { generation: PoolGeneration },
+
+    /// Describes a point in time after the initial hello has completed, but before authentication.
```
patrickfreed, line 603: maybe instead of saying before authentication, we can say "before the entire handshake (e.g. including authentication) completes", since this phase includes errors due to authentication (and any other steps in the future).

src/error.rs
```diff
@@ -584,3 +584,26 @@ pub(crate) fn convert_bulk_errors(error: Error) -> Error {
         _ => error,
     }
 }
+
+/// Flag a load-balanced mode mismatch.  In tests, it will panic; outside of tests, it will
```
patrickfreed, line 588: I like the idea of conditionally panicking, but what if we gated it on debug vs release (i.e. `#[cfg(debug_assertions)]`)? That way user test harnesses that may not be using cargo's test framework will also leverage this.

Also, what about returning in this macro instead of just evaluating to an expression? This would match the `todo!()` and `unimplemented!()` macros and would avoid accidentally progressing through an inconsistent state.

src/cmap/establish/mod.rs
```diff
@@ -38,26 +46,74 @@ impl ConnectionEstablisher {
...
+                .await
+                .map_err(|e| EstablishError::post_hello(e, connection.generation.clone()))?
         }
 
         Ok(connection)
     }
 }
+
+#[derive(Debug, Clone, Error)]
+#[error("{cause}")]
```
patrickfreed, line 100: `std::result::Result` doesn't actually require that the `Err` case be an `Error`, so I think we can drop the `Error` derivation here to simplify things a little (unless we need it to be an `Error` for some reason I missed of course).

patrickfreed (commented):
Looking good! Just have a few minor suggestions / questions.

src/lib.rs
abr-egn: Ah, didn't know about that, much nicer - done.

src/sdam/state/mod.rs
abr-egn: Done; also realized that `Monitor` was already storing the `ClientOptions`, so `Monitor::start` didn't need the `is_load_balanced` parameter.

src/sdam/state/mod.rs
abr-egn: Done.

src/error.rs
abr-egn: Done.

src/cmap/establish/mod.rs
abr-egn: Done.

isabelatkinson (approved):

src/cmap/worker.rs
```diff
@@ -189,7 +207,17 @@ impl ConnectionPoolWorker {
             .unwrap_or(MAINTENACE_FREQUENCY);
 
         #[cfg(not(test))]
-        let (state, maintenance_frequency) = (PoolState::New, MAINTENACE_FREQUENCY);
+        let (mut state, maintenance_frequency) = (PoolState::New, MAINTENACE_FREQUENCY);
+
+        if options
```
patrickfreed, line 212: we can use `is_load_balanced` here to avoid having to go through the options again.

patrickfreed (commented):
One tiny nitpick otherwise LGTM!

src/cmap/worker.rs
abr-egn: Done.

patrickfreed (approved):

PR #416, opened 2021-08-04
RUST-944 Integration tests for `observeSensitiveCommands`

abr-egn:
RUST-944

This pulls in the new integration tests for `observeSensitiveCommands`; it also updates the implementation of that flag because it turned out to be broken in exactly the way that spurred those tests.


src/test/spec/unified_runner/entity.rs
```diff
@@ -43,28 +45,26 @@ impl ClientEntity {
...
+            observe_sensitive_commands,
         }
     }
 
     /// Gets a list of all of the events of the requested event types that occurred on this client.
     /// Ignores any event with a name in the ignore list. Also ignores all configureFailPoint
     /// events.
-    pub fn get_filtered_events(
-        &self,
-        observe_events: &Option<Vec<String>>,
```
abr-egn, line 60: I don't know why these were being passed as parameters, since the values at the one call site were just referencing the `ClientEntity`'s own state, so I updated this while I was here.

isabelatkinson (approved):

patrickfreed (approved):

PR #419, opened 2021-08-09
RUST-913 Add versioned API migration example for `count`

abr-egn:


src/test/documentation_examples.rs
```diff
@@ -1414,9 +1422,74 @@ async fn versioned_api_examples() -> Result<()> {
...
+        doc! { "_id" : 2, "item" : "jkl", "price" : 20, "quantity" : 1, "date" : iso_date("2021-02-03T09:00:00Z")? },
+        doc! { "_id" : 4, "item" : "abc", "price" : 10, "quantity" : 10, "date" : iso_date("2021-02-15T08:00:00Z")? },
+        doc! { "_id" : 5, "item" : "xyz", "price" : 5, "quantity" : 10, "date" : iso_date("2021-02-15T09:05:00Z")? },
+        doc! { "_id" : 6, "item" : "xyz", "price" : 5, "quantity" : 5, "date" : iso_date("2021-02-15T12:05:10Z")? },
+        doc! { "_id" : 7, "item" : "xyz", "price" : 5, "quantity" : 10, "date" : iso_date("2021-02-15T14:12:12Z")? },
+        doc! { "_id" : 8, "item" : "abc", "price" : 10, "quantity" : 5, "date" : iso_date("2021-03-16T20:20:13Z")? }
+    ], None).await?;
+    // End Versioned API Example 5
+
+    let result = db
```
patrickfreed, line 1447: I think we need to include this bit in example 6 to demonstrate how `count` can be run in the driver.

src/test/documentation_examples.rs
```diff
@@ -1414,9 +1422,74 @@ async fn versioned_api_examples() -> Result<()> {
...
+        .run_command(
+            doc! {
+                "count": "sales"
+            },
+            None,
+        )
+        .await;
+    // Start Versioned API Example 6
+    let err = CommandError {
```
patrickfreed, line 1456: rather than creating an error, in Swift we opted to print the result and add a comment indicating what would be printed. I think that would be a bit more natural for an example. https://github.com/mongodb/mongo-swift-driver/blob/main/Examples/Docs/Sources/AsyncExamples/main.swift#L316

src/test/documentation_examples.rs
```diff
@@ -1414,9 +1422,74 @@ async fn versioned_api_examples() -> Result<()> {
...
+        code: 323,
+        code_name: "APIStrictError".to_string(),
+    };
+    // End Versioned API Example 6
+    let got_err = if let ErrorKind::Command(ref e) = *result.as_ref().unwrap_err().kind {
+        e
+    } else {
+        panic!("invalid result {:?}", result);
+    };
+    assert_eq!(err, *got_err);
```
patrickfreed, line 1467: rather than using eq here (which impacts the public API), I think it's sufficient to compare code and codename.

src/test/documentation_examples.rs
```diff
@@ -1414,9 +1422,74 @@ async fn versioned_api_examples() -> Result<()> {
...
+        .version(ServerApiVersion::V1)
+        .strict(true)
+        .build();
+    options.server_api = Some(server_api);
+    let client = Client::with_options(options)?;
+    let db = client.database("versioned-api-migration-examples");
+    db.collection::<Document>("sales").drop(None).await?;
+    // Start Versioned API Example 5
+    db.collection("sales").insert_many(vec![
+        doc! { "_id" : 1, "item" : "abc", "price" : 10, "quantity" : 2, "date" : iso_date("2021-01-01T08:00:00Z")? },
```
patrickfreed, line 1436: ideally this would be copy-pastable for users, so I think we want leverage the `chrono` interop here and do:
```rust
chrono::DateTime::parse_from_rfc3339("...")?
```
Since this is feature gated, we'll need to add a comment at the top of the example indicating that users need to enable the `bson-chrono-0_4` feature flag.

You might have to do some hackery to get the `?` operator to work, maybe we could update this function to return `Result<(), Box<dyn Error>>` or something to ease that.

src/test/documentation_examples.rs
abr-egn: Ah, I was modeling this from [Python's implementation](https://github.com/mongodb/mongo-python-driver/blob/master/test/test_examples.py#L1169), which uses runCommand, but doesn't include it.  Updated.

src/test/documentation_examples.rs
abr-egn: Done; opted to only print `Error::kind`, since `tags` isn't public.

src/test/documentation_examples.rs
abr-egn: Done.

src/test/documentation_examples.rs
abr-egn: `bson-chrono-0_4` isn't enabled by default*, so doing it this way required adding `bson` with that feature as a `dev-dependency` (and also updating to `resolver = "2"` so the features didn't get merged into the regular build).  It's a little unfortunate to have to update the `bson` version in two places; the other option would be to keep the conversion function (which can be made much shorter thanks to `chrono::DateTime::parse_from_rfc3339`) and pull it into the example with a note that it wouldn't be needed if `bson-chrono-0_4` was enabled.

\*  Side note: why not? `chrono`'s pulled in either way.

abr-egn: Correction: `resolver = "2"` isn't in our MSRV, so I switched to the other option.

src/test/documentation_examples.rs
```diff
@@ -1414,6 +1422,75 @@ async fn versioned_api_examples() -> Result<()> {
...
+        .version(ServerApiVersion::V1)
+        .strict(true)
+        .build();
+    options.server_api = Some(server_api);
+    let client = Client::with_options(options)?;
+    let db = client.database("versioned-api-migration-examples");
+    db.collection::<Document>("sales").drop(None).await?;
+
+    use std::{error::Error, result::Result};
+    // Start Versioned API Example 5
```
isabelatkinson, line 1436: It might be more useful to the docs team if we number these examples Count Migration 1-4 to stay in line with the ticket description.

abr-egn: There was a [comment](https://jira.mongodb.org/browse/DRIVERS-1846?focusedCommentId=3950144&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3950144) in the drivers ticket that these should be "Versioned API Example 5" and onwards, did I miss a convention for them to be elsewise?

isabelatkinson: Nope, I missed that comment!

src/test/documentation_examples.rs
patrickfreed: Ah yeah that is a bit tricky. I think your solution of including the conversion function in the example is a good choice for now. It made me realize that we probably want to have a `parse_from_rfc3339` method on `bson::DateTime` even without the feature flag, so I filed RUST-977 for that.

>Side note: why not? chrono's pulled in either way.

The feature flag is to protect us from instability in `chrono`'s API rather than from it as a dependency. See https://github.com/mongodb/bson-rust/pull/260#pullrequestreview-676430020 and the [C-STABLE](https://rust-lang.github.io/api-guidelines/necessities.html#c-stable) Rust API guideline for more detail.

patrickfreed (approved):
LGTM!

isabelatkinson (approved):
sorry i didn't realize i hadn't approved this one already!

PR #420, opened 2021-08-09
RUST-947 Implement Index Management API

NBSquare:
Implemented the asynchronous version of the Index management API. Summary of changes:

- Added `src/index` directory containing generic models for indexes.
- Added methods to `Collection` in order to support Index CRUD.
- Implemented a number of unit tests verifying behavior of operations and of the API.

Thanks to Stargateur for starting the work on this project! Next up is sync API and session support.


Cargo.toml
```diff
@@ -67,6 +67,10 @@ version_check = "0.9.1"
 webpki = "0.21.0"
 webpki-roots = "0.21.0"
 
+[dependencies.either]
```
isabelatkinson, line 70: Does this get used anywhere?

src/coll/mod.rs
```diff
@@ -313,6 +320,55 @@ impl<T> Collection<T> {
...
+        index: IndexModel,
+        options: impl Into<Option<CreateIndexesOptions>>,
+    ) -> Result<String> {
+        let mut result = self
+            .create_indexes_common(vec![index], options, None)
+            .await?;
+        if result.len() != 1 {
+            return Err(ErrorKind::InvalidResponse {
+                message: format!(
+                    "Expected exactly one index to be created, received {:?}",
```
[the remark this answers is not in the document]
NBSquare: You know, I added this as a sanity check but I don't think it's possible for this error to be returned. I'll remove it.

src/index/mod.rs
```diff
@@ -0,0 +1,50 @@
+pub mod options;
+
+use crate::bson::Document;
+
+use self::options::*;
+use serde::{Deserialize, Serialize};
+
+use typed_builder::TypedBuilder;
+
+/// Specifies the options to a
```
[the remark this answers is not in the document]
NBSquare: The doc comment on `IndexOptions` is correct but the doc on `IndexModel` itself is wrong. Fixing.

src/index/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+    #[serde(rename = "key")]
+    keys: Document,
+
+    #[serde(flatten)]
+    options: Option<IndexOptions>,
+}
+
+impl IndexModel {
+    /// If the client did not specify a name, generate and set it. Otherwise, do nothing.
+    pub(crate) fn update_name(&mut self) {
```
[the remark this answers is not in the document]
NBSquare: So by default collections have an index with key `{_id: 1}` which has the name `_id_`, which is what the spec is referring to. This is created along with the collection, so there's no need for the driver to handle this.

Of course, you could create an index which has `_id` as a key along with something else, but from what I can tell, expected behavior is to name that using the normal logic. And this is what I saw going on in some of the other drivers, too.

src/index/options.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub unique: Option<bool>,
+
+    /// Specify the version number of the index.
+    /// Starting in MongoDB 3.2, Version 0 indexes are not allowed.
+    #[serde(rename = "v")]
+    pub version: Option<IndexVersion>,
+
+    /// For text indexes, the language that determines the list of stop words and the
+    /// rules for the stemmer and tokenizer.
+    #[serde(rename = "default_language")]
```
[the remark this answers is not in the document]
NBSquare: This is to override the `camelCase` rename. Is there a better way to do that?

src/operation/create_indexes/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl Operation for CreateIndexes {
+    type O = Vec<String>;
+    type Command = Document;
+    type Response = CommandResponse<EmptyBody>;
+    const NAME: &'static str = "createIndexes";
+
+    fn build(&mut self, _description: &StreamDescription) -> Result<Command> {
+        self.indexes.iter_mut().for_each(|i| i.update_name()); // Generate names for unnamed indexes.
```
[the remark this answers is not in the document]
NBSquare: `createIndexes` command requires the name field to be specified 😞 
```
> db.runCommand({createIndexes:"foo", indexes:[{key: {x:1}}]})
{
	"ok" : 0,
	"errmsg" : "Error in specification { key: { x: 1.0 } } :: caused by :: The 'name' field is a required property of an index specification",
	"code" : 9,
	"codeName" : "FailedToParse"
}
```

Cargo.toml
NBSquare: Nope, good catch.

src/coll/mod.rs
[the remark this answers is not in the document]
isabelatkinson: cool, can we also add a quick comment above the `unwrap` explaining why it's safe?

src/index/mod.rs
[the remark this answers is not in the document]
NBSquare: Actually, duplicate indexes is one of the places where the spec falls short. If a user creates a duplicate index (i.e. the same keys as an index on the collection), the operation succeeds, but the server won't modify any of the collections.

However, in a situation where a user creates several indexes, and _some_ of them are duplicates, the server doesn't really inform the client _which_ of the indexes were created and which were duplicates.
```
{
        "createdCollectionAutomatically" : false,
        "numIndexesBefore" : 3,
        "numIndexesAfter" : 4,
        "note" : "index already exists",
        "ok" : 1
}
```
So, the duplicate name issue is a more insidious problem. If I have an index `foo` with key `{x: 1}` on a collection, and I try `create_index` with index `bar` with the same key, the collection will *keep the index `foo`, but tell me that my operation succeded*. So I'd still get back the wrong name.

Again, the server does tell me whether any indexes already existed on the collection, but not which ones, so it's completely unhelpful in a case where I'm creating a bunch of indexes.

The spec doesn't really talk about this either. IMO the best approach here is to log the message somewhere, or maybe return some other string in this situation.

src/operation/create_indexes/mod.rs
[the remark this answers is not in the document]
NBSquare: Agreed that it's not super clear, and doesn't really explain what's going on with `_id` at all. I can update that along with my other DRIVERS-side work.

src/test/coll.rs
```diff
@@ -1105,6 +1123,77 @@ async fn index_management_creates() {
     assert_eq!(names, vec!["_id_", "a_1_b_-1", "c_1", "customname"]);
 }
 
+// Test that creating a duplicate index orks as expected.
```
isabelatkinson, line 1126:
```suggestion
// Test that creating a duplicate index works as expected.
```

src/coll/mod.rs
```diff
@@ -313,6 +327,46 @@ impl<T> Collection<T> {
...
+        index: IndexModel,
+        options: impl Into<Option<CreateIndexOptions>>,
+    ) -> Result<CreateIndexResult> {
+        let response = self
+            .create_indexes_common(vec![index], options, None)
+            .await?;
+        Ok(response.into())
+    }
+
+    /// Create several indexes.
```
isabelatkinson, line 361:
```suggestion
    /// Creates the given indexes on this collection.
```

src/coll/mod.rs
```diff
@@ -313,6 +327,46 @@ impl<T> Collection<T> {
...
+
+        let indexes: Vec<IndexModel> = indexes.into_iter().collect();
+
+        let create_indexes = CreateIndexes::new(self.namespace(), indexes, options);
+        self.client()
+            .execute_operation(create_indexes, session)
+            .await
+    }
+
+    /// Convenience method for creating a single index.
```
isabelatkinson, line 349:
```suggestion
    /// Creates the given index on this collection.
```

src/coll/mod.rs
```diff
@@ -423,6 +477,81 @@ impl<T> Collection<T> {
...
+        resolve_write_concern_with_session!(self, options, session.as_ref())?;
+
+        // If there is no provided name, that means we should drop all indexes.
+        let index_name = name.into().unwrap_or("*").to_string();
+
+        let drop_index = DropIndexes::new(self.namespace(), index_name, options);
+        self.client().execute_operation(drop_index, session).await
+    }
+
+    /// Drop the index specified by `name` from the collection.
```
isabelatkinson, line 498:
```suggestion
    /// Drops the index specified by `name` from this collection.
```

src/coll/mod.rs
```diff
@@ -423,6 +477,81 @@ impl<T> Collection<T> {
...
+                message: "Cannot pass name \"*\" to drop_index since more than one index would be \
+                          dropped."
+                    .to_string(),
+            }
+            .into());
+        }
+        self.drop_index_common(name, options, None).await
+    }
+
+    /// Drop all indexes associated with the collection.
```
isabelatkinson, line 516:
```suggestion
    /// Drops all indexes associated with this collection.
```

src/coll/mod.rs
```diff
@@ -423,6 +477,81 @@ impl<T> Collection<T> {
...
+    pub async fn drop_indexes(&self, options: impl Into<Option<DropIndexOptions>>) -> Result<()> {
+        self.drop_index_common(None, options, None).await
+    }
+
+    /// List all indexes on the collection.
```
isabelatkinson, line 521:
```suggestion
    /// Lists all indexes on this collection.
```

src/coll/mod.rs
```diff
@@ -423,6 +477,81 @@ impl<T> Collection<T> {
...
+        &self,
+        cursor: impl TryStreamExt<Ok = IndexModel, Error = Error>,
+    ) -> Result<Vec<String>> {
+        cursor
+            .try_filter_map(|index| future::ok(index.get_name()))
+            .try_collect()
+            .await
+    }
+
+    /// Gets the names of all indexes on the collection.
```
isabelatkinson, line 544:
```suggestion
    /// Gets the names of all indexes on this collection.
```

src/operation/list_indexes/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+            ns: Namespace {
+                db: String::new(),
+                coll: String::new(),
+            },
+            options: None,
+        }
+    }
+}
+
+impl Operation for ListIndexes {
```
isabelatkinson, line 39: According to the [retryable reads spec](https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.rst#supported-read-operations), this operation is read-retryable, so we should implement the `retryability` method here to return `Retryability::Read`. This also reminded me that we're currently skipping index management operations in the v2 and unified runners, so we'll need to implement those operations. We don't need to do that as part of this PR, though; I filed RUST-987 to do as add-on work.

NBSquare: Sounds good, I also overrode `Operation::selection_criteria` since the [enumerate indexes spec](https://github.com/mongodb/specifications/blob/master/source/enumerate-indexes.rst#replica-sets) says

> Drivers MUST run `listIndexes` on the primary node when in a replica set topology, unless directly connected to a secondary node in Single topology.

src/index/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct IndexModel {
+    /// Specifies the index’s fields. For each field, specify a key-value pair in which the key is
+    /// the name of the field to index and the value is index type.
+    #[serde(rename = "key")]
+    keys: Document,
+
+    /// The options for the index.
+    #[serde(flatten)]
```
abr-egn, line 22: I'm curious why `IndexOptions` are split out as their own struct; is it ever useful to be dealing with the options distinct from the index keys?

src/operation/create_indexes/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+impl<'de> Deserialize<'de> for Response {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct ResponseBody {
```
abr-egn, line 125: `ResponseBody` and `Response` look to be identical - is there a need for the former?

NBSquare: It's necessary because the response after running `createIndexes` depends on the toplogy.
For example, running `createIndexes` on a single toplogy might yield
```
{ numIndexesBefore: 1, numIndexesAfter: 2 }
```
but running on a sharded topology could yield 
```
{ raw: { mongodb://localhost:27017/: { numIndexesBefore: 1, numIndexesAfter: 2 } } }
```
Because the internals of the response are the same, we have to do a custom deserialization in order to account for the wrapper around the "relevant" part of the response.

src/index/mod.rs
NBSquare: I think it's just a convenience - `keys` is the only part of the command which the driver strictly requires. To be honest with you I don't love it because any index returned from the server will include the `name` option, so `IndexOptions` will never be `None` in practice (except if the user is creating an index with no name specified).

But, the `keys` / `index_options` pattern is what the specification requires so I'd argue for keeping it at the end of the day

src/operation/list_indexes/mod.rs
isabelatkinson: I think you accidentally added it to `CreateIndexes` -- that one isn't retryable

src/index/options.rs
```diff
@@ -0,0 +1,227 @@
...
+    ///
+    /// See the [documentation](https://docs.mongodb.com/manual/core/index-sparse/)
+    /// for more information on how to use this option.
+    pub sparse: Option<bool>,
+
+    /// Allows users to configure the storage engine on a per-index basis when creating
+    /// an index.
+    pub storage_engine: Option<Document>,
+
+    /// Forces the index to be unique. The default value is false.
```
isabelatkinson, line 51: Can we update this to better describe what "unique" means in this context? Specifically that the collection will not accept duplicate index keys

src/test/coll.rs
```diff
@@ -1049,3 +1053,258 @@ async fn collection_generic_bounds() {
         .collection(function_name!());
     let _result = coll.insert_one(Bar {}, None).await;
 }
+
```
isabelatkinson, line 1056: In addition to these tests we should also make sure we're implementing the [prose tests](https://github.com/mongodb/specifications/blob/master/source/enumerate-indexes.rst#tests) described at the end of the enumerate indexes spec; these tests already encompass some of what's described but we should make sure we're covering all of those cases.

isabelatkinson (commented):
looks good! just a few more comments

src/operation/create_indexes/mod.rs
abr-egn: Oh, I see, thank you - I'd missed the need for two different implementations of `Deserialize`.  Could also achieve that with a newtype, but that's not a must.

abr-egn (approved):

src/test/coll.rs
NBSquare: Okay - I had created a ticket for this but I think it makes sense to do this on this PR for the sake of correctness, so I'll take care of that.

isabelatkinson: oh didn't realize there was already a ticket for it, but feel free to do here if you think that's the right move

NBSquare: After looking into it closely it seems like it's really a 15 line modification to the current prose testing, since I had handled almost everything in `index_management_lists`. Gonna make that change and merge that here.

isabelatkinson (approved):
lgtm! great job with this!

abr-egn (approved):

src/operation/create_indexes/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        match ResponseHelper::deserialize(deserializer)? {
+            ResponseHelper::PlainResponse(body) => Ok(Response(body)),
+            ResponseHelper::ShardedResponse { raw } => {
+                let len = raw.values().count();
+                if len != 1 {
+                    return Err(Error::invalid_length(len, &"a single result"));
+                }
+
+                let v = raw.values().last().unwrap(); // Safe unwrap because of length check above.
```
isabelatkinson, line 134: to avoid needing to clone `v` here we can use the `into_iter` method on `Document`, i.e.

`let (_, v) = raw.into_iter().next().unwrap();`

isabelatkinson (approved):
looks good! one small comment but I don't need to re-review after that change

PR #423, opened 2021-08-12
RUST-926 Allow direct connections to hidden nodes

patrickfreed:
RUST-926

This PR fixes a bug where the driver would always time out trying to determine if sessions were supported when directly connected to a hidden node.

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -574,6 +574,17 @@ impl Client {
         }
     }
 
+    async fn select_data_bearing_server(&self) -> Result<()> {
+        let topology_type = self.inner.topology.topology_type().await;
+        let criteria = SelectionCriteria::Predicate(Arc::new(move |server_info| {
+            let server_type = server_info.server_type();
+            (matches!(topology_type, TopologyType::Single) && server_type.is_available())
+                || server_type.is_data_bearing()
+        }));
+        let _: SelectedServer = self.select_server(Some(&criteria)).await?;
+        Ok(())
+    }
+
     /// Gets whether the topology supports sessions, and if so, returns the topology's logical
     /// session timeout. If it has yet to be determined if the topology supports sessions, this
     /// method will perform a server selection that will force that determination to be made.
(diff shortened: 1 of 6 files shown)


src/client/executor.rs
```diff
@@ -583,8 +584,11 @@ impl Client {
         // sessions are supported or not.
         match initial_status {
             SessionSupportStatus::Undetermined => {
+                let topology_type = self.inner.topology.topology_type().await;
                 let criteria = SelectionCriteria::Predicate(Arc::new(move |server_info| {
-                    server_info.server_type().is_data_bearing()
+                    let server_type = server_info.server_type();
+                    (matches!(topology_type, TopologyType::Single) && server_type.is_available())
```
patrickfreed, line 590: for direct connections and standalones, just try to select any server. for all other topologies, wait for a data bearing server.

src/sdam/description/topology/mod.rs
```diff
@@ -272,6 +272,11 @@ impl TopologyDescription {
     /// Updates the topology's logical session timeout value based on the server's value for it.
     fn update_session_support_status(&mut self, server_description: &ServerDescription) {
         if !server_description.server_type.is_data_bearing() {
+            if let TopologyType::Single = self.topology_type {
+                self.session_support_status = SessionSupportStatus::Unsupported {
```
patrickfreed, line 276: if the only server in the topology isn't data bearing, just mark it as not supporting sessions.

daleione:
The document of  `src/client/options/mod.rs#L373` is still direct=true.
It's actually `directConnection`.
Please remember to fix it.

patrickfreed:
@daleione fixed, thanks for the reminder!

abr-egn (approved):

src/client/executor.rs
isabelatkinson: We use this same logic in `transaction_support_status` directly below, so if a user is trying to execute a transaction on a direct connection to a hidden node I believe they'll encounter the same problem (although this is a much more niche use case). Can we make these changes for `TransactionSupportStatus` as well?

patrickfreed: Good catch, fixed.

isabelatkinson (approved):

daleione:
with this commit, query with hide node is also failed.
New error message is 
`Error: Error { kind: Command(CommandError { code: 13, code_name: "Unauthorized", message: "command find requires authentication" }), labels: {} }`.

The db link work well with mongosh. Please check, if there something missing to fix.

PR #428, opened 2021-08-18
RUST-626 Run benchmarks in Evergreen

patrickfreed:
RUST-626

This PR adds an Evergreen config that will be used by the `mongo-rust-driver-benchmarking` project to run benchmarks and record the data using https://github.com/10gen/signal-processing-service. Some updates to the benchmark runner were required to properly calculate the various composite scores.

### DIFF
--- a/.evergreen/benchmarks.yml
+++ b/.evergreen/benchmarks.yml
@@ -13,8 +13,10 @@ stepback: true
 command_type: system
 
 # Protect ourself against rogue test case, or curl gone wild, that runs forever
-# 12 minutes is the longest we'll ever run
-exec_timeout_secs: 3600 # 12 minutes is the longest we'll ever run
+# 60 minutes is the longest we'll ever run
+exec_timeout_secs: 3600
+
+pre_error_fails_task: true
 
 # What to do when evergreen hits the timeout (`post:` tasks are run automatically)
 timeout:
@@ -173,7 +175,7 @@ functions:
         working_dir: "src"
         script: |
           ${PREPARE_SHELL}
-          ASYNC_RUNTIME=${ASYNC_RUNTIME} RUST_VERSION=${RUST_VERSION} .evergreen/compile-only.sh
+          ASYNC_RUNTIME=${ASYNC_RUNTIME} .evergreen/run-compile-benchmarks.sh
 
   "upload-mo-artifacts":
     - command: shell.exec
@@ -273,8 +275,6 @@ pre:
 
 post:
   - func: "stop mongo orchestration"
-  - func: "upload-mo-artifacts"
-  - func: "upload benchmark results"
   - func: "cleanup"
 
 tasks:
@@ -286,6 +286,7 @@ tasks:
           MONGODB_VERSION: "4.0"
           TOPOLOGY: "server"
       - func: "run driver benchmarks"
+      - func: "upload benchmark results"
 
   - name: "benchmark-4.0-replica_set"
     tags: ["4.0", "replica_set"]
@@ -295,6 +296,7 @@ tasks:
           MONGODB_VERSION: "4.0"
           TOPOLOGY: "replica_set"
       - func: "run driver benchmarks"
+      - func: "upload benchmark results"
 
   - name: "benchmark-4.0-sharded_cluster"
     tags: ["4.0", "sharded_cluster"]
@@ -304,6 +306,7 @@ tasks:
           MONGODB_VERSION: "4.0"
           TOPOLOGY: "sharded_cluster"
       - func: "run driver benchmarks"
+      - func: "upload benchmark results"
 
   - name: "benchmark-5.0-standalone"
     tags: ["5.0", "standalone"]
@@ -313,6 +316,7 @@ tasks:
           MONGODB_VERSION: "5.0"
           TOPOLOGY: "server"
       - func: "run driver benchmarks"
+      - func: "upload benchmark results"
 
   - name: "benchmark-5.0-replica_set"
     tags: ["5.0", "replica_set"]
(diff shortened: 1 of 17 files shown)


.evergreen/benchmarks.yml
```diff
@@ -0,0 +1,457 @@
...
+  - name: "benchmark-compile"
+    commands:
+      - func: "compile only"
+      - func: "upload benchmark results"
+
+axes:
+  - id: "mongodb-version"
+    display_name: MongoDB Version
+    values:
+      - id: "5.0"
```
patrickfreed, line 353: I opted for 5.0 and 4.2 instead of 5.0 and 3.6 as the design recommended to avoid having to deal with the involved database dropping mechanism required on sharded clusters. see: https://docs.mongodb.com/manual/reference/method/db.dropDatabase/#replica-set-and-sharded-clusters for more info

.evergreen/benchmarks.yml
```diff
@@ -0,0 +1,457 @@
...
+      - id: "async-std"
+        display_name: "async-std"
+        variables:
+          ASYNC_RUNTIME: "async-std"
+
+  # Note that drivers-evergreen-tools expects `SSL` as the environmental
+  # variable, not `TLS`, so we have to use that for the actual value used in the
+  # script; we use `TLS` for the metadata that isn't used by the actual shell
+  # scripts.
+  - id: "auth-and-tls"
```
patrickfreed, line 394: I enabled auth and TLS for the benchmarks since that's how our users will typically use the driver and so we'll want benchmarking coverage for our implementations of those too.

benchmarks/src/bench/json_multi_export.rs
```diff
@@ -52,24 +61,28 @@ impl Benchmark for JsonMultiExportBenchmark {
...
-        let mut tasks = FuturesUnordered::new();
+        let mut tasks = Vec::new();
 
         for i in 0..TOTAL_FILES {
-            let coll_ref = self.coll.clone();
+            let coll_ref = self.coll.clone_with_type::<Tweet>();
             let path = std::env::temp_dir();
 
-            tasks.push(async move {
+            tasks.push(crate::spawn(async move {
```
patrickfreed, line 85: these weren't actually running in parallel since it was the same task moving them all forward. This has been updated to spawn a new task for each worker, which greatly sped up this benchmark.

benchmarks/src/bench/mod.rs
```diff
@@ -124,3 +130,34 @@ pub async fn run_benchmark<B: Benchmark + Send + Sync>(
     test_durations.sort();
     Ok(test_durations)
 }
+
+pub async fn drop_database(uri: &str, database: &str) -> Result<()> {
```
patrickfreed, line 134: this was needed to avoid running into `NamespaceNotFound` issues in the benchmarks on sharded clusters. See the link in the comment below for more info.

benchmarks/src/main.rs
```diff
@@ -15,106 +14,148 @@ macro_rules! spawn_blocking_and_await {
...
+    },
+    fs::File,
+    score::{score_test, BenchmarkResult, CompositeScore},
 };
 
 lazy_static! {
     static ref DATA_PATH: PathBuf = Path::new(env!("CARGO_MANIFEST_DIR")).join("data");
 }
 
-fn get_nth_percentile(durations: &[Duration], n: f64) -> Duration {
```
patrickfreed, line 46: this deleted stuff was moved to `score.rs`

benchmarks/src/main.rs
```diff
@@ -15,106 +14,148 @@ macro_rules! spawn_blocking_and_await {
...
+const SMALL_DOC_INSERT_ONE_BENCH: &'static str = "Small doc insertOne";
+const LARGE_DOC_INSERT_ONE_BENCH: &'static str = "Large doc insertOne";
+const SMALL_DOC_BULK_INSERT_BENCH: &'static str = "Small doc bulk insert";
+const LARGE_DOC_BULK_INSERT_BENCH: &'static str = "Large doc bulk insert";
+const GRIDFS_UPLOAD_BENCH: &'static str = "GridFS upload";
+const LDJSON_MULTI_IMPORT_BENCH: &'static str = "LDJSON multi-file import";
+const GRIDFS_MULTI_UPLOAD_BENCH: &'static str = "GridFS multi-file upload";
+
+/// Benchmarks included in the "BSONBench" composite.
+const BSON_BENCHES: &[&'static str] = &[
```
patrickfreed, line 90: the breakdown of composite scores that we were using wasn't fully accurate, so I sorted that out here with these constants.

benchmarks/src/score.rs
```diff
@@ -0,0 +1,206 @@
+use std::time::Duration;
```
patrickfreed, line 1: We needed some more abstract types than integers to be able to support the scoring hierarchy required by the spec, so I introduced a few here. We also needed types to model the JSON format that signal-processing expects as well.

.evergreen/benchmarks.yml
```diff
@@ -0,0 +1,456 @@
...
+  - id: "os"
+    display_name: OS
+    values:
+      - id: ubuntu-18.04
+        display_name: "Ubuntu 18.04"
+        run_on: ubuntu1804-test
+        variables:
+          PYTHON: "/opt/mongodbtoolchain/v3/bin/python"
+          VENV_BIN_DIR: "bin"
+      # TODO: RUST-990 re-enable macOS benchmarks
```
patrickfreed, line 412: I opted to just skip macOS benchmarks for now since the timeouts don't seem to be exclusive to the Rust driver (e.g. see the Swift waterfall) and that macOS performance monitoring isn't as high of a priority as the other platforms.

PR #430, opened 2021-08-20
RUST-958/RUST-957 Implement index management on sessions and sync

NBSquare:
As in title. Split into two commits implementing sessions and sync, respectively.


abr-egn (approved):
LGTM!

src/coll/mod.rs
```diff
@@ -543,15 +614,19 @@ impl<T> Collection<T> {
...
-        let client = self.client();
-        let cursor: Cursor<IndexModel> = client
-            .execute_cursor_operation(list_indexes)
-            .await
-            .map(|(spec, session)| Cursor::new(client.clone(), spec, session))?;
+        let cursor = self.list_indexes(None).await?;
         self.list_index_names_common(cursor).await
     }
 
+    /// Gets the names of all indexes on the collection using an explicit session.
```
isabelatkinson, line 621:
```suggestion
    /// Gets the names of all indexes on the collection using the provided `ClientSession`.
```

src/sync/coll.rs
```diff
@@ -210,6 +221,52 @@ impl<T> Collection<T> {
         ))
     }
 
+    /// Convenience method for creating a single index.
```
isabelatkinson, line 224: Can we update the comments on these methods to match the ones in the async API?

isabelatkinson (commented):
looks good! just a few documentation requests

isabelatkinson:
Oh one thing I just noticed as well -- we need to mark the `CreateIndexResult` and `CreateIndexesResult` structs `#[non_exhaustive]` in case we want to add fields to them in the future. We can either just add that in here or file a separate ticket for it.

isabelatkinson (approved):

PR #441, opened 2021-08-27
RUST-987 Add index management operations to test runners

isabelatkinson:


src/coll/mod.rs
```diff
@@ -578,7 +578,7 @@ impl<T> Collection<T> {
     /// Lists all indexes on this collection.
     pub async fn list_indexes(
         &self,
-        options: impl Into<Option<ListIndexOptions>>,
+        options: impl Into<Option<ListIndexesOptions>>,
```
isabelatkinson, line 581: Renamed this options struct to be consistent with other names.

src/test/spec/v2_runner/operation.rs
```diff
@@ -1327,6 +1355,178 @@ impl TestOperation for AssertCollectionNotExists {
     }
 }
 
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(super) struct CreateIndex {
+    keys: Document,
+    name: Option<String>,
```
isabelatkinson, line 1362: Normally for test operations we add the top-level arguments as fields and then flatten the rest into an options struct, but that isn't feasible here because we have two possible options structs: the options for the `IndexModel` and the options for the operation. Instead of trying to parse all of that out I just included the two fields that are actually used by tests and added `deny_unknown_fields` so that we can update this if new spec tests include other fields.

patrickfreed (approved):

abr-egn (approved):

PR #442, opened 2021-08-30
RUST-994 Only return index names from CreateIndex(es)Result

isabelatkinson:


src/operation/create_indexes/mod.rs
```diff
@@ -77,19 +71,11 @@ impl Operation for CreateIndexes {
 
     fn handle_response(
         &self,
-        response: Response,
+        _response: WriteConcernOnlyBody,
```
patrickfreed, line 74: We'll need a call to `response.validate()` in this method somewhere to check for write concern errors.

abr-egn (approved):
LGTM  modulo Patrick's comment.

patrickfreed (approved):

PR #443, opened 2021-08-30
RUST-991 Perform wire check when commitQuorum is provided to createIndexes

isabelatkinson:


src/test/coll.rs
```diff
@@ -1050,282 +1049,3 @@ async fn collection_generic_bounds() {
         .collection(function_name!());
     let _result = coll.insert_one(Bar {}, None).await;
 }
-
```
isabelatkinson, line 1053: Moved these tests into their own file.

src/test/index_management.rs
```diff
@@ -0,0 +1,304 @@
...
+    coll.drop_indexes(None)
+        .await
+        .expect("Drop indexes op failed");
+    assert_eq!(client.get_command_started_events(&["dropIndexes"]).len(), 2);
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+#[function_name::named]
+async fn commit_quorum_error() {
```
isabelatkinson, line 284: This is the only new test added to this file.

patrickfreed (approved):

abr-egn (approved):

src/lib.rs
```diff
@@ -50,14 +50,22 @@
 //!
```
patrickfreed, line 50: Ah we ran into this issue with rustfmt messing up the markdown table in the bson repo too. We'll need to add [`ignore = ["src/lib.rs"]`](https://github.com/mongodb/bson-rust/blob/master/rustfmt.toml#L10) to our `rustfmt.toml` to prevent this. As part of that, can you revert the formatting changes to this file?

PR #446, opened 2021-08-30
RUST-954 Pin cursor connections in load balancer mode

abr-egn:
RUST-954

This implements the cursor portion of connection pinning; transactions and checkout type tracking will come in future PRs.

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -7,7 +7,7 @@ use std::{collections::HashSet, sync::Arc, time::Instant};
 use super::{session::TransactionState, Client, ClientSession};
 use crate::{
     bson::Document,
-    cmap::{conn::PinHandle, Connection, RawCommand, RawCommandResponse},
+    cmap::{conn::PinnedConnectionHandle, Connection, RawCommand, RawCommandResponse},
     cursor::{session::SessionCursor, Cursor, CursorSpecification},
     error::{
         Error,
@@ -74,14 +74,14 @@ impl Client {
     ) -> Result<T::O> {
         self.execute_operation_with_details(op, session)
             .await
-            .map(|details| details.output)
+            .map(|details| details.output.operation_output)
     }
 
     async fn execute_operation_with_details<T: Operation>(
         &self,
         op: T,
         session: impl Into<Option<&mut ClientSession>>,
-    ) -> Result<OperationDetails<T>> {
+    ) -> Result<ExecutionDetails<T>> {
         Box::pin(async {
             // TODO RUST-9: allow unacknowledged write concerns
             if !op.is_acknowledged() {
@@ -119,14 +119,13 @@ impl Client {
                     implicit_session.as_mut()
                 }
             };
-            let mut details = self.execute_operation_with_retry(op, session).await?;
-            details.implicit_session = implicit_session;
-            Ok(details)
+            let output = self.execute_operation_with_retry(op, session).await?;
+            Ok(ExecutionDetails { output, implicit_session })
         })
         .await
     }
 
-    /// Execute the given operation, returning cursor created by the operation.
+    /// Execute the given operation, returning the cursor created by the operation.
     ///
     /// Server selection be will performed using the criteria specified on the operation, if any.
     pub(crate) async fn execute_cursor_operation<Op, T>(&self, op: Op) -> Result<Cursor<T>>
@@ -136,10 +135,10 @@ impl Client {
     {
         Box::pin(async {
             let mut details = self.execute_operation_with_details(op, None).await?;
-            let pinned = self.pin_connection_for_cursor(&mut details)?;
+            let pinned = self.pin_connection_for_cursor(&mut details.output)?;
             Ok(Cursor::new(
                 self.clone(),
-                details.output,
+                details.output.operation_output,
                 details.implicit_session,
                 pinned,
             ))
@@ -157,19 +156,19 @@ impl Client {
         T: DeserializeOwned + Unpin + Send + Sync,
     {
         let mut details = self.execute_operation_with_details(op, session).await?;
-        let pinned = self.pin_connection_for_cursor(&mut details)?;
-        Ok(SessionCursor::new(self.clone(), details.output, pinned))
+        let pinned = self.pin_connection_for_cursor(&mut details.output)?;
+        Ok(SessionCursor::new(self.clone(), details.output.operation_output, pinned))
     }
 
     fn pin_connection_for_cursor<Op, T>(
         &self,
-        details: &mut OperationDetails<Op>,
-    ) -> Result<Option<PinHandle>>
+        details: &mut ExecutionOutput<Op>,
+    ) -> Result<Option<PinnedConnectionHandle>>
     where
         Op: Operation<O = CursorSpecification<T>>,
     {
         let is_load_balanced = self.inner.options.load_balanced.unwrap_or(false);
-        if is_load_balanced && details.output.info.id != 0 {
+        if is_load_balanced && details.operation_output.info.id != 0 {
             Ok(Some(details.connection.pin()?))
         } else {
             Ok(None)
@@ -182,7 +181,7 @@ impl Client {
         &self,
         mut op: T,
         mut session: Option<&mut ClientSession>,
-    ) -> Result<OperationDetails<T>> {
+    ) -> Result<ExecutionOutput<T>> {
         // If the current transaction has been committed/aborted and it is not being
         // re-committed/re-aborted, reset the transaction's state to TransactionState::None.
         if let Some(ref mut session) = session {
@@ -252,10 +251,9 @@ impl Client {
             )
             .await
         {
-            Ok(output) => Ok(OperationDetails {
-                output,
+            Ok(operation_output) => Ok(ExecutionOutput {
+                operation_output,
                 connection: conn,
-                implicit_session: None, // populated in execute_operation_with_details
             }),
             Err(mut err) => {
                 // Retryable writes are only supported by storage engines with document-level
@@ -302,7 +300,7 @@ impl Client {
         session: &mut Option<&mut ClientSession>,
         txn_number: Option<i64>,
         first_error: Error,
-    ) -> Result<OperationDetails<T>> {
+    ) -> Result<ExecutionOutput<T>> {
         op.update_for_retry();
 
         let server = match self.select_server(op.selection_criteria()).await {
@@ -329,10 +327,9 @@ impl Client {
             .execute_operation_on_connection(op, &mut conn, session, txn_number, &retryability)
             .await
         {
-            Ok(output) => Ok(OperationDetails {
-                output,
+            Ok(operation_output) => Ok(ExecutionOutput {
+                operation_output,
                 connection: conn,
-                implicit_session: None, // populated in execute_operation_with_details
             }),
             Err(err) => {
                 self.inner
@@ -825,8 +822,12 @@ struct CommandResult<T> {
     deserialized: T,
 }
 
-struct OperationDetails<T: Operation> {
-    output: T::O,
-    connection: Connection,
+struct ExecutionDetails<T: Operation> {
+    output: ExecutionOutput<T>,
     implicit_session: Option<ClientSession>,
+}
+
+struct ExecutionOutput<T: Operation> {
+    operation_output: T::O,
+    connection: Connection,
 }
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -210,8 +210,8 @@ impl Client {
         };
 
         let mut owned_conn = None;
-        let mut conn = match pinned_connection {
-            Some(ref mut l) => &mut **l,
+        let mut conn = match &mut pinned_connection {
+            Some(l) => &mut **l,
             None => match server.pool.check_out().await {
                 Ok(c) => {
                     owned_conn = Some(c);
@@ -312,9 +312,16 @@ impl Client {
             }
         };
 
-        let mut conn = match server.pool.check_out().await {
-            Ok(c) => c,
-            Err(_) => return Err(first_error),
+        let mut owned_conn = None;
+        let conn = match pinned_connection {
+            Some(c) => c,
+            None => match server.pool.check_out().await {
+                Ok(c) => {
+                    owned_conn = Some(c);
+                    owned_conn.as_mut().unwrap()
+                },
+                Err(_) => return Err(first_error),
+            }
         };
 
         let retryability = self.get_retryability(&conn, op, session).await?;
@@ -323,10 +330,10 @@ impl Client {
         }
 
         match self
-            .execute_operation_on_connection(op, &mut conn, session, txn_number, &retryability)
+            .execute_operation_on_connection(op, conn, session, txn_number, &retryability)
             .await
         {
-            Ok(result) => Ok((result, Some(conn))),
+            Ok(result) => Ok((result, owned_conn)),
             Err(err) => {
                 self.inner
                     .topology
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -7,7 +7,7 @@ use std::{collections::HashSet, sync::Arc, time::Instant};
 use super::{session::TransactionState, Client, ClientSession};
 use crate::{
     bson::Document,
-    cmap::{Connection, RawCommand, RawCommandResponse},
+    cmap::{Connection, conn::PinHandle, RawCommand, RawCommandResponse},
     cursor::{session::SessionCursor, Cursor, CursorSpecification},
     error::{
         Error,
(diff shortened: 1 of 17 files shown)


src/client/executor.rs
```diff
@@ -70,6 +72,26 @@ impl Client {
         op: T,
         session: impl Into<Option<&mut ClientSession>>,
     ) -> Result<T::O> {
+        self.execute_operation_pinned(op, session, None).await
+    }
+
+    pub(crate) async fn execute_operation_pinned<T: Operation>(
```
abr-egn, line 78: I'm not thrilled with `_pinned` (use passed-in connection) vs `_pinnable` (return checked-out connection), but this at least avoids the churn of updating all unrelated `execute_operation` calls.

src/cursor/common.rs
```diff
@@ -214,3 +249,43 @@ pub(crate) struct CursorInformation {
     pub(crate) batch_size: Option<u32>,
     pub(crate) max_time: Option<Duration>,
 }
+
+#[derive(Clone, Debug)]
+pub(crate) struct PinnedConnection {
+    connection: Option<Arc<Mutex<Connection>>>,
```
abr-egn, line 255: It's hypothetically possible that this doesn't need to be an `Arc<Mutex<_>>`, but could instead use a similar trick to the `Session` associated type in `GetMoreProviderResult`.  However, couldn't get that to line up well after a day or so of work - particular challenges were the manual polling and state management in the cursor lifecycle, the drop impls for the cursors, and not having lifetimes leak out into the public cursor api - so I went with this solution.

There is a runtime cost incurred by the reference counting and locking, but it's unlikely to be a significant issue in practice because the cursors buffer values returned, so there's not going to be a hot loop involving those operations.

src/cursor/common.rs
```diff
@@ -214,3 +249,43 @@ pub(crate) struct CursorInformation {
...
+
+    pub(super) async fn lock(&self) -> Option<MutexGuard<'_, Connection>> {
+        match &self.connection {
+            Some(c) => Some(c.lock().await),
+            None => None,
+        }
+    }
+}
+
+pub(super) fn kill_cursor(
```
abr-egn, line 275: I opted to make this a free function rather than a method on `PinnedConnection` because it doesn't feel like it belongs to the connection as such.

src/cursor/common.rs
patrickfreed: Yeah I agree this is probably fine, and it's simple which is really nice considering the complexity of the existing session stuff. The runtime cost from the locking at least will be insignificant I think since cursors themselves can't be used concurrently.

src/cursor/common.rs
```diff
@@ -214,3 +249,43 @@ pub(crate) struct CursorInformation {
     pub(crate) batch_size: Option<u32>,
     pub(crate) max_time: Option<Duration>,
 }
+
+#[derive(Clone, Debug)]
+pub(crate) struct PinnedConnection {
```
patrickfreed, line 254: What do you think about modeling this as an enum instead? It avoids certain cases like `connection: None` + `valid: false`, which is nice.

e.g.
```rust
enum PinnedConnection {
    ValidPinned(Arc<Mutex<Connection>>),
    InvalidPinned(Arc<Mutex<Connection>>),
    Unpinned
}
```

src/client/executor.rs
abr-egn: That cleans things up but runs into lifetime issues - with that signature it mutably borrows the `Operation` for the span that the returned connection is used, which conflicts with the rest of the execution.  `Operation` can be parameterized to give the return a `&'conn mut Connection` lifetime to disconnect the borrows, but that has substantial ripple effects on the rest of the driver code.  Sadly, default lifetimes never made it in, because `Operation<'conn = 'static>` would be ideal here.

abr-egn:
I've updated this to use the discussed pin-by-id setup; the key point this enables is that a connection can be pinned while mutably borrowed rather than just owned, which will be crucial for transactions.  As a side benefit, it also enables passing the pin via the `Operation` as Patrick suggested.

This comes at the cost of losing the potential performance improvements of directly owned connections - pinned will now have to incur the same channel round-trip as checked out - and some increased potential contention on the connection pool worker from handling pin takes / unpins.  I think the latter is unlikely to be substantial (or even really measurable above normal noise), but if it is it could probably be mitigated by moving the pinned connections to a different structure with its own worker task and channel.

src/cursor/common.rs
```diff
@@ -88,6 +103,18 @@ where
...
+                        if exhausted {
+                            // If the cursor is exhausted, the driver must return the pinned
+                            // connection to the pool.
+                            self.pinned_connection = PinnedConnection::new(None);
+                        }
+                        if let Err(e) = &result {
+                            if e.is_network_error() {
+                                // Flag the connection as invalid, preventing a killCursors command,
+                                // but leave the connection pinned.
+                                self.pinned_connection.invalidate();
```
patrickfreed, line 115: Once the connection is invalidated, we will probably want to return `Poll::Ready(None)` in all subsequent polls after returning the error.

We don't currently handle this properly in the regular code path either (see the `?` below without any additional handling). We can fix that if necessary in a separate piece of work though.

src/cursor/common.rs
```diff
@@ -214,3 +246,63 @@ pub(crate) struct CursorInformation {
...
+
+impl PinnedConnection {
+    pub(super) fn new(handle: Option<PinHandle>) -> Self {
+        match handle {
+            Some(h) => Self::Valid(h),
+            None => Self::Unpinned,
+        }
+    }
+
+    fn handle(&self) -> Option<&PinHandle> {
```
patrickfreed, line 265: This method might be a little bit dangerous to use (see my comment above about stopping after invalidation) since we may only want to use the handle if the connection is thought to be valid. What about splitting this up into `valid_handle()` and `any_handle()` or something like that?

src/cmap/conn/mod.rs
```diff
@@ -75,6 +75,8 @@ pub(crate) struct Connection {
 
     stream: AsyncStream,
 
+    pinned: bool,
```
patrickfreed, line 78: can we add a docstring to this field? ditto for the `pin` method and `unpin` method. For `unpin`, can we also mention that the connection pool is in charge of calling that method?

src/cmap/conn/mod.rs
```diff
@@ -340,6 +377,21 @@ impl Drop for Connection {
     }
 }
 
+#[derive(Clone, Debug)]
+pub(crate) struct PinHandle {
```
patrickfreed, line 381: ditto docstring for this type and methods.

Also, I wonder if the name for this type may be confused a bit with the `Pin` type from `std`. What do you think about a name like `PinnedConnectionHandle`?

src/cmap/conn/mod.rs
```diff
@@ -340,6 +377,21 @@ impl Drop for Connection {
...
+    connection_id: u32,
+    pool_manager: PoolManager,
+}
+
+impl PinHandle {
+    pub(crate) async fn take_connection(&self) -> Result<Connection> {
+        self.pool_manager.take_pinned(self.connection_id).await
+    }
+    pub(crate) fn unpin(&self) {
```
patrickfreed, line 390: can we add the `_connection` suffix here to match `take_connection`?

patrickfreed (commented):
Looks great! I mostly have comments about naming/bikeshedding and documentation.

src/cmap/manager.rs
```diff
@@ -59,6 +63,46 @@ impl PoolManager {
...
+            .sender
+            .send(PoolManagementRequest::StorePinned(connection))
+        {
+            let conn = request.0.unwrap_check_in();
+            return Err(conn);
+        }
+        Ok(())
+    }
+
+    pub(super) async fn take_pinned(&self, id: u32) -> Result<Connection> {
```
isabelatkinson, line 80: (For my own understanding) what is the difference between taking a pinned connection and unpinning?

src/client/executor.rs
```diff
@@ -200,7 +252,11 @@ impl Client {
             )
             .await
         {
-            Ok(result) => Ok(result),
+            Ok(output) => Ok(OperationDetails {
+                output,
+                connection: conn,
+                implicit_session: None, // populated in execute_operation_with_details
```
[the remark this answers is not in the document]
abr-egn: Done.

src/cmap/manager.rs
```diff
@@ -59,6 +63,46 @@ impl PoolManager {
         Ok(())
     }
 
+    pub(super) fn store_pinned(
```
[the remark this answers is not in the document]
abr-egn: Done.

src/cursor/common.rs
```diff
@@ -88,6 +103,18 @@ where
                     Poll::Ready(get_more_result) => {
                         let exhausted = get_more_result.exhausted();
                         let (result, session) = get_more_result.into_parts();
+                        if exhausted {
+                            // If the cursor is exhausted, the driver must return the pinned
+                            // connection to the pool.
+                            self.pinned_connection = PinnedConnection::new(None);
```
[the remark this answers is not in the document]
abr-egn: Done.

src/cursor/common.rs
```diff
@@ -214,3 +246,63 @@ pub(crate) struct CursorInformation {
...
+    }
+
+    fn handle(&self) -> Option<&PinHandle> {
+        match self {
+            Self::Valid(h) | Self::Invalid(h) => Some(h),
+            Self::Unpinned => None,
+        }
+    }
+
+    fn is_valid(&self) -> bool {
```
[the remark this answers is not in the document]
abr-egn: Switched to `is_invalid` - I try to avoid double negatives in boolean conditions, but in this case I see the confusion  with the `is_valid` name.

src/cursor/common.rs
```diff
@@ -88,6 +103,18 @@ where
                     Poll::Ready(get_more_result) => {
                         let exhausted = get_more_result.exhausted();
                         let (result, session) = get_more_result.into_parts();
+                        if exhausted {
```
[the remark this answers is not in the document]
abr-egn: As far as I can tell, the spec says connections need to be checked back in for _cursor-initiating commands_ on command failure; I don't see anything about command failure for `getMore` beyond the network error clause.

src/cursor/common.rs
abr-egn: Updated the poll to return `Poll::Ready(None)` for an invalidated connection; would prefer to leave handling general errors to a different PR.

src/cursor/common.rs
abr-egn: I think we want to continue to provide the handle for a pinned connection even if it's invalidated; for `start_get_more` and `kill_cursor`, it would otherwise check a fresh connection out and potentially cause bad things to happen if the cursor id happened to collide with one on a different server behind the load balancer, and for `unpin` it doesn't matter if it's valid or not.

src/cmap/conn/mod.rs
abr-egn: Done.

src/cmap/conn/mod.rs
abr-egn: Added docstrings, and `PinnedConnectionHandle` it is.

src/cmap/conn/mod.rs
abr-egn: Done.

src/client/executor.rs
```diff
@@ -99,41 +122,81 @@ impl Client {
...
+                .execute_operation_with_retry(op, session, pinned_connection)
+                .await?;
+            details.implicit_session = implicit_session;
+            Ok(details)
         })
         .await
     }
 
-    /// Execute the given operation, returning the implicit session created for it if one was.
+    /// Execute the given operation, returning cursor created by the operation.
```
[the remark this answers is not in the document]
abr-egn: Applied.

src/cmap/manager.rs
abr-egn: Pinning a connection removes it from the normal connection pool and makes it available to retrieve by id (wrapped by `PinnedConnectionHandle`).  Taking a pinned connection is that retrieval; it enforces ownership mechanics at runtime by only allowing a connection to be taken by id if that connection had previously been pinned and dropped.  Unpinning returns the connection to the normal connection pool, either immediately (if nothing's using it) or when it gets dropped.

Hopefully the docstrings added will help make this clearer :)

src/cmap/worker.rs
```diff
@@ -483,6 +508,28 @@ impl ConnectionPoolWorker {
         }
     }
 
+    fn store_pinned(&mut self, mut conn: Connection) {
+        if self.pending_unpins.remove(&conn.id) {
+            conn.unpin();
+            self.check_in(conn);
+        } else {
+            self.pinned_connections.insert(conn.id, conn);
```
patrickfreed, line 516: We're going to need to check if a connection was dropped while it was reading/writing and, if so, check it back in (which will close it) rather than keeping it pinned. Otherwise the next time it's used, it may read data from a previous operation. See line 504 for where this check happens for normal check ins.

src/cursor/common.rs
patrickfreed: Oh good point, my suggestion may have been even more dangerous...

src/cursor/common.rs
```diff
@@ -104,7 +131,7 @@ where
                     }
                     return Poll::Ready(Some(Ok(doc)));
                 }
-                None if !self.exhausted => {
+                None if !self.exhausted && !self.pinned_connection.is_invalid() => {
```
patrickfreed, line 134: this condition needs to be added a few lines above too. Alternatively, we can just remove that call to `self.start_get_more()` in the branch where a doc is found, since it doesn't actually do anything until it gets polled.

src/cmap/conn/mod.rs
```diff
@@ -340,6 +377,21 @@ impl Drop for Connection {
     }
 }
 
+#[derive(Clone, Debug)]
```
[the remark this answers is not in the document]
abr-egn: Blergh, well spotted.

For the code as currently written, I _think_ this can only happen via `timeout` or some similar future cancellation; if the `GetMoreProvider` future is polled to completion, it'll return the pinned connection and it'll be available for the cursor's `drop` invocation of `kill_cursors`.  Concurrent _execution_ of `getMore` with a cursor's `drop` can't happen, I think, because a normal `Cursor` is its own stream and a `SessionCursorStream` can't outlive the `SessionCursor`.

Given that we've already documented that cancelling futures returned by the driver isn't supported, I think getting an internal error if you do isn't a terrible outcome.

For the more general issue, yeah, it's definitely unfortunately easy to accidentally clone and overuse a handle.  I updated the `PinnedConnection[Handle]` types to not implement `Clone` and instead have a `replicate()` method that does the same thing, with a comment on what it's for; that should at least prevent accidentally cloning via including in other structs or the like.

src/cmap/worker.rs
abr-egn: Done.

src/cursor/common.rs
abr-egn: Took the second approach - I think it makes the flow a little easier to follow.

src/cmap/conn/mod.rs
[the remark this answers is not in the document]
patrickfreed: I actually think we may have races here even without cancellation, e.g.
```rust
cursor.next().await; // after this poll completes, connection is sent back over a channel via `store_pinned`
cursor.next().await; // `take_connection` races with the return of the connection
```
I think we may need to make returning a pinned connection an acknowledged operation, so that we can be sure the next time the connection is used it will be available. Alternatively, we can make `take_connection` block until its available or perished.

>For the more general issue, yeah, it's definitely unfortunately easy to accidentally clone and overuse a handle. I updated the PinnedConnection[Handle] types to not implement Clone and instead have a replicate() method that does the same thing, with a comment on what it's for; that should at least prevent accidentally cloning via including in other structs or the like.

This seems like a nice improvement :+1:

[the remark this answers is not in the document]
abr-egn: I realized that pinning can use a channel directly between the handle and the connection itself, removing the pool worker from the path and simplifying things all around - this should fix that race since `take_connection` now blocks.  The downside of this is that calling `take_connection` in the same task that's holding the pinned `Connection` will deadlock rather than causing an internal error; I think this is okay because that'll show up immediately in spec tests.

src/cmap/conn/mod.rs
```diff
@@ -340,6 +396,40 @@ impl Drop for Connection {
...
+        receiver
+            .recv()
+            .await
+            .ok_or_else(|| Error::internal("cannot take connection after unpin"))
+    }
+
+    /// Return the pinned connection to the normal connection pool.
+    pub(crate) async fn unpin_connection(&self) {
+        let mut receiver = self.receiver.lock().await;
+        receiver.close();
```
patrickfreed, line 429: Per the [tokio docs](https://docs.rs/tokio/1.11.0/tokio/sync/mpsc/struct.Receiver.html), this won't drop the connection if its in the channel, so we need to drain it to ensure the connection gets dropped and returned to the pool.

abr-egn: Done.

patrickfreed (approved):
LGTM, great job with this! it was a really tricky one

isabelatkinson (approved):

PR #450, opened 2021-09-02
RUST-1012 remove `Clone` impl from `ClientSession`

abr-egn:
RUST-1012

This was included for testing, i.e. so that `Entity` could continue to impl `Clone`; however, cloning a session is somewhere between meaningless and actively harmful.  Updated the few places where `Entity` was actually cloned to use one-off projections for the actual subtypes needed.

### DIFF
--- a/src/test/spec/unified_runner/operation.rs
+++ b/src/test/spec/unified_runner/operation.rs
@@ -597,7 +597,9 @@ impl TestOperation for Aggregate {
                             )
                             .await?
                         }
-                        AggregateEntity::Other(debug) => panic!("Cannot execute aggregate on {}", &debug),
+                        AggregateEntity::Other(debug) => {
+                            panic!("Cannot execute aggregate on {}", &debug)
+                        }
                     };
                     cursor
                         .stream(session)
--- a/src/test/spec/unified_runner/operation.rs
+++ b/src/test/spec/unified_runner/operation.rs
@@ -35,6 +35,8 @@ use crate::{
     },
     selection_criteria::ReadPreference,
     test::FailPoint,
+    Collection,
+    Database,
     IndexModel,
     RUNTIME,
 };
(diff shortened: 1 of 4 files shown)


src/test/spec/unified_runner/operation.rs
```diff
@@ -566,10 +568,19 @@ impl TestOperation for Aggregate {
         async move {
             let result = match &self.session {
                 Some(session_id) => {
-                    let entity = test_runner.entities.get(id).unwrap().clone();
+                    enum AggregateEntity {
```
patrickfreed, line 571: Are these changes necessary? It looks functionally pretty similar to the previous implementation.

abr-egn: Yup, it's needed - pre-change, the entity retrieved was unconditionally cloned; with this change, only a collection or database is cloned, other entities are formatted for later debug output.

src/test/spec/unified_runner/operation.rs
```diff
@@ -566,10 +568,19 @@ impl TestOperation for Aggregate {
...
+                        Collection(Collection<Document>),
+                        Database(Database),
+                        Other(String),
+                    }
+                    let entity = match test_runner.entities.get(id).unwrap() {
```
isabelatkinson, line 576: Instead of going through this helper struct can we match on `test_runner.entities.get(id).unwrap()` and clone within the branches? i.e.
```
match test_runner.entities.get(id).unwrap() {
	Entity::Collection(c) => {
		let c = c.clone();
		...
	}
	...
}
```

patrickfreed (commented):
Given that this has API implications, could you file a ticket for it? Then I can add it to the release notes.

src/test/spec/unified_runner/operation.rs
abr-egn: Not easily, unfortunately - `test_runner.entities.get(id).unwrap()` borrows from `test_runner.entities`, so matching on it would do so for the span of the match, and that would conflict with the exclusive borrow needed for `let session = test_runner.get_mut_session(session_id);`.

abr-egn:
> Given that this has API implications, could you file a ticket for it? Then I can add it to the release notes.

Good call, filed RUST-1012 and updated the PR title/description.

src/test/spec/unified_runner/operation.rs
isabelatkinson: Ah I do remember that being an issue when I implemented this, I think that's why I had that line of code on 569 in the first place.

isabelatkinson (approved):

patrickfreed (approved):

PR #457, opened 2021-09-09
minor: use new rustdoc:: warn prefix

abr-egn:
The `missing_crate_level_docs` lint was renamed to `rustdoc::missing_crate_level_docs` in rustc 1.55; however, our MSRV doesn't recognize the latter.

Unfortunately, `cfg_attr` doesn't have compiler-provided attributes for compiler version, and the `rustversion` crate that provides macros for such can't be used because macro attributes aren't allowed in inner attribute position, so we're left with suppressing the warning.  I re-enabled it for modules in the crate so the potential surface for problems should be small.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -315,7 +315,6 @@ macro_rules! define_if_single_runtime_enabled {
 // and warnings other than our custom ones.
 #[warn(renamed_and_removed_lints)]
 define_if_single_runtime_enabled! {
-
     #[macro_use]
     pub mod options;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -282,7 +282,7 @@
 //! it will only happen in a minor or major version release.
 
 #![warn(missing_docs)]
-#![warn(missing_crate_level_docs)]
+#![warn(rustdoc::missing_crate_level_docs)]
 #![cfg_attr(
     feature = "cargo-clippy",
     allow(
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -284,9 +284,8 @@
 #![warn(missing_docs)]
 // `missing_crate_level_docs` was renamed with a `rustdoc::` prefix in rustc 1.55, but isn't
 // supported in the MSRV.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -282,6 +282,7 @@
 //! it will only happen in a minor or major version release.
 
 #![warn(missing_docs)]
+#![allow(renamed_and_removed_lints)]  // `missing_crate_level_docs` was renamed with a `rustdoc::` prefix
```
isabelatkinson, line 285: Can we add a TODO here to update this when we bump our MSRV to a version that supports the new name?

abr-egn: Done.

isabelatkinson (commented):
lgtm!

isabelatkinson (approved):

src/lib.rs
```diff
@@ -282,6 +282,10 @@
 //! it will only happen in a minor or major version release.
 
 #![warn(missing_docs)]
+// `missing_crate_level_docs` was renamed with a `rustdoc::` prefix in rustc 1.55, but isn't
+// supported in the MSRV.
+// TODO: remove this if/when the MSRV is 1.55+.
+#![allow(renamed_and_removed_lints)]
```
patrickfreed, line 288: I think one possible way to avoid this would be to use `#![cfg_attr(docsrs, warn(rustdoc::missing_crate_level_docs))]`, and then to add `--cfg docsrs` to the end of the invocations in `check-rustdoc.sh`. We'd also need to use `+nightly` in that script too once we do that, since we use some nightly only stuff for our docs generation. This would be a general improvement anyways I think, since it would bring our docs linting closer to what docs.rs actually does to generate the documentation and also [how we lint in the BSON library](https://github.com/mongodb/bson-rust/blob/master/.evergreen/check-rustdoc.sh#L6).

abr-egn: Ah, that is nicer, thank you - done.

patrickfreed (approved):

PR #459, opened 2021-09-10
RUST-967 Fold `modifiers` subfield into parent struct

abr-egn:
RUST-967

The associated spec change updates the legacy test to drop use of the obsolete and otherwise-entirely-removed `modifiers` key, allowing the test parsing to fold those fields into the main `Find` struct.

I ran into an unclear MATCH failure during implementation, so I also updated that machinery to emit more precise failure information.


src/test/spec/command_monitoring/operation.rs
```diff
@@ -110,42 +110,6 @@ impl TestOperation for DeleteOne {
...
-        options.max_time = self.max_time;
-        options.min = self.min.clone();
-        options.max = self.max.clone();
-        options.return_key = self.return_key;
-        options.show_record_id = self.show_disk_loc;
-    }
-}
-
 #[derive(Debug, Default, Deserialize)]
 pub(super) struct Find {
```
isabelatkinson, line 114: Can we add `#[serde(deny_unknown_fields)]` to this to ensure we don't miss any new fields?

isabelatkinson (commented):
LGTM mod one small suggestion. I like the new matching logic -- it would be nice to extend it to the unified test runner's matcher in the future

src/test/spec/command_monitoring/operation.rs
abr-egn: Done.

isabelatkinson (approved):

patrickfreed (approved):

PR #460, opened 2021-09-10
RUST-856 Fix race between server selection and server monitoring

patrickfreed:
RUST-856

This PR fixes a race between server selection requesting an update to the topology and a monitor subscribing to those requests, which occasionally would result in the `min_heartbeat_frequency` test failing.

### DIFF
--- a/src/sdam/monitor.rs
+++ b/src/sdam/monitor.rs
@@ -122,12 +122,14 @@ impl HeartbeatMonitor {
                 None => break,
             };
 
+            // subscribe to check requests before performing the check in case one comes in
+            // after the check completes
+            let mut topology_check_requests_subscriber =
+                topology.subscribe_to_topology_check_requests();
+
             if self.check_server(&topology, &server).await {
                 topology.notify_topology_changed();
             }
-
-            let mut topology_check_requests_subscriber =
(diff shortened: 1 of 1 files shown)


src/sdam/monitor.rs
```diff
@@ -122,13 +122,15 @@ impl HeartbeatMonitor {
...
+            // subscribe to check requests before performing the check in case one comes in
+            // after the check completes
+            let mut topology_check_requests_subscriber =
+                topology.subscribe_to_topology_check_requests();
+
             if self.check_server(&topology, &server).await {
                 topology.notify_topology_changed();
             }
 
-            let mut topology_check_requests_subscriber =
```
patrickfreed, line 129: My theory is that the following sequence happens:
- check completes
- monitor notifies that topology changed
- operation sleeping in server selection gets notification and quickly fails server selection, requesting another check
- monitor starts subscribing to request checks, missing the previous one
- monitor sleeps until heartbeatFrequencyMS is hit

This fixes that by ensuring any requests that come in after the check will be recorded too, hopefully fixing this case.

abr-egn (approved):

isabelatkinson (approved):

PR #461, opened 2021-09-14
RUST-954 Pin connections for transactions when connected to a load balancer

abr-egn:
RUST-954

Transaction connection pinning [follows the same rules](https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst#behaviour-with-transactions) as mongos pinning for sharded transactions, so this CL augments that machinery to be able to track a `PinnedConnectionHandle`.

The load balancer spec requires the drivers to track whether a connection is checked out for a cursor vs a transaction vs anything else for the purpose of augmenting the `WaitQueueTimeoutError` message; however, the Rust driver does not support wait queue timeouts, so I've omitted that tracking.

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -160,7 +160,8 @@ impl Client {
     {
         let mut details = self.execute_operation_with_details(op, session).await?;
         let pinned = if details.output.connection.is_pinned() {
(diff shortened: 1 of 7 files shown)


patrickfreed (approved):
LGTM!

PR #465, opened 2021-09-20
RUST-978 Unified test format additions in support of load balanced tests

abr-egn:
RUST-978

This adds several new constructs to the unified test format (and tests for those) that will be used for the load balancer unified tests.  This also covers RUST-984.


src/cursor/mod.rs
```diff
@@ -107,9 +109,18 @@ where
                 PinnedConnection::new(pin),
                 provider,
             ),
+            kill_watcher: None,
             _phantom: Default::default(),
         }
     }
+
+    #[allow(dead_code)]
```
abr-egn, line 117: This (and the equivalent method on `SessionCursor`) trigger a false-positive dead code warning.

src/operation/count/mod.rs
```diff
@@ -47,8 +47,10 @@ impl Operation for Count {
     const NAME: &'static str = "count";
 
     fn build(&mut self, description: &StreamDescription) -> Result<Command> {
+        let mut name = Self::NAME.to_string();
         let mut body = match description.max_wire_version {
             Some(v) if v >= SERVER_4_9_0_WIRE_VERSION => {
+                name = "aggregate".to_string();
```
abr-egn, line 53: This has actually been broken for a while; the test that should have caught it was passing because the `command_name` field didn't have a camelCase rename and the unparsed `commandName` field was being ignored as an unknown field.

src/test/spec/unified_runner/matcher.rs
```diff
@@ -78,21 +104,75 @@ pub fn events_match(
...
+            ExpectedCmapEvent::ConnectionCheckOutFailed {
+                reason: actual_reason,
+            },
+        ) => {
+            if expected_reason.is_some() && actual_reason != expected_reason {
+                return false;
+            }
+            true
+        }
+        _ => actual == expected,
```
abr-egn, line 172: The rest of the `ExpectedCmapEvent` variants have zero fields, so `==` is conveniently applicable.

src/test/spec/unified_runner/operation.rs
```diff
@@ -92,6 +95,22 @@ impl<'de> Deserialize<'de> for OperationObject {
     }
 }
 
+#[derive(Debug)]
+pub enum Expectation {
```
abr-egn, line 99: With the new `ignoreResultAndError` case it seemed worthwhile to capture the valid states as an enum rather than continuing to add optional fields.

src/test/spec/unified_runner/operation.rs
```diff
@@ -92,6 +95,22 @@ impl<'de> Deserialize<'de> for OperationObject {
...
+    Result {
+        value: Option<Bson>,
+        save_as_entity: Option<String>,
+    },
+    Error(ExpectError),
+    Ignore,
+}
+
+fn deserialize_op<'de, 'a, T: 'a + Deserialize<'de> + TestOperation>(
```
abr-egn, line 108: I factored this out of the match in `Operation::deserialize` below to make it more clear there that it's just the type parameter that varies.

src/test/spec/unified_runner/operation.rs
```diff
@@ -343,39 +309,65 @@ impl TestOperation for DeleteOne {
 }
 
 #[derive(Debug, Default, Deserialize)]
-#[serde(rename_all = "camelCase", deny_unknown_fields)]
```
abr-egn, line 346: Removing `deny_unknown_fields` is unfortunate but this was already in an unsupported situation by using it with a `flatten`ed field, and outright broke when it was itself used as such.

src/test/spec/unified_runner/operation.rs
```diff
@@ -343,39 +309,65 @@ impl TestOperation for DeleteOne {
...
                 }
-                None => {
-                    let cursor = collection
-                        .find(self.filter.clone(), self.options.clone())
-                        .await?;
+                FindCursor::Normal(cursor) => {
+                    let cursor = cursor.into_inner();
                     cursor.try_collect::<Vec<Document>>().await?
                 }
+                FindCursor::Closed => panic!("get_cursor returned a closed cursor"),
```
abr-egn, line 370: `Find::get_cursor` *could* return a different enum that doesn't have the `Closed` variant and impls `Into<FindCursor>`, but that didn't seem worth the type proliferation for a single use.

src/test/spec/unified_runner/test_event.rs
```diff
@@ -1,38 +1,191 @@
...
+            Event::Cmap(sub) => ExpectedEvent::Cmap(sub.into()),
+            Event::Sdam(_) => ExpectedEvent::Sdam,
+        }
+    }
+}
+
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub enum ExpectedCommandEvent {
+    #[serde(rename = "commandStartedEvent", rename_all = "camelCase")]
```
abr-egn, line 29: To my surprise, `rename_all = "camelCase"` on an `enum` only applies to the names of the variants, not the fields of the individual variants.

src/test/spec/unified_runner/test_event.rs
```diff
@@ -1,38 +1,191 @@
...
+    Failed {
+        command_name: Option<String>,
+        has_service_id: Option<bool>,
     },
-    #[serde(rename = "commandFailedEvent")]
-    Failed { command_name: Option<String> },
 }
 
-impl From<CommandEvent> for TestEvent {
+impl From<CommandEvent> for ExpectedCommandEvent {
```
abr-egn, line 49: Idea for future cleanup: it might be worthwhile to remove the `From<Event>` machinery and directly match `ExpectedEvent` against `Event`.

src/test/spec/unified_runner/test_event.rs
```diff
@@ -1,38 +1,191 @@
...
+                    has_service_id: None,
+                }
+            }
+            CmapEvent::ConnectionCheckedIn(_) => Self::ConnectionCheckedIn {},
+        }
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Deserialize)]
+pub enum ObserveEvent {
```
abr-egn, line 133: I opted to keep this a single `enum` rather than splitting into `ObserveCommandEvent` / `ObserveCmapEvent` to keep this as close to the spec descriptions as possible.

src/test/spec/unified_runner/test_event.rs
patrickfreed: Yeah I agree. As it is right now it's a little confusing that the `actual_event` in `events_match` is an `ExpectedEvent` type.

src/test/spec/unified_runner/matcher.rs
```diff
@@ -78,21 +104,75 @@ pub fn events_match(
...
+    match (actual, expected) {
+        (
+            ExpectedCmapEvent::PoolCleared {
+                has_service_id: expected_has_service_id,
+            },
+            ExpectedCmapEvent::PoolCleared {
+                has_service_id: actual_has_service_id,
+            },
+        ) => {
+            if expected_has_service_id.is_some() && actual_has_service_id != expected_has_service_id
```
patrickfreed, line 140: these `if`/`else`s in the various branches can be simplified to just be the actual condition (i.e. `expected_has_service_id.is_none() || actual_has_service_id == expected_has_service_id`)

src/test/spec/unified_runner/operation.rs
patrickfreed: technically it's a requirement of the unified test format that we error on unrecognized fields, though it would be quite a bit of boilerplate to adhere to that here, so not sure if it's worth it or not.

src/test/util/event.rs
```diff
@@ -120,7 +120,9 @@ pub struct EventHandler {
     command_events: EventQueue<CommandEvent>,
     pub pool_cleared_events: EventQueue<PoolClearedEvent>,
     sdam_events: EventQueue<SdamEvent>,
+    cmap_events: EventQueue<CmapEvent>,
```
patrickfreed, line 123: It looks like some of the new tests are pretty flaky. This might be due to the use of the `EventQueue` rather than the channel, since the former is more of a point-in-time measurement that can be easily susceptible to races. In general, I've been trying to move towards using channels for all the event assertions for this reason, since you can just listen to them (with a set timeout) until the expected number of events have come in.

Not sure if this is the actual problem here, but something to consider.

src/test/util/event.rs
```diff
@@ -120,7 +120,9 @@ pub struct EventHandler {
     command_events: EventQueue<CommandEvent>,
     pub pool_cleared_events: EventQueue<PoolClearedEvent>,
```
patrickfreed, line 121: I think we can just drop this field now that we have full support for CMAP events.

abr-egn: Done.

src/test/util/event.rs
abr-egn: I went a different direction here and introduced a sync point to ensure that all queued worker requests have been processed before checking the events.  Waiting on a channel seems like it has a few downsides that avoids, in that it'll still catch failure cases of too many events, and it doesn't require timeouts.

I wasn't able to reproduce these failures locally, though, so I'll have to wait for evergreen to see if it fixed the flakes; if not I'll shift to a channel.

src/test/spec/unified_runner/matcher.rs
abr-egn: Done.

src/test/spec/unified_runner/mod.rs
```diff
@@ -140,69 +143,57 @@ pub async fn run_unified_format_test(test_file: TestFile) {
...
-                        });
-                        if let Some(ref expect_result) = operation.expect_result {
-                            match result {
-                                ResultEntity::Bson(ref result) => {
-                                    assert!(
-                                        results_match(
-                                            Some(result),
+                    match &operation.expectation {
+                        Expectation::Result {
+                            value,
```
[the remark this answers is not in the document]
abr-egn: Done.

src/test/spec/unified_runner/matcher.rs
```diff
@@ -78,21 +104,75 @@ pub fn events_match(
...
+        }
         _ => false,
     }
 }
 
+fn cmap_events_match(actual: &ExpectedCmapEvent, expected: &ExpectedCmapEvent) -> bool {
+    match (actual, expected) {
+        (
+            ExpectedCmapEvent::PoolCleared {
+                has_service_id: expected_has_service_id,
```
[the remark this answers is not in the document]
abr-egn: Eep, good catch.

src/test/spec/unified_runner/operation.rs
abr-egn: I'd personally go with "not" - I think not having to deal with keeping multiple duplicate data structures in sync is the lesser evil here - but I don't know what precedent is like.

src/cursor/mod.rs
abr-egn: Done.

src/test/spec/unified_runner/operation.rs
abr-egn: Updated.

PR #469, opened 2021-09-21
RUST-956 Include `service_id` in relevant events

abr-egn:
RUST-956

Some events need the `service_id` of the backend behind the load balancer; this adds those.  For the `Command*` events, this would have fit better in the `ConnectionInfo` struct, but unfortunately that's not marked `non_exhaustive` so that would be a breaking change :(

Whichever of this or https://github.com/mongodb/mongo-rust-driver/pull/465 is merged later will need to include the (trivial) event matching code for this field.

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -478,6 +478,7 @@ impl Client {
         }
 
         let connection_info = connection.info();
+        let service_id = connection.service_id();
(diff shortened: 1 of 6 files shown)


patrickfreed (approved):
Ah, too bad about `ConnectionInfo` not being `non_exhaustive`, we must have missed that one when combing over the public structs unfortunately.

PR #472, opened 2021-09-23
RUST-1020 Track `bson` master branch

patrickfreed:
RUST-1020

This PR updates the `Cargo.toml` to track the master branch of the `bson` repo. This will allow our tests and benchmarks to pull in the latest changes from `bson` since the last release. 

I think in general we should have a policy of always tracking the master of bson right up until the release to ensure we have full coverage. Once we need to do a release, we can switch to the latest `bson` release, and then immediately afterwards we can go back to tracking master.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ aws-auth = ["reqwest"]
 async-trait = "0.1.42"
 base64 = "0.13.0"
 bitflags = "1.1.0"
-bson = "2.0.0"
+bson = { git = "https://github.com/mongodb/bson-rust", branch = "master" }
 chrono = "0.4.7"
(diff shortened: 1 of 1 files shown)


abr-egn:
The policy here sounds fine to me, just need to be sure to document it.

abr-egn (approved):

isabelatkinson (approved):
lgtm, there's a rustfmt failure in CI but it doesn't look related to your changes

patrickfreed:
> The policy here sounds fine to me, just need to be sure to document it.

Good idea, updated the [release checklist](https://wiki.corp.mongodb.com/display/DRIVERS/Rust+Release+Checklist).

PR #473, opened 2021-09-23
RUST-979 Non-unified spec tests for load balancer support behavior

abr-egn:
RUST-979

This pulls in changes and additions to non-unified spec tests (and, in one case, implements a prose test) that validate various behaviors in support of load balancer functionality.


src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -23,162 +24,194 @@ struct TestFile {
...
+        return;
+    }
 
-        let result = if cfg!(target_os = "windows") {
-            ClientOptions::parse_with_resolver_config(&test_file.uri, ResolverConfig::cloudflare())
-                .await
-        } else {
-            ClientOptions::parse(&test_file.uri).await
-        };
+    // TODO RUST-979 unskip these tests
```
abr-egn, line 64: This will require an evergreen config for running tests against a loadbalancer, which will come in a follow-up PR.

src/sdam/srv_polling/test.rs
```diff
@@ -102,3 +104,24 @@ async fn timeout_error() {
 async fn no_results() {
     run_test(Ok(Vec::new()), DEFAULT_HOSTS.iter().cloned().collect()).await;
 }
+
+// SRV polling is not done for load-balanced clusters.
```
patrickfreed, line 108: Can we update this comment to refer to the spec that contains this prose test?

src/sdam/state/mod.rs
```diff
@@ -168,8 +94,6 @@ impl Topology {
             handler.handle_topology_opening_event(event);
         }
 
-        let hosts: Vec<_> = options.hosts.drain(..).collect();
```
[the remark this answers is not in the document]
patrickfreed: Oh wow good catch. And yeah that seems fine.

patrickfreed (approved):
One request for a minor comment, but otherwise LGTM! (don't need another round of review after the comment is added)

PR #474, opened 2021-09-24
RUST-781 Remove unwraps from examples

abr-egn:
RUST-781

This removes uses of `unwrap` from example code; only one of these was actually in the example portion.  There are a bunch of remaining `unwrap`s but they're in various places (macros, closures) that would be nontrivial to remove and not produce any benefit.

### DIFF
--- a/src/test/documentation_examples.rs
+++ b/src/test/documentation_examples.rs
@@ -34,7 +34,7 @@ macro_rules! run_on_each_doc {
 }
 
 async fn insert_examples(collection: &Collection<Document>) -> Result<()> {
-    collection.drop(None).await.unwrap();
+    collection.drop(None).await?;
 
     // Start Example 1
     collection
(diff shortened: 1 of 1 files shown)


patrickfreed (commented):
To clarify: did you also check the examples included in the docstrings of modules / types? The changes here are good, but the scope of the ticket also includes those examples, since they'll be included in the docs.rs page for the driver. These ones actually won't be on docs.rs, but they should be included in the MongoDB documentation. I'm noticing now that a ticket may never have been filed for some of the older ones unfortunately, so once this PR goes in I'll go ahead and do that.

abr-egn:
Apologies, I should have mentioned - yes, I looked for docstrings with `unwrap` and couldn't find any.  There are a few comments that reference unwrap but they're generally of the "// This unwrap is okay because ..." form.

patrickfreed (approved):

PR #476, opened 2021-09-29
RUST-54 Support for OP_COMPRESSED

awitten1:
Provides support for reading and writing OP_COMPRESSED as specified [here](https://github.com/mongodb/specifications/blob/master/source/compression/OP_COMPRESSED.rst). 

The following is a snippet of the server logs when running `cargo test compression::test::ping_server_with_zlib_compression`, which connects to a server on localhost and sends a ping (see src/compression/test.rs).  I have highlighted the most relevant parts.  (The exact output will be server version dependent).

Screen Shot 2021-09-29 at 10 31 04 AM


awitten1:
Not sure what went wrong with evergreen.  For now I submitted a patch [here](https://spruce.mongodb.com/patch/61548343a4cf4749c061e91a/configure/tasks).

src/cmap/establish/handshake/mod.rs
```diff
@@ -229,6 +244,32 @@ impl Handshaker {
                 .map(|server_first| client_first.into_first_round(server_first))
         });
 
+        // Check that master reply has a compressor list and unpack it
+        if let Some(ref server_compressors) = is_master_reply.command_response.compressors {
+            // Check that client has a compressor list and unpack it
+            if let Some(ref client_compressors) = self.compressors {
```
abr-egn, line 250: I tend to bundle pattern matches like this via tuples, i.e.
```
if let (Some(ref server_compressors), Some(ref client_compressors)) = (is_master_reply.command_response.compressors, self.compressors) {
  ...
```

src/cmap/establish/handshake/mod.rs
```diff
@@ -229,6 +244,32 @@ impl Handshaker {
...
+                // Use the Client's first compressor choice that the server supports
+                for client_compressor in client_compressors {
```
abr-egn, line 252: I think this would be clearer using iterators, i.e. something like
```
  if let Some(compressor) = client_compressors.iter().find(|c| server_compressors.contains(c)) {
    conn.compressor = compressor.parse().ok();
    ...
  }
```

src/is_master.rs
```diff
@@ -144,6 +144,10 @@ pub(crate) struct IsMasterCommandResponse {
     /// The address of the server that returned this `IsMasterCommandResponse`.
     pub me: Option<String>,
 
+    #[serde(rename = "compression")]
+    // The list of compatible compressors that the server returned.
+    pub compressors: Option<Vec<String>>,
```
abr-egn, line 149: If this is a fixed set, can this be an enum rather than a string?

src/cmap/establish/handshake/mod.rs
awitten1: Nice, will do.  That was bothering me.

Cargo.toml
```diff
@@ -42,6 +42,7 @@ bitflags = "1.1.0"
 bson = "2.0.0"
 chrono = "0.4.7"
 derivative = "2.1.1"
+flate2 = "1.0"
```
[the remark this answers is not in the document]
patrickfreed: Yep I agree.

patrickfreed (commented):
Overall looks really good! I have some API suggestions but otherwise mostly minor comments.

src/client/options/mod.rs
```diff
@@ -349,7 +349,10 @@ pub struct ClientOptions {
     #[builder(default)]
     pub app_name: Option<String>,
 
-    #[builder(default, setter(skip))]
+    /// The compressors that the Client is willing to use in the order they are specified
+    /// in the configuration.  The Client sends this list of compressors to the server.
+    /// The server responds with the intersection of its supported list of compressors.
+    #[builder(default)]
     pub(crate) compressors: Option<Vec<String>>,
```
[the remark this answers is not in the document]
awitten1: Agreed on all points except one.  I don't think `HashSet` will maintain ordering on the compressors, but the ordering is significant here (it indicates a preference between compressors).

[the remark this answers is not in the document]
patrickfreed: Oh very good point, yeah then let's stick with `Vec` then.

Cargo.toml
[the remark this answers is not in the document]
isabelatkinson: agreed

src/client/options/mod.rs
```diff
@@ -1670,7 +1679,7 @@ impl ClientOptionsParser {
                 self.auth_mechanism_properties = Some(doc);
             }
             "compressors" => {
-                self.compressors = Some(value.split(',').map(String::from).collect());
+                self.compressors = Some(value.split(',').map(str::to_lowercase).collect());
```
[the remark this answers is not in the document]
awitten1: I don't think it specifies.  I'm fine either way.  What do you think?

src/compression/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            Compressor::Zlib { level: _ } => CompressorID::ZlibID,
+            Compressor::Snappy => CompressorID::SnappyID,
+        }
+    }
+
+    pub(crate) fn to_encoder(&self) -> Result<Encoder> {
+        match *self {
+            Compressor::Zstd { level } => {
+                let encoder =
+                    zstd::Encoder::new(vec![], level.try_into().unwrap()).map_err(|e| {
```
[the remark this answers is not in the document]
awitten1: I switched to using an `as` to cast the integer.  It should be safe because all the options are validated.

src/compression/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+    feature = "zstd-compression",
+    feature = "zlib-compression",
+    feature = "snappy-compression"
+))]
+/// Enum representing supported compressor algorithms.
+/// Used for compressing messages sent to the server.
+/// For compressors that take a level, use None to indicate the default level.
+/// Requires one of zstd-compression, zlib-compression, or snappy-compression
+/// feature flags.
+pub enum Compressor {
```
abr-egn, line 79: I think a lot of the feature flag annotations could be simplified by including `Noop` in this enum and unconditionally including the machinery elsewhere, relying on there being only a single enum value when none are enabled.

Alternatively, the entire `compression` module could be put behind a
```
#[cfg(any(
    feature = "zstd-compression",
    feature = "zlib-compression",
    feature = "snappy-compression"
))]
```
clause in src/lib.rs.

src/compression/mod.rs
```diff
@@ -0,0 +1,394 @@
+#[cfg(test)]
```
abr-egn, line 1: This can be put behind an additional
```
#[cfg(any(
    feature = "zstd-compression",
    feature = "zlib-compression",
    feature = "snappy-compression"
))]
```

to remove the need for those within the test module.

src/cmap/conn/wire/message.rs
```diff
@@ -84,12 +91,108 @@ impl Message {
...
+            feature = "snappy-compression"
+        ))]
+        if header.op_code == OpCode::Compressed {
+            return Self::read_from_op_compressed(reader, &header).await;
+        }
+
+        Err(Error::new(
+            ErrorKind::InvalidResponse {
+                message: format!(
+                    "Invalid op code, expected {} or {} and got {}",
```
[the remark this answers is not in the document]
awitten1: Good point.  It is confusing because if none of the feature flags are enabled then the client is not expecting `OpCode::Compressed`, so we shouldn't say that it is expecting `OpCode::Compressed`.  Although the server will never send compressed messages if none of the feature flags are enabled (because the server only sends compressed messages if we tell it we can parse them).

src/compression/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+#[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
+#[cfg(not(any(
+    feature = "zstd-compression",
+    feature = "zlib-compression",
+    feature = "snappy-compression"
+)))]
+pub(crate) enum Compressor {}
+
+impl FromStr for Compressor {
```
patrickfreed, line 104: Instead of implementing `FromStr` (which will be present in the public API), can we instead make this a `pub(crate)` method on `Compressor`? That way, users won't accidentally miss the level option or misspell the compressor name.

src/compression/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+                                ),
+                            })
+                        })?;
+
+                Ok(Encoder::Zstd { encoder })
+            }
+            #[cfg(feature = "zlib-compression")]
+            Compressor::Zlib { level } => {
+                let level = match level {
+                    Some(level) if level != -1 => Compression::new(level as u32),
```
patrickfreed, line 231: If a user provides a negative value that isn't `-1`, this cast will likely produce an incorrect value. Instead, can we use `TryFrom` here and map the error to our `InvalidArgument` error?

src/lib.rs
```diff
@@ -58,6 +58,9 @@
 //! | `aws-auth`          | Enable support for the MONGODB-AWS authentication mechanism.                                                                          | `reqwest` 0.11                      | no      |
 //! | `bson-uuid-0_8`     | Enable support for v0.8 of the [`uuid`](docs.rs/uuid/0.8) crate in the public API of the re-exported `bson` crate.                    | n/a                                 | no      |
 //! | `bson-chrono-0_4`   | Enable support for v0.4 of the [`chrono`](docs.rs/chrono/0.4) crate in the public API of the re-exported `bson` crate.                | n/a                                 | no      |
+//! | `zlib-compression`  | Enable support for compressing messages with zlib                                                                                     | `flate2`                            | no      |
```
patrickfreed, line 61: can we linkify `zlib`, `zstd`, and `snappy`? Linking to their homepages rather than the rust crates would be good.

```suggestion
//! | `zlib-compression`  | Enable support for compressing messages with zlib.                                                                                     | `flate2`                            | no      |
```

src/lib.rs
```diff
@@ -58,6 +58,9 @@
...
+//! | `zstd-compression`  | Enable support for compressing messages with zstd.  This flag requires rust version 1.54.                                             | `zstd`                              | no      |
```
patrickfreed, line 62:
```suggestion
//! | `zstd-compression`  | Enable support for compressing messages with zstd.  This flag requires Rust version 1.54.                                             | `zstd`                              | no      |
```

src/compression/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+#[cfg(any(
+    feature = "zstd-compression",
+    feature = "zlib-compression",
+    feature = "snappy-compression"
+))]
+/// Enum representing supported compressor algorithms.
+/// Used for compressing messages sent to the server.
+/// For compressors that take a level, use None to indicate the default level.
+/// Requires one of zstd-compression, zlib-compression, or snappy-compression
+/// feature flags.
```
patrickfreed, line 78:
```suggestion
/// Enum representing supported compressor algorithms.
/// Used for compressing messages sent to and read from the server.
/// For compressors that take a level, use `None` to indicate the default level.
/// Requires one of `zstd-compression`, `zlib-compression`, or `snappy-compression`
/// feature flags.
```

src/compression/mod.rs
patrickfreed, line 79: When this enum is public, make sure to export it from `src/options.rs`. At the top of that file you can see some example of other options exports.

src/compression/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+pub enum Compressor {
+    /// Zstd compressor.  Level 0 is the default.  Requires Rust version 1.54.
+    /// See Zstd [manual](http://facebook.github.io/zstd/zstd_manual.html) for more information
+    #[cfg(feature = "zstd-compression")]
+    Zstd { level: Option<i32> },
+    /// Zlib compressor.  Levels -1 through 9 are valid.
+    /// Level -1 indicates default level.
```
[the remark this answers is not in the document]
awitten1: Not even for the purpose of documenting this public type?  The specs indicate that -1 should be handled specially.

src/compression/mod.rs
awitten1: We can do that, although any invalid values would have been caught by now.

abr-egn (approved):
LGTM!

src/test/spec/json/uri-options/compression-options.json
```diff
@@ -7,12 +7,7 @@
       "warning": false,
       "hosts": null,
       "auth": null,
-      "options": {
-        "compressors": [
-          "zlib"
-        ],
-        "zlibCompressionLevel": 9
-      }
+      "options": {}
```
isabelatkinson, line 10: Why were these removed -- was there an update to the spec tests that we pulled in here?

src/compression/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+))]
+/// Enum representing supported compressor algorithms.
+/// Used for compressing messages sent to the server.
+/// For compressors that take a level, use None to indicate the default level.
+/// Requires one of zstd-compression, zlib-compression, or snappy-compression
+/// feature flags.
+pub enum Compressor {
+    /// Zstd compressor.  Level 0 is the default.  Requires Rust version 1.54.
+    /// See Zstd [manual](http://facebook.github.io/zstd/zstd_manual.html) for more information
+    #[cfg(feature = "zstd-compression")]
```
patrickfreed, line 64: Can we update the attributes on these variants to be the following:
```rust
#[cfg(any(feature = "<compression feature here>", docsrs))]
#[cfg_attr(docsrs, doc(cfg(feature = "<compression feature here>")))]
```
This allows the little badges indicating the feature flag to show up in the generated docs.

src/compression/test.rs
```diff
@@ -0,0 +1,127 @@
+// Tests OP_COMPRESSED.  To actually test compression you need to look at
```
isabelatkinson, line 1: Are the tests in this module actually being run in CI? I believe we'd need to update our evergreen config to enable the compression feature flags, from looking at the evergreen logs it doesn't look like these ones show up in the test output.

isabelatkinson (commented):
looks good, just a few more comments!

src/test/spec/json/uri-options/compression-options.json
awitten1: The reason is that we only store compressors that are valid and enabled (via its corresponding feature flag).  When these tests are run none of the feature flags are set, so none of these compressors get stored.  As a result, when `ClientOptions` is deserialized, all these fields are empty.  We probably want to also run these tests with the feature flags set.

src/compression/test.rs
awitten1: No they aren't being run right now.  They only run if the feature flags are set.  I was going to update the evergreen configs in a separate PR.

src/test/spec/json/uri-options/compression-options.json
awitten1: I'll just disable these tests if not all of the feature flags are set.  I didn't realize that we don't manually edit these.

PR #477, opened 2021-09-29
RUST-980 Run load balancer tests on evergreen, and update existing tests 

abr-egn:
RUST-980

This adds evergreen configuration to run the tests against a load-balancer, updates several existing tests to run on that topology, and fixes several minor bugs that those tests revealed.

A few unrelated changes got picked up in test syncs: serverless run configuration (which the runner already understood) and language changes.


src/cmap/conn/mod.rs
```diff
@@ -313,6 +313,7 @@ impl Connection {
         let (tx, rx) = mpsc::channel(1);
         self.pinned_sender = Some(tx);
         Ok(PinnedConnectionHandle {
+            id: self.id,
```
abr-egn, line 316: This isn't needed for functionality, but it's very handy for debugging.

src/cmap/test/integration.rs
```diff
@@ -78,6 +78,10 @@ async fn concurrent_connections() {
     let _guard = LOCK.run_exclusively().await;
 
     let mut options = CLIENT_OPTIONS.clone();
+    if options.load_balanced.unwrap_or(false) {
```
abr-egn, line 81: `direct_connection` is explicitly incompatible with `load_balanced`.

src/cmap/worker.rs
```diff
@@ -513,13 +513,11 @@ impl ConnectionPoolWorker {
                 handler.handle_pool_cleared_event(event);
             });
 
-            if !self.generation.is_load_balanced() {
```
abr-egn, line 516: I swear that some part of the spec said that the queue shouldn't be drained in load-balancing mode, but I was apparently hallucinating that.

src/test/util/mod.rs
```diff
@@ -267,19 +270,19 @@ impl TestClient {
...
+        self.base_topology() == Topology::Single
     }
 
     pub fn is_replica_set(&self) -> bool {
-        self.options.repl_set_name.is_some()
+        self.base_topology() == Topology::ReplicaSet
     }
 
     pub fn is_sharded(&self) -> bool {
-        self.server_info.msg.as_deref() == Some("isdbgrid")
```
abr-egn, line 278: This returns `true` for load-balanced toplogies, so to avoid having to manually maintain a bunch of fiddly boolean exclusions I refactored these.

src/cmap/worker.rs
abr-egn: Yep - https://github.com/mongodb/mongo-rust-driver/blob/master/src/test/spec/retryable_reads.rs#L74 failed (the second checkout succeeds rather than fails), and the load balancer spec [specifically calls out](https://github.com/mongodb/specifications/blob/master/source/load-balancers/tests/README.rst#id5) retryable reads tests.

src/sdam/state/mod.rs
```diff
@@ -139,9 +140,11 @@ impl Topology {
         if is_load_balanced {
             for server_address in &options.hosts {
                 // Load-balanced clients don't have a heartbeat monitor, so we synthesize
-                // updating the server to `ServerType::LoadBalancer`.
+                // updating the server to `ServerType::LoadBalancer` with an RTT of 0 so it'll
+                // be selected.
                 let new_desc = ServerDescription {
                     server_type: ServerType::LoadBalancer,
+                    average_round_trip_time: Some(Duration::from_nanos(0)),
```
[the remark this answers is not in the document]
abr-egn: Nope - `retain_servers_within_latency_window` [filters out](https://github.com/mongodb/mongo-rust-driver/blob/master/src/sdam/description/topology/server_selection/mod.rs#L157) any with unset `average_round_trip_time`.

src/cmap/worker.rs
patrickfreed: Ah okay, so that test is specifically verifying the pool paused behavior, which shouldn't apply to load balancers. The spec doesn't currently clarify this so I filed DRIVERS-1942 to update it. In the meantime, I think we can preserve the existing behavior of not clearing the WaitQueue on pool clear if in load balanced mode and just skip that individual test.

abr-egn: Updated - needed to skip the equivalent test for retryable writes as well.

src/test/mod.rs
```diff
@@ -66,12 +59,31 @@ lazy_static! {
...
+
+pub(crate) fn client_options_for_uri(uri: &str) -> ClientOptions {
+    let mut options = ClientOptions::parse_without_srv_resolution(uri).unwrap();
+    options.max_pool_size = Some(MAX_POOL_SIZE);
+    options.server_api = SERVER_API.clone();
+    if LOAD_BALANCED_SINGLE_URI
+        .as_ref()
+        .map_or(false, |uri| !uri.is_empty())
+    {
+        options.test_options_mut().mock_service_id = true;
```
patrickfreed, line 75: Why is mocking needed only for the single URI?

patrickfreed (approved):
LGTM! I just have one question but I imagine it won't result in any further changes.

PR #491, opened 2021-10-14
RUST-807 Disallow maxPoolSize=0

abr-egn:
RUST-807

This disallows `maxPoolSize=0` and merges the relevant updated spec tests for `minPoolSize`/`maxPoolSize`, skipping the one for `maxPoolSize=0`.


kmahar:
I think it's reasonable to take the same route we did in Swift and not treat zero as unlimited.

We did discuss at some point recently-ish what Rust should do on invalid/unsupported URI options and there were some points made in both directions for erroring vs. just logging a warning. I don't remember if we reached a conclusion... Regardless, it would seem we might want to apply similar logic to how we handle invalid values for valid options, such as 0 in this case.

PR #504, opened 2021-10-28
RUST-656 Run tests against serverless in evergreen

abr-egn:
RUST-656

This adds evergreen configuration to run a subset of tests on serverless, and fixes various test runner bugs that were preventing those tests from passing.

This also closes RUST-983.


src/client/options/mod.rs
```diff
@@ -1190,6 +1190,9 @@ impl ClientOptions {
     /// Applies the options in other to these options if a value is not already present
     #[cfg(test)]
     pub(crate) fn merge(&mut self, other: ClientOptions) {
+        if self.hosts.is_empty() {
```
abr-egn, line 1193: I'm not super thrilled about the behavior here but I can't think of a clean solution.

* `TestClient` and `EventClient` have a `with_additional_options` constructor fn that, semantically, takes a passed-in `ClientOptions` and merges in the set of "test-default" options there for any unset fields, with preference given to the passed-in fields where both were set.
* Prior to this change, `hosts` was not included in the merge, so the passed-in `ClientOptions` would always determine `hosts`.  This caused  the `label_not_added_first_read_error` / `label_not_added_second_read_error` tests to break on serverless because those tests were not setting `hosts` and the default value was not valid in those cases.
* Ideally, the semantics for `hosts` would be the same as for the rest of the fields - if unset, it would use the "test default".  However, because it's populated in the builder by default, there's no way to actually determine "if unset"; testing against the default value causes *other* tests to fail because those are specifically constructing a hosts vec that, coincidentally, is equal to the default value, and should not be overridden by the test default.
* Again ideally, this could be fixed by changing `hosts` to an `Option<Vec<_>>` or not populating it by default so `vec[]` would be a reasonable signal for "unset", but either of those would be a breaking change.
* Calling `with_additional_options` with a `ClientOptions` constructed via builder with the default hosts is a nasty pitfall that's easy to hit - from the call site it looks like you'll be constructing a client with only the options specified, and the default `hosts` value will coincidentally work in most but not all testing environments.
* Existing tests with the exception of the two that broke avoid this by either explicitly populating `hosts` or copying the `ClientOptions` from the `CLIENT_OPTIONS` value.

.evergreen/config.yml
```diff
@@ -377,6 +377,32 @@ functions:
...
             .evergreen/run-tests.sh
 
+  "run serverless tests":
+    - command: shell.exec
+      type: test
+      params:
+        shell: bash
+        working_dir: "src"
+        script: |
+          # Exported before PREPARE_SHELL to avoid echo with xtrace
```
patrickfreed, line 387: I think it might be worth just disabling xtrace for these tests in case one of the later scripts accidentally echos these. We can still run `${PREPARE_SHELL}` but just disable it afterwards; I believe that's what we do in Swift.

.evergreen/run-serverless-tests.sh
```diff
@@ -0,0 +1,43 @@
...
+
+if [ "$ASYNC_RUNTIME" = "async-std" ]; then
+    FEATURE_FLAGS="${FEATURE_FLAGS},async-std-runtime"
+    DEFAULT_FEATURES="--no-default-features"
+elif [ "$ASYNC_RUNTIME" != "tokio" ]; then
+    echo "invalid async runtime: ${ASYNC_RUNTIME}" >&2
+    exit 1
+fi
+
+. ~/.cargo/env
```
patrickfreed, line 18: I think this should get run as part of `env.sh` and isn't required here.

src/error.rs
```diff
@@ -473,6 +487,10 @@ pub struct WriteConcernError {
     /// A document identifying the write concern setting related to the error.
     #[serde(rename = "errInfo")]
     pub details: Option<Document>,
+
+    /// Labels categorizing the error.
+    #[serde(rename = "errorLabels", default)]
+    pub labels: Vec<String>,
```
patrickfreed, line 493: Error labels are only ever actually returned at the top level of an error response, and it's merely a quirk / bug in the spec tests that they appear in the write concern error. @isabelatkinson actually just put up a PR to fix that: https://github.com/mongodb/specifications/pull/1086. 

For that reason, I don't think we should include public API support for this case, and if need be we can just skip the affected tests until the PR is merged. I think I remember you saying we may have to wait for a serverless change on this too to support the labels being specified in the correct place.

abr-egn: The serverless proxy apparently enforces that they appear here; I've removed this field from the public API but I'd prefer to leave it and the related logic in place so the tests can continue to pass.

.evergreen/run-serverless-tests.sh
abr-egn: Done.

.evergreen/config.yml
abr-egn: Done.

src/error.rs
isabelatkinson: have we confirmed that this behavior is correct/desired on the serverless end? it seems like a weird inconsistency. I think it's fine to have it within the driver right now so that we can run the tests but in the long run it would be nice not to have to maintain a special case like this

src/test/spec/retryable_writes/mod.rs
```diff
@@ -179,6 +179,8 @@ async fn run_spec_tests() {
                     .await
                     .unwrap();
             }
+
+            let _ = client.database(&db_name).drop(None).await;
```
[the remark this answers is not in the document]
abr-egn: Done.

src/client/options/mod.rs
```diff
@@ -1232,6 +1236,10 @@ impl ClientOptions {
                 original_uri
             ]
         );
+        #[cfg(test)]
```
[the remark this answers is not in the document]
abr-egn: Derp, fixed.

src/error.rs
abr-egn: I'm in the process of following up about this on the serverless side - it looks like this isn't expected behavior.  I'll make sure to send a follow-up PR for the driver (even if it's just a TODO with a bug reference) once that gets sorted out.

patrickfreed: SGTM to leave this in temporarily until serverless fixes the issue on their end.

patrickfreed (approved):
LGTM!

src/error.rs
abr-egn: I've added todos here referencing CLOUDP-105256.

isabelatkinson (approved):

PR #508, opened 2021-11-03
RUST-537 Aggregation, runCommand, and index management examples

abr-egn:


src/test/documentation_examples/aggregation_data.rs
```diff
@@ -0,0 +1,258 @@
+use bson::doc;
+
+use crate::{error::Result, Database};
+
+pub(crate) async fn populate(db: &Database) -> Result<()> {
+    fn iso_date(text: &str) -> bson::DateTime {
```
isabelatkinson, line 6: Can we update this to return a result rather than unwrapping? Also, we should consider waiting til RUST-977 is merged so that we can use the method implemented there rather than going through `chrono`.

abr-egn: Updated to use `parse_rfc3339_str`.

src/test/documentation_examples/mod.rs
```diff
@@ -1497,6 +1501,242 @@ async fn versioned_api_examples() -> std::result::Result<(), Box<dyn std::error:
...
+                },
+                doc! {
+                    "$sort": { "numberSold": 1 }
+                },
+            ],
+            None,
+        )
+        .await?;
+    // End Aggregation Example 2
+    let out: Vec<_> = cursor.try_collect().await?;
```
patrickfreed, line 1557: Per [this comment](https://jira.mongodb.org/browse/DRIVERS-448?focusedCommentId=1810127&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1810127), it looks like we should include iterating the cursor in the example too. As part of that we'll want to make sure to include a `use` for `TryStreamExt` within the example. Ditto for other examples.

src/test/documentation_examples/mod.rs
```diff
@@ -1497,6 +1501,242 @@ async fn versioned_api_examples() -> std::result::Result<(), Box<dyn std::error:
...
+    let db = client.database("aggregation_examples");
+    db.drop(None).await?;
+    aggregation_data::populate(&db).await?;
+
+    // Start Aggregation Example 1
+    let cursor = db
+        .collection::<Document>("sales")
+        .aggregate(
+            vec![
+                doc! { "$match" : { "items.fruit":"banana" } },
```
patrickfreed, line 1515: I know this format matches the example provided in the DRIVERS ticket, but could we use `<key>: <value>` spacing for the key value pairs here and in other `doc!` invocations? That's what we use in our other existing existing examples.

src/test/documentation_examples/mod.rs
abr-egn: Updated to include the `use` and `try_collect` in the examples.  Gave each example its own scope to allow the otherwise-redundant `use` clauses, which is a bit of a hack but I didn't see a better way.

patrickfreed (approved):

patrickfreed:
Oh it looks like lint broke with the latest changes, so once thats fixed LGTM (no need for me to review again).

isabelatkinson (approved):

PR #513, opened 2021-11-05
RUST-829 Remove empty write concerns from commands

karmenliang:
[RUST-829](https://jira.mongodb.org/browse/RUST-829)

This PR updates write concern serialization in options structs across all operations such that empty write concerns are not sent to the server. We do this by checking if all the `Option`al fields of a `WriteConcern` are `None`—if so, we leave it out of the command bson.


abr-egn (commented):
It looks like there are a bunch more operations that include a write concern (i.e. `CommitTransaction`) - those will need to be included as well.

src/operation/abort_transaction/test.rs
```diff
@@ -0,0 +1,66 @@
...
+    client::session::TransactionPin,
+    cmap::StreamDescription,
+    concern::{Acknowledgment, WriteConcern},
+    operation::{AbortTransaction, Operation},
+    selection_criteria::{ReadPreference::Primary, SelectionCriteria::ReadPreference},
+};
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn build() {
```
abr-egn, line 14: It looks like these (and similar tests for other operations) don't need to be async, which means they can just be annotated with `#[test]` instead of the runtime test whatnots.

src/operation/commit_transaction/mod.rs
```diff
@@ -33,7 +36,13 @@ impl Operation for CommitTransaction {
             Self::NAME: 1,
         };
 
-        append_options(&mut body, self.options.as_ref())?;
+        let mut options = self.options.clone().unwrap_or_default();
```
abr-egn, line 39: It would be good to capture this pattern - either via a new trait implemented by the various `*Options` that gives access to a `&mut WriteConcern`, or a simple macro.

abr-egn (commented):
Looks good overall!  A few relatively minor comments.

src/operation/abort_transaction/test.rs
karmenliang: Every test for all operations include these features—should I re-annotate them all with `#[test]` and remove the `async`?

abr-egn: I hadn't noticed that there were a bunch of existing operation tests that were async!  I think to avoid churn the best way here would be to update the tests that you've added in this PR and then file a ticket with the greenerbuild tag to update the rest of them.

src/operation/abort_transaction/test.rs
```diff
@@ -0,0 +1,66 @@
+use pretty_assertions::assert_eq;
```
isabelatkinson, line 1: I'm having trouble finding a paper trail for this but I remember having a discussion a while back about moving away from these types of tests -- IIRC we agreed that in general it's more preferable to test the driver's behavior rather than the lower-level details of how operations are being constructed. To that end, I'd prefer to test this behavior with command monitoring rather than via operation building tests. It would even be fairly simple to write unified tests to do this, although that may be outside the scope of this work. WDYT @abr-egn @patrickfreed ?

isabelatkinson (commented):
looking good! there are also some TODO comments for deleting unnecessary code in the V2 test runner; we'll want to address those in this PR as well. to find them you can grep for "TODO RUST-829" in `src/test/spec/v2_runner/operation.rs`

src/operation/abort_transaction/test.rs
patrickfreed: Yeah I don't think we formally discussed it, but I agree that we should probably not write any new operation unit tests, since in my experience they've been more work to maintain than they've been worth vs regular integration tests.

Regarding spec tests, it looks like they actually [already exist](https://github.com/mongodb/mongo-rust-driver/pull/513), albeit in the v2 format rather than unified, and in fact we're [already running them](https://github.com/mongodb/mongo-rust-driver/blob/master/src/test/spec/read_write_concern/operation.rs#L11), so maybe no new testing is necessary? Though I'm curious why these tests weren't failing if there are code changes in the actual driver code in this PR. Are those changes necessary? Or are there holes in the read-write-concern spec tests?

isabelatkinson: I wrote some logic in the v2 runner to strip out empty write concerns when I first noticed this issue/filed the ticket: https://github.com/mongodb/mongo-rust-driver/blob/b843bfab8560525ae47095efab7fa7318d98243c/src/test/spec/v2_runner/operation.rs#L149
My overall comment from my review notes that we should delete this logic. I thought I remembered that this was behavior in an unrelated test, but given that there are dedicated tests for this I think removing the TODOs should be sufficient.

src/operation/commit_transaction/mod.rs
karmenliang: Wrote a macro in https://github.com/mongodb/mongo-rust-driver/pull/513/commits/1f81c084161c53fafef01c2fff489a19bfda7ff2. `run_command`, `abort_transaction`, and `drop_database` go about this slightly differently, so I didn't use it in those three operations.

src/operation/aggregate/mod.rs
```diff
@@ -53,7 +53,14 @@ impl Operation for Aggregate {
             "pipeline": bson_util::to_bson_array(&self.pipeline),
             "cursor": {}
         };
-        append_options(&mut body, self.options.as_ref())?;
+
+        let mut options = self.options.clone().unwrap_or_default();
+        if let Some(write_concern) = self.write_concern() {
+            if *write_concern == Default::default() {
```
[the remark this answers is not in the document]
karmenliang: Good point—done in https://github.com/mongodb/mongo-rust-driver/pull/513/commits/1f81c084161c53fafef01c2fff489a19bfda7ff2.

src/operation/abort_transaction/test.rs
karmenliang: Sounds good, filed [RUST-1102](https://jira.mongodb.org/browse/RUST-1102) to track this.

src/operation/mod.rs
```diff
@@ -434,3 +434,15 @@ pub(crate) enum Retryability {
     Read,
     None,
 }
+
+macro_rules! remove_empty_write_concern {
```
isabelatkinson, line 438: Because all of the options structs passed into this macro contain a `write_concern` field, we don't need to pass the actual write concern into this method -- instead, we can access it directly on the options via `$opts.write_concern`. (The compiler enforces that this field exists when it expands the macro calls.) In addition, I think it would be better to treat `$opts` as an `Option<*Options>` here, i.e.:

```
if let Some(ref mut options) = $opts {
    if let Some(ref write_concern) = options.write_concern {
        if write_concern.is_empty() {
            options.write_concern = None;
        }
    }
}
```

and call as such:

```
remove_empty_write_concern!(self.options);
```

This way we can eliminate the need to construct an empty options struct when the value for `options` is `None`.

src/operation/insert/mod.rs
```diff
@@ -110,17 +110,15 @@ impl<'a, T: Serialize> Operation for Insert<'a, T> {
             }
             .into());
         }
-
-        let mut options = self.options.clone().unwrap_or_default();
-        options.ordered = Some(self.is_ordered());
```
isabelatkinson, line 115: In this case I think we need to keep around these two deleted lines to ensure `ordered` is set properly.

src/operation/run_command/mod.rs
```diff
@@ -64,10 +64,22 @@ impl<'conn> Operation for RunCommand<'conn> {
                 message: "an empty document cannot be passed to a run_command operation".into(),
             })?;
 
+        let mut command = self.command.clone();
```
isabelatkinson, line 67: Based on [this language](https://github.com/mongodb/specifications/blob/5334c6c33de34949d39eed913f325ad82dbbcf2a/source/read-write-concern/read-write-concern.rst#generic-command-method-1) from the read/write concern spec, I don't think we need to strip out the empty write concern here. Generally we avoid validation on the documents provided to `run_command` (except for some conditions within a transaction).

isabelatkinson (commented):
Just a few more comments, also was there a reason for keeping around the operation test files for `abort_transaction` and `commit_transaction`?

src/operation/insert/mod.rs
karmenliang: Not sure I understand why we need to set `ordered` on a copy of `self.options` if we'll be using `self.options` directly in the rest of this method. Should I be passing the copied `options` to `remove_empty_write_concern` and the `InsertCommand` builder instead?

isabelatkinson: Yeah, prior to these changes the copied `options` with the updated `ordered` field was being passed to `InsertCommand`, so we'll want to continue to do that. Essentially the only change here should be calling `remove_empty_write_concern` on the copied options.

src/operation/run_command/mod.rs
karmenliang: Removed this in https://github.com/mongodb/mongo-rust-driver/pull/513/commits/6833ab8273d127d549f3fdf105dcd06e7d63fb20.

src/operation/insert/mod.rs
karmenliang: Done in https://github.com/mongodb/mongo-rust-driver/pull/513/commits/baa9a519f1eadef8168a9a64b08ee2b1bf859aea.

src/operation/abort_transaction/test.rs
karmenliang: Removed the unit tests I added in https://github.com/mongodb/mongo-rust-driver/pull/513/commits/1d1c8f7bb26fb46c040618b08678647cfa0ca60f.

abr-egn (approved):
LGTM!  Thank you for working through all the complications!

patrickfreed (approved):
LGTM!

src/operation/create/mod.rs
```diff
@@ -48,7 +48,9 @@ impl Operation for Create {
         let mut body = doc! {
             Self::NAME: self.ns.coll.clone(),
         };
-        append_options(&mut body, self.options.as_ref())?;
+
+        remove_empty_write_concern!(self.options);
+        append_options(&mut body, Some(&self.options))?;
```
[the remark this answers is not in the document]
karmenliang: Fixed this in https://github.com/mongodb/mongo-rust-driver/pull/513/commits/6b7ebc1dae273efc5301c68c787d841ef3c30117. Thanks for the help!

isabelatkinson (commented):
lgtm!

PR #518, opened 2021-11-10
RUST-519 Implement (but do not expose) the public change streams API

abr-egn:
RUST-519

This is largely a directly transliteration of the [spec](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst), and is heavily based on [the PR](https://github.com/mongodb/mongo-rust-driver/commit/2f16ab6600457cbf6d24a90603301b3554f5d71f) implementing this in the change-streams branch.  Notable changes from that PR:
* `ChangeStreamEvent` is generic over the document type.
* `ChangeStream` is generic over the item type.
* `watch` assumes a `ChangeStreamEvent<_>` as the returned stream item type, and notes in documentation that if the structure is changed it'll need a `with_type` call.

### DIFF
--- a/src/change_stream/event.rs
+++ b/src/change_stream/event.rs
@@ -1,4 +1,4 @@
-//! Contains documents related to a ChangeStream event.
+//! Contains the types related to a `ChangeStream` event.
 use crate::coll::Namespace;
 use bson::{Bson, Document};
 use serde::{Deserialize, Serialize};
@@ -37,10 +37,10 @@ pub struct ChangeStreamEvent<T> {
     /// Describes the type of operation represented in this change notification.
     pub operation_type: OperationType,
 
-    /// Identifies which collection or database where the event occurred.
+    /// Identifies the collection or database on which the event occurred.
     pub ns: Option<ChangeStreamEventSource>,
 
-    /// The new name for the ns collection.  Only included for `OperationType::Rename`.
+    /// The new name for the `ns` collection.  Only included for `OperationType::Rename`.
     pub to: Option<Namespace>,
 
     /// For unsharded collections this contains a single field, id, with the value of the id of the
@@ -48,7 +48,8 @@ pub struct ChangeStreamEvent<T> {
     /// shard key in order, followed by the id if the id isn’t part of the shard key.
     pub document_key: Option<Document>,
 
-    /// Contains a description of updated and removed fields in this operation.
+    /// A `Document` describing the fields that were updated or removed by the update operation.
+    /// Only specified if `operation_type` is `OperationType::Update`.
     pub update_description: Option<UpdateDescription>,
 
     /// For operations of type "insert" and "replace", this key will contain the document being
@@ -105,7 +106,7 @@ pub enum OperationType {
     Invalidate,
 }
 
-/// Identifies which collection or database where an event occurred.
+/// Identifies the collection or database on which an event occurred.
 #[derive(Deserialize, Debug)]
 #[serde(untagged)]
 #[non_exhaustive]
@@ -114,6 +115,6 @@ pub enum ChangeStreamEventSource {
     /// the change happened.
     Namespace(Namespace),
 
-    // Contains the name of the dabatase in which the change happened.
+    // Contains the name of the database in which the change happened.
     Database(String),
 }
--- a/src/change_stream/event.rs
+++ b/src/change_stream/event.rs
@@ -4,12 +4,11 @@ use bson::{Bson, Document};
 use serde::{Deserialize, Serialize};
 
 /// An opaque token used for resuming an interrupted
-/// [`ChangeStream`](../struct.ChangeStream.html).
+/// [`ChangeStream`](crate::change_stream::ChangeStream).
 ///
 /// When starting a new change stream,
-/// [`start_after`](../option/struct.ChangeStreamOptions.html#structfield.start_after)
-/// and [`resume_after`](../option/struct.ChangeStreamOptions.html#structfield.resume_after) fields
-/// on [`ChangeStreamOptions`](../option/struct.ChangeStreamOptions.html) can be specified
+/// [`crate::options::ChangeStreamOptions::start_after`] and
+/// [`crate::options::ChangeStreamOptions::resume_after`] fields can be specified
 /// with instances of `ResumeToken`.
 ///
 /// See the documentation
@@ -109,6 +108,7 @@ pub enum OperationType {
 /// Identifies which collection or database where an event occurred.
 #[derive(Deserialize, Debug)]
 #[serde(untagged)]
+#[non_exhaustive]
 pub enum ChangeStreamEventSource {
     /// Contains two fields: "db" and "coll" containing the database and collection name in which
     /// the change happened.
--- a/src/change_stream/event.rs
+++ b/src/change_stream/event.rs
@@ -118,6 +118,6 @@ pub enum ChangeStreamEventSource {
     /// The [`Namespace`] containing the database and collection in which the change occurred.
     Namespace(Namespace),
 
-    // Contains the name of the database in which the change happened.
+    /// Contains the name of the database in which the change happened.
     Database(String),
 }
--- a/src/change_stream/event.rs
+++ b/src/change_stream/event.rs
@@ -1,0 +1,119 @@
+//! Contains documents related to a ChangeStream event.
+use crate::coll::Namespace;
+use bson::{Bson, Document};
+use serde::{Deserialize, Serialize};
+
+/// An opaque token used for resuming an interrupted
+/// [`ChangeStream`](../struct.ChangeStream.html).
+///
+/// When starting a new change stream,
+/// [`start_after`](../option/struct.ChangeStreamOptions.html#structfield.start_after)
+/// and [`resume_after`](../option/struct.ChangeStreamOptions.html#structfield.resume_after) fields
+/// on [`ChangeStreamOptions`](../option/struct.ChangeStreamOptions.html) can be specified
+/// with instances of `ResumeToken`.
+///
+/// See the documentation
+/// [here](https://docs.mongodb.com/manual/changeStreams/#change-stream-resume-token) for more
+/// information on resume tokens.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct ResumeToken(pub(crate) Bson);
+
+/// A `ChangeStreamEvent` represents a
+/// [change event](https://docs.mongodb.com/manual/reference/change-events/) in the associated change stream.
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct ChangeStreamEvent<T> {
+    /// An opaque token for use when resuming an interrupted `ChangeStream`.
+    ///
+    /// See the documentation
+    /// [here](https://docs.mongodb.com/manual/changeStreams/#change-stream-resume-token) for
+    /// more information on resume tokens.
+    ///
+    /// Also see the documentation on [resuming a change
+    /// stream](https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream).
+    #[serde(rename = "_id")]
+    pub id: ResumeToken,
+
+    /// Describes the type of operation represented in this change notification.
+    pub operation_type: OperationType,
+
+    /// Identifies which collection or database where the event occurred.
+    pub ns: Option<ChangeStreamEventSource>,
+
+    /// The new name for the ns collection.  Only included for `OperationType::Rename`.
+    pub to: Option<Namespace>,
+
+    /// For unsharded collections this contains a single field, id, with the value of the id of the
+    /// document updated.  For sharded collections, this will contain all the components of the
+    /// shard key in order, followed by the id if the id isn’t part of the shard key.
+    pub document_key: Option<Document>,
+
+    /// Contains a description of updated and removed fields in this operation.
+    pub update_description: Option<UpdateDescription>,
+
+    /// For operations of type "insert" and "replace", this key will contain the document being
+    /// inserted, or the new version of the document that is replacing the existing
+    /// document, respectively.
+    ///
+    /// For operations of type "update", when the `ChangeStream's` full document type is
+    /// `UpdateLookup`, this key will contain a copy of the full version of the document from
+    /// some point after the update occurred. If the document was deleted since the updated
+    /// happened, it will be `None`.
+    pub full_document: Option<T>,
+}
+
+/// Describes which fields have been updated or removed from a document.
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct UpdateDescription {
+    /// A `Document` containing key:value pairs of names of the fields that were changed, and the
+    /// new value for those fields.
+    pub updated_fields: Document,
+
+    /// An array of field names that were removed from the `Document`.
+    pub removed_fields: Vec<String>,
+}
+
+/// The operation type represented in a given change notification.
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub enum OperationType {
+    /// See [insert-event](https://docs.mongodb.com/manual/reference/change-events/#insert-event)
+    Insert,
+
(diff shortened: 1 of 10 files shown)


src/coll/mod.rs
```diff
@@ -788,6 +789,54 @@ impl<T> Collection<T> {
...
+    ///
+    /// Change streams require either a "majority" read concern or no read concern. Anything else
+    /// will cause a server error.
+    ///
+    /// Also note that using a `$project` stage to remove any of the `_id`, `operationType` or `ns`
+    /// fields will cause an error. The driver requires these fields to support resumability. For
+    /// more information on resumability, see the documentation for
+    /// [`ChangeStream`](change_stream/struct.ChangeStream.html)
+    #[allow(unused)]
+    pub(crate) async fn watch_with_pipeline<D>(
```
patrickfreed, line 830: I like the idea of having separate methods here, though I think we should think about what we want to do with regular `aggregate` as well when deciding how to do this here for the sake of consistency. Something like `aggregate_with_type` seems like it would work well there as well, what do you think?

The main drawback in both cases is the lack of turbofish as you point out, but hopefully that won't turn out to be to much of a pain point.

src/change_stream/options.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use crate::{change_stream::event::ResumeToken, collation::Collation, options::AggregateOptions};
+
+#[skip_serializing_none]
+#[derive(Clone, Debug, Default, Deserialize, Serialize, TypedBuilder)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+/// These are the valid options that can be passed to the `watch` method for creating a
+/// [`ChangeStream`](../struct.ChangeStream.html).
+pub struct ChangeStreamOptions {
```
patrickfreed, line 15: One thing the original PR did not account for is sessions, which we'll need to do now. I think we should try to mimic the API of cursors here where we have a `ChangeStream` type and a `SessionChangeStream`. We may need to introduce an internal change stream type that both can wrap as part of that.

src/change_stream/event.rs
```diff
@@ -0,0 +1,119 @@
...
+    ///
+    /// Also see the documentation on [resuming a change
+    /// stream](https://docs.mongodb.com/manual/changeStreams/#resume-a-change-stream).
+    #[serde(rename = "_id")]
+    pub id: ResumeToken,
+
+    /// Describes the type of operation represented in this change notification.
+    pub operation_type: OperationType,
+
+    /// Identifies which collection or database where the event occurred.
```
isabelatkinson, line 40:
```suggestion
    /// Identifies the collection or database on which the event occurred.
```

src/change_stream/event.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub ns: Option<ChangeStreamEventSource>,
+
+    /// The new name for the ns collection.  Only included for `OperationType::Rename`.
```
isabelatkinson, line 43:
```suggestion
    /// The new name for the `ns` collection.  Only included for `OperationType::Rename`.
```

src/change_stream/event.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub to: Option<Namespace>,
+
+    /// For unsharded collections this contains a single field, id, with the value of the id of the
+    /// document updated.  For sharded collections, this will contain all the components of the
+    /// shard key in order, followed by the id if the id isn’t part of the shard key.
+    pub document_key: Option<Document>,
```
isabelatkinson, line 49: The doc for this is a little wordy; I like what the [manual](https://docs.mongodb.com/manual/reference/change-events/) has, can we replace it with something similar?

> A document that contains the `_id` of the document created or modified by the insert, replace, delete, update operations (i.e. CRUD operations). For sharded collections, also displays the full shard key for the document. The `_id` field is not repeated if it is already a part of the shard key.

src/change_stream/event.rs
```diff
@@ -0,0 +1,119 @@
+//! Contains documents related to a ChangeStream event.
+use crate::coll::Namespace;
+use bson::{Bson, Document};
+use serde::{Deserialize, Serialize};
+
+/// An opaque token used for resuming an interrupted
+/// [`ChangeStream`](../struct.ChangeStream.html).
+///
+/// When starting a new change stream,
+/// [`start_after`](../option/struct.ChangeStreamOptions.html#structfield.start_after)
```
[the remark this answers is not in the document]
abr-egn: Updated.

src/change_stream/event.rs
```diff
@@ -0,0 +1,119 @@
...
+    DropDatabase,
+
+    /// See [invalidate-event](https://docs.mongodb.com/manual/reference/change-events/#invalidate-event)
+    Invalidate,
+}
+
+/// Identifies which collection or database where an event occurred.
+#[derive(Deserialize, Debug)]
+#[serde(untagged)]
+pub enum ChangeStreamEventSource {
```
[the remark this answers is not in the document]
abr-egn: Done.

src/change_stream/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+where
+    T: DeserializeOwned + Unpin + Send + Sync,
+{
+    /// Returns the cached resume token that can be used to resume after the most recently returned
+    /// change.
+    ///
+    /// See the documentation
+    /// [here](https://docs.mongodb.com/manual/changeStreams/#change-stream-resume-token) for more
+    /// information on change stream resume tokens.
+    pub fn resume_token(&self) -> Option<ResumeToken> {
```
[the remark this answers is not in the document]
abr-egn: Done.

src/change_stream/options.rs
```diff
@@ -0,0 +1,77 @@
...
+use typed_builder::TypedBuilder;
+
+use crate::{change_stream::event::ResumeToken, collation::Collation, options::AggregateOptions};
+
+#[skip_serializing_none]
+#[derive(Clone, Debug, Default, Deserialize, Serialize, TypedBuilder)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+/// These are the valid options that can be passed to the `watch` method for creating a
+/// [`ChangeStream`](../struct.ChangeStream.html).
```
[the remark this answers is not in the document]
abr-egn: Done.

src/change_stream/event.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    /// The new name for the ns collection.  Only included for `OperationType::Rename`.
+    pub to: Option<Namespace>,
+
+    /// For unsharded collections this contains a single field, id, with the value of the id of the
+    /// document updated.  For sharded collections, this will contain all the components of the
+    /// shard key in order, followed by the id if the id isn’t part of the shard key.
+    pub document_key: Option<Document>,
+
+    /// Contains a description of updated and removed fields in this operation.
```
isabelatkinson, line 51:
```suggestion
    /// A `Document` describing the fields that were updated or removed by the update operation.
    /// Only specified if `operation_type` is `OperationType::Update`.
```

src/change_stream/event.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    /// For operations of type "insert" and "replace", this key will contain the document being
+    /// inserted, or the new version of the document that is replacing the existing
+    /// document, respectively.
+    ///
+    /// For operations of type "update", when the `ChangeStream's` full document type is
+    /// `UpdateLookup`, this key will contain a copy of the full version of the document from
+    /// some point after the update occurred. If the document was deleted since the updated
+    /// happened, it will be `None`.
+    pub full_document: Option<T>,
```
isabelatkinson, line 62: I think this comment could also be improved. Here is the description from the manual:

> The document created or modified by the insert, replace, delete, update operations (i.e. CRUD operations).
>
> For insert and replace operations, this represents the new document created by the operation.
>
> For delete operations, this field is omitted as the document no longer exists.
>
> For update operations, this field only appears if you configured the change stream with fullDocument set to updateLookup. This field then represents the most current majority-committed version of the document modified by the update operation. This document may differ from the changes described in updateDescription if other majority-committed operations modified the document between the original update operation and the full document lookup.

We probably don't need all the details in the last paragraph, but I think including at least the first line as a summary would be nice.

src/change_stream/event.rs
```diff
@@ -0,0 +1,119 @@
...
+    Rename,
+
+    /// See [dropdatabase-event](https://docs.mongodb.com/manual/reference/change-events/#dropdatabase-event)
+    DropDatabase,
+
+    /// See [invalidate-event](https://docs.mongodb.com/manual/reference/change-events/#invalidate-event)
+    Invalidate,
+}
+
+/// Identifies which collection or database where an event occurred.
```
isabelatkinson, line 108:
```suggestion
/// Identifies the collection or database on which an event occurred.
```

src/change_stream/event.rs
```diff
@@ -0,0 +1,119 @@
...
+#[derive(Deserialize, Debug)]
+#[serde(untagged)]
+#[non_exhaustive]
+pub enum ChangeStreamEventSource {
+    /// Contains two fields: "db" and "coll" containing the database and collection name in which
+    /// the change happened.
+    Namespace(Namespace),
```
isabelatkinson, line 115: I think the documentation here was written with a `Document` in mind as the type for `Namespace`; instead, I think we can simplify this to say something like "The `Namespace` containing the database and collection in which the change occurred."

src/change_stream/event.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    // Contains the name of the dabatase in which the change happened.
```
isabelatkinson, line 117:
```suggestion
    // Contains the name of the database in which the change happened.
```

src/coll/mod.rs
abr-egn: Extending `aggregate` API to offer a similar typed interface runs into the same issue that this does with sessions, see below.

src/change_stream/options.rs
abr-egn: Sessions hadn't occurred to me and the api surface gets ugly with the `_with_session` x `_with_pipeline` matrix - currently stepping back to see if I can figure out a non-terrible way to expose this.

src/change_stream/mod.rs
```diff
@@ -0,0 +1,149 @@
+//! Contains the functionality for ChangeStreams.
```
isabelatkinson, line 1:
```suggestion
//! Contains the functionality for change streams.
```

src/change_stream/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+    Collection,
+    Cursor,
+    Database,
+};
+
+/// A `ChangeStream` streams the ongoing changes of its associated collection, database or
+/// deployment. `ChangeStream` instances should be created with method `watch` or
+/// `watch_with_pipeline` against the relevant target.
+///
+/// `ChangeStream`'s are "resumable", meaning that they can be restarted at a given place in the
```
isabelatkinson, line 33:
```suggestion
/// `ChangeStream`s are "resumable", meaning that they can be restarted at a given place in the
```

src/change_stream/options.rs
```diff
@@ -0,0 +1,79 @@
+//! Contains options for ChangeStreams.
+use serde::{Deserialize, Serialize};
+use serde_with::skip_serializing_none;
+use std::time::Duration;
+use typed_builder::TypedBuilder;
+
+use crate::{change_stream::event::ResumeToken, collation::Collation, options::AggregateOptions};
+
+/// These are the valid options that can be passed to the `watch` method for creating a
+/// [`ChangeStream`](crate::change_stream::ChangeStream).
```
isabelatkinson, line 10: We can also remove this link (and others in this file)

src/change_stream/options.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// Specifies a collation.
+    #[builder(default)]
+    #[serde(skip_serializing)]
+    pub collation: Option<Collation>,
+
+    /// The change stream will only provide changes that occurred at or after the specified
+    /// timestamp. Any command run against the server will return an operation time that can be
+    /// used here.
+    #[builder(default)]
+    pub start_at_operation_time: Option<i64>,
```
isabelatkinson, line 52: This value should never be negative, right? If so we should use a `u64`. (Alternatively we could use the `bson::Timestamp` type here; the spec isn't super prescriptive and just indicates `Timestamp` for this type.)

isabelatkinson (commented):
API looks good, mostly documentation suggestions. we'll also need to add the sync API, but I think it would be fine to file a separate ticket for that work

src/change_stream/options.rs
patrickfreed: Ah yeah, hadn't thought about that. One possible solution is to move the problem to `Cursor` / `ChangeStream` and have a `fn with_type<T>(self) -> Cursor<T>` method. Then it would look like:
```rust
let cs = collection.watch(vec![...], None).await?.with_type::<MyType>();
```
Not super pretty either but it does generalize well.

abr-egn: How would that be implemented, though?  `Cursor` has an internal buffer that can't be trivially converted.

patrickfreed: It just so happens that the ticket I'm working on now is to change the backing buffer of `Cursor` to `RawDocumentBuf`, which would make this change pretty easy. It used to be this way (with `Document` though) but was updated to use `T` as part of the perf work. Now that we have a more expressive raw API, we don't need to deserialize everything up front anymore.

src/change_stream/options.rs
abr-egn: Updated to use `bson::Timestamp`.

src/change_stream/options.rs
abr-egn: Not sure why?

src/change_stream/event.rs
abr-egn: Updated.

src/change_stream/event.rs
abr-egn: Updated.

src/change_stream/event.rs
abr-egn: Done.

src/coll/mod.rs
abr-egn: With `ChangeStream` now using `with_type`, the equivalent for `aggregate` will be `Cursor::with_type`.

> API looks good, mostly documentation suggestions. we'll also need to add the sync API, but I think it would be fine to file a separate ticket for that work

Filed RUST-1104 to track the sync API.

src/change_stream/options.rs
abr-egn: Updated the API to expose `ChangeStream::with_type` and added session methods and types.

src/change_stream/session.rs
```diff
@@ -0,0 +1,137 @@
...
+    ///     other_coll.insert_one_with_session(doc, None, &mut session).await?;
+    /// }
+    /// # Ok::<(), mongodb::error::Error>(())
+    /// # };
+    /// # }
+    /// ```
+    pub fn stream<'session>(
+        &mut self,
+        session: &'session mut ClientSession,
+    ) -> SessionCursorStream<'_, 'session, T> {
```
patrickfreed, line 107: We may want to introduce a new type here since the change stream's stream will need to be able to resume.

patrickfreed (commented):
API looks mostly good to me, just have one question about the type returned from `SessionChangeStream::stream`. Also, it looks like some of the doctests may need fixing up.

src/change_stream/session.rs
abr-egn: Done.  Named it 'SessionChangeStreamValues` to avoid the confusing `StreamStream` suffix it would otherwise have had.  Open to better naming suggestions if inspiration strikes.

abr-egn:
Fixed the doctests, but since they're testing an API that's not visible right now had to tag them with `ignore`.  Filed RUST-1106 so I don't forget to enable them as part of making the API visible when it's ready.

src/change_stream/session.rs
patrickfreed: Yeah I can't really think of a better name either at the moment.

patrickfreed (approved):
LGTM!

src/change_stream/options.rs
isabelatkinson: oh sorry misread, the types aren't in scope so we do need the links

isabelatkinson (approved):

PR #519, opened 2021-11-12
RUST-1039 Support deserializing `RawDocumentBuf` in cursors

patrickfreed:
RUST-1039

This PR updates the internals of the driver to support using `RawDocumentBuf` as the generic parameter of a cursor, thereby providing support for raw BSON cursors. From initial benchmarking, these cursors are _really_ fast, nearly C fast. Furthermore, users can now perform borrowed deserialization from the values retrieved from these cursors, which seems to be much faster than owned deserialization. I filed a follow up ticket for creating benchmarks for these cases to ensure we have an actual measurement of the performance differences.

It also seems that regular `Document` cursors may have been sped up by this change, though it's not immediately clear why. My guess is that we stopped using `#[serde(flatten)]`, which may be doing extra copies under the hood.

Note that the way this is implemented involves copying the results, so it's not truly zero-copy (the deserialization part from `RawDocumentBuf` to `T` can be, but iterating the cursor involves copies). I opted for this more straightforward path after doing some benchmarking that showed a true-zero-copy approach netted very little benefit when doing no deserialization at all and essentially zero benefit when doing borrowed deserialization. If we want to revisit this in the future we certainly can, though.

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -1,22 +1,50 @@
-use bson::{RawBson, RawDocument, Timestamp, doc};
+use bson::{doc, RawBson, RawDocument, Timestamp};
 use lazy_static::lazy_static;
 use serde::de::DeserializeOwned;
 
 use std::{collections::HashSet, sync::Arc, time::Instant};
 
 use super::{session::TransactionState, Client, ClientSession};
-use crate::{ClusterTime, bson::Document, cmap::{
-        conn::PinnedConnectionHandle, Connection, ConnectionPool, RawCommand, RawCommandResponse,
-    }, cursor::{session::SessionCursor, Cursor, CursorSpecification}, error::{
-        Error, ErrorKind, Result, RETRYABLE_WRITE_ERROR, TRANSIENT_TRANSACTION_ERROR,
+use crate::{
+    bson::Document,
+    cmap::{
+        conn::PinnedConnectionHandle,
+        Connection,
+        ConnectionPool,
+        RawCommand,
+        RawCommandResponse,
+    },
+    cursor::{session::SessionCursor, Cursor, CursorSpecification},
+    error::{
+        Error,
+        ErrorKind,
+        Result,
+        RETRYABLE_WRITE_ERROR,
+        TRANSIENT_TRANSACTION_ERROR,
         UNKNOWN_TRANSACTION_COMMIT_RESULT,
-    }, event::command::{CommandFailedEvent, CommandStartedEvent, CommandSucceededEvent}, operation::{
-        AbortTransaction, CommandErrorBody, CommandResponse, CommitTransaction, Operation,
-        Response, Retryability,
-    }, options::SelectionCriteria, sdam::{
-        HandshakePhase, SelectedServer, ServerType, SessionSupportStatus, TopologyType,
+    },
+    event::command::{CommandFailedEvent, CommandStartedEvent, CommandSucceededEvent},
+    operation::{
+        AbortTransaction,
+        CommandErrorBody,
+        CommandResponse,
+        CommitTransaction,
+        Operation,
+        Response,
+        Retryability,
+    },
+    options::SelectionCriteria,
+    sdam::{
+        HandshakePhase,
+        SelectedServer,
+        ServerType,
+        SessionSupportStatus,
+        TopologyType,
         TransactionSupportStatus,
-    }, selection_criteria::ReadPreference};
+    },
+    selection_criteria::ReadPreference,
+    ClusterTime,
+};
 
 lazy_static! {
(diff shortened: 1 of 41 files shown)


src/client/executor.rs
```diff
@@ -530,82 +523,79 @@ impl Client {
         let start_time = Instant::now();
         let command_result = match connection.send_raw_command(raw_cmd, request_id).await {
             Ok(response) => {
-                match T::Response::deserialize_response(&response) {
```
patrickfreed, line 533: Because `#[serde(flatten)]` doesn't play nicely with deserializing raw BSON values (see https://github.com/serde-rs/serde/issues/2106), response deserialization was changed back to being handled in the individual `handle_response` implementations of the various operations like it was before the performance improvements work. However, this time we can use `RawDocument` to extract the various things we need here that are not operation specific (e.g. operationTime) without having to deserialize into `Document` first, which was a big bottleneck before.

This also helps enable the `Cursor/ChangeStream::with_type` API being discussed in #518.

src/client/executor.rs
```diff
@@ -530,82 +523,79 @@ impl Client {
...
-                                .map(|error_response| error_response.into())
-                                .unwrap_or_else(|e| {
-                                    Error::from(ErrorKind::InvalidResponse {
-                                        message: format!(
-                                            "error deserializing command error: {}",
-                                            e
-                                        ),
-                                    })
-                                }))
+                async fn handle_response<T: Operation>(
```
patrickfreed, line 526: I wrote a little function here to enable the use of `?`.

src/client/executor.rs
```diff
@@ -530,82 +523,79 @@ impl Client {
...
                         }
+                    };
+
+                    let cluster_time: Option<ClusterTime> = raw_doc
+                        .get("$clusterTime")?
+                        .and_then(RawBson::as_document)
+                        .map(|d| bson::from_slice(d.as_bytes()))
+                        .transpose()?;
+
+                    let at_cluster_time = op.extract_at_cluster_time(raw_doc)?;
```
patrickfreed, line 558: Unfortunately, the `atClusterTime` field is not always in the same spot (`find` and `aggregate` return it within the `cursor` subdocument, `distinct` has it as a top level field), so we need to introduce this new operation method to handle that.

src/client/executor.rs
```diff
@@ -530,82 +523,79 @@ impl Client {
...
+                async fn handle_response<T: Operation>(
+                    client: &Client,
+                    op: &T,
+                    session: &mut Option<&mut ClientSession>,
+                    is_sharded: bool,
+                    response: RawCommandResponse,
+                ) -> Result<RawCommandResponse> {
+                    let raw_doc = RawDocument::new(response.as_bytes())?;
+
+                    let ok = match raw_doc.get("ok")? {
```
patrickfreed, line 535: doing linear scans of the command result for each of these keys shouldn't be a performance concern, since the top level of a command response rarely has many keys to sift through.

src/cursor/common.rs
```diff
@@ -28,25 +30,26 @@ use crate::{
...
 where
-    P: GetMoreProvider<DocumentType = T>,
+    P: GetMoreProvider,
 {
     #[derivative(Debug = "ignore")]
     provider: P,
     client: Client,
     info: CursorInformation,
-    buffer: VecDeque<T>,
+    buffer: VecDeque<RawDocumentBuf>,
```
patrickfreed, line 39: Since deserialization is no longer eager, we needed to update the stored buffer for the cursors to just contain raw BSON. Deserialization will happen as the cursor is iterated.

A cascading effect of this is that all of the cursor machinery (traits, result types, etc.) could have their generic parameters removed.

src/operation/mod.rs
```diff
@@ -183,42 +186,6 @@ impl<T: CommandBody> Command<T> {
     }
 }
 
-/// Trait modeling the behavior of a command response to a server operation.
-pub(crate) trait Response: Sized {
```
patrickfreed, line 187: This trait was no longer needed now that we extract these fields manually via the `RawDocument` API.

src/operation/run_command/mod.rs
```diff
@@ -106,55 +118,3 @@ pub(crate) struct Response {
     cluster_time: Option<ClusterTime>,
     recovery_token: Option<Document>,
 }
```
abr-egn, line 120: This doesn't seem to be used anywhere now.

patrickfreed: Good catch, removed. Not sure why clippy didn't pick up on this though.

abr-egn (approved):
LGTM!

isabelatkinson (approved):

PR #522, opened 2021-11-17
RUST-520 Implement `watch` for change streams

abr-egn:
RUST-520

This implements `watch[_with_session]` for `Client`, `Database`, and `Collection`.  It's implemented by directly executing an `Aggregate` operation rather than creating a new wrapping operation as the ticket suggests; I couldn't find any reason that would require the latter.

This also implements `with_type` for `[Session]Cursor` and `[Session]ChangeStream`; thanks to #519 this is a nearly trivial operation.


src/client/mod.rs
```diff
@@ -415,6 +445,29 @@ impl Client {
         }
     }
 
+    pub(crate) fn build_watch(
```
isabelatkinson, line 448: I think it would make more sense to move some of this work to `Aggregate`, i.e. with a `new_watch` function that does the work currently in this function to build the aggregation pipeline and create a new `Aggregate`. That way we could probably eliminate this method as initializing the `ChangeStreamData` is just one line, and the `watch` methods would look something like this:

```
// resolve options
let aggregate = Aggregate::new_watch(target, pipeline, options)?;
let cursor = client.execute_operation(aggregate)?;
Ok(ChangeStream::new(
    cursor,
    ChangeStreamData::new(pipeline, client, target, options),
))
```
In general I think it would be a bit more clear to associate the work in this function with the `Aggregate` struct rather than the `Client` struct. What do you think?

src/change_stream/options.rs
```diff
@@ -44,29 +67,40 @@ pub struct ChangeStreamOptions {
...
-    /// timestamp. Any command run against the server will return an operation time that can be
-    /// used here.
+    /// The read concern to use for the operation.
+    ///
+    /// If none is specified, the read concern defined on the object executing this operation will
+    /// be used.
     #[builder(default)]
-    pub start_at_operation_time: Option<Timestamp>,
+    #[serde(skip_serializing)]
+    pub read_concern: Option<ReadConcern>,
```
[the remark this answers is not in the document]
abr-egn: The spec says "The helper methods MUST determine a read concern for the operation in accordance with the Read and Write Concern specification.", which I took to mean resolution following the same pattern as `aggregate` / `find` / etc.

src/client/mod.rs
abr-egn: Good thought, done.

src/cursor/session.rs
```diff
@@ -184,6 +184,24 @@ where
...
 
+    /// Update the type streamed values will be parsed as.
+    pub fn with_type<D>(mut self) -> SessionCursor<D>
+    where
+        D: DeserializeOwned + Unpin + Send + Sync,
+    {
+        let out = SessionCursor {
+            client: self.client.clone(),
+            info: self.info.clone(),
+            buffer: self.buffer.drain(..).collect(),
```
patrickfreed, line 195: I think [`std::mem::replace`](https://doc.rust-lang.org/std/mem/fn.replace.html) might be more efficient here.

src/cursor/mod.rs
```diff
@@ -105,18 +107,32 @@ where
...
             _phantom: Default::default(),
         }
     }
 
+    /// Update the type streamed values will be parsed as.
+    pub fn with_type<D>(mut self) -> Cursor<D>
+    where
+        D: DeserializeOwned + Unpin + Send + Sync,
+    {
+        Cursor {
```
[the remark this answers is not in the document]
abr-egn: Unfortunately, it looks like `transmute` wouldn't be valid here - there's no guarantee that the monomorphized fields of `Cursor<T>` are laid out in the same order as the ones for `Cursor<D>`.  We could add `#[repr(C)]` to force that, but introducing that and `unsafe` doesn't seem worth saving an `unwrap`, especially since this is unlikely to be a performance bottleneck.

src/cursor/session.rs
abr-egn: Good call (and clippy prompted me to replace that with `std::mem::take`).

patrickfreed (approved):

isabelatkinson (approved):

PR #524, opened 2021-11-19
RUST-844 Fix flaky speculative authentication test

patrickfreed:
RUST-844

This PR fixes an issue which I think occurs when the test attempts to authenticate against a node that the credentials haven't replicated to yet. The fix is to just always run the test against the primary on replica sets.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -12,31 +12,19 @@ use derivative::Derivative;
 
 #[cfg(test)]
 use crate::options::ServerAddress;
-use crate::{
-    bson::Document,
-    change_stream::{
+use crate::{ClientSession, bson::Document, change_stream::{
         event::ChangeStreamEvent,
         options::ChangeStreamOptions,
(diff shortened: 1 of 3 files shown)


abr-egn (approved):

isabelatkinson (commented):
code changes lgtm, is the flakiness in this test any cause for concern in the driver? i.e. could the user creation not propagating affect users?

patrickfreed:
I think this is only an issue with this test and wouldn't be seen by users in practice really. The issue here is that we were creating a connection (rather than a `Client`) to whatever host happened to be the first in the list, and sometimes this one is a secondary that hasn't had the write acknowledged yet (we don't use `w: majority` when we create the user). Even if we did use a majority wc, there is still a _slight_ possibility we select a secondary that hasn't been replicated to yet, so it seemed most robust to just always use the primary.

isabelatkinson (approved):

PR #531, opened 2021-12-02
RUST-521 Implement naive streaming and resume token caching for change streams

abr-egn:
RUST-521

The `Stream` impl just forwards to the underlying cursor; this will need to be updated for RUST-522 to resume on error, but that seemed like a big enough chunk of work to be worth doing separately.

The token-tracking machinery is included in the base `Cursor` behavior but only exposed via `ChangeStream`; the other option here would have been introducing a new type parallel to `GenericCursor` and wrapping that directly in `ChangeStream`.  That didn't seem like a good way to go since it would either involve a lot of duplication of things like the buffering and stream impl, or adding another nested doll to encapsulate the shared logic (i.e. something like `Cursor` -> `CursorCommon` -> `GenericCursor` and `ChangeStream` -> `CursorCommon` -> `GenericCursorWithToken`).


src/cursor/common.rs
```diff
@@ -37,8 +38,10 @@ where
     client: Client,
     info: CursorInformation,
     buffer: VecDeque<RawDocumentBuf>,
+    post_batch_resume_token: Option<ResumeToken>,
```
abr-egn, line 41: Two tokens need to be tracked - the one returned by the most recent `getMore` call, and the one that will be returned to users via `resume_token()`.  The two are only the same at the end of a batch.

src/cursor/common.rs
```diff
@@ -136,14 +148,25 @@ where
...
                 }
             }
 
             match self.buffer.pop_front() {
                 Some(doc) => {
+                    self.cached_resume_token =
+                        if self.buffer.is_empty() && self.post_batch_resume_token.is_some() {
+                            self.post_batch_resume_token.clone()
+                        } else {
+                            doc.get("_id")?.map(|val| ResumeToken(val.to_raw_bson()))
```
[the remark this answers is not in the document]
abr-egn: Updated!  I also realized that I hadn't implemented any of this for the session types, so added that in, and refactored things a bit to be able to share the bulk of the implementation.

src/cursor/common.rs
```diff
@@ -98,63 +105,131 @@ where
...
+                    return Poll::Ready(Some(Ok(bson::from_slice(doc.as_bytes())?)))
+                }
+                BatchValue::Empty => continue,
+                BatchValue::Exhausted => return Poll::Ready(None),
+            },
+        }
+    }
+}
+
+pub(crate) struct NextInBatchFuture<'a, T>(&'a mut T);
```
abr-egn, line 199: It seems awkward to have to have a one-off struct to go from the desugared `fn() -> Poll<T>` form back to `.await`, but I guess this is the way until `poll_fn` stabilizes.

src/cursor/common.rs
```diff
@@ -98,63 +105,131 @@ where
...
+            post_batch_resume_token: self.post_batch_resume_token,
             info: self.info,
             pinned_connection: self.pinned_connection,
             _phantom: Default::default(),
         }
     }
 }
 
-impl<P, T> Stream for GenericCursor<P, T>
+pub(crate) trait CursorStream {
```
abr-egn, line 116: This trait allows the implementation of `poll_next` to be shared via the `stream_poll_next` fn; open to suggestions for a more descriptive name.

src/cursor/common.rs
```diff
@@ -98,63 +105,131 @@ where
...
+    fn poll_next_in_batch(&mut self, cx: &mut Context<'_>) -> Poll<Result<BatchValue>>;
```
abr-egn, line 117: I used the `next_in_batch` suggestion for internal methods but named the external `next_if_any` since the batching behavior is not the main point of the method from a user's perspective.

src/client/executor.rs
```diff
@@ -192,6 +203,88 @@ impl Client {
...
+    {
+        if let Some(handle) = session.transaction.pinned_connection() {
+            // Cursor operations on a transaction share the same pinned connection.
+            Ok(Some(handle.replicate()))
+        } else {
+            self.pin_connection_for_cursor(details)
+        }
+    }
+
+    pub(crate) async fn execute_watch<T>(
```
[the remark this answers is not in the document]
abr-egn: Unfortunately we need the `CursorSpecification` in order to get the initial resume token, and that's not preserved or exposed by `Cursor`.

src/change_stream/event.rs
```diff
@@ -15,7 +23,31 @@ use serde::{Deserialize, Serialize};
 /// [here](https://docs.mongodb.com/manual/changeStreams/#change-stream-resume-token) for more
 /// information on resume tokens.
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct ResumeToken(pub(crate) Bson);
+pub struct ResumeToken(pub(crate) RawBson);
+
+impl ResumeToken {
+    pub(crate) fn initial(
```
isabelatkinson, line 29: nit: can we update this so that both values don't always need to be created? e.g.
```
match spec.post_batch_resume_token {
    Some(token) if spec.initial_buffer.is_empty() => token,
    None => // token from options
}
```

isabelatkinson (approved):
just one small suggestion! adding my LGTM now so as not to block this while I'm gone

patrickfreed (commented):
looks good! just have one suggestion

src/change_stream/mod.rs
```diff
@@ -100,16 +113,48 @@ where
...
+    /// most up to date token is received, e.g.
+    ///
+    /// ```ignore
+    /// # use mongodb::{Client, error::Result};
+    /// # async fn func() -> Result<()> {
+    /// # let client = Client::with_uri_str("mongodb://example.com").await?;
+    /// # let coll = client.database("foo").collection("bar");
+    /// let mut change_stream = coll.watch(None, None).await?;
+    /// let mut resume_token = None;
+    /// loop {
```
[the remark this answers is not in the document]
abr-egn: Ah, good thought, done.

src/change_stream/event.rs
abr-egn: That's much nicer, thank you!

patrickfreed (approved):
LGTM!

PR #532, opened 2021-12-02
RUST-1122 Fix x509 auth for pkcs8 keys and Atlas free tier

abr-egn:


src/client/auth/x509.rs
```diff
@@ -58,7 +58,7 @@ pub(super) async fn authenticate_stream(
             .auth_response_body("MONGODB-X509")?,
     };
 
-    if server_response.get_str("dbname") != Ok("$external") {
+    if server_response.get_i32("ok") != Ok(1) {
```
patrickfreed, line 61: sometimes the server can return different types for this integer (and sometimes even a double), so we use `bson_util::get_int` for these kinds of checks to ensure all types are covered.

src/client/options/mod.rs
```diff
@@ -846,22 +846,40 @@ impl TlsOptions {
             };
 
             file.seek(SeekFrom::Start(0))?;
```
patrickfreed, line 848: since we're only taking the first one, it may be a bit clearer to perform this whole process in a single loop which can exit early using `read_one`. The `rustls` [example code](https://github.com/rustls/rustls/blob/main/rustls-mio/examples/tlsclient.rs#L420) actually has a good snippet for this:
```rust
fn load_private_key(filename: &str) -> rustls::PrivateKey {
    let keyfile = fs::File::open(filename).expect("cannot open private key file");
    let mut reader = BufReader::new(keyfile);

    loop {
        match rustls_pemfile::read_one(&mut reader).expect("cannot parse private key .pem file") {
            Some(rustls_pemfile::Item::RSAKey(key)) => return rustls::PrivateKey(key),
            Some(rustls_pemfile::Item::PKCS8Key(key)) => return rustls::PrivateKey(key),
            None => break,
            _ => {}
        }
    }

    panic!(
        "no keys found in {:?} (encrypted keys not supported)",
        filename
    );
}
```
I looks like we may need to add the `rustls-pemfile` crate as a dependency for this though, since it seems that `rustls` broke out the logic for reading these files into a separate crate. It's likely we have it as a transitive dependency already.

src/client/auth/x509.rs
abr-egn: It turns out that `ok: 1` is itself not always present, and apparently the thing to do is just not inspect the server response at all (i.e. https://github.com/mongodb/mongo-python-driver/blob/89f41cfbd2560b41592dadf6e1b59f80c33ec207/pymongo/auth.py#L424).

src/client/options/mod.rs
abr-egn: Updated.

src/coll/mod.rs
```diff
@@ -1140,6 +1140,7 @@ where
             let docs: Vec<&T> = ds.iter().skip(n_attempted).map(Borrow::borrow).collect();
             let insert = Insert::new(self.namespace(), docs, options.clone());
 
+            #[allow(clippy::needless_option_as_deref)]
```
abr-egn, line 1143: This appears to be a clippy false-positive; the suggestion (just `session`) doesn't work because this is in a loop, which I think is context clippy isn't looking at.

patrickfreed: Yep, see https://github.com/rust-lang/rust-clippy/issues/8047. I actually already [pushed a fix](https://github.com/mongodb/mongo-rust-driver/commit/91987a04e4671fddf3a54376000d0a30c9f30436) for these errors last night, so you can omit these changes from here.

src/client/auth/x509.rs
patrickfreed: Every command should always return an `ok` value, so that sounds like a bug in the server / proxy. What circumstances could trigger that? (I think in the Python example, the `command` method probably will throw an exception if `ok: 0` is encountered). I think we need to check _some_ field here to properly determine if authentication worked or not.

abr-egn: This is observed behavior in the Evergreen x509 tests (i.e. [here](https://evergreen.mongodb.com/lobster/evergreen/task/mongo_rust_driver_x509_auth__async_runtime~tokio_auth_and_tls~auth_and_tls_os~macos_10.14_test_x509_auth_patch_f93fb4057b30490d9fb407b23451daa37646c6c4_61aa31932a60ed5e66cfdcbb_21_12_03_15_02_47/0/task#bookmarks=0%2C1413&shareLine=1105)); it looks like those are running against mongodb 4.4 in a standalone configuration.

Updated the check to look for either `ok: 1` or `dbname: $external`.

src/coll/mod.rs
abr-egn: Ah, thanks; removed.

src/client/auth/x509.rs
patrickfreed: Ah I think I understand now. The `ok` is omitted when the authentication is performed speculatively (i.e. we include the first step of authentication in our original handshake command, and the server includes their response to it in their initial handshake response). This happens because the authentication response is returned as a subdocument of a larger command response, and we extract it and pass it down through `authenticate_stream`. I see now that in the Python example, it returns early in that case. I think we can take a similar approach here to Python: if speculative auth happened (i.e. we have a server first value), just return `Ok(())`. Otherwise, do the next round and ensure `ok: 1`.

abr-egn: Ah, I see, thanks.  Updated.

patrickfreed (approved):
LGTM!

Zageron:
Hype

isabelatkinson (approved):

PR #535, opened 2021-12-13
RUST-1162 Fix "unpin after transient error within a txn and commit" test failure

patrickfreed:
RUST-1162

This PR fixes a race in the v2 runner between events being emitted and the runner asserting that they have been received. It does so by waiting on a channel with a given timeout instead of checking once that all the events are there after executing all the operations.


patrickfreed:
This PR is waay out of date so I'm going to close it.

PR #539, opened 2021-12-16
RUST-1103 Update microbenchmarks to use raw BSON

patrickfreed:
RUST-1103

This PR introduces two new benchmarks: find many (raw BSON) and find many (serde struct). It also updates the multi export benchmark to use raw BSON, resulting in a speedup.

I didn't include any new versions of the find one by id benchmark, since using raw BSON didn't actually change anything there. I believe this is due to the test being bottlenecked by network I/O rather than BSON serialization. 

I also didn't include any borrowed deserialization benchmarks for any case, since it also didn't make any noticeable difference. I think we'd need to use a data set that has documents much larger than the tweet one provided with the microbenchmarks.

Lastly, I didn't update the multi-import benchmark to use raw documents, since it currently doesn't result in a speedup due to RUST-1132.

example patch: https://evergreen.mongodb.com/version/61bbc5421e2d177b9452baf4

Some numbers from my machine:
| benchmark                | median iteration time |
|:------------------------:|:---------------------:|
| find many (Document)     | 0.166s                |
| find many (raw BSON)     | 0.022s                |
| find many (serde struct) | 0.068s                |
| find many (C, raw BSON)            | 0.019                 |
| find many (Go, raw BSON)           | 0.012                 |
| find many (Go, bson.M)   | 0.366                 |

So overall, performance is quite good (especially serde struct, which used to be 0.15ish but improved due to the internal use of `RawDocumentBuf`), but we are a bit slower than Go and C in the raw BSON case. This is likely due to extra copies incurred by going through `serde` and could be avoided, but it didn't seem like a high enough priority to focus on now.

### DIFF
--- a/benchmarks/src/main.rs
+++ b/benchmarks/src/main.rs
@@ -151,7 +151,7 @@ const BSON_BENCHES: &[&'static str] = &[
 /// Benchmarkes included in the "SingleBench" composite.
 /// This consists of all the single-doc benchmarks except Run Command.
 const SINGLE_BENCHES: &[&'static str] = &[
-    FIND_MANY_BENCH_RAW,
+    FIND_ONE_BENCH,
     SMALL_DOC_INSERT_ONE_BENCH,
     LARGE_DOC_INSERT_ONE_BENCH,
 ];
@@ -175,7 +175,7 @@ const PARALLEL_BENCHES: &[&'static str] = &[
 
 /// Benchmarks included in the "ReadBench" composite.
 const READ_BENCHES: &[&'static str] = &[
-    FIND_MANY_BENCH_RAW,
+    FIND_ONE_BENCH,
     FIND_MANY_BENCH_RAW,
     GRIDFS_DOWNLOAD_BENCH,
     LDJSON_MULTI_EXPORT_BENCH,
--- a/benchmarks/src/main.rs
+++ b/benchmarks/src/main.rs
@@ -193,7 +193,7 @@ const WRITE_BENCHES: &[&'static str] = &[
     GRIDFS_MULTI_UPLOAD_BENCH,
 ];
 
-const MAX_ID: usize = 17;
+const MAX_ID: u8 = BenchmarkId::FindManySerde as u8;
 
 async fn run_benchmarks(
     uri: &str,
@@ -529,6 +529,12 @@ fn parse_ids(matches: ArgMatches) -> HashSet<BenchmarkId> {
 #[cfg_attr(feature = "tokio-runtime", tokio::main)]
 #[cfg_attr(feature = "async-std-runtime", async_std::main)]
 async fn main() {
+    // ensure MAX_ID is kept up to date.
+    assert!(
+        BenchmarkId::try_from(MAX_ID + 1).is_err(),
+        "MAX_ID not up to date"
+    );
+
     let matches = App::new("RustDriverBenchmark")
         .version(env!("CARGO_PKG_VERSION"))
         .about("Runs performance micro-bench")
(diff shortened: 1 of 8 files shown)


benchmarks/src/main.rs
```diff
@@ -86,6 +92,52 @@ const GRIDFS_UPLOAD_BENCH: &'static str = "GridFS upload";
 const LDJSON_MULTI_IMPORT_BENCH: &'static str = "LDJSON multi-file import";
 const GRIDFS_MULTI_UPLOAD_BENCH: &'static str = "GridFS multi-file upload";
 
+#[derive(Copy, Clone, num_enum::TryFromPrimitive, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[repr(u8)]
+enum BenchmarkId {
```
patrickfreed, line 97: the previous approach of using a vec of bool values was very error prone and hard to extend, so I updated it to use an enum instead.

benchmarks/src/main.rs
```diff
@@ -141,304 +193,334 @@ const WRITE_BENCHES: &[&'static str] = &[
...
-            num_iter: 10_000,
-            path: DATA_PATH.join("extended_bson").join("full_bson.json"),
-        };
-        println!("Running {}...", FULL_BSON_ENCODING);
-        let bson_full_encode =
-            bench::run_benchmark::<BsonEncodeBenchmark>(bson_full_encode_options).await?;
-
-        comp_score += score_test(bson_full_encode, FULL_BSON_ENCODING, 57.34, more_info);
+    let mut sorted_ids: Vec<BenchmarkId> = ids.into_iter().collect();
+    sorted_ids.sort();
```
abr-egn, line 207: It might be worth randomizing the id order rather than sorting to eliminate accidental cache priming effects and the like.

abr-egn (approved):

isabelatkinson (approved):

benchmarks/src/main.rs
```diff
@@ -141,304 +193,334 @@ const WRITE_BENCHES: &[&'static str] = &[
     GRIDFS_MULTI_UPLOAD_BENCH,
 ];
 
-async fn run_benchmarks(uri: &str, more_info: bool, ids: &[bool]) -> Result<CompositeScore> {
+const MAX_ID: usize = 17;
```
abr-egn, line 196: Rather than a hand-maintained constant, this could be a terminal arm of the `BenchmarkId` enum.

patrickfreed: I updated this to be defined in terms of the enum, but I stopped short of adding a new, terminal variant (if that's what you were suggesting) so that we wouldn't have to handle it everywhere we match on `BenchmarkId`. Instead, I added an assertion that `MAX_ID` is in fact the max element in the enum, which should ensure this doesn't accidentally become out of sync.

benchmarks/src/main.rs
patrickfreed: I think the ergonomics of having a consistent order for when doing manual benchmarking / debugging outweigh the potential risks of cache priming, though that's in part because I don't have a ton of knowledge as to how severe these effects can be. What do you think?

abr-egn: I don't have a whole lot of knowledge here myself, so if there's good concrete benefit to the sorting keeping it that way sounds fine.

benchmarks/src/main.rs
abr-egn: Yup, a terminal variant is what I was thinking of, but your way covers it nicely as well.

abr-egn (approved):

PR #541, opened 2021-12-21
Updated readme reference to driver version to 2.1.0

pkdone:
The latest version always needs to be reflected in the README because, for better or worse, the [MongoDB Rust Driver official documentation section "Installation"](https://docs.mongodb.com/drivers/rust/#installation), delegates to this repo's "master" branch README for user installation instructions and we of course don't want to be encouraging users to use old versions of the driver by default.


abr-egn (approved):
Thanks!

PR #546, opened 2022-01-04
RUST-1133 Update driver to use the raw BSON API

patrickfreed:
RUST-1133

We implemented some of the raw BSON functionality in the driver itself as a temporary measure until the BSON library had its own support for it. Now that it does, we can remove these driver-only implementations and update to use the BSON library's versions instead.

### DIFF
--- a/src/operation/insert/mod.rs
+++ b/src/operation/insert/mod.rs
@@ -3,7 +3,7 @@ mod test;
 
 use std::{collections::HashMap, convert::TryInto};
 
-use bson::{oid::ObjectId, Bson, RawArrayBuf};
+use bson::{oid::ObjectId, Bson, RawArrayBuf, RawDocumentBuf};
 use serde::Serialize;
 
 use crate::{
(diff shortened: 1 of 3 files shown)


src/operation/insert/mod.rs
```diff
@@ -66,31 +65,32 @@ impl<'a, T: Serialize> Operation for Insert<'a, T> {
             .take(description.max_write_batch_size as usize)
             .enumerate()
         {
-            let mut doc = bson::to_vec(d)?;
-            let id = match bson_util::raw_get(doc.as_slice(), "_id")? {
-                Some(b) => b,
+            let mut doc = bson::to_raw_document_buf(d)?;
+            let id = match doc.get("_id")? {
```
patrickfreed, line 69: I ended up using basically the same prepend `_id` logic here.

abr-egn (approved):
LGTM!

isabelatkinson (approved):

PR #552, opened 2022-01-10
minor: fix external crate links

isabelatkinson:
Fixes the links to external crates on our [docs.rs page](https://docs.rs/mongodb/latest/mongodb/) that are currently broken.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,8 +1,8 @@
 //! This crate contains the officially supported MongoDB Rust driver, a
 //! client side library that can be used to interact with MongoDB deployments
 //! in Rust applications. It uses the [`bson`] crate for BSON support.
-//! The driver contains a fully async API that supports either [`tokio`] (default)
-//! or [`async_std`], depending on the feature flags set. The driver also has
+//! The driver contains a fully async API that supports either [`tokio`](https://docs.rs/tokio) (default)
+//! or [`async_std`](https://docs.rs/async_std), depending on the feature flags set. The driver also has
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1,8 +1,8 @@
 //! This crate contains the officially supported MongoDB Rust driver, a
 //! client side library that can be used to interact with MongoDB deployments
-//! in Rust applications. It uses the [`bson`](docs.rs/bson) crate for BSON support.
```
isabelatkinson, line 3: Links aren't required for external crates specified in this format. `docs.rs/bson` was being treated as a relative link (i.e. `https://docs.rs/mongodb/latest/mongodb/docs.rs/bson`); omitting it fixes the issue and properly links to the bson crate.

src/lib.rs
```diff
@@ -1,8 +1,8 @@
...
-//! The driver contains a fully async API that supports either [`tokio`](docs.rs/tokio) (default)
-//! or [`async-std`](docs.rs/async-std), depending on the feature flags set. The driver also has
+//! in Rust applications. It uses the [`bson`] crate for BSON support.
+//! The driver contains a fully async API that supports either [`tokio`] (default)
+//! or [`async-std`](https://docs.rs/async_std), depending on the feature flags set. The driver also has
```
isabelatkinson, line 5: Using an absolute link here because async-std is an optional dependency, which causes docs tests to fail when the async-std feature is not enabled.

abr-egn (approved):

patrickfreed (approved):

PR #558, opened 2022-01-12
println!({:?}) instead of println!("{})

Paulo-21:
I had this error : 
error[E0277]: `Vec<std::string::String>` doesn't implement `std::fmt::Display


abr-egn:
Can you give more information about how you got that error?  In the example you've updated in this PR, `db_name` is a `String`, which implements `Display`, so using the debug formatter isn't needed.

PR #561, opened 2022-01-18
RUST-1149 Prose tests for change streams.

abr-egn:
RUST-1149

This implements the prose tests for change streams, with the exception of test 6, which is (a) tricky to observe and (b) essentially guaranteed to be correct by the design of the Rust driver.  Some relatively minor updates were required to the change stream implementation as a result of bugs found by these tests.

### DIFF
--- a/src/change_stream/event.rs
+++ b/src/change_stream/event.rs
@@ -22,7 +22,7 @@ use serde::{Deserialize, Serialize};
 /// See the documentation
 /// [here](https://docs.mongodb.com/manual/changeStreams/#change-stream-resume-token) for more
 /// information on resume tokens.
-#[derive(Clone, Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct ResumeToken(pub(crate) RawBson);
 
 impl ResumeToken {
@@ -52,7 +52,7 @@ impl ResumeToken {
 
 /// A `ChangeStreamEvent` represents a
 /// [change event](https://docs.mongodb.com/manual/reference/change-events/) in the associated change stream.
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, PartialEq)]
 #[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct ChangeStreamEvent<T> {
@@ -104,7 +104,7 @@ pub struct ChangeStreamEvent<T> {
 }
 
 /// Describes which fields have been updated or removed from a document.
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, PartialEq)]
 #[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct UpdateDescription {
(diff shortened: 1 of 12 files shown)


src/change_stream/mod.rs
```diff
@@ -236,50 +240,56 @@ where
...
+                    let args = self.args.clone();
+                    let mut data = self.data.take();
+                    data.implicit_session = self.cursor.take_implicit_session();
+                    self.pending_resume = Some(Box::pin(async move {
+                        let new_stream: Result<ChangeStream<ChangeStreamEvent<()>>> = client
+                            .execute_watch(args.pipeline, args.options, args.target, Some(data))
+                            .await;
+                        new_stream.map(|cs| cs.with_type::<T>())
+                    }));
+                    // Iterate the loop so the new future gets polled and can register wakers.
```
abr-egn, line 286: This was the point of wrapping the logic in a loop - there needs to be an underlying future that's been polled rather than just returning `Poll::Pending` directly because with the latter this future will never be polled again due to the lack of registered wake conditions.

src/change_stream/event.rs
```diff
@@ -142,13 +147,11 @@ pub enum OperationType {
...
 
 /// Identifies the collection or database on which an event occurred.
-#[derive(Deserialize, Debug)]
-#[serde(untagged)]
-#[non_exhaustive]
-pub enum ChangeStreamEventSource {
-    /// The [`Namespace`] containing the database and collection in which the change occurred.
-    Namespace(Namespace),
+#[derive(Deserialize, Debug, PartialEq, Eq)]
+pub struct ChangeStreamEventSource {
```
isabelatkinson, line 151: is it possible this could change in the future? if so, we should keep it `non_exhaustive`

src/test/change_stream.rs
```diff
@@ -0,0 +1,656 @@
...
+
+/// Prose test 1: ChangeStream must continuously track the last seen resumeToken
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn tracks_resume_token() -> Result<()> {
+    let _guard = LOCK.run_concurrently().await;
+
+    let (client, coll, mut stream) = match init_stream("track_resume_token", false).await? {
+        Some(t) => t,
+        None => return Ok(()),
```
isabelatkinson, line 62: can we log that we're skipping due to topology here? (in the spirit of #523)

src/change_stream/event.rs
abr-egn: Good call, done.

src/test/change_stream.rs
abr-egn: `init_stream` logs when it returns None :)

patrickfreed (commented):
overall looks really good, just have a few minor comments

src/test/change_stream.rs
```diff
@@ -0,0 +1,656 @@
...
+
+    let _getmore_guard = FailPoint::fail_command(
+        &["getMore"],
+        FailPointMode::Times(1),
+        FailCommandOptions::builder().error_code(43).build(),
+    )
+    .enable(&client, None)
+    .await?;
+
+    let _killcursors_guard = FailPoint::fail_command(
```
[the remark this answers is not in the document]
abr-egn: Ah, I see - fixed.

isabelatkinson (approved):

patrickfreed (approved):
LGTM!

PR #564, opened 2022-01-25
RUST-523 Support change streams in unified spec tests

abr-egn:
RUST-523

This pulls in the unified spec test for change streams and adds support to the Rust unified runner for creating and iterating change streams.  I also added more detailed mismatch reporting to the runner as part of debugging initial failures.


src/change_stream/event.rs
```diff
@@ -114,10 +114,25 @@ pub struct UpdateDescription {
...
+
+    /// Arrays that were truncated in the `Document`.
+    pub truncated_arrays: Vec<TruncatedArray>,
+}
+
+/// Describes an array that has been truncated.
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct TruncatedArray {
```
patrickfreed, line 126: Is there a ticket for adding support for this (from a DRIVERS one) that we ought to close out?

src/test/spec/json/change-streams/legacy/change-streams-errors.json
```diff
@@ -0,0 +1,153 @@
+{
```
patrickfreed, line 1: we're not actually running any of the legacy ones right? i think we ought to just leave them out then to avoid any red-herring greps and what not.

patrickfreed (commented):
Overall looks good! I like the added detail to the matching. Only have a few minor questions/suggestions

src/change_stream/event.rs
abr-egn: It looks like this was https://jira.mongodb.org/browse/RUST-659, which was closed out as "Will handle in our initial implementation of Change Streams."

src/test/spec/json/change-streams/legacy/change-streams-errors.json
abr-egn: Good call, done.

src/test/spec/unified_runner/entity.rs
```diff
@@ -44,14 +45,21 @@ pub struct ClientEntity {
     observe_sensitive_commands: bool,
 }
 
+impl ClientEntity {
```
isabelatkinson, line 48: I don't think this method is necessary -- `ClientEntity` implements `Deref` with a target type of `Client` so you should be able to call `watch` directly on a `ClientEntity` (where this method is used).

isabelatkinson (commented):
looks good! nice to see some upgrades to the matcher

patrickfreed (approved):
LGTM mod Isabel's comment

src/test/spec/unified_runner/entity.rs
abr-egn: I missed that, thank you!  Removed.

isabelatkinson (approved):
lgtm! looks like there's something going on with evergreen but i don't think we need another full CI run here

PR #567, opened 2022-01-28
RUST-1077 Update read/write concern document tests

isabelatkinson:


src/test/spec/read_write_concern/document.rs
```diff
@@ -15,80 +12,108 @@ struct TestFile {
...
-            ("wtimeoutMS", Bson::Int32(i)) if i > 0 => {
-                write_concern.w_timeout = Some(Duration::from_millis(i as u64));
+        if let Some(specified_write_concern_document) = test_case.write_concern {
+            let specified_write_concern =
+                match bson::from_document::<WriteConcern>(specified_write_concern_document) {
+                    Ok(write_concern) => {
+                        let is_valid = write_concern.validate().is_ok();
+                        assert_eq!(
+                            is_valid, test_case.valid,
+                            "Write concern deserialization/validation should fail: {}",
```
abr-egn, line 38: This message seems like it could be confusing for the case where the test case is expected to be valid but isn't.

src/test/spec/read_write_concern/document.rs
```diff
@@ -15,80 +12,108 @@ struct TestFile {
...
-            _ => {}
-        };
-    }
 
-    Some(write_concern)
-}
+            let actual_write_concern_document = bson::to_document(&specified_write_concern)
+                .unwrap_or_else(|_| {
+                    panic!(
+                        "Write concern serialization should succeed: {}",
```
abr-egn, line 78: Might be good to include the error in the output here for diagnostics (likewise for the read concern doc).

patrickfreed (approved):

src/test/spec/read_write_concern/document.rs
isabelatkinson: @abr-egn looks like I refactored to address this comment -- now `test_case.valid` is always expected to be true, so the failure message should work in all cases

abr-egn (approved):
LGTM!

PR #568, opened 2022-01-28
RUST-812 Reduce flakiness of `in_window::load_balancing_test`

patrickfreed:
RUST-812

This PR makes the following changes to hopefully eliminate the remaining flakiness in the `in_window::load_balancing_test`, which verifies that the operation-count tracking server selection algorithm successfully routes requests away from busy servers:
- Add a 30 second local threshold config to the client, ensuring both mongoses are always in the latency window
- Use SDAM monitoring events to ensure both mongoses are discovered before starting the test

From testing, it seems that this (in conjunction with skipping on auth + tls) eliminates the random failures, but we'll have to wait and see if they crop up again. Should these changes prove successful, I'll update the spec to require them so other drivers can avoid similar errors (DRIVERS-1926)

### DIFF
--- a/src/bson_util/mod.rs
+++ b/src/bson_util/mod.rs
@@ -191,7 +191,6 @@ where
 
 /// The size in bytes of the provided document's entry in a BSON array at the given index.
 pub(crate) fn array_entry_size_bytes(index: usize, doc_len: usize) -> u64 {
-    // 
     //   * type (1 byte)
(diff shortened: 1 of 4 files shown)


patrickfreed:
The lint errors appear  / are fixed on master already

abr-egn (approved):
LGTM!

isabelatkinson (approved):

PR #570, opened 2022-02-03
DRIVERS-1713 Convert change stream spec tests to the unified format.

abr-egn:
DRIVERS-1713

This was a semi-automated conversion; some of the tests needed minor tweaks to account for things like differences in matching semantics.  This PR also includes the conversion script.


abr-egn:
Good to go for review; the remaining evergreen failures are known flakes or fallout from RUST-1163.

src/test/spec/unified_runner/entity.rs
```diff
@@ -45,6 +45,12 @@ pub struct ClientEntity {
     observe_sensitive_commands: bool,
 }
 
+impl ClientEntity {
```
patrickfreed, line 48: `ClientEntity` implements `Deref` for `Client`, so I think we can omit this.

etc/change_stream_legacy/src/main.rs
```diff
@@ -0,0 +1,416 @@
...
+        visit_mut(map, &|_, val| match val {
+            Value::Number(num) if num.as_i64() == Some(42) => *val = exists(),
+            Value::String(s) if s == "42" => *val = exists(),
+            _ => (),
+        });
+    }
+
+    fn parse_success(file: &legacy::File, mut suc: serde_yaml::Mapping) -> Operation {
+        visit_mut(&mut suc, &|key, val| {
+            if key == "fullDocument" {
```
patrickfreed, line 326: Why is this extra step for `fullDocument` necessary? I noticed it adds an assertion that didn't always exist in the legacy version (e.g. in "$changeStream must be the first stage in a change stream pipeline sent to the server").

etc/change_stream_legacy/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
```
patrickfreed, line 1: The code for this is great, and it was also super helpful for reviewing this PR. That said, I don't think we'll want to actually check this in when merging this PR, since I don't believe we'll need to use it again now that all the tests have been converted. It's good to keep around for when submitting these conversions to the specs repo though.

etc/change_stream_legacy/src/main.rs
abr-egn: The legacy test matching was subset at all levels, whereas the unified is more strict and doesn't allow unexpected keys in most places.  Practically, what this meant is that doing the test without adding some `{ $$exists: true }` clauses would fail because the server output had fields not listed in the test.  The `fullDocument._id` case showed up in most of the tests, so it was worth automating.

etc/change_stream_legacy/Cargo.toml
abr-egn: Is there any downside to leaving it checked in?  I've found it very useful to have artifacts like this easily present when trying to track down change history.

patrickfreed: I was thinking it might add some noise to greps through the code base, but that's basically it. It's fine with me to keep it checked in if you think it's worth it.

patrickfreed (approved):
LGTM!

PR #572, opened 2022-02-10
RUST-43 Add change streams examples for documentation

abr-egn:


src/change_stream/options.rs
```diff
@@ -114,6 +114,7 @@ pub enum FullDocumentType {
     /// The
     /// [`ChangeStreamEvent::full_document`](crate::change_stream::event::ChangeStreamEvent::full_document)
     /// field will be populated with a copy of the entire document that was updated.
+    #[serde(rename = "updateLookup")]
```
patrickfreed, line 117: This seems like something that ought to have been caught by testing, are the spec tests missing coverage of this? If so, we should file a DRIVERS ticket.

src/test/documentation_examples/mod.rs
```diff
@@ -1755,6 +1754,72 @@ async fn index_examples() -> Result<()> {
...
+            }
+            Result::Ok(())
+        })
+        .unwrap();
+
+    #[allow(unused_variables)]
+    {
+        // Start Changestream Example 1
+        let mut stream = inventory.watch(None, None).await?;
+        let next = stream.next().await.transpose()?;
```
patrickfreed, line 1795: Rather than using `next()` and `transpose`, we can also import `TryStreamExt` and then just `try_next()`. I think that's what we have in our existing examples. 

On a related note, I think we might want to include the `(Try)StreamExt` import in the examples, since users will probably just copy/paste them and figuring out they need to import that trait could be difficult. That said, they'll also need to figure out they need a `futures` dependency, so maybe it doesn't really matter.

src/change_stream/options.rs
abr-egn: Filed DRIVERS-2201.

src/test/documentation_examples/mod.rs
abr-egn: Updated to use `TryStreamExt` and include that in the examples.  I don't think the `futures` dependency needs to be worried about; anyone more than passingly familiar with Rust will know that `use foo::bar;` means you need to pull in crate `foo`.

isabelatkinson (commented):
Not necessarily something we need to address completely in this PR, but the `documentation_examples.rs` file is getting pretty long. When we introduced the transactions examples we did so in the top-level `tests` directory rather than `src/test` and I think it would be nice to continue doing that with documentation examples going forward (incl. here). This helps to ensure that we're not accidentally using anything in the doc examples that isn't part of the public API.

src/test/documentation_examples/mod.rs
abr-egn: Derp, done.

patrickfreed (approved):

isabelatkinson (approved):

PR #578, opened 2022-02-17
RUST-995 Use the tokio runtime for the sync API

isabelatkinson:


src/runtime/mod.rs
```diff
@@ -74,29 +72,17 @@ impl AsyncRuntime {
...
-            match TokioCallingContext::current() {
-                TokioCallingContext::Async(_handle) => {
-                    tokio::task::block_in_place(|| futures::executor::block_on(fut))
-                }
-                TokioCallingContext::Sync => {
-                    panic!("block_on called from tokio outside of async context")
-                }
-            }
-        }
+        match self {
```
abr-egn, line 80: I don't think this (or `spawn`, for that matter) need the `match self` - and, thinking about it for a little more, I don't think `Runtime` needs to be an enum at all.  Most of the methods ignore the value in favor of `cfg` blocks, and it'll only ever have a single actual discriminant.

abr-egn (commented):
Looks good!  A few smallish comments.

src/runtime/mod.rs
isabelatkinson: yeah that's a good point, it's also a little confusing as the syntax of calling these methods (i.e. `RUNTIME.method()`) makes it appear that they're being called on the async runtimes directly. I refactored a bit to remove the enum and make the methods static functions within the `runtime` module.

abr-egn: Very nice!

abr-egn (approved):
LGTM!

.evergreen/run-tokio-tests.sh
```diff
@@ -15,4 +15,10 @@ FEATURE_FLAGS="zstd-compression,snappy-compression,zlib-compression"
 echo "cargo test options: --features $FEATURE_FLAGS ${OPTIONS}"
 
 RUST_BACKTRACE=1 cargo test --features $FEATURE_FLAGS $OPTIONS | tee results.json
-cat results.json | cargo2junit > results.xml
+cat results.json | cargo2junit > async-tests.xml
+RUST_BACKTRACE=1 cargo test sync --features sync $FEATURE_FLAGS $OPTIONS | tee sync-tests.json
+cat sync-tests.json | cargo2junit > sync-tests.xml
+RUST_BACKTRACE=1 cargo test --doc sync --features sync $FEATURE_FLAGS $OPTIONS | tee sync-doc-tests.json
```
patrickfreed, line 21: I'm not exactly sure what's wrong here, but this test invocation seems to have some formatting errors per the evg output: https://evergreen.mongodb.com/task_log_raw/mongo_rust_driver_tests__async_runtime~tokio_auth_and_tls~auth_and_tls_compressor~no_compression_os~ubuntu_18.04_arm64_test_5.0_replica_set_patch_3b3e6f91372dc174a7b654509aa9963042bcd9de_621d44d632f417468148eef2_22_02_28_21_55_35/0?type=T#L2142

What's weird is that it doesn't appear to happen every time :thinking: 
There's also a junit-report-merger error there too that seems unrelated but also worth looking into.

src/lib.rs
```diff
@@ -367,20 +367,13 @@ compile_error!(
      `async-std-runtime` or set `default-features = false` in your Cargo.toml"
 );
 
-#[cfg(all(feature = "tokio-runtime", feature = "sync"))]
+#[cfg(all(feature = "async-std-runtime", feature = "sync"))]
 compile_error!(
-    "`tokio-runtime` and `sync` can't both be enabled; either disable `sync` or set \
-     `default-features = false` in your Cargo.toml"
+    "`async-std-runtime` and `sync` can't both be enabled"
```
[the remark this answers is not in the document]
isabelatkinson: I was making the code changes for this but think doing this will cause breaking changes as well... currently, the `sync` feature in our `Cargo.toml` enables `async-std-runtime`, and I swapped that out for `tokio-runtime` for this implementation. But if we want to allow for either, I think we'd want something like
```
[features]
...
sync = []
```
in `Cargo.toml` so as not to enable one runtime feature when the other is intended for use. However, this will break compilation for users who currently have
```
mongodb = { version = "2.1.0", default-features = false, features = ["sync"] }
```
in their `Cargo.toml` as they'll no longer have a runtime specified.

Is there a clean way around this that I'm missing? If we leave `async-std-runtime` enabled for `sync` then users will have to pull in both runtimes when they want to use tokio which isn't great. Alternatively we could make a `tokio-sync` feature for using sync with tokio and clean this up if we ever make breaking changes down the line.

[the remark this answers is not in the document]
patrickfreed: Ah good point. Hm, I'm not sure there _is_ a good option here besides the `tokio-sync` feature you suggest, since if we have any runtime in the `sync = [...]` definition, then the other won't be able to be used. This would present issues for RUST-488, since we'd ideally want to allow users to use whichever runtime they want.

So yeah, I guess our choices are the following:
- have `sync = ["tokio-runtime"]` per this PR, but don't allow both `sync` and `async-std-runtime` when we do RUST-488
    - pros: no new feature flag, sync users get tokio under the hood automatically
    - cons: technically breaking change, no async-std-runtime + sync
- introduce `tokio-sync` = ["tokio-runtime"]
    - pros: can use sync API with either tokio or async-std once we do RUST-488, no breaking changes
    - cons: new feature flag to maintain, existing sync users don't get benefits

I think I personally lean towards the new flag to avoid any potential (albeit rare) breaking changes and also to avoid another mutual exclusion problem between feature flags. I guess we'd probably want `sync` and `tokio-sync` to be mutually exclusive, or we could just use tokio if they're both enabled.

WDYT? cc @abr-egn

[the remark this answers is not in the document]
abr-egn: Agreed that the `tokio-sync` flag seems like the least bad option here; it seems particularly worth avoiding the "can't use async-std-runtime + sync" problem.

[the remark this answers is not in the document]
isabelatkinson: updated to introduce a `tokio-sync` feature flag

.evergreen/check-rustdoc.sh
```diff
@@ -5,4 +5,5 @@ set -o errexit
 . ~/.cargo/env
 cargo +nightly rustdoc -- -D warnings --cfg docsrs
 cargo +nightly rustdoc --no-default-features --features async-std-runtime -- -D warnings --cfg docsrs
-cargo +nightly rustdoc --no-default-features --features sync -- -D warnings --cfg docsrs
+cargo +nightly rustdoc --features sync -- -D warnings --cfg docsrs
```
patrickfreed, line 8: I think we still need `--no-default-features` here for the async-std sync variant. It looks like the `check-rustdoc.sh` test is failing as a result.

patrickfreed (commented):
Changes LGTM! Just need to sort out the lint failures and this should be good to go

.evergreen/check-rustdoc.sh
isabelatkinson: oh whoops, good catch

patrickfreed (approved):

PR #579, opened 2022-02-22
RUST-1175 Support zero-copy deserialization from cursors

patrickfreed:
RUST-1175

This PR adds new methods to `Cursor` and `SessionCursor` to allow for zero-copy deserialization. As part of that, the trait requirements on the cursor types were relaxed, and they no longer require `DeserializeOwned`. The `Stream` implementations still do require all of the trait requirements though.

example usage:

``` rust
#[derive(Debug, Deserialize)]
struct Stuff<'a> {
    #[serde(borrow)]
    name: &'a str,
    #[serde(borrow)]
    bio: &'a str
}

let coll = db.collection::<Stuff>("stuff");
let mut cursor = coll.find(None, None).await?;
while cursor.advance()?.await {
    println!("{:?}", cursor.deserialize_current()?);
}
```


src/cursor/common.rs
```diff
@@ -37,17 +37,15 @@ where
     provider: P,
     client: Client,
     info: CursorInformation,
-    buffer: VecDeque<RawDocumentBuf>,
```
patrickfreed, line 40: I moved the stuff that needed to be preserved between `GenericCursor` constructions in a `CursorState` struct to ensure nothing gets forgotten.

src/cursor/mod.rs
```diff
@@ -90,10 +91,7 @@ pub(crate) use common::{
 /// # }
 /// ```
 #[derive(Debug)]
-pub struct Cursor<T>
-where
-    T: DeserializeOwned + Unpin + Send + Sync,
```
patrickfreed, line 95: Since we no longer buffer `T`, we can drop the `Unpin + Send + Sync` parts of this, and because we now support borrowed deserialization, we can drop `DeserializeOwned` too.

src/operation/aggregate/test.rs
```diff
@@ -188,59 +186,6 @@ async fn op_selection_criteria() {
     });
 }
 
-#[cfg_attr(feature = "tokio-runtime", tokio::test)]
-#[cfg_attr(feature = "async-std-runtime", async_std::test)]
-async fn handle_success() {
```
patrickfreed, line 193: Rather than continuing to maintain these tests that break a lot and don't offer much additional coverage over our integration tests, I decided just to delete the ones that broke. We discussed this a bit in slack, but in retrospect these tests are probably far too close to the actual implementation and are probably more effort than they're worth. Thoughts?

isabelatkinson: agreed, I don't think these are particularly valuable

isabelatkinson (commented):
looks good, just a few comments. do we also want to add this functionality to the sync API?

src/cursor/common.rs
```diff
@@ -89,33 +145,67 @@ where
...
-        let client = self.client.clone();
-        self.provider
-            .start_execution(info, client, self.pinned_connection.handle());
+    fn mark_exhausted(&mut self) {
+        self.state_mut().exhausted = true;
+        self.state_mut().pinned_connection = PinnedConnection::Unpinned;
+    }
+
+    fn handle_get_more_result(&mut self, get_more_result: Result<GetMoreResult>) -> Result<()> {
+        self.state_mut().exhausted = get_more_result
```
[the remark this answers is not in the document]
patrickfreed: nope, good catch. removed

src/cursor/session.rs
```diff
@@ -235,23 +340,17 @@ where
...
 }
 
-impl<T> SessionCursor<T>
-where
-    T: DeserializeOwned + Unpin,
-{
+impl<T> SessionCursor<T> {
     fn mark_exhausted(&mut self) {
-        self.info.id = 0;
+        self.state.as_mut().unwrap().exhausted = true;
```
patrickfreed, line 345: now that state is the single source of truth, this needed to be updated to use `exhausted` instead of the spec.

patrickfreed (commented):
Ah yep, totally forgot the sync API. Added it it now

src/cursor/common.rs
```diff
@@ -57,23 +55,81 @@ where
...
+        Ok(true)
+    }
+
+    pub(super) fn take_state(&mut self) -> CursorState {
+        self.state.take().unwrap()
     }
 
     pub(super) fn is_exhausted(&self) -> bool {
-        self.exhausted
+        self.state.as_ref().unwrap().exhausted
```
abr-egn, line 132: For this and similar calls, I think it would be slightly better to centralize the `unwrap` in the `state()` accessor and write this as `self.state().exhausted`.

src/operation/aggregate/test.rs
abr-egn: SGTM.

abr-egn (approved):
LGTM!  Minor comments only.

src/cursor/common.rs
patrickfreed: Updated. I had to leave the one on line 245 in to satisfy the borrow checker, not sure why though.

PR #583, opened 2022-02-28
RUST-1186 Convert sessions tests to unified test format

isabelatkinson:


src/test/spec/sessions.rs
```diff
@@ -6,20 +6,13 @@ use crate::{
...
 async fn run_unified() {
     let _guard: RwLockWriteGuard<()> = LOCK.run_exclusively().await;
-    run_spec_test(&["sessions", "unified"], run_unified_format_test).await;
-}
-
-#[cfg_attr(feature = "tokio-runtime", tokio::test(flavor = "multi_thread"))]
-#[cfg_attr(feature = "async-std-runtime", async_std::test)]
-async fn run_legacy() {
-    let _guard: RwLockWriteGuard<()> = LOCK.run_exclusively().await;
-    run_spec_test(&["sessions", "legacy"], run_v2_test).await;
```
isabelatkinson, line 22: retryable reads/transactions are still using the V2 runner so we can't delete any runner code yet

abr-egn (approved):

src/client/mod.rs
```diff
@@ -250,7 +250,11 @@ impl Client {
     }
 
     /// Starts a new `ClientSession`.
-    pub async fn start_session(&self, options: Option<SessionOptions>) -> Result<ClientSession> {
+    pub async fn start_session(
+        &self,
+        options: impl Into<Option<SessionOptions>>,
```
patrickfreed, line 255: I think this is technically a breaking change if someone were relying on the type inference for the options, but I don't really think it's a cause for concern.

patrickfreed (approved):

PR #585, opened 2022-03-01
RUST-1173 Replace "Versioned API" references with "Stable API"

isabelatkinson:
Grepped for "versioned" in the driver and replaced all references with "stable". The only terminology I didn't change was in the spec tests as the directory in the specifications repo hasn't been renamed.

### DIFF
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -322,8 +322,8 @@ impl<'de> Deserialize<'de> for ServerApiVersion {
     }
 }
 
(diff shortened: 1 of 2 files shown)


isabelatkinson:
The Atlas connectivity tests are failing but they're also failing on the 1.49 PR so I believe they're unrelated to these/those changes.

abr-egn (approved):

patrickfreed (approved):

PR #587, opened 2022-03-03
RUST-1124 Pin version of clippy used in CI

isabelatkinson:
This PR fixes our existing clippy errors and pins clippy to the latest version (1.59.0).

### DIFF
--- a/.evergreen/install-dependencies.sh
+++ b/.evergreen/install-dependencies.sh
@@ -24,6 +24,9 @@ source ~/.cargo/env
 # Install nightly rustfmt
 rustup toolchain install nightly -c rustfmt
 
+# Pin clippy version to the latest Rust version. This should be updated with each new Rust release.
+rustup toolchain install 1.59.0 -c clippy
+
 # Install tool for converting cargo test output to junit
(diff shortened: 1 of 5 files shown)


abr-egn (approved):

.evergreen/install-dependencies.sh
```diff
@@ -24,6 +24,9 @@ source ~/.cargo/env
 # Install nightly rustfmt
 rustup toolchain install nightly -c rustfmt
 
+# Pin clippy version to the latest Rust version. This should be updated with each new Rust release.
+rustup toolchain install 1.59.0 -c clippy
```
patrickfreed, line 28: I think we may need to update `check-clippy.sh` to select `1.59.0` as the toolchain to use. I think otherwise it'll just use stable by default.

isabelatkinson: Removed this in favor of running `rustup default 1.59.0` in `check-clippy.sh`. I tested this out locally by changing the default to a few different versions and running `cargo clippy` and it seems to do the trick.

patrickfreed (approved):

PR #590, opened 2022-03-08
RUST-1083 Add openssl as an optional TLS provider

abr-egn:
RUST-1083

This adds the `openssl-tls` feature, which enables TLS connections via openssl rather than rustls.  In service of that, I refactored the TLS-handling code into its own module which is conditionally renamed based on the feature flag, making conditional surface area minimal.

Unfortunately, for backwards compatibility rustls can't be made optional, so it's still linked in even when it's not used.

The evergreen auth tests have been updated to run against both TLS crates; in the process of doing that I discovered that the tests previously were ignoring the runtime variable, so I picked that up as well.


src/error.rs
```diff
@@ -474,6 +488,11 @@ pub enum ErrorKind {
     /// No resume token was present in a change stream document.
     #[error("Cannot provide resume functionality when the resume token is missing")]
     MissingResumeToken,
+
+    /// Wrapper around [`openssl::ssl::Error`].
+    #[cfg(feature = "openssl-tls")]
+    #[error("{0}")]
+    OpenSSL(Arc<openssl::ssl::Error>),
```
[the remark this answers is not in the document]
abr-egn: Good catch, fixed.  I folded this into the existing `Authentication` error type rather than introduce a new one.

Cargo.toml
```diff
@@ -58,10 +59,12 @@ hex = "0.4.0"
 hmac = "0.11"
 lazy_static = "1.4.0"
 md-5 = "0.9.1"
+openssl = { version = "0.10.38", optional = true }
+openssl-probe = { version = "0.1.5", optional = true }
 os_info = { version = "3.0.1", default-features = false }
 percent-encoding = "2.0.0"
 rand = { version = "0.8.3", features = ["small_rng"] }
-rustls-pemfile = "0.2.1"
+rustls-pemfile =  "0.2.1"
```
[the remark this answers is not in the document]
abr-egn: Whoops, done.

abr-egn:
Okay, good to go.

src/runtime/tls_openssl.rs
```diff
@@ -0,0 +1,95 @@
...
+        tcp_stream: AsyncTcpStream,
+        cfg: TlsOptions,
+    ) -> Result<Self> {
+        init_trust();
+
+        let connector = make_openssl_connector(cfg)?;
+        let ssl = connector
+            .configure()?
+            .use_server_name_indication(true)
+            .verify_hostname(false)
```
patrickfreed, line 33: why is `verify_hostname` always false?

src/error.rs
```diff
@@ -388,6 +388,24 @@ impl From<std::io::ErrorKind> for ErrorKind {
     }
 }
 
+#[cfg(feature = "openssl-tls")]
```
patrickfreed, line 391: I think we'll want to match the error types that the existing rustls implementaiton uses as best we can here. So for example, any errors returned as part of `make_ssl_connector` and the subsequent configuration seem to map nicely to `InvalidTlsConfig`, and everything after that would be a custom `std::io::Error`, since that appears to be [what `rustls` does (for better or worse)](https://docs.rs/tokio-rustls/0.23.2/src/tokio_rustls/lib.rs.html#127). I'm not sure if a single `From` implementation will be able to handle this correctly unfortunately, so we may have to do it at the OpenSSL call sites. 

More generally, `ErrorKind::Authentication` has thus far been reserved for MongoDB authentication failures (including x509 auth failures, which should be fine without this `From` I believe), not TLS verification failures.

patrickfreed (commented):
looks good! just have one question and one comment on error types

src/runtime/tls_openssl.rs
abr-egn: Whoops.  Updated this to an option that's only present when `openssl-tls` is flagged on, since there's no equivalent in rustls.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ tokio-runtime = ["tokio/macros", "tokio/net", "tokio/rt", "tokio/time", "serde_b
 async-std-runtime = ["async-std", "async-std/attributes", "async-std-resolver", "tokio-util/compat"]
 sync = ["async-std-runtime"]
 tokio-sync = ["tokio-runtime"]
+openssl-tls = ["openssl", "openssl-probe", "tokio-openssl"]
```
patrickfreed, line 28: can we update the feature flag table in the readme and `lib.rs` documentation to include this flag?

patrickfreed (commented):
code LGTM, just need some docs updates

Cargo.toml
abr-egn: Done.

patrickfreed (approved):
LGTM!

PR #595, opened 2022-03-11
RUST-1202 Sync latest initial DNS seedlist discovery tests

kmahar:
* Sync latest copy of initial DNS seedlist discovery spec tests
* Updates the test runner to support some file format adjustments, unskip new tests we can now run, and skip some new ones we can't run yet

there are some failures on Evergreen but they all seem unrelated to these changes.

### DIFF
--- a/src/test/spec/initial_dns_seedlist_discovery.rs
+++ b/src/test/spec/initial_dns_seedlist_discovery.rs
@@ -211,9 +211,10 @@ async fn replica_set() {
              requirement ({})",
             skip
         ));
-    } else {
-        run_spec_test(&["initial-dns-seedlist-discovery", "replica-set"], run_test).await;
+        return;
     }
+
+    run_spec_test(&["initial-dns-seedlist-discovery", "replica-set"], run_test).await;
 }
 
 #[cfg_attr(feature = "tokio-runtime", tokio::test)]
@@ -228,7 +229,6 @@ async fn load_balanced() {
         );
         return;
     }
-
     run_spec_test(
         &["initial-dns-seedlist-discovery", "load-balanced"],
         run_test,
--- a/src/test/spec/initial_dns_seedlist_discovery.rs
+++ b/src/test/spec/initial_dns_seedlist_discovery.rs
@@ -108,7 +108,7 @@ async fn run_test(mut test_file: TestFile) {
 
     let options = result.unwrap();
 
-    if let Some(mut expected_seeds) = test_file.seeds {
+    if let Some(ref mut expected_seeds) = test_file.seeds {
         let mut actual_seeds = options
             .hosts
             .iter()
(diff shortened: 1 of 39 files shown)


src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -56,26 +57,14 @@ struct ParsedOptions {
 }
 
 async fn run_test(mut test_file: TestFile) {
-    // TODO DRIVERS-796: unskip this test
```
kmahar, line 59: we already did the associated Rust ticket (RUST-479), so this test can be run now.

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -31,6 +31,7 @@ struct ResolvedOptions {
     ssl: bool,
     load_balanced: Option<bool>,
     direct_connection: Option<bool>,
+    srv_max_hosts: Option<u32>,
```
kmahar, line 34: my sync brought in a bunch of tests for srvMaxHosts. this addition was necessary for the test files to be deserialized successfully, though I added logic below to avoid actually running the new tests.

not sure u32 was the right type choice here, but it seemed consistent with other options like this in `ClientOptions`?

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -122,20 +111,8 @@ async fn run_test(mut test_file: TestFile) {
         None => true,
     };
     let client = TestClient::new().await;
-    let skip = if !client.is_replica_set() {
```
kmahar, line 125: these checks needed to move into the topology-specific test methods since we need to validate things differently for each topology type.

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -104,38 +108,30 @@ async fn run_test(mut test_file: TestFile) {
 
     let options = result.unwrap();
 
-    let mut expected_seeds = test_file.seeds.split_off(0);
```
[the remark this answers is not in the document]
abr-egn: The use of `split_off` confuses me as well - what you're doing looks right.

abr-egn (approved):
LGTM!

isabelatkinson (approved):
looks good! just a few small suggestions, no need for another round of review

src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -104,38 +108,30 @@ async fn run_test(mut test_file: TestFile) {
...
+            .map(|address| address.to_string())
+            .collect::<Vec<_>>();
 
-    expected_seeds.sort();
-    actual_seeds.sort();
+        expected_seeds.sort();
+        actual_seeds.sort();
 
-    assert_eq!(expected_seeds, actual_seeds,);
+        assert_eq!(*expected_seeds, actual_seeds,);
```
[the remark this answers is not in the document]
kmahar: huh, weird. removed!

PR #591, opened 2022-03-09
bump  trust-dns-proto trust-dns-resolver

attila-lin:

PR #596, opened 2022-03-15
RUST-1220 Bump outdated dependencies

isabelatkinson:
Bumps all of our outdated dependencies (excluding the few that are covered in #591). Also increases the MSRV to 1.51.0 (new MSRV for `pbkdf2`). The most significant breaking changes were due to updates in rustls; you can reference the changelog [here](https://github.com/rustls/rustls/blob/main/RELEASE_NOTES.md).


src/client/options/mod.rs
```diff
@@ -871,11 +870,24 @@ impl TlsOptions {
                 }
             };
 
+            ClientConfig::builder()
```
isabelatkinson, line 873: `ClientConfig` was made mostly opaque in favor of a builder, documentation [here](https://docs.rs/rustls/latest/rustls/struct.ConfigBuilder.html). This branch calls `with_single_cert` on the builder which corresponds to the `set_single_client_cert` method we were calling on the config previously. The `else` branch uses the defaults suggested in the docs, which are also in line with the now-removed [`new`](https://docs.rs/rustls/0.18.0/rustls/struct.ClientConfig.html#method.new) function we were calling previously.

Cargo.toml
```diff
@@ -57,43 +57,42 @@ futures-io = "0.3.14"
...
 socket2 = "0.4.0"
 stringprep = "0.1.2"
 strsim = "0.10.0"
 take_mut = "0.2.2"
 thiserror = "1.0.24"
 tokio-openssl = { version = "0.6.3", optional = true }
-trust-dns-proto = "0.20.0"
-trust-dns-resolver = "0.20.0"
-typed-builder = "0.9.0"
+trust-dns-proto = "0.21.1"
```
isabelatkinson, line 79: I ended up needing to bump these rather than waiting for #591 because the older versions were incompatible with the new version of `async-std-resolver`.

patrickfreed (approved):

abr-egn (approved):
LGTM!

PR #598, opened 2022-03-17
RUST-1064 Retry on handshake failure

abr-egn:
RUST-1064

This updates operation logic to retry if a connection fails with a network or authentication error.  The test sync for this also brought a split to `legacy` and `unified`; none of the tests in `legacy` were modified.


kmahar (commented):
it looks like from the ticket there are also new retryable writes and txns tests to sync (though we are maybe supposed to skip the txn tests for now)

patrickfreed (approved):
LGTM! glad to see the existing pool cleared error retry mechanism ended up being the same for the more general one

patrickfreed:
From looking at the test failures, it looks like a load balancer test (the "errors during the initial connection hello are ignored" test in `sdam-error-handling.yml`) needs to be updated, though I'm not sure it actually ever was as part of the spec changes. We may need to file a DRIVERS ticket about that.

kmahar:
> From looking at the test failures, it looks like a load balancer test (the "errors during the initial connection hello are ignored" test in sdam-error-handling.yml) needs to be updated, though I'm not sure it actually ever was as part of the spec changes. We may need to file a DRIVERS ticket about that.

it seems strange to me that Python, Java, and .NET are all still passing that test -- at least I'm guessing they are based on the fact they have merged in PRs and closed their tickets for this. I suggest asking in #driver-devs to try to figure out why only we have run into this so far

abr-egn:
My theory is that I need to refine the Rust driver error-handling machinery to make a more precise distinction between error types.

src/client/auth/sasl.rs
```diff
@@ -98,12 +99,20 @@ fn validate_command_success(auth_mechanism: &str, response: &Document) -> Result
...
     match bson_util::get_int(ok) {
         Some(1) => Ok(()),
-        Some(_) => Err(Error::authentication_error(
-            auth_mechanism,
-            response
-                .get_str("errmsg")
-                .unwrap_or("Authentication failure"),
-        )),
+        Some(_) => {
+            let source = bson::from_bson::<CommandResponse<CommandErrorBody>>(Bson::Document(
```
abr-egn, line 103: When implementing this, I was a little surprised to learn that the SASL commands use bespoke parsing and validation rather than going through Serde integration like the rest; if there's no strong reason for that, it might be a good candidate for future cleanup.

src/client/executor.rs
```diff
@@ -331,16 +331,24 @@ impl Client {
             Ok(c) => c,
             Err(mut err) => {
                 err.add_labels_and_update_pin(None, &mut session, None)?;
+                if err.is_read_retryable() {
```
abr-egn, line 334: This seems a little loopy but:
* the spec mandates that the `RETRYABLE_WRITE_ERROR` label be added for retryable errors from authentication and
* `is_write_retryable()` is implemented as a label check, so it can't be used to add the label here in the first place

Note that the csharp logic, which is [called out](https://github.com/mongodb/specifications/blob/082306075eb2d125c47503e94fabc42f00a16784/source/retryable-reads/retryable-reads.rst#reference-implementation) in the spec as a reference implementation, ends up doing the [same thing](https://github.com/mongodb/mongo-csharp-driver/blob/21f8c1615dcfec501b66bc19a43e185e13fdad46/src/MongoDB.Driver.Core/Core/Operations/RetryableWriteOperationExecutor.cs#L139).

src/client/executor.rs
```diff
@@ -824,35 +832,49 @@ impl Client {
     }
 
     /// Returns the retryability level for the execution of this operation.
-    async fn get_retryability<T: Operation>(
```
abr-egn, line 827: It turns out this didn't need to be async.

abr-egn (commented):
Updated, PTAL.

src/client/executor.rs
patrickfreed: `err.add_labels_and_update_pin` calls through to `should_add_retryable_write_label`, which _should_ take care of this for us, but it doesn't currently for two reasons:
  - we always pass in `None` for retryability
      - This can be fixed by creating a new `Retryability` variant called `Always` or something
  - more importantly, we don't have a connection and therefore don't know the max wire version of the server, so we don't know if we're supposed to add labels to it or should leave it to the server to do so
      - @matthewdale points out the same issue as experienced in the Go driver [here](https://jira.mongodb.org/browse/GODRIVER-2191?focusedCommentId=4422059&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4422059).

I think what we ideally need here is a single set of criteria for retrying handshake errors that is independent of whether the operation is a read or write or what the server version might be. It feels like it'll end up just being network errors + the union of the pre-4.4 retryable write + retryable read error codes. but I'm not sure.

With all that being said, I think that the solution you have here seems good until the spec gets that all fleshed out.

patrickfreed (commented):
The code looks good to me as-is, but I'm wondering if we should hold off for a bit pending some further spec changes that might come as a result of the issues we've faced here. I left a comment on DRIVERS-746 about it.

src/client/executor.rs
kmahar: should we be checking here if the client has `retryWrites` set to true before we add the label?

spec: "When the driver encounters a network error establishing an initial connection to a server,
  it MUST add a RetryableWriteError label to that error if the MongoClient performing
  the operation has the retryWrites configuration option set to true."

and I see C# checks `context.RetryRequested` there.

do any of the new tests cover retryWrites=false? (I'm not seeing any that do)

I guess either way it seems like we only actually do retry if the client has it set given the check in `get_op_retryability`, but I'm wondering if there could be any ill effects or user confusion from adding the label unconditionally here -- the retryable writes spec discusses that a bit [here](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#why-does-the-driver-only-add-the-retryablewriteerror-label-to-errors-that-occur-on-a-mongoclient-with-retrywrites-set-to-true)

patrickfreed (approved):

src/client/executor.rs
abr-egn: Good catch - updated to check for `retryWrites` here.

kmahar (approved):

isabelatkinson (approved):

PR #599, opened 2022-03-18
RUST-1014 Test against all Atlas test instances

kmahar:
Previously we could not test against all of the drivers test instances because some of them were on MongoDB < 4.2, which we cannot connect to on Atlas due to RUST-272.
They have all been upgraded now, so this expands our testing to cover all available instances.

Corresponding private variables have been added to the Evergreen project.


src/test/atlas_connectivity.rs
```diff
@@ -65,9 +65,78 @@ async fn atlas_serverless() {
 #[cfg_attr(feature = "tokio-runtime", tokio::test)]
 #[cfg_attr(feature = "async-std-runtime", async_std::test)]
 async fn atlas_serverless_srv() {
+    run_test("MONGO_ATLAS_SERVERLESS_URI_SRV", None).await;
```
kmahar, line 68: for the other existing SRV test above for free tier, we test both with cloudflare and with no resolver config specified. I used the same pattern for the new tests below so I figured I should add a corresponding check to this test.

patrickfreed: Given that serverless and atlas free tier are a bit different on the cloud side of things, I think I lean towards more coverage here, especially since almost all users will be using the default resolver config.

abr-egn (approved):

kmahar:
@patrickfreed any more comments or good to merge?

patrickfreed (approved):
nope, no more comments, LGTM!

PR #600, opened 2022-03-21
RUST-803 Conditionally use `hello` for monitoring

patrickfreed:
RUST-803

This PR updates the monitors in the driver to make use of `helloOk`, a new (in 4.4.5) flag to the legacy hello command which can be used to determine if the server supports the new `hello` command. If it echos back `helloOk: true`, all subsequent monitor checks should use `hello` instead of the legacy hello command. The initial check needs to always be legacy though in case the server doesn't support the newer one.

The SDAM tests were synced as part of this.

There are a few remaining instances of the legacy hello command in the driver, all used for initial information requests or for deserialization purposes.


src/hello.rs
```diff
@@ -0,0 +1,273 @@
+use std::{
```
patrickfreed, line 1: This file is largely the same, but git's diff algorithm couldn't detect it as a rename. There are some minor changes, but it's mostly just search and replace.

src/test/util/mod.rs
```diff
@@ -82,14 +82,10 @@ impl TestClient {
             .await;
         session.mark_dirty();
 
-        let is_master =
-            RunCommand::new("admin".into(), doc! { "isMaster":  1 }, None, None).unwrap();
+        let hello = RunCommand::new("admin".into(), doc! { "isMaster":  1 }, None, None).unwrap();
```
abr-egn, line 85: Should this be using `"hello"`?

patrickfreed: This one I thought was required, since it's essentially an initial handshake. I realized that I can just reuse the `hello_command` method though, so I've updated it to that.

src/cmap/conn/wire/test.rs
```diff
@@ -1,50 +0,0 @@
...
-use crate::{
-    bson::{doc, Bson},
-    cmap::options::StreamOptions,
-    runtime::AsyncStream,
-    test::{log_uncaptured, CLIENT_OPTIONS, LOCK},
-};
-
-#[cfg_attr(feature = "tokio-runtime", tokio::test)]
-#[cfg_attr(feature = "async-std-runtime", async_std::test)]
-async fn basic() {
```
patrickfreed, line 14: this test seemed to be sufficiently covered by our end-to-end integration tests and is a bit too specific to the implementation, so I just decided to remove it rather than try to accommodate any more changes to it.

kmahar (commented):
couple small things but LGTM otherwise

src/cmap/test/integration.rs
```diff
@@ -107,7 +107,7 @@ async fn concurrent_connections() {
     let failpoint = doc! {
         "configureFailPoint": "failCommand",
         "mode": "alwaysOn",
-        "data": { "failCommands": [ "isMaster" ], "blockConnection": true, "blockTimeMS": 1000 }
+        "data": { "failCommands": [ "isMaster", "hello" ], "blockConnection": true, "blockTimeMS": 1000 }
```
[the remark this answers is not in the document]
patrickfreed: good idea, done

abr-egn (approved):
LGTM!

kmahar (approved):
lgtm!

isabelatkinson (approved):

PR #601, opened 2022-03-23
RUST-886 Use lossy UTF-8 decoding for responses to insert and update commands

kmahar:
These changes work around a somewhat commonly encountered issue in the server where long error messages containing multi-byte characters are truncated incorrectly such that they are no longer valid UTF-8.

From digging through all of the related tickets for the server and drivers, this only ever seems to be observed in duplicate key errors, which I believe can only result from `insert` and `update` commands, so I've limited this PR to those two operations. It seems possible it could come up elsewhere, but I figured we could start with this and expand our usage of the lossy decoding later if the need arises.

Before my changes, all of the errors below, which are now duplicate key errors, would instead be something like
```
Error { kind: InvalidResponse { message: "invalid utf-8 sequence of 1 bytes from index 264" }, labels: {}, wire_version: Some(15) }
```

### DIFF
--- a/src/test/coll.rs
+++ b/src/test/coll.rs
@@ -19,6 +19,7 @@ use crate::{
         FindOneOptions,
         FindOptions,
         Hint,
+        IndexOptions,
         InsertManyOptions,
         ReadConcern,
         ReadPreference,
@@ -35,6 +36,7 @@ use crate::{
         LOCK,
     },
     Collection,
+    IndexModel,
 };
 
 #[cfg_attr(feature = "tokio-runtime", tokio::test)]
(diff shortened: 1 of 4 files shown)


src/test/coll.rs
```diff
@@ -1098,3 +1100,104 @@ async fn cursor_batch_size() {
...
+        .keys(doc! {"name": 1})
+        .options(IndexOptions::builder().unique(true).build())
+        .build();
+    coll.create_index(index_model, None)
+        .await
+        .expect("creating an index should succeed");
+
+    // a document containing a long string with multi-byte unicode characters. taken from a user
+    // repro in RUBY-2560.
+    let long_unicode_str_doc = doc! {"name":  "(╯°□°)╯︵ ┻━┻(╯°□°)╯︵ ┻━┻(╯°□°)╯︵ ┻━┻(╯°□°)╯︵ ┻━┻(╯°□°)╯︵ ┻━┻(╯°□°)╯︵ ┻━┻"};
```
abr-egn, line 1126: ┬─┬ノ( º _ ºノ)

abr-egn (approved):
LGTM!

patrickfreed (approved):
LGTM!

isabelatkinson (approved):

PR #603, opened 2022-03-23
RUST-1228 Create a skeleton for a mdbook manual

abr-egn:
Happily, [mdbook](https://rust-lang.github.io/mdBook/index.html) is quite straightforward.


kmahar (approved):
LGTM - have you thought about how we host/serve this? if GH pages, we should think about whether we want to use a separate `gh-pages` branch for the book

abr-egn:
> LGTM - have you thought about how we host/serve this? if GH pages, we should think about whether we want to use a separate `gh-pages` branch for the book

From our in-person discussion, GH pages seems like the clear way to go, and at least initially we won't be using a separate branch.  I've updated this PR to point the output of mdbook at `docs/manual` and include the generated files.  I also added a `.gitattributes` file so the contents of that are marked as generated, which should mean they don't count towards language stats for the repo.  Unfortunately, I couldn't find any way to exclude it from the contribution graph.

manual/src/SUMMARY.md
```diff
@@ -0,0 +1,18 @@
...
+
+- [Introduction](README.md)
+
+# Basic Usage
+
+- [Installation](basic/installation.md)
+- [Feature Choices](basic/feature_choices.md)
+- [Connection and Operation](./basic/connection_and_operation.md)
+
+# Advanced
```
patrickfreed, line 11: IMO we can drop the "basic" vs "advanced" distinction. I'm a bit worried that users may potentially skip over the advanced stuff when getting started, even though ideally all users, from beginners to experts, should be using serde, for example.

manual/src/SUMMARY.md
```diff
@@ -0,0 +1,18 @@
...
+
```
patrickfreed, line 12: I think some more good sections to have would be "Cursors", "Sessions and Transactions", "Change Streams", and "Monitoring" (for stuff like command, CMAP, and SDAM monitoring). Perhaps, we'd want "Reading from the Database" and "Writing to the Database" sections too, and "Cursors" could just be a subsection of the former.

patrickfreed (commented):
format looks great! I just have some comments on the proposed section breakdown.

manual/src/SUMMARY.md
abr-egn: Done.

manual/src/SUMMARY.md
abr-egn: Done.

patrickfreed (approved):

isabelatkinson (approved):

PR #617, opened 2022-03-29
RUST-1229 Add install/features and connecting sections to the manual

abr-egn:
RUST-1229

This PR adds an introduction, installation and features, and connecting to the db sections to the manual.  The first two were taken nearly verbatim from README.md; the last is mostly fresh.

This also includes some code examples, so I set up the mdbook compile-testing machinery for that and added that as an evergreen task.  This required setting up a dummy `deps` rust library to depend on the mongodb library, which is kind of a hack, but this area is a [known deficiency](https://github.com/rust-lang/mdBook/issues/706) of mdbook at the moment.


patrickfreed (approved):
LGTM!

isabelatkinson (approved):

PR #618, opened 2022-03-30
RUST-1229 Add a "Reading from the Database" manual section

abr-egn:
RUST-1229

The newly added section covers the basics I could think of - handle characteristics, `Collection` parameterization, and cursors.  I also the warning about cancellation and the MSRV to the introduction.


manual/src/reading.md
```diff
@@ -0,0 +1,104 @@
+# Reading From the Database
+
+## Database and Collection Handles
+
+Once you have a `Client`, you can call [`Client::database`](https://docs.rs/mongodb/latest/mongodb/struct.Client.html#method.database) to create a handle to a particular databse on the server, and [`Database::collection`](https://docs.rs/mongodb/latest/mongodb/struct.Database.html#method.collection) to create a handle to a particular collection in that database.  [`Database`](https://docs.rs/mongodb/latest/mongodb/struct.Database.html) and [`Collection`](https://docs.rs/mongodb/latest/mongodb/struct.Collection.html) handles are lightweight - creating them requires no IO, `clone`ing them is cheap, and they can be safely shared across threads or async tasks.  For example:
```
isabelatkinson, line 5:
```suggestion
Once you have a `Client`, you can call [`Client::database`](https://docs.rs/mongodb/latest/mongodb/struct.Client.html#method.database) to create a handle to a particular database on the server, and [`Database::collection`](https://docs.rs/mongodb/latest/mongodb/struct.Database.html#method.collection) to create a handle to a particular collection in that database.  [`Database`](https://docs.rs/mongodb/latest/mongodb/struct.Database.html) and [`Collection`](https://docs.rs/mongodb/latest/mongodb/struct.Collection.html) handles are lightweight - creating them requires no IO, `clone`ing them is cheap, and they can be safely shared across threads or async tasks.  For example:
```

manual/src/reading.md
```diff
@@ -0,0 +1,104 @@
...
+
+        // Do something with the collection
+    });
+}
+#
+# Ok(())
+# }
+```
+
+A `Collection` can be parameterized with any type that implements the `Serialize` and `Deserialize` traits from the [`serde`](https://serde.rs/) crate. This includes but is not limited to just `Document`. The various methods that accept or return instances of the documents in the collection will accept/return instances of the generic parameter (e.g. [`Collection::insert_one`](https://docs.rs/mongodb/latest/mongodb/struct.Collection.html#method.insert_one) accepts it as an argument, [`Collection::find_one`](https://docs.rs/mongodb/latest/mongodb/struct.Collection.html#method.find_one) returns an `Option` of it). It is recommended to define types that model your data which you can parameterize your `Collection`s with instead of `Document`, since doing so eliminates a lot of boilerplate deserialization code and is often more performant.
```
isabelatkinson, line 30: technically `Serialize` and `Deserialize` don't both need to be implemented depending upon how the collection is being used. I don't think we need to get into the nitty gritty of that here, but can we update this language to reflect that the `T` for `Collection` doesn't have any specific requirements upon creation?

manual/src/reading.md
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Parameterize our collection with the model.
+let coll = client.database("items").collection::<Item>("in_stock");
+
+for i in 0..5 {
+    let coll_ref = coll.clone();
+
+    // Spawn several tasks that operate on the same collection concurrently.
+    task::spawn(async move {
```
isabelatkinson, line 61: I think it might be simpler to leave out the clone/spawn stuff in this snippet since it's already covered in the previous section

manual/src/connecting.md
```diff
@@ -21,7 +21,7 @@ options.app_name = Some("My App".to_string());
 ```
 
 ## Creating a `Client`
-The `Client` struct is the main entry point for the driver.  You can create one from a `ClientOptions` struct:
+The [`Client`](https://docs.rs/mongodb/latest/mongodb/struct.Client.html) struct is the main entry point for the driver.  You can create one from a `ClientOptions` struct:
```
kmahar, line 24: it's sort of performance related so we would probably talk about it more extensively there, but I think it could still be worth briefly mentioning here because it's so important that for most uses cases users should create and persist a single client for their entire app. 
this has been a point of confusion in the past (not just in Rust but for MongoDB drivers in general) by people coming from different DB drivers where the top level object is a single connection

abr-egn: Done.

manual/src/README.md
```diff
@@ -2,4 +2,36 @@
...
+
+## Warning about timeouts / cancellation
+
+In async Rust, it is common to implement cancellation and timeouts by dropping a future after a certain period of time instead of polling it to completion. This is how [`tokio::time::timeout`](https://docs.rs/tokio/1.10.1/tokio/time/fn.timeout.html) works, for example. However, doing this with futures returned by the driver can leave the driver's internals in an inconsistent state, which may lead to unpredictable or incorrect behavior (see RUST-937 for more details). As such, it is **_highly_** recommended to poll all futures returned from the driver to completion. In order to still use timeout mechanisms like `tokio::time::timeout` with the driver, one option is to spawn tasks and time out on their [`JoinHandle`](https://docs.rs/tokio/1.10.1/tokio/task/struct.JoinHandle.html) futures instead of on the driver's futures directly. This will ensure the driver's futures will always be completely polled while also allowing the application to continue in the event of a timeout.
```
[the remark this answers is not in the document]
abr-egn: Done.

manual/src/reading.md
abr-egn: Good call, done.

manual/src/reading.md
abr-egn: Done.

manual/src/reading.md
abr-egn: Done.

isabelatkinson (approved):

patrickfreed (approved):
LGTM!

PR #621, opened 2022-04-11
RUST-1108 Add a manual section on client best practices

abr-egn:
RUST-1108

This covers client lifetime, pooling, runtime pitfalls, and encouraged parallelism.


manual/src/performance.md
```diff
@@ -0,0 +1,120 @@
...
+    // Do something with the client
+    Ok(())
+}
+```
+
+This is especially noticeable when using a framework that provides connection pooling; because `Client` does its own pooling internally, attempting to maintain a pool of `Client`s will (somewhat counter-intuitively) result in worse performance than using a single one.
+
+### Runtime
+
+A `Client` is implicitly bound to the instance of the `tokio` or `async-src` runtime in which it was created.  Attempting to execute operations on a different runtime instance will cause incorrect behavior and unpredictable failures.  This is easy to accidentally invoke when testing, as the `tokio::test` or `async_std::test` helper macros create a new runtime for each test.
```
kmahar, line 32: async-src -> async-std?

kmahar (approved):
LGTM mod one possible typo

manual/src/performance.md
abr-egn: Derp, finger macro failure.  Fixed.

manual/src/performance.md
```diff
@@ -0,0 +1,120 @@
+# Performance
+
+## `Client` Best Practices
+
+The [`Client`](https://docs.rs/mongodb/latest/mongodb/struct.Client.html) handles many aspects of databse connection behind the scenes that can require manual management for other database drivers; it discovers server topology, monitors it for any changes, and maintains an internal connection pool.  This has implications for how a `Client` should be used for best performance.
```
isabelatkinson, line 5:
```suggestion
The [`Client`](https://docs.rs/mongodb/latest/mongodb/struct.Client.html) handles many aspects of database connection behind the scenes that can require manual management for other database drivers; it discovers server topology, monitors it for any changes, and maintains an internal connection pool.  This has implications for how a `Client` should be used for best performance.
```

isabelatkinson (approved):
lgtm! just one typo fix

manual/src/performance.md
abr-egn: Good catch, fixed.

manual/src/performance.md
```diff
@@ -0,0 +1,120 @@
...
+use once_cell::sync::Lazy;
+
+static CLIENT_RUNTIME: Lazy<(Client, Runtime)> = Lazy::new(|| {
+    let rt = Runtime::new().unwrap();
+    let client = rt.block_on(async {
+        Client::with_uri_str("mongodb://example.com").await.unwrap()
+    });
+    (client, rt)
+});
+
```
patrickfreed, line 73: can we add `#[test]` here to emphasize that `tokio::test` shouldn't be used with this scheme.

patrickfreed (commented):
one small suggestion but otherwise LGTM

patrickfreed (approved):

PR #622, opened 2022-04-13
RUST-1268 Correctly report rustc version used for compilation in handshake

kmahar:
RUST-1268

Switches the driver to use `rustc_version_runtime` to retrieve the rustc version used for compilation for inclusion in the driver handshake metadata.

Previously, on a machine where rustc was available the platforms field looked like this:
```
"rustc 1.60.0 stable (2022-04-04) with tokio" // stable release
"rustc 1.62.0 nightly (2022-04-11) with tokio" // nightly release
```

When rustc is not available, this field was null.

Now, it looks like this (slightly different due to the formatting of `VersionMeta.short_version_string`, but the relevant info is still there and in a similar format):
```
"rustc 1.60.0 (7737e0b5c 2022-04-04) with tokio" // stable release
"rustc 1.62.0-nightly (90ca44752 2022-04-11) with tokio" // nightly release
```

It doesn't seem like we test this at all right now, but I did test manually to get the results I pasted above. I don't think a test would have caught the bug this fixes anyway though, since we always have rustc available when we are running tests. 

Feel free to check yourselves as well.  Note one quirk is that if you switch Rust versions and recompile, it seems `rustc_version_runtime` doesn't get rebuilt, so you have to `cargo clean --package rustc_version_runtime` and rebuild to get that updated.

### DIFF
--- a/src/cmap/establish/handshake/mod.rs
+++ b/src/cmap/establish/handshake/mod.rs
@@ -35,7 +35,7 @@ struct ClientMetadata {
     application: Option<AppMetadata>,
     driver: DriverMetadata,
     os: OsMetadata,
-    platform: Option<String>,
+    platform: String,
 }
 
 #[derive(Clone, Debug)]
@@ -74,10 +74,7 @@ impl From<ClientMetadata> for Bson {
         );
 
         metadata_doc.insert("os", metadata.os);
-
-        if let Some(platform) = metadata.platform {
-            metadata_doc.insert("platform", platform);
-        }
+        metadata_doc.insert("platform", metadata.platform);
 
         Bson::Document(metadata_doc)
     }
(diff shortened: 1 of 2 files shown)


src/cmap/establish/handshake/mod.rs
```diff
@@ -35,7 +35,7 @@ struct ClientMetadata {
     application: Option<AppMetadata>,
     driver: DriverMetadata,
     os: OsMetadata,
-    platform: Option<String>,
```
kmahar, line 38: this only seemed to be an option to allow it to be initially set to `None` in `BASE_CLIENT_METADATA` below, and then conditionally set if `version_check::triple` returned a value. 
since we can unconditionally set that value now in one line, and the fact that we allowed this value to be `None` is kind of how this bug existed in the first place, it seems preferable to make it non-optional.

patrickfreed (approved):

kmahar:
just realized I made the base branch here main rather than 2.2.x, but I think these changes can be cherry-picked cleanly onto that branch, so I'll do that after this gets merged

abr-egn (approved):
LGTM!

isabelatkinson (approved):

PR #623, opened 2022-04-13
RUST-1168 Future compatibility for change streams

abr-egn:
RUST-1168

This PR syncs tests that check handling of new operation types and new fields in various places.  New operation types required setting up some serde helper types, and passing a test that needed a field preserved required switching from `ChangeStream<ChangeStreamEvent<Document>>` to `ChangeStream<Document>` in test contexts.

### DIFF
--- a/src/change_stream/event.rs
+++ b/src/change_stream/event.rs
@@ -192,7 +192,7 @@ impl<'a> From<&'a OperationType> for OperationTypeWrapper<'a> {
             OperationType::Rename => Self::Known(OperationTypeHelper::Rename),
             OperationType::DropDatabase => Self::Known(OperationTypeHelper::DropDatabase),
             OperationType::Invalidate => Self::Known(OperationTypeHelper::Invalidate),
-            OperationType::Other(s) => Self::Unknown(&s),
+            OperationType::Other(s) => Self::Unknown(s),
         }
     }
 }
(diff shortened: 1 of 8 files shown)


kmahar (commented):
the changes LGTM, though it's unfortunate that this means we lose all the unified test runner coverage of our `ChangeStreamEvent` type. however, I can't really think of a way around that besides maybe maintaining some manual list of tests that require a `ChangeStream<Document>`, and I don't know whether that would be worth it. any thoughts? do we have other coverage of `ChangeStreamEvent` already?

abr-egn:
> the changes LGTM, though it's unfortunate that this means we lose all the unified test runner coverage of our `ChangeStreamEvent` type. however, I can't really think of a way around that besides maybe maintaining some manual list of tests that require a `ChangeStream<Document>`, and I don't know whether that would be worth it. any thoughts? do we have other coverage of `ChangeStreamEvent` already?

Plumbing through whether or not to use `ChangeStreamEvent` looks like it'd be a giant hassle - fortunately, the prose tests use it pretty extensively, so I think it's well covered even without that.

kmahar:
> Plumbing through whether or not to use `ChangeStreamEvent` looks like it'd be a giant hassle - fortunately, the prose tests use it pretty extensively, so I think it's well covered even without that.

Cool, seems fine to me then

kmahar (approved):

patrickfreed (approved):

isabelatkinson (approved):

PR #625, opened 2022-04-13
minor: add links to the manual and docs page

abr-egn:


kmahar (approved):
maybe we want to update the homepage on 2.2.x before release too so that gets pushed to crates.io sooner?

isabelatkinson (approved):

patrickfreed (approved):

PR #628, opened 2022-04-14
RUST-585 Refactor Topology implementation to actor model

patrickfreed:
RUST-585

This PR refactors the Topology implementation to follow an actor model (similar to the connection pool) rather than a lock based one, hopefully making the code easier to maintain and reason about.


src/sdam/message_manager.rs
```diff
@@ -1,64 +0,0 @@
-use std::time::Duration;
```
patrickfreed, line 1: this is no longer needed now that we have the new handles

src/sdam/topology.rs
```diff
@@ -0,0 +1,949 @@
...
+        self.receiver
+            .borrow()
+            .description
+            .get_server_description(address)
+            .cloned()
+    }
+
+    /// Clone the latest state, marking it as seen.
+    pub(crate) fn clone_latest(&mut self) -> TopologyState {
+        self.receiver.borrow_and_update().clone()
```
abr-egn, line 778: I don't understand the usage of the receiver's "seen" flag.  It's set in `clone_latest` but not `borrow_latest` or `server_description`?

patrickfreed: The "seen" flag determines whether the currently observed `TopologtState` would be considered "new" or not for the purposes of `wait_for_update`. If we've called `clone_latest` and then immediately call `wait_for_update`, it'll block. If we haven't, it'll return immediately.

Not setting it in `borrow_latest` and `server_description` was more of an ergonomic choice, since it would require those methods to be `&mut self`.

To make this a bit clearer, I renamed the methods to `observe_latest` (formerly `clone_latest`) and `peek_latest` (formerly `borrow_latest`). Let me know if you think these are more helpful / if there are better ones.

src/sdam/topology.rs
```diff
@@ -0,0 +1,949 @@
...
+            check_requester: update_requester,
+            _worker_handle: worker_handle,
+        })
+    }
+
+    /// Begin watching for changes in the topology.
+    pub(crate) fn watch(&self) -> TopologyWatcher {
+        let mut watcher = self.watcher.clone();
+        // mark the latest topology as seen
+        watcher.receiver.borrow_and_update();
```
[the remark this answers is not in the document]
patrickfreed: This is to ensure that any calls to `wait_for_update` on the returned `TopologyWatcher` will block until a new state is published and not return immediately.

src/client/mod.rs
```diff
@@ -391,40 +385,30 @@ impl Client {
...
-                .unwrap_or_else(|| Duration::from_millis(0));
-
-            let change_occurred = topology_change_subscriber
-                .wait_for_message(time_remaining)
-                .await;
+            self.inner.topology.request_update();
 
+            let change_occurred = start_time.elapsed() < timeout
+                && watcher
+                    .wait_for_update(timeout - start_time.elapsed())
```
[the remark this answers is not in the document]
patrickfreed: each iteration of the loop marks the topology as "seen" in the `clone_latest` call, so any update after that point (potentially before the `request_update` call actually) will be accounted for in `wait_for_update`.

src/sdam/topology.rs
abr-egn: Much clearer, thank you :)

abr-egn (approved):
LGTM!  Definitely much easier to follow, and thank you for explaining the bits I didn't get.

kmahar (commented):
few minor questions but besides those LGTM!

src/runtime/acknowledged_message.rs
```diff
@@ -14,22 +14,41 @@ impl<M, R> AcknowledgedMessage<M, R> {
...
+                acknowledger: AcknowledgmentSender { sender },
             },
             AcknowledgmentReceiver { receiver },
         )
     }
 
-    /// Borrow the message.
-    pub(crate) fn message(&self) -> &M {
+    /// Send acknowledgement to the receiver.
+    #[allow(dead_code)]
```
isabelatkinson, line 24: any reason for keeping this around even though it's unused?

isabelatkinson (commented):
looks good, just one question

patrickfreed (commented):
To try to preserve the review comment / commit history, I did a regular merge commit to fix the merge conflicts. Will still squash it all down to a single commit at the end though.

kmahar (approved):

isabelatkinson (approved):

PR #630, opened 2022-04-18
RUST-1017 Reduce flakiness of SDAM `heartbeat_events` test

patrickfreed:
RUST-1017

This PR attempts to reduce the flakiness of the `heartbeat_events` test by scoping the failpoint only to commands issued with a given appname to prevent collisions with other tests. I *think* this is the cause of the occasional failures, but we'll have to see.

### DIFF
--- a/src/sdam/description/topology/test/sdam.rs
+++ b/src/sdam/description/topology/test/sdam.rs
@@ -603,6 +603,7 @@ async fn heartbeat_events() {
     let mut options = CLIENT_OPTIONS.clone();
     options.hosts.drain(1..);
     options.heartbeat_freq = Some(Duration::from_millis(50));
+    options.app_name = "heartbeat_events".to_string().into();
 
     let event_handler = EventHandler::new();
     let mut subscriber = event_handler.subscribe();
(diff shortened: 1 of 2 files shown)


abr-egn (approved):

isabelatkinson (approved):

kmahar:
I'm not sure I follow what the theoretical problem is. we don't run this test in parallel, right? how would other tests be hitting the failpoint before we do?

patrickfreed:
We actually do run them in parallel since that's Cargo's default (except on macOS where we were hitting thread / file limits a lot back in the day). That said, it's been a source of pain for a while imo, and I'd personally be open to trying out running them sequentially if there isn't a serious impact on test times. From looking at the waterfall, there might not be? Some macOS runs seem to take the same amount of time as the equivalent Ubuntu ones.

kmahar:
ah I see, we talked about this offline, I was thinking the test lock completely prevented any interaction between the tests but that's not the case. this solution seems fine to me then, agree it may be worth exploring sequential test runs

kmahar (approved):

PR #632, opened 2022-04-19
RUST-1283 Simulate docs.rs build environment on Evergreen

kmahar:
Fixes our broken docs.rs build by bumping to the latest version of `rustc_version_runtime`, and updates our check-rustdoc.sh script to try to more closely simulate docs.rs builds.

Here's an example of check-rustdoc.sh failing with these changes when using the old version of `rustc_version_runtime`: https://spruce.mongodb.com/task/mongo_rust_driver_lint_check_rustdoc_patch_43723f9cf53fa91243dd066c7b28fa5fda8b504d_625f057e0305b9378a239ee0_22_04_19_18_54_58/logs?execution=0

The error is slightly different because the error is due to a being in a read-only directory rather than in an entirely read-only file system, but I think it probably gives close enough coverage.

### DIFF
--- a/.evergreen/check-rustdoc.sh
+++ b/.evergreen/check-rustdoc.sh
@@ -2,7 +2,7 @@
 
 set -o errexit
 
-. ~/.cargo/env
+source ./.evergreen/env.sh
 
 # docs.rs builds the driver on a read-only file system. to create a more realistic environment, we first
 # build the driver to ensure we have all the deps already in src, and then limit the permissions on that directory.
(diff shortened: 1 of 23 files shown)


.evergreen/check-rustdoc.sh
```diff
@@ -3,6 +3,16 @@
...
 . ~/.cargo/env
+
+# docs.rs builds the driver on a read-only file system. to create a more realistic environment, we first
+# build the driver to ensure we have all the deps already in src, and then limit the permissions on that directory.
+# this is to help us avoid introducing problems like those described here 
+# https://docs.rs/about/builds#read-only-directories where we or a dependency modify source code during the
+# build process.
+cargo +nightly build
+cargo clean
+chmod -R 555 ~/.cargo/registry/src
```
kmahar, line 14: so you don't have to look it up, this gives read and execute, but not write, permission, to everyone.

PR #633, opened 2022-04-20
RUST-1166 Add support for pre-images in change streams.

abr-egn:
RUST-1166

This adds the options and output fields to allow pre-images to be reported by change streams.

### DIFF
--- a/src/change_stream/options.rs
+++ b/src/change_stream/options.rs
@@ -122,7 +122,7 @@ pub enum FullDocumentType {
     /// event is available.
     WhenAvailable,
 
-    /// The same behavior as 'whenAvailable' except that an error is raised if the post-image is
+    /// The same behavior as `WhenAvailable` except that an error is raised if the post-image is
     /// not available.
     Required,
 
@@ -141,7 +141,7 @@ pub enum FullDocumentBeforeChangeType {
     /// for this event is available.
     WhenAvailable,
 
-    /// The same behavior as 'whenAvailable' except that an error is raised if the pre-image is
+    /// The same behavior as `WhenAvailable` except that an error is raised if the pre-image is
     /// not available.
     Required,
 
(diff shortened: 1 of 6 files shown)


src/change_stream/options.rs
```diff
@@ -103,20 +108,46 @@ impl ChangeStreamOptions {
...
+    /// The field will be populated for replace, update, and delete change events if the pre-image
+    /// for this event is available.
+    WhenAvailable,
+
+    /// The same behavior as `WhenAvailable` except that an error is raised if the pre-image is
+    /// not available.
+    Required,
+
+    /// Do not send a value.
+    Off,
```
patrickfreed, line 149: this doesn't seem particularly useful but I see the spec has it so I guess we might as well too

patrickfreed (approved):
LGTM!

src/change_stream/options.rs
abr-egn: Yeah, I initially left it out but the spec tests use it.

src/db/options.rs
```diff
@@ -103,6 +103,9 @@ pub struct CreateCollectionOptions {
         serialize_with = "bson_util::serialize_duration_option_as_int_secs"
     )]
     pub expire_after_seconds: Option<Duration>,
+
+    /// Options for supporting change stream pre- and post-images.
+    pub change_stream_pre_and_post_images: Option<ChangeStreamPreAndPostImages>,
```
kmahar, line 108: are we actually testing / exercising this option anywhere? it doesn't look like it from the spec tests, which always use collMod to configure the option after the fact.

abr-egn: Nope, it's not covered by the spec tests :(

kmahar: can we add a test of our own then?

abr-egn: Done.

kmahar (approved):

isabelatkinson (approved):

PR #634, opened 2022-04-21
RUST-1272 Bump rustc_version_runtime to fix docs.rs build

kmahar:
Bump to a newer version of rustc_version_runtime which does not need to write to the `src` directory of the crate. docs.rs builds the docs on a read-only file system which does not permit that.


isabelatkinson (approved):

kmahar:
equivalent PR targeting 2.2 branch: #634 

will merge both and release 2.2.1 pending evergreen success

alekseysidorov:
I also have an issue like this when I am trying to build a project by `nix build`, because nix mounts the source directory as read-only.

patrickfreed (approved):

kmahar:
@alekseysidorov, we should have a fix released shortly

PR #636, opened 2022-04-21
RUST-616 Error if message is too long

abr-egn:
RUST-616

This adds support for the `maxMessageSizeBytes` hello response field, and validates message header `length` against that maximum.

### DIFF
--- a/src/cmap/conn/stream_description.rs
+++ b/src/cmap/conn/stream_description.rs
@@ -37,7 +37,7 @@ pub(crate) struct StreamDescription {
     pub(crate) hello_ok: bool,
 
     /// The maximum permitted size of a BSON wire protocol message.
-    pub max_message_size_bytes: Option<i32>,
+    pub(crate) max_message_size_bytes: i32,
 }
 
 impl StreamDescription {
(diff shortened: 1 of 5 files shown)


src/cmap/conn/stream_description.rs
```diff
@@ -35,6 +35,9 @@ pub(crate) struct StreamDescription {
 
     /// Whether the server associated with this connection supports the `hello` command.
     pub(crate) hello_ok: bool,
+
+    /// The maximum permitted size of a BSON wire protocol message.
+    pub max_message_size_bytes: Option<i32>,
```
kmahar, line 40: why `pub` and not `pub(crate)` like the other properties on this type? (...does that distinction even matter since the type itself is `pub(crate)`?)

src/hello.rs
```diff
@@ -231,6 +231,9 @@ pub(crate) struct HelloCommandResponse {
 
     /// For internal use.
     pub topology_version: Option<Document>,
+
+    /// The maximum permitted size of a BSON wire protocol message.
+    pub max_message_size_bytes: Option<i32>,
```
kmahar, line 236: is this field ever actually not present in a hello response? or could it be non-optional here and in `StreamDescription`?

kmahar (commented):
looks good, just a couple small questions!

src/hello.rs
abr-egn: Looks like it's always there, updated.

kmahar (approved):

patrickfreed (approved):

isabelatkinson (approved):

PR #640, opened 2022-04-22
RUST-1078 Fix and start running test for clientMetadata in handshake

patrickfreed:
RUST-1078

### DIFF
--- a/src/test/client.rs
+++ b/src/test/client.rs
@@ -15,55 +15,74 @@ use crate::{
 };
 
 #[derive(Debug, Deserialize)]
-struct Metadata {
-    #[serde(rename = "clientMetadata")]
-    pub client: ClientMetadata,
-}
-
-#[derive(Debug, Deserialize)]
-#[allow(dead_code)]
 struct ClientMetadata {
     pub driver: DriverMetadata,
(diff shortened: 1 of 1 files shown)


isabelatkinson (approved):

src/test/client.rs
```diff
@@ -15,55 +15,81 @@ use crate::{
...
-struct OsMetadata {
-    #[serde(rename = "type")]
-    pub os_type: String,
-    pub architecture: String,
-}
-
-// TODO RUST-1078: fix and run this test. Once this test is fully implemented we can remove the
-// #[allow(dead_code)] tags on the above structs.
-//
-// This test currently doesn't pass on replica sets and sharded clusters consistently due to
```
kmahar, line 48: the comment implies that this test doesn't pass consistently on replica sets, but I see we are testing on replica sets, so was this wrong?

kmahar (approved):

src/test/client.rs
kmahar: I meant to delete this comment because I realized your use of command.currentOp fixes the heartbeat problem, but it seems like the deletion didn't actually succeed

PR #642, opened 2022-04-25
RUST-1193 Provide a public `ConnectionString` type

abr-egn:
RUST-1193

This renames and makes public the existing `ClientOptionsParser` struct as `ConnectionString` to allow for parsing and modeling connection strings without requiring IO for DNS lookup.  I was deliberately conservative in this addition - there's no builder type, so the only way to construct one is from a string, and only those fields that were already directly exposed in `ClientOptions` are also exposed here.

### DIFF
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -794,7 +794,7 @@ pub struct ConnectionString {
     pub load_balanced: Option<bool>,
 
     /// Amount of time spent attempting to send or receive on a socket before timing out; note that
-    /// this only applies to application operations, not SDAM.
+    /// this only applies to application operations, not server discovery and monitoring.
     pub socket_timeout: Option<Duration>,
 
     /// Default read preference for the client.
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -2145,6 +2145,13 @@ impl ConnectionString {
     }
 }
 
+impl FromStr for ConnectionString {
+    type Err = Error;
+    fn from_str(s: &str) -> Result<Self> {
+        ConnectionString::parse(s)
+    }
+}
+
 impl<'de> Deserialize<'de> for ConnectionString {
     fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
     where
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -718,10 +718,6 @@ pub struct ConnectionString {
     /// ReadConcern type documentation for more details.
     pub read_concern: Option<ReadConcern>,
 
-    /// The default selection criteria for operations performed on the Client. See the
-    /// SelectionCriteria type documentation for more details.
-    pub selection_criteria: Option<SelectionCriteria>,
-
     /// The name of the replica set that the Client should connect to.
     pub repl_set_name: Option<String>,
 
@@ -804,10 +800,6 @@ pub struct ConnectionString {
     /// 9 signifies the best compression.
     pub zlib_compression: Option<i32>,
 
-    /// The maximum replication lag, in wall clock time, that a secondary can suffer and still be
-    /// eligible for server selection.
-    pub max_staleness: Option<Duration>,
-
     /// The authentication mechanism method to use for connection to the server.
     pub auth_mechanism: Option<AuthMechanism>,
 
@@ -831,6 +823,7 @@ pub struct ConnectionString {
 #[derive(Default)]
 struct ConnectionStringParts {
     read_preference_tags: Option<Vec<TagSet>>,
+    max_staleness: Option<Duration>,
 }
 
 /// Specifies whether TLS configuration should be used with the operations that the
@@ -966,7 +959,7 @@ impl From<ConnectionString> for ClientOptions {
             heartbeat_freq: parser.heartbeat_freq,
             local_threshold: parser.local_threshold,
             read_concern: parser.read_concern,
-            selection_criteria: parser.selection_criteria,
+            selection_criteria: parser.read_preference.map(Into::into),
             repl_set_name: parser.repl_set_name,
             write_concern: parser.write_concern,
             max_pool_size: parser.max_pool_size,
@@ -1644,7 +1637,7 @@ impl ConnectionString {
             };
         }
 
-        if let Some(max_staleness) = self.max_staleness.take() {
+        if let Some(max_staleness) = parts.max_staleness {
             self.read_preference = match self.read_preference.take() {
                 Some(read_pref) => Some(read_pref.with_max_staleness(max_staleness)?),
                 None => {
@@ -1657,8 +1650,6 @@ impl ConnectionString {
                 }
             };
         }
-
-        self.selection_criteria = self.read_preference.take().map(Into::into);
 
         if let Some(true) = self.direct_connection {
             if self.srv {
@@ -1846,7 +1837,7 @@ impl ConnectionString {
                     Ordering::Greater => Duration::from_secs(max_staleness_seconds as u64),
                 };
 
-                self.max_staleness = Some(max_staleness);
+                parts.max_staleness = Some(max_staleness);
             }
             k @ "maxpoolsize" => {
                 self.max_pool_size = Some(get_u32!(value, k));
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -1172,6 +1172,18 @@ impl ClientOptions {
         Ok(options)
     }
 
+    /// Creates a `ClientOptions` from the given `ConnectionString`.
+    ///
+    /// In the case that "mongodb+srv" is used, SRV and TXT record lookups will be done using the
+    /// provided `ResolverConfig` as part of this method.
+    #[cfg(any(feature = "sync", feature = "tokio-sync"))]
+    pub fn parse_connection_string_sync(
+        conn_str: ConnectionString,
+        resolver_config: impl Into<Option<ResolverConfig>>,
+    ) -> Result<Self> {
+        crate::runtime::block_on(Self::parse_connection_string(conn_str, resolver_config))
+    }
+
     #[cfg(test)]
     pub(crate) fn parse_without_srv_resolution(s: &str) -> Result<Self> {
         let mut conn_str = ConnectionString::parse(s)?;
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -835,7 +835,7 @@ pub enum HostInfo {
     /// A set of addresses.
     HostIdentifiers(Vec<ServerAddress>),
     /// A DNS record for SRV lookup.
-    DnsRecord(String)
+    DnsRecord(String),
 }
 
 impl Default for HostInfo {
@@ -848,7 +848,7 @@ impl Into<Vec<ServerAddress>> for HostInfo {
     fn into(self) -> Vec<ServerAddress> {
         match self {
             Self::HostIdentifiers(hosts) => hosts,
-            Self::DnsRecord(host) => vec![ServerAddress::Tcp { host, port: None }]
+            Self::DnsRecord(host) => vec![ServerAddress::Tcp { host, port: None }],
         }
     }
 }
@@ -1479,7 +1479,8 @@ impl ConnectionString {
             None => (None, None),
         };
 
-        let host_list: Result<Vec<_>> = hosts_section.split(',').map(ServerAddress::parse).collect();
+        let host_list: Result<Vec<_>> =
+            hosts_section.split(',').map(ServerAddress::parse).collect();
 
         let host_list = host_list?;
 
@@ -1713,7 +1714,12 @@ impl ConnectionString {
         Ok(parts)
     }
 
-    fn parse_option_pair(&mut self, parts: &mut ConnectionStringParts, key: &str, value: &str) -> Result<()> {
+    fn parse_option_pair(
+        &mut self,
+        parts: &mut ConnectionStringParts,
+        key: &str,
+        value: &str,
+    ) -> Result<()> {
         macro_rules! get_bool {
             ($value:expr, $option:expr) => {
                 match $value {
@@ -1937,7 +1943,8 @@ impl ConnectionString {
                         .collect()
                 };
 
-                parts.read_preference_tags
+                parts
+                    .read_preference_tags
                     .get_or_insert_with(Vec::new)
                     .push(tags?);
             }
@@ -2127,7 +2134,8 @@ impl ConnectionString {
 impl<'de> Deserialize<'de> for ConnectionString {
     fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
     where
-        D: Deserializer<'de> {
+        D: Deserializer<'de>,
+    {
         deserializer.deserialize_str(ConnectionStringVisitor)
     }
 }
@@ -2143,7 +2151,8 @@ impl<'de> serde::de::Visitor<'de> for ConnectionStringVisitor {
 
     fn visit_str<E>(self, v: &str) -> std::result::Result<Self::Value, E>
     where
-            E: serde::de::Error, {
+        E: serde::de::Error,
+    {
         ConnectionString::parse(v).map_err(serde::de::Error::custom)
     }
 }
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -250,6 +250,12 @@ impl ServerAddress {
         }
     }
 
+    pub(crate) fn into_host(self) -> String {
+        match self {
+            Self::Tcp { host, .. } => host,
+        }
+    }
+
     pub(crate) fn port(&self) -> Option<u16> {
         match self {
             Self::Tcp { port, .. } => *port,
@@ -684,7 +690,7 @@ pub struct ConnectionString {
     /// Note that by default, the driver will autodiscover other nodes in the cluster. To connect
     /// directly to a single server (rather than autodiscovering the rest of the cluster), set the
     /// `direct_connection` field to `true`.
-    pub hosts: Vec<ServerAddress>,
+    pub hosts: HostInfo,
 
     /// The application name that the Client will send to the server as part of the handshake. This
     /// can be used in combination with the server logs to determine which Client is connected to a
@@ -806,7 +812,6 @@ pub struct ConnectionString {
     /// Default read preference for the client.
     pub read_preference: Option<ReadPreference>,
 
-    pub(crate) srv: bool,
     wait_queue_timeout: Option<Duration>,
     tls_insecure: Option<bool>,
     original_uri: String,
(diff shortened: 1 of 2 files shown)


src/client/options/mod.rs
```diff
@@ -675,40 +675,139 @@ impl Serialize for ClientOptions {
...
+    pub(crate) wait_queue_timeout: Option<Duration>,
+    pub(crate) socket_timeout: Option<Duration>,
+    pub(crate) zlib_compression: Option<i32>,
     max_staleness: Option<Duration>,
     tls_insecure: Option<bool>,
     auth_mechanism: Option<AuthMechanism>,
     auth_source: Option<String>,
     auth_mechanism_properties: Option<Document>,
     read_preference: Option<ReadPreference>,
     read_preference_tags: Option<Vec<TagSet>>,
```
isabelatkinson, line 810: I think we should make some of these public (specifically the ones that are outlined in the [uri-options](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst) spec). Basically anything that is a key in a connection string can be a public option here.

src/client/options/mod.rs
```diff
@@ -995,10 +1094,21 @@ impl ClientOptions {
...
-        let auth_source_present = parser.auth_source.is_some();
-        let mut options: Self = parser.into();
+        Self::resolve(ConnectionString::parse(uri)?, resolver_config).await
+    }
+
+    /// Creates a `ClientOptions` from the given `ConnectionString`.
+    ///
+    /// In the case that "mongodb+srv" is used, SRV and TXT record lookups will be done using the
+    /// provided `ResolverConfig` as part of this method.
+    pub async fn resolve(
```
isabelatkinson, line 1104: can we indicate in this name somehow that the connection string is being resolved to `ClientOptions`? it might be slightly unreadable in user code to determine what the output of this method is if someone isn't super familiar with the interaction between the two types

src/client/options/mod.rs
abr-egn: Done.

src/client/options/mod.rs
abr-egn: Good thought; renamed this to `ClientOptions::resolve_from`, which I think makes this more clear.

src/client/options/mod.rs
```diff
@@ -675,40 +675,170 @@ impl Serialize for ClientOptions {
...
+
+    /// Default read preference tags for the client; only valid if the read preference mode is not
+    /// primary.
+    ///
+    /// The order of the tag sets in the read preference is the same as the order they are
+    /// specified in the URI.
+    pub read_preference_tags: Option<Vec<TagSet>>,
+
+    /// Amount of time spent attempting to check out a connection from a server's connection pool
+    /// before timing out.  Not supported by the Rust driver.
```
kmahar, line 834: why include this and the next one that are unsupported? what will happen if users try to use them?

abr-egn: It seemed useful to include them for domain modeling, i.e. even if the driver doesn't support them customers could use this to parse connection strings and look at those values.  I've moved them to private fields with read-only accessors to make it more clear they shouldn't be set.

Current behavior is the driver ignores those fields.  I'd prefer it if attempting to use them caused an invalid argument error, but that would be a breaking change :(

src/client/options/mod.rs
```diff
@@ -675,43 +675,181 @@ impl Serialize for ClientOptions {
...
+
+    /// The TLS configuration for the Client to use in its connections with the server.
+    ///
+    /// By default, TLS is disabled.
     pub tls: Option<Tls>,
+
+    /// The amount of time each monitoring thread should wait between performing server checks.
+    ///
+    /// The default value is 10 seconds.
     pub heartbeat_freq: Option<Duration>,
```
patrickfreed, line 704: I know the `ClientOptions` analogue for this is also `heartbeat_freq`, but given that this type is mirror a connection string, I feel like it might be better to name this `heartbeat_frequency` to match the actual option. Wdyt?

src/client/options/mod.rs
```diff
@@ -675,43 +675,181 @@ impl Serialize for ClientOptions {
...
+    pub(crate) srv: bool,
+    wait_queue_timeout: Option<Duration>,
     tls_insecure: Option<bool>,
-    auth_mechanism: Option<AuthMechanism>,
-    auth_source: Option<String>,
-    auth_mechanism_properties: Option<Document>,
-    read_preference: Option<ReadPreference>,
-    read_preference_tags: Option<Vec<TagSet>>,
-    load_balanced: Option<bool>,
     original_uri: String,
```
patrickfreed, line 836: I think storing this here might actually be a bit dangerous, since users could mutate this connection string via its fields and create a disconnect. I think we'll instead want to implement `Display` on `ConnectionString` and use that implementation if we need to get the original / current URI.

Do you have any idea how much work that would constitute @isabelatkinson? If it's a lot, we could break it (and `Serialize`) implementations out into a separate ticket, seeing as this is a quick win.

src/client/options/mod.rs
```diff
@@ -675,43 +675,181 @@ impl Serialize for ClientOptions {
     }
 }
 
+/// Contains the options that can be set via a MongoDB connection string.
+///
+/// The format of a MongoDB connection string is described [here](https://docs.mongodb.com/manual/reference/connection-string/#connection-string-formats).
 #[derive(Debug, Default, PartialEq)]
```
patrickfreed, line 681: I think it would be nice to have a `Deserialize` implementation that parses a string. See my other comment below about going back to strings which we'd use for a `Display` and `Serialize` implementation.

src/client/options/mod.rs
```diff
@@ -675,43 +675,181 @@ impl Serialize for ClientOptions {
...
+impl ConnectionString {
+    /// Amount of time spent attempting to check out a connection from a server's connection pool
+    /// before timing out.  Not supported by the Rust driver.
+    pub fn wait_queue_timeout(&self) -> Option<Duration> {
+        self.wait_queue_timeout
+    }
+
+    /// Relax TLS constraints as much as possible (e.g. allowing invalid certificates or hostname
+    /// mismatches).  Not supported by the Rust driver.
+    pub fn tls_insecure(&self) -> Option<bool> {
```
patrickfreed, line 848: Now that we have OpenSSL support, I _think_ we can / do support this? If not, maybe we should file a ticket about adding support, at least for OpenSSL only.

src/client/options/mod.rs
```diff
@@ -675,43 +675,181 @@ impl Serialize for ClientOptions {
...
+    /// The default value is 15 ms.
     pub local_threshold: Option<Duration>,
+
+    /// Specifies the default read concern for operations performed on the Client. See the
+    /// ReadConcern type documentation for more details.
     pub read_concern: Option<ReadConcern>,
+
+    /// The default selection criteria for operations performed on the Client. See the
+    /// SelectionCriteria type documentation for more details.
     pub selection_criteria: Option<SelectionCriteria>,
```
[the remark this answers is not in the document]
abr-egn: As long as the transformations to create them from the raw options are lossless parsing and aggregation, I'd be inclined to go with the higher-level structs and remove or hide the components (e.g. `read_preference_tags`).  That feels more idiomatic for Rust - see for example clap, which allows mapping fairly [complex nested structures](https://github.com/clap-rs/clap/blob/v3.1.14/examples/git-derive.rs) to the cli args.

I'd be fine going the other way if consensus disagrees with me, though :)  @isabelatkinson @kmahar thoughts?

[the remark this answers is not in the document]
kmahar: I think I'm slightly partial to the consolidated types because they better express the relationship between the individual options. the fact that connection strings have e.g. 3 top-level options around read preference feels mainly like a shortcoming of what can be expressed with a string. I think it could be sufficient to note in the docstring for this property which connection string options can be used together to specify the read preference. I don't feel that strongly either way though.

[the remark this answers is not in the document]
isabelatkinson: Yeah, I would prefer the higher level types for the reasons Kaitlin and Abraham outlined.

src/client/options/mod.rs
isabelatkinson: [This](https://github.com/mongodb/mongo-swift-driver/blob/08d68a5c863d10da7221880a6a98e758a99bb63d/Sources/MongoSwift/MongoConnectionString.swift#L931) is the equivalent implementation in Swift -- it's not a ton of work but it is pretty tedious. I think it would be fine to do in a separate ticket (I'm pretty sure I did that implementation as its own PR) as long as it gets done prior to us releasing this work.

src/client/options/mod.rs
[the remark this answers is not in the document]
patrickfreed: Consolidated types SGTM then. Since we're going that route, should make sure to remove the fields duplicated by them.

src/client/options/mod.rs
abr-egn: Filed [RUST-1309](https://jira.mongodb.org/browse/RUST-1309) to track this.

src/client/options/mod.rs
abr-egn: Done.

src/client/options/mod.rs
abr-egn: Makes sense; I filed [RUST-1308](https://jira.mongodb.org/browse/RUST-1308) for that.  `original_uri` is used in a lot of tests so I didn't want to remove it outright, but I added `#[cfg(test)]` to enforce that it's not part of internal or external API.

src/client/options/mod.rs
abr-egn: Done.

src/client/options/mod.rs
```diff
@@ -675,43 +675,181 @@ impl Serialize for ClientOptions {
...
 #[derive(Debug, Default, PartialEq)]
-struct ClientOptionsParser {
+#[non_exhaustive]
+pub struct ConnectionString {
+    /// The initial list of seeds that the Client should connect to.
+    ///
+    /// Note that by default, the driver will autodiscover other nodes in the cluster. To connect
+    /// directly to a single server (rather than autodiscovering the rest of the cluster), set the
+    /// `direct_connection` field to `true`.
     pub hosts: Vec<ServerAddress>,
```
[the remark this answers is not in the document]
abr-egn: `HostInfo` makes sense, done.  I don't think there's a need for a `ConnectionStringScheme` enum when `HostInfo` can just be directly matched against, though, so left that out.  If a new scheme is added that introduces ambiguity with that matching it'll be easy to introduce a field at that point.

src/client/options/mod.rs
```diff
@@ -675,43 +675,181 @@ impl Serialize for ClientOptions {
...
+    pub read_preference: Option<ReadPreference>,
+
+    /// Default read preference tags for the client; only valid if the read preference mode is not
+    /// primary.
+    ///
+    /// The order of the tag sets in the read preference is the same as the order they are
+    /// specified in the URI.
+    pub read_preference_tags: Option<Vec<TagSet>>,
+
+    pub(crate) srv: bool,
```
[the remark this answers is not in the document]
abr-egn: See below.

src/client/options/mod.rs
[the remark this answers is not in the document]
abr-egn: Updated; it looks like only `ReadPreference` and `Credential` needed consolidation, the others already did it inline.

src/client/options/mod.rs
```diff
@@ -673,41 +680,177 @@ impl Serialize for ClientOptions {
...
-    pub hosts: Vec<ServerAddress>,
-    pub srv: bool,
+#[non_exhaustive]
+pub struct ConnectionString {
+    /// The initial list of seeds that the Client should connect to.
+    ///
+    /// Note that by default, the driver will autodiscover other nodes in the cluster. To connect
+    /// directly to a single server (rather than autodiscovering the rest of the cluster), set the
+    /// `direct_connection` field to `true`.
+    pub hosts: HostInfo,
```
patrickfreed, line 694: can we update the docstring to indicate that this can also contain a DNS name used for SRV lookups for the initial seed list? Also, I think renaming the field `host_info` might be more appropriate given that.

src/client/options/mod.rs
```diff
@@ -673,41 +680,177 @@ impl Serialize for ClientOptions {
...
+    /// The default value is 15 ms.
     pub local_threshold: Option<Duration>,
+
+    /// Specifies the default read concern for operations performed on the Client. See the
+    /// ReadConcern type documentation for more details.
     pub read_concern: Option<ReadConcern>,
-    pub selection_criteria: Option<SelectionCriteria>,
+
+    /// The name of the replica set that the Client should connect to.
     pub repl_set_name: Option<String>,
```
patrickfreed, line 729: Along the lines of the other option, should we name this `replica_set` to match the actual URI option? We could also just document that it corresponds to the `replicaSet` option.

src/client/options/mod.rs
```diff
@@ -673,41 +680,177 @@ impl Serialize for ClientOptions {
...
+    pub load_balanced: Option<bool>,
+
+    /// Amount of time spent attempting to send or receive on a socket before timing out; note that
+    /// this only applies to application operations, not SDAM.
+    pub socket_timeout: Option<Duration>,
+
+    /// Specifies the level of compression when using zlib to compress wire protocol messages; -1
+    /// signifies the default level, 0 signifies no compression, 1 signifies the fastest speed, and
+    /// 9 signifies the best compression.
+    pub zlib_compression: Option<i32>,
```
patrickfreed, line 808: The zlib compressor enum variant contains a field for this, so I think we can omit it.

src/client/options/mod.rs
```diff
@@ -673,41 +680,177 @@ impl Serialize for ClientOptions {
...
+
+    /// The database that connections should authenticate against.
+    pub auth_source: Option<String>,
```
patrickfreed, line 811: this is included in `Credential` so it can be omitted too

src/client/options/mod.rs
```diff
@@ -834,38 +977,39 @@ pub struct DriverInfo {
     pub platform: Option<String>,
 }
 
-impl From<ClientOptionsParser> for ClientOptions {
-    fn from(parser: ClientOptionsParser) -> Self {
+impl From<ConnectionString> for ClientOptions {
```
patrickfreed, line 980: I think we actually don't want this implemented (and didn't before either really), since it bypasses the SRV resolution. Before it was somewhat okay since `ClientOptionsParser` was `pub(crate)`, but now that `ConnectionString` is public we'll probably need to remove this.

Instead, we can just update the parts in `ClientOptions::parse` that use this to first resolve the hosts (either by just getting them directly from the `ConnectionString` or by doing SRV lookup) and then basically doing the same thing that's done here but inline.

src/client/options/mod.rs
```diff
@@ -673,41 +680,177 @@ impl Serialize for ClientOptions {
...
+}
+
+impl Default for HostInfo {
+    fn default() -> Self {
+        Self::HostIdentifiers(vec![])
+    }
+}
+
+impl HostInfo {
+    fn into_addresses(self) -> Vec<ServerAddress> {
```
patrickfreed, line 848: Similar to my comment about about the `From` implementation, this kind of discards the distinction between an actual host and a DNS name used for SRV lookup (which contains actual hosts), so I think we should avoid having a method like this.

src/client/options/mod.rs
```diff
@@ -250,6 +250,12 @@ impl ServerAddress {
         }
     }
 
+    pub(crate) fn into_host(self) -> String {
+        match self {
+            Self::Tcp { host, .. } => host,
```
patrickfreed, line 255: A method like this could lead to accidentally discarding the port, so I think we'll want to omit it.

abr-egn: Done.

src/client/options/mod.rs
abr-egn: Inlined this into `parse_without_srv_resolution`.

src/client/options/mod.rs
abr-egn: Ah, didn't consider visibility - it's a pity there's no non-awkward way to have a non-pub `impl`.  Updated, although I had to preserve the struct-translation as its own method so it could be shared with `parse_without_srv_resolution`.

src/client/options/mod.rs
```diff
@@ -993,10 +1137,21 @@ impl ClientOptions {
...
-        let auth_source_present = parser.auth_source.is_some();
-        let mut options: Self = parser.into();
+        Self::resolve_from(ConnectionString::parse(uri)?, resolver_config).await
+    }
+
+    /// Creates a `ClientOptions` from the given `ConnectionString`.
+    ///
+    /// In the case that "mongodb+srv" is used, SRV and TXT record lookups will be done using the
+    /// provided `ResolverConfig` as part of this method.
+    pub async fn resolve_from(
```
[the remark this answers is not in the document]
abr-egn: Updated to `parse_connection_string`.

src/client/options/mod.rs
abr-egn: Done.  This did require passing through `auth_source_present` as a non-public field so that `ClientOptions` can know whether to override it or not.

src/client/options/mod.rs
abr-egn: Done.

src/client/options/mod.rs
abr-egn: Renamed.

src/client/options/mod.rs
abr-egn: Done.

src/client/options/mod.rs
```diff
@@ -993,70 +1101,126 @@ impl ClientOptions {
...
     #[cfg(test)]
     pub(crate) fn parse_without_srv_resolution(s: &str) -> Result<Self> {
-        let parser = ClientOptionsParser::parse(s)?;
-        let options: Self = parser.into();
+        let mut conn_str = ConnectionString::parse(s)?;
+        let host_info = std::mem::take(&mut conn_str.host_info);
+        let mut options = Self::from_connection_string(conn_str);
+        options.hosts = match host_info {
+            HostInfo::HostIdentifiers(hosts) => hosts,
+            HostInfo::DnsRecord(host) => vec![ServerAddress::Tcp { host, port: None }],
```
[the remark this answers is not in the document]
abr-egn: Done.

src/client/options/mod.rs
abr-egn: Updated.

src/client/options/mod.rs
```diff
@@ -1221,8 +1385,11 @@ fn validate_userinfo(s: &str, userinfo_type: &str) -> Result<()> {
     Ok(())
 }
 
-impl ClientOptionsParser {
-    fn parse(s: &str) -> Result<Self> {
+impl ConnectionString {
+    /// Parses a MongoDB connection string into a [`ConnectionString`] struct. If the string is
+    /// malformed or one of the options has an invalid value, an error will be returned.
+    pub fn parse(s: impl AsRef<str>) -> Result<Self> {
```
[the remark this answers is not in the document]
abr-egn: Done.

kmahar (approved):
couple small comments but LGTM otherwise, don't need to re-review.

src/client/options/mod.rs
```diff
@@ -993,70 +1100,147 @@ impl ClientOptions {
...
-        if srv {
-            let mut resolver = SrvResolver::new(resolver_config.map(|config| config.inner)).await?;
-            let mut config = resolver
-                .resolve_client_options(options.hosts[0].host())
-                .await?;
+    /// Creates a `ClientOptions` from the given `ConnectionString`.
+    ///
+    /// In the case that "mongodb+srv" is used, SRV and TXT record lookups will be done using the
+    /// provided `ResolverConfig` as part of this method.
+    pub async fn parse_connection_string(
```
kmahar, line 1110: #666 got me thinking, maybe we need a sync version of this method, too?

PR #643, opened 2022-04-26
RUST-690 Run `cargo deny` via evergreen

abr-egn:
RUST-690

This adds configuration for running `cargo deny` against the driver; this will give us errors on any security advisories, unknown licenses, or crate sources other than crates.io.


deny.toml
```diff
@@ -0,0 +1,213 @@
...
+    # the actual valid features supported by the target architecture.
+    #{ triple = "wasm32-unknown-unknown", features = ["atomics"] },
+]
+
+# This section is considered when running `cargo deny check advisories`
+# More documentation for the advisories section can be found here:
+# https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html
+[advisories]
+# The path where the advisory database is cloned/fetched into
+db-path = "~/.cargo/advisory-db"
```
patrickfreed, line 35: @kmahar will this need to change if/when we change CARGO_HOME for CI?

deny.toml
```diff
@@ -0,0 +1,213 @@
...
+    "MPL-2.0",
+]
+# List of explicitly disallowed licenses
+# See https://spdx.org/licenses/ for list of possible licenses
+# [possible values: any SPDX 3.11 short identifier (+ optional exception)].
+deny = [
+    #"Nokia",
+]
+# Lint level for licenses considered copyleft
+copyleft = "warn"
```
patrickfreed, line 92: From my understanding, we may need to use "deny" here, since if one of or dependencies is GPL, for example, our crate would then need to be GPL too.

deny.toml
```diff
@@ -0,0 +1,213 @@
...
+# The path where the advisory database is cloned/fetched into
+db-path = "~/.cargo/advisory-db"
+# The url(s) of the advisory databases to use
+db-urls = ["https://github.com/rustsec/advisory-db"]
+# The lint level for security vulnerabilities
+vulnerability = "deny"
+# The lint level for unmaintained crates
+unmaintained = "warn"
+# The lint level for crates that have been yanked from their source registry
+yanked = "warn"
```
patrickfreed, line 43: this seems like it would be good to "deny" on.

deny.toml
```diff
@@ -0,0 +1,213 @@
...
+# https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html
+[advisories]
+# The path where the advisory database is cloned/fetched into
+db-path = "~/.cargo/advisory-db"
+# The url(s) of the advisory databases to use
+db-urls = ["https://github.com/rustsec/advisory-db"]
+# The lint level for security vulnerabilities
+vulnerability = "deny"
+# The lint level for unmaintained crates
+unmaintained = "warn"
```
patrickfreed, line 41: If this doesn't current emit any warnings, it'd probably be better to have this (and anything we don't think is too noisy) to be "deny" instead of "warn". My concern is just that the warnings won't be visible enough to us if they're not causing CI to fail.

deny.toml
```diff
@@ -0,0 +1,213 @@
...
+# The lint level for crates that have been yanked from their source registry
+yanked = "warn"
+# The lint level for crates with security notices. Note that as of
+# 2019-12-17 there are no security notice advisories in
+# https://github.com/rustsec/advisory-db
+notice = "warn"
```
patrickfreed, line 47: ditto deny if we can.

.evergreen/config.yml
```diff
@@ -1675,6 +1689,14 @@ buildvariants:
   tasks:
     - "compile-only"
 
+- matrix_name: "cargo-deny"
```
patrickfreed, line 1692: Since we don't need any special matrix configuration, would it be fine to just include a new task in the `Lint` matrix?

deny.toml
kmahar: yeah I think we would want to put this in $CARGO_HOME rather than ~/.cargo.

deny.toml
abr-egn: Updated.

deny.toml
abr-egn: Updated.

deny.toml
abr-egn: Updated.

deny.toml
abr-egn: Updated.

deny.toml
abr-egn: It looks like cargo-deny [defaults](https://github.com/EmbarkStudios/cargo-deny/blob/main/src/advisories/helpers.rs#L51) to using `$CARGO_HOME` if the option isn't set, so updated the config to that.

.evergreen/config.yml
abr-egn: Good call, done.

kmahar (approved):
LGTM assuming we hear back that all the listed licenses are OK

isabelatkinson (approved):
lgtm (assuming the license list is fine with the legal team)!

patrickfreed (approved):
ditto LGTM pending license follow up

PR #644, opened 2022-04-26
RUST-1216 Always use `count` for `estimated_document_count`

patrickfreed:
RUST-1216

This PR updates the `estimated_document_count` method to always use the `count` server command rather than `aggregate` with a `$collStats` stage, fixing the regression on views introduced with the original change to `$collStats`.

For some rationale on why this is being introduced in a minor version, see the [CRUD spec Q&A](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#q-a).


patrickfreed:
The serverless failures are unrelated I believe--they're failing on the waterfall too.

src/test/spec/unified_runner/mod.rs
```diff
@@ -105,17 +105,24 @@ pub async fn run_unified_format_test_filtered(
             continue;
         }
 
-        if test_case
+        if let Some(op) = test_case
             .operations
             .iter()
-            .any(|op| SKIPPED_OPERATIONS.contains(&op.name.as_str()))
+            .find(|op| SKIPPED_OPERATIONS.contains(&op.name.as_str()))
```
kmahar, line 111: this isn't exactly related to your changes, but this `SKIPPED_OPERATIONS` list is a little weird. I see `watch` on there, but `watch` isn't actually a valid unified operation (see [here](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#watch)) and has been superseded by `createChangeStream`. I also see `count` on there, which I would assume refers to the deprecated `count` API, but I don't think any unified tests use that either... 

edit: I now see that this seems like it was copy-pasted from the v2 runner. so I think this _might_ mean we are unfortunately skipping a bunch of change stream retryable reads tests that still use that runner... any idea if we run those @abr-egn? basically all the change stream related ones here: https://github.com/mongodb/specifications/tree/master/source/retryable-reads/tests/legacy

kmahar (approved):
changes LGTM, left a comment about an unrelated possible issue I noticed during review

src/test/spec/unified_runner/mod.rs
abr-egn: Looks like they're skipped (and silently, too).  I'll put together a PR to fix that.

abr-egn (approved):

isabelatkinson (approved):

PR #645, opened 2022-04-27
RUST-488 Allow usage of both async and sync apis

abr-egn:
RUST-488

This removes the restriction on having both async and sync apis enabled.  It is still required to use the same runtime for both, i.e. either `sync,async-std-runtime` or `tokio-sync,tokio-runtime`.


manual/src/installation_features.md
```diff
@@ -19,22 +19,23 @@ features = ["async-std-runtime"]
 ```
 
 ## Enabling the sync API
-The driver also provides a blocking sync API. To enable this, add the `"sync"` feature to your `Cargo.toml`:
+The driver also provides a blocking sync API. To enable this, add the `"sync"` or `"tokio-sync"` feature to your `Cargo.toml`:
 ```toml
 [dependencies.mongodb]
```
patrickfreed, line 24: this guidance can be simplified to omit `default-features = false` now if we use `tokio-sync`. We may still want to include more details about how to use async-std with sync too though. Also, make sure to update the README to reflect these changes too.

This just gave me the thought: should we add an `async-std-sync` feature and just remove mentions of `sync` from our docs? The old feature will continue to work, but this new one might be easier to understand.

In 3.0, I think we'd want to go back to just having a single `sync` feature flag and delegate the runtime choice to the `async-std-runtime` or `tokio-runtime` feature flags though.

manual/src/installation_features.md
```diff
@@ -19,22 +19,23 @@ features = ["async-std-runtime"]
...
 version = "2.1.0"
```
patrickfreed, line 25: sort of unrelated: this should be `2.2.0`. We'll probably want to update the release checklist to include the manual as another place needing updating when doing releases.

manual/src/installation_features.md
abr-egn: Updated the documentation here, in the README, and in `lib.rs`.

I agree on how things should work in 3.0.  For now, however, given that there'd still need to be the `default-features` distinction between `tokio-sync` and `async-std-sync`, I don't think adding the latter would be much of a gain in clarity or worth the churn.

kmahar (approved):

patrickfreed (approved):
LGTM!

PR #647, opened 2022-04-29
RUST-1035 Fix flaky server selection test

patrickfreed:
RUST-1035

This PR makes some updates to the load balancing server selection prose test to make it hopefully less flaky.

Relevant test here: https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection-tests.rst#prose-test

### DIFF
--- a/src/sdam/description/topology/server_selection/test/in_window.rs
+++ b/src/sdam/description/topology/server_selection/test/in_window.rs
@@ -256,7 +256,7 @@ async fn load_balancing_test() {
                 client
                     .database("load_balancing_test")
                     .collection::<Document>("load_balancing_test")
-                    .find(doc! { "$where": "sleep(500) && true" }, options)
+                    .find(doc! { "$where": "sleep(1000) && true" }, options)
                     .await
                     .unwrap();
             });
@@ -271,6 +271,7 @@ async fn load_balancing_test() {
             .await
             .expect("timed out waiting for both pools to be saturated");
         conns += 1;
+        println!("conns: {}", conns);
     }
     drop(subscriber);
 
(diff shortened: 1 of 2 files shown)


abr-egn (approved):

src/sdam/description/topology/server_selection/test/in_window.rs
```diff
@@ -211,39 +216,54 @@ async fn load_balancing_test() {
...
+            let client = client.clone();
+            let selector = selector.clone();
+            runtime::execute(async move {
+                let options = FindOptions::builder()
+                    .selection_criteria(SelectionCriteria::Predicate(selector))
+                    .build();
+                client
+                    .database("load_balancing_test")
+                    .collection::<Document>("load_balancing_test")
+                    .find(doc! { "$where": "sleep(500) && true" }, options)
```
kmahar, line 239: why do we use the sleep(500)?

kmahar (approved):
couple questions for my own understanding, but LGTM

isabelatkinson (approved):

src/sdam/description/topology/server_selection/test/in_window.rs
```diff
@@ -170,19 +172,22 @@ async fn load_balancing_test() {
...
                 for _ in 0..iterations {
-                    let _ = collection.find_one(None, None).await;
+                    collection.find_one(None, None).await?;
                 }
+                Ok(())
             }))
         }
 
-        futures::future::join_all(handles).await;
+        for handle in handles {
```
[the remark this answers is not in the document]
patrickfreed: `join_all` looks like it will await all the futures concurrently, wheras this does them in order. Since the futures being awaited here are just task handles, there isn't really a difference as they're all running in the background anyways. I think I changed this to iterate over the handles directly so I could unwrap all of them to ensure no errors occurred. This could've been done with `join_all` too, but I guess I figured I'd just do it directly instead of in two steps.

src/sdam/description/topology/server_selection/test/in_window.rs
patrickfreed: This was done to ensure the connections weren't returned to the pools immediately and reused. I think this could've been achieved with failpoints and blockConnection too, but this was a bit easier.

PR #648, opened 2022-04-29
RUST-332 Return an error for replica set name mismatch

abr-egn:
RUST-332 / RUST-358

This causes servers that don't match the repl set name given in the connection string to be marked as unknown, so any operation will return an error.

### DIFF
--- a/src/sdam/test.rs
+++ b/src/sdam/test.rs
@@ -478,7 +478,7 @@ async fn repl_set_name_mismatch() -> crate::error::Result<()> {
     options.hosts.drain(1..);
     options.direct_connection = Some(true);
     options.repl_set_name = Some("invalid".to_string());
-    let client = TestClient::with_options(Some(options)).await;
+    let client = Client::with_options(options)?;
     let result = client.list_database_names(None, None).await;
     assert!(match result {
         Err(Error { ref kind, .. }) => match **kind {
--- a/src/sdam/test.rs
+++ b/src/sdam/test.rs
@@ -482,10 +482,7 @@ async fn repl_set_name_mismatch() -> crate::error::Result<()> {
     let result = client.list_database_names(None, None).await;
     assert!(
         match result {
-            Err(Error { ref kind, .. }) => match **kind {
-                ErrorKind::ServerSelection { .. } => true,
-                _ => false,
-            },
+            Err(Error { ref kind, .. }) => matches!(**kind, ErrorKind::ServerSelection { .. }),
             _ => false,
         },
         "Unexpected result {:?}",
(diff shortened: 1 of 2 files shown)


src/sdam/state/mod.rs
```diff
@@ -204,12 +204,31 @@ impl Topology {
...
-
-        server_selection::attempt_to_select_server(
+        let selected = server_selection::attempt_to_select_server(
             criteria,
             &topology_state.description,
             &topology_state.servers,
-        )
+        )?;
+        let opts = &self.common.options;
+        if let (Some(true), Some(set_name), Some(sel_server)) =
```
abr-egn, line 213: This would be _way_ easier to write and read if https://rust-lang.github.io/rfcs/2497-if-let-chains.html was implemented :(

src/sdam/test.rs
```diff
@@ -451,3 +451,35 @@ async fn hello_ok_true() {
...
+    if !client.is_replica_set() {
+        log_uncaptured("skipping repl_set_name_mismatch due to non-replica set topology");
+        return Ok(());
+    }
+
+    let mut options = CLIENT_OPTIONS.clone();
+    options.hosts.drain(1..);
+    options.direct_connection = Some(true);
+    options.repl_set_name = Some("invalid".to_string());
+    let client = Client::with_options(options)?;
```
abr-egn, line 470: This needs to be a `Client` rather than a `TestClient` because the latter executes a few operations with `unwrap()`s, which fail and crash before the test can check anything.

kmahar:
I'm sorry we didn't discover this sooner, but I just went to look at the relevant section in the SDAM spec and noticed the language referenced in the ticket has now changed to:
> A client MAY allow the user to supply a setName with an initial TopologyType of Single. In this case, if the ServerDescription's setName is null or wrong, the ServerDescription MUST be replaced with a default ServerDescription of type Unknown.

It looks like shortly after RUST-332 was filed, DRIVERS-980 got filed which resulted in the language being changed. The relevant Rust ticket spun out from that is RUST-358. So I think we'll want to adopt the newer behavior here instead.

As an aside, this is probably going to cause merge conflicts with #628 so we should keep that in mind for whichever of these PRs gets merged in later

abr-egn:
> It looks like shortly after [RUST-332](https://jira.mongodb.org/browse/RUST-332) was filed, DRIVERS-980 got filed which resulted in the language being changed. The relevant Rust ticket spun out from that is [RUST-358](https://jira.mongodb.org/browse/RUST-358). So I think we'll want to adopt the newer behavior here instead.

Whoops, missed that.  Shouldn't be too hard to update to follow that.

kmahar (approved):

patrickfreed (approved):
LGTM!

isabelatkinson (approved):

PR #653, opened 2022-05-06
RUST-1280 Run tests sequentially in CI

isabelatkinson:
Updates the tests to run sequentially rather than in parallel (thereby eliminating the need for the test lock). The first CI run made for this patch took 273 hours which is in line with the time spent for patches on the waterfall, usually ranging in the mid-to-high 200s. These changes also did not cause any performance regressions when I ran the test suite locally against a standalone (avg of 4:08 for both sequential and parallel runs).


isabelatkinson:
Closing as this work is infeasible; see RUST-1280 for details.

PR #654, opened 2022-05-10
WIP: tracing/logging implementation

kmahar:
Putting this up to get some initial feedback.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -12,6 +12,8 @@ use derivative::Derivative;
 
 #[cfg(test)]
 use crate::options::ServerAddress;
+#[cfg(feature = "tracing-unstable")]
+use crate::trace::{debug_tracing_or_log_enabled, CommandTracingEventEmitter};
 use crate::{
     bson::Document,
     change_stream::{
@@ -130,7 +132,7 @@ impl Client {
 
     pub(crate) fn emit_command_event(&self, generate_event: impl FnOnce() -> CommandEvent) {
         #[cfg(feature = "tracing-unstable")]
-        let should_emit =  self.inner.options.command_event_handler.is_some() || tracing::enabled!(tracing::Level::DEBUG);
+        let should_emit =  self.inner.options.command_event_handler.is_some() || debug_tracing_or_log_enabled!();
         #[cfg(not(feature = "tracing-unstable"))]
         let should_emit =  self.inner.options.command_event_handler.is_some();
 
@@ -143,8 +145,8 @@ impl Client {
             }
             
             #[cfg(feature = "tracing-unstable")]
-            if tracing::enabled!(tracing::Level::DEBUG) {
-                let tracing_emitter = crate::trace::CommandTracingEventEmitter::new(
+            if debug_tracing_or_log_enabled!() {
+                let tracing_emitter = CommandTracingEventEmitter::new(
                     self.inner.options.tracing_max_document_length_bytes
                 );
                 handle_command_event(&tracing_emitter, event.clone());
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -130,28 +130,19 @@ impl Client {
         Ok(Self { inner })
     }
 
-    pub(crate) fn emit_command_event(&self, generate_event: impl FnOnce() -> CommandEvent) {
-        #[cfg(feature = "tracing-unstable")]
-        let should_emit =
-            self.inner.options.command_event_handler.is_some() || debug_tracing_or_log_enabled!();
-        #[cfg(not(feature = "tracing-unstable"))]
-        let should_emit = self.inner.options.command_event_handler.is_some();
-
-        // TODO: this is probably not ideal because we always clone the event, even if we aren't
-        // actually going to use it twice... any nice way to solve that?
-        if should_emit {
+    pub(crate) fn emit_command_event(&self, generate_event: impl Fn() -> CommandEvent) {
+        if let Some(ref handler) = self.inner.options.command_event_handler {
             let event = generate_event();
-            if let Some(ref handler) = self.inner.options.command_event_handler {
-                handle_command_event(handler.as_ref(), event.clone());
-            }
+            handle_command_event(handler.as_ref(), event);
+        }
 
-            #[cfg(feature = "tracing-unstable")]
-            if debug_tracing_or_log_enabled!() {
-                let tracing_emitter = CommandTracingEventEmitter::new(
-                    self.inner.options.tracing_max_document_length_bytes,
-                );
-                handle_command_event(&tracing_emitter, event);
-            }
+        #[cfg(feature = "tracing-unstable")]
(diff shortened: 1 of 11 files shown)


src/client/mod.rs
```diff
@@ -134,9 +136,19 @@ impl Client {
         Ok(Self { inner })
     }
 
-    pub(crate) fn emit_command_event(&self, emit: impl FnOnce(&Arc<dyn CommandEventHandler>)) {
+    pub(crate) fn emit_command_event(&self, generate_event: impl Fn() -> CommandEvent) {
```
kmahar, line 139: I spent a while trying to figure out a nice way to keep this as a `FnOnce` (which would save the `clone()`s I had to add in the operation executor code that emits events), but I had a hard time figuring out how to only clone the resulting event if it was actually being passed to `handle_command_event` twice (the code looked like [this](https://github.com/kmahar/mongo-rust-driver/blob/4d5b402ab8db969a2953dc0cb8932b880d8d5db8/src/client/mod.rs#L139-L162)), short of branching on whether one or both of monitoring and tracing were enabled and having near-duplicate code under each branch to create/unwrap the appropriate handler(s) and pass the event/event clone to them. 

in the end I got to thinking that the `clone()`s I had to add were likely pretty cheap, and that it seems rare someone would actually use both monitoring and tracing given they provide really similar functionality, so that paying the penalty of re-generating the event in those cases is probably fine. 

that said, if there is a nicer way to do what I was trying to do I'm not opposed to reworking this. thoughts?

src/trace.rs
```diff
@@ -0,0 +1,136 @@
...
+
+impl TracingRepresentation for bson::Document {
+    fn tracing_representation(self) -> String {
+        Bson::Document(self).into_canonical_extjson().to_string()
+    }
+}
+
+impl TracingRepresentation for crate::error::Error {
+    fn tracing_representation(self) -> String {
+        self.to_string()
```
kmahar, line 109: probably want to do something nicer here (or update the `Display` implementation to) but I haven't looked at it much yet

src/trace.rs
```diff
@@ -0,0 +1,136 @@
...
+        // is_char_boundary returns true when the provided value == the length of the string, so
+        // if we reach the end of the string this loop will terminate.
+        while !s.is_char_boundary(truncate_index) {
+            truncate_index += 1;
+        }
+        s.truncate(truncate_index);
+    }
+}
+
+macro_rules! debug_tracing_or_log_enabled {
```
kmahar, line 131: this is a placeholder for fixing the issue in tracing linked above, where the log level is not consulted by the macro.

src/trace.rs
patrickfreed: Yeah, for server errors we could just want the extjson form. For the more basic ones we can just do the message. For structured rust ones, I'm not totally sure what to do, though the `Display` implementation will be at least decent for those too.

src/trace.rs
```diff
@@ -0,0 +1,136 @@
+use crate::event::command::{
+    CommandEventHandler,
+    CommandFailedEvent,
+    CommandStartedEvent,
+    CommandSucceededEvent,
+};
+use bson::Bson;
+
+pub(crate) struct CommandTracingEventEmitter {
```
patrickfreed, line 9: Is the plan to pass the events through this handler after doing tracing if the user also configured an event handler? If so, we'd need something like an `Option<Arc<dyn CommandEventHandler>>` field here.

patrickfreed (commented):
Approach LGTM to me so far! Just have some minor comments / questions

src/trace.rs
```diff
@@ -0,0 +1,136 @@
...
+    max_document_length_bytes: usize,
+}
+
+impl CommandTracingEventEmitter {
+    pub(crate) fn new(max_document_length_bytes: Option<usize>) -> CommandTracingEventEmitter {
+        CommandTracingEventEmitter {
+            max_document_length_bytes: max_document_length_bytes.unwrap_or(1000),
```
abr-egn, line 16: Nit - I think it'd be a little better to have the default be a named constant.

src/trace.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    fn serialize_command_or_reply(&self, doc: bson::Document) -> String {
+        let mut ext_json = doc.tracing_representation();
+        truncate_on_char_boundary(&mut ext_json, self.max_document_length_bytes);
+        ext_json
+    }
+}
+
+impl CommandEventHandler for CommandTracingEventEmitter {
+    fn handle_command_started_event(&self, event: CommandStartedEvent) {
```
abr-egn, line 28: It seems like it might be worth recording command started -> command succeeded/failed as a span?

abr-egn (commented):
Agreed, approach LGTM as well!

src/trace.rs
patrickfreed: Ah I missed that bit in `emit_command_event`, LGTM. And yeah, creating a new one on the fly seems fine since it doesn't have any state.

src/test/util/trace.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/// A tracing subscriber for use in tests.
+#[derive(Clone)]
+pub struct TracingSubscriber {
+    /// The maximum verbosity level which this subscriber will
+    /// collect events for.
+    max_verbosity_level: Level,
+    /// Stores events the subscriber has collected.
+    events: Arc<RwLock<VecDeque<TracingEvent>>>,
```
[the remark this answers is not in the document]
patrickfreed: LGTM!

src/trace.rs
kmahar: I agree defining some spans of our own is a good idea, though for the purpose of this PR and the initial spec implementation I think I'd like to hold off on introducing the additional complexity of defining and testing spans until everything else is in place. it seems like something to revisit shortly after this work is finished though, and perhaps even before we initially release it. I've filed RUST-1390 about this.

src/trace.rs
kmahar: good point. I've added a constant for this on my current working branch.

src/trace.rs
```diff
@@ -0,0 +1,136 @@
...
+
+impl TracingRepresentation for crate::error::Error {
+    fn tracing_representation(self) -> String {
+        self.to_string()
+    }
+}
+
+/// Truncates the given string at the closest UTF-8 character boundary >= the provided length.
+/// If the new length is >= the current length, does nothing.
+pub(crate) fn truncate_on_char_boundary(s: &mut String, new_len: usize) {
```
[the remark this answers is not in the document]
kmahar: for this particular version of truncation, I haven't really been able to find anything besides the nightly-only method we can't use yet. that said, the particular implementation of truncation is currently up for discussion on the spec design doc so we may not need to end up doing this exact thing anyway. for now I've added a TODO on my working branch about this.

kmahar:
per my slack message, closing out this draft, actual support will come once the spec is closer to finalized in a series of PRs.

PR #656, opened 2022-05-10
RUST-1283 Build docs in read-only directory and install cargo in project directory on Evergreen

kmahar:
Attempts to better simulate the docs.rs build environment in our check-rustdoc test by making the cargo registry directory read-only.
Since the `chmod` makes the cargo registry unusable for future tasks that need to e.g. download different dependencies than the docs one, this also required changing how we install rustup cargo so that it happens in the project directory and is cleaned up at the end of each task, which is the "right" thing to do on Evergreen anyway rather than writing to the shared home directory.

Here's a[ patch](https://spruce.mongodb.com/task/mongo_rust_driver_lint_check_rustdoc_patch_279322cae0fe4f2302ed14fd541de640eebd75c4_627af92132f4173983980fd9_22_05_10_23_45_53/logs?execution=0) showing that check-rustdoc now fails if using the old version of `rustc_version_runtime` that we ran into issues with. 

I was concerned this change might lead to our tasks taking a lot longer since we are starting from scratch with rustup and cargo each time now, but the difference between my patch on this PR and the latest commit on main in terms of total runtime across all tasks is 293 hours (mine) vs 283 hours (main) so it doesn't seem to make a huge difference.

### DIFF
--- a/.evergreen/check-all.sh
+++ b/.evergreen/check-all.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -o errexit
 
--- a/.evergreen/check-rustdoc.sh
+++ b/.evergreen/check-rustdoc.sh
@@ -1,8 +1,23 @@
-#!/bin/sh
+#!/bin/bash
 
 set -o errexit
 
-. ~/.cargo/env
+source ./.evergreen/configure-rust.sh
+
+# docs.rs builds the driver on a read-only file system. to create a more realistic environment, we first
+# build the driver to ensure we have all the deps already in src, and then limit the permissions on that directory
+# and rebuild the docs.
+# this is to help us avoid introducing problems like those described here 
+# https://docs.rs/about/builds#read-only-directories where we or a dependency modify source code during the
+# build process.
+cargo +nightly build
+cargo +nightly build --no-default-features --features async-std-runtime
+cargo +nightly build --no-default-features --features sync
+cargo +nightly build --features tokio-sync
+cargo clean
+
+chmod -R 555 ${CARGO_HOME}/registry/src
+
 cargo +nightly rustdoc -- -D warnings --cfg docsrs
 cargo +nightly rustdoc --no-default-features --features async-std-runtime -- -D warnings --cfg docsrs
 cargo +nightly rustdoc --no-default-features --features sync -- -D warnings --cfg docsrs
--- a/.evergreen/check-rustdoc.sh
+++ b/.evergreen/check-rustdoc.sh
@@ -10,15 +10,18 @@ source ./.evergreen/configure-rust.sh
 # this is to help us avoid introducing problems like those described here 
 # https://docs.rs/about/builds#read-only-directories where we or a dependency modify source code during the
 # build process.
-cargo +nightly build
-cargo +nightly build --no-default-features --features async-std-runtime
-cargo +nightly build --no-default-features --features sync
-cargo +nightly build --features tokio-sync
+
+source ./.evergreen/feature-combinations.sh
+
+# build with all available features to ensure all optional dependencies are brought in too.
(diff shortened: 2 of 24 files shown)


.evergreen/check-all.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
kmahar, line 1: the configure-rust script uses bash-only syntax; I took the opportunity to standardize all of our scripts to use bash (even if they don't call configure-rust.sh) to try to avoid any problems around this in the future

.evergreen/check-rustdoc.sh
```diff
@@ -1,8 +1,23 @@
...
-. ~/.cargo/env
+source ./.evergreen/configure-rust.sh
+
+# docs.rs builds the driver on a read-only file system. to create a more realistic environment, we first
+# build the driver to ensure we have all the deps already in src, and then limit the permissions on that directory
+# and rebuild the docs.
+# this is to help us avoid introducing problems like those described here 
+# https://docs.rs/about/builds#read-only-directories where we or a dependency modify source code during the
+# build process.
+cargo +nightly build
```
kmahar, line 13: these might not all be strictly necessary but I just copied the exact invocations that we run `cargo rustdoc` with below for completeness

.evergreen/config.yml
```diff
@@ -431,6 +432,7 @@ functions:
           export MONGO_ATLAS_TLS11_URI_SRV='${MONGO_ATLAS_TLS11_URI_SRV}'
           export MONGO_ATLAS_TLS12_URI='${MONGO_ATLAS_TLS12_URI}'
           export MONGO_ATLAS_TLS12_URI_SRV='${MONGO_ATLAS_TLS12_URI_SRV}'
+          export PROJECT_DIRECTORY='${PROJECT_DIRECTORY}'
```
kmahar, line 435: this is needed because configure-rust.sh uses it. for all of the other tasks we run $PREPARE_SHELL which already does this for us.

.evergreen/check-rustdoc.sh
abr-egn: Might be worth putting these into their own little script?

abr-egn (approved):
LGTM!  One minor suggestion.

patrickfreed (approved):

isabelatkinson (approved):

kmahar:
@abr-egn @patrickfreed I gave up on figuring out what docs.rs does, but I did add a new script `feature-combinations.sh` that generates a list of feature combinations we can iterate through, and updated the relevant places (check-rustdoc.sh, check-clippy.sh) to do so. we still have to manually update this when we add new features but at least it is all in one place now.
here's a patch I ran with xtrace on for the relevant tasks where you can see the correct flag combinations are indeed tested: https://spruce.mongodb.com/version/62841c215623430430b68009/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC 

lmk what you think

.evergreen/check-rustdoc.sh
kmahar: I didn't end up doing this since we don't use it anywhere else and the number of scripts we have already feels a little out of hand, but per my comment below I did abstract out the "writing out feature combinations to test" part

abr-egn:
> @abr-egn @patrickfreed I gave up on figuring out what docs.rs does, but I did add a new script `feature-combinations.sh` that generates a list of feature combinations we can iterate through, and updated the relevant places (check-rustdoc.sh, check-clippy.sh) to do so. we still have to manually update this when we add new features but at least it is all in one place now. here's a patch I ran with xtrace on for the relevant tasks where you can see the correct flag combinations are indeed tested: https://spruce.mongodb.com/version/62841c215623430430b68009/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
> 
> lmk what you think

I like this approach!  LGTM.

PR #657, opened 2022-05-17
RUST-645 / RUST-1245 Add Atlas planned maintenance testing workload executor

isabelatkinson:
Workload executor specification: https://mongodb-labs.github.io/drivers-atlas-testing/spec-workload-executor.html

There are still a few kinks in the implementation that I'm working through in testing, but the bulk of the code should be ready for review.

### DIFF
--- a/src/event/sdam/mod.rs
+++ b/src/event/sdam/mod.rs
@@ -4,10 +4,11 @@ mod topology_description;
 
 use std::time::Duration;
 
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 
 use crate::{
     bson::{oid::ObjectId, Document},
+    bson_util::serialize_error_as_string,
     error::Error,
     options::ServerAddress,
 };
@@ -20,7 +21,8 @@ pub use topology_description::TopologyDescription;
 pub type ServerDescription = crate::sdam::public::ServerInfo<'static>;
 
 /// Published when a server description changes.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct ServerDescriptionChangedEvent {
     /// The address of the server.
@@ -48,7 +50,8 @@ impl ServerDescriptionChangedEvent {
 }
 
 /// Published when a server is initialized.
-#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct ServerOpeningEvent {
     /// The address of the server.
@@ -60,7 +63,8 @@ pub struct ServerOpeningEvent {
 }
 
 /// Published when a server is closed.
-#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct ServerClosedEvent {
     /// The address of the server.
@@ -72,7 +76,8 @@ pub struct ServerClosedEvent {
 }
 
 /// Published when a topology description changes.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct TopologyDescriptionChangedEvent {
     /// The ID of the topology.
@@ -86,7 +91,8 @@ pub struct TopologyDescriptionChangedEvent {
 }
 
 /// Published when a topology is initialized.
-#[derive(Clone, Debug, Deserialize)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct TopologyOpeningEvent {
     /// The unique ID of the topology.
@@ -96,7 +102,8 @@ pub struct TopologyOpeningEvent {
 
 /// Published when a topology is closed. Note that this event will not be published until the client
 /// associated with the topology is dropped.
-#[derive(Clone, Debug, Deserialize)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct TopologyClosedEvent {
     /// The unique ID of the topology.
@@ -105,7 +112,8 @@ pub struct TopologyClosedEvent {
 }
 
 /// Published when a server monitor's `hello` or legacy hello command is started.
-#[derive(Clone, Debug, Deserialize)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct ServerHeartbeatStartedEvent {
     /// The address of the server.
@@ -114,7 +122,8 @@ pub struct ServerHeartbeatStartedEvent {
 }
 
 /// Published when a server monitor's `hello` or legacy hello command succeeds.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub struct ServerHeartbeatSucceededEvent {
     /// The execution time of the event.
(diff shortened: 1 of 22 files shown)


src/test/atlas_planned_maintenance_testing/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    let test_file: TestFile = from_value(workload).unwrap();
+    let description = test_file.description.clone();
+    if AssertUnwindSafe(test_runner.run_test(test_file, |_| true))
+        .catch_unwind()
+        .await
+        .is_err()
+    {
+        execution_errors.push(
+            doc! {
+                "error": format!("Unexpected error occurred while running {}", description),
```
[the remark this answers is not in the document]
kmahar: is the problem that `run_test` uses `panic` to report failures? the Swift runner has a couple custom error types we throw for unexpected failures, I wonder how much work it would be to switch to something like that so `run_test` could return a `Result`? otherwise I imagine this is going to make it really difficult to debug any failures we see in Evergreen

src/test/spec/unified_runner/operation.rs
```diff
@@ -1919,6 +1938,195 @@ impl TestOperation for RenameCollection {
...
+                    };
+                    match result {
+                        Ok(entity) => match operation.expectation {
+                            Expectation::Result {
+                                ref expected_value,
+                                ref save_as_entity,
+                            } => {
+                                if let Some(expected_value) = expected_value {
+                                    if let Some(Entity::Bson(ref actual_value)) = entity {
+                                        if actual_value == expected_value {
```
kmahar, line 1997: should we be using the test runner "matching" logic here or is this always an exact match?

kmahar (commented):
looking good! few comments and questions

src/test/spec/unified_runner/operation.rs
```diff
@@ -63,7 +75,13 @@ pub trait TestOperation: Debug {
         _id: &'a str,
         _test_runner: &'a mut TestRunner,
     ) -> BoxFuture<'a, Result<Option<Entity>>> {
-        todo!()
+        async move {
+            Err(ErrorKind::InvalidArgument {
+                message: "execute_entity_operation called on unsupported operation".into(),
```
abr-egn, line 80: Is it useful for this to be a failure rather than an immediate `panic!`?

src/test/spec/unified_runner/operation.rs
```diff
@@ -1919,6 +1938,195 @@ impl TestOperation for RenameCollection {
...
+                                                &operation.name,
+                                                actual_value,
+                                                expected_value,
+                                                test_runner,
+                                            );
+                                            // If an unexpected failure is encountered, the
+                                            // sub-operation loop should be terminated.
+                                            break;
+                                        }
+                                    }
```
abr-egn, line 2010: Should there be handling for expected value but the entity is `None` / not `Bson`?

abr-egn (commented):
Overall LG!  Just a few small comments beyond Kaitlin's.

src/test/atlas_planned_maintenance_testing/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+    {
+        errors.append(&mut operation_errors);
+    }
+    events.errors = errors;
+    if let Some(Entity::Bson(Bson::Array(failures))) = test_runner.entities.remove("failures") {
+        events.failures = failures;
+    }
+    let mut path =
+        env::var("WORKLOAD_EXECUTOR_WORKING_DIRECTORY").expect("No working directory specified");
+    path.push_str("/events.json");
```
patrickfreed, line 75: It might be better to first create a `PathBuf` from the environment variable here and then use `path.push("events.json")` to prevent any issues with Windows.

src/test/spec/unified_runner/operation.rs
```diff
@@ -1919,6 +1938,195 @@ impl TestOperation for RenameCollection {
...
+                for operation in &self.operations {
+                    let result = match operation.object {
+                        OperationObject::TestRunner => {
+                            panic!("Operations within a loop must be entity operations")
+                        }
+                        OperationObject::Entity(ref id) => {
+                            operation.execute_entity_operation(id, test_runner).await
+                        }
+                    };
+                    match result {
```
patrickfreed, line 1989: since we have a lot of nested matches and if lets, this might be more readable as one larger, flat match statement,
e.g.
```rust
match (result, operation.expectation) {
    // Have an entity result, expected one, need to check if equal
    (Ok(Entity::Bson(entity)), Expectation::Result { Some(expected_value), save_as_entity }) => { ... },

    // Got an ok response and expected one, but don't care about what it is, always success
    (Ok(_), Expectation::Result { None, _ }) => { ... }

    // Got an ok result, but expected an error, always a failure
    (Ok(entity), Expectation::Error(expected_error) => { ... }

    // ignore, always success
    (Ok(_), Expectation::Ignore) => { ... }

    // Got an error, but expected a result
    (Err(e), Expectation::Result { ... } | Expectation::Ignore) => { ... }

    // Got an error, expected an error
    (Err(e), Expectation::Error(expected_error)) => { ... }
}
```
To me this is a bit easier to reason about and also has the benefit of having the compiler verify that we've handled all cases properly. Also could combine the first two cases and keep a nested `if let`. Anyways, this is just my preference, feel free to leave it as-is if you think the current version is more readable.

src/test/atlas_planned_maintenance_testing/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use crate::{
+    bson::{doc, Bson},
+    test::spec::unified_runner::{entity::Entity, test_file::TestFile, test_runner::TestRunner},
+};
+
+use json_models::{Events, Results};
+
+#[test]
+fn get_exe_name() {
```
isabelatkinson, line 23: this is a slightly more convenient way to get the name of the test executable (as opposed to parsing through `cargo` output). all of the tests in the crate are run via one executable that can be filtered by passing test names as arguments. `std::env::current_exe()` returns the name of the executable that's currently running, so when it's called within a test it'll return the name of that test executable. this test method then writes that name to a file that the workload executor can invoke via `$(cat exe_name.txt workload_executor)`. (I used `into_os_string` on the `PathBuf` returned rather than the `Debug` implementation because the latter includes quotations around the path that would need to be stripped out.)

src/test/atlas_planned_maintenance_testing/mod.rs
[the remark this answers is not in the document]
kmahar: thanks for the context! 

in Swift, the top-level method `UnifiedTestRunner.runTests` is marked `throws`, so we can propagate errors all the way up to that level pretty easily, which leads the calling test case to throw an error and fail. 
for operations that expect errors we use a `do`/`catch` around the appropriate code, and for some unexpected errors we catch them and re-throw them with extra context added. some unexpected errors just get directly propagated up though, like anything during test setup I believe.

I suppose this all ends up being more straightforward and concise in Swift than Rust since the Swift compiler does not care about what type of error a method throws, so we don't have to worry about converting errors all to some consistent type. also, we can just `try` things that throw in Swift, but since we can't use `?` on async return values working with the results is a lot more verbose here. 🤔

src/event/sdam/mod.rs
```diff
@@ -156,6 +166,13 @@ pub(crate) enum SdamEvent {
     ServerHeartbeatFailed(ServerHeartbeatFailedEvent),
 }
 
+fn serialize_error_as_string<S: Serializer>(
```
abr-egn, line 169: Maybe this should go in some common place so it can be shared?

src/test/atlas_planned_maintenance_testing/mod.rs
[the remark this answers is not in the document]
abr-egn: I think it's reasonable to take this approach here; it should work for what's needed and it keeps this PR from ballooning.

I do think there should be a bug to track converting the test code to return errors rather than using `panic!`.  I'm not sure how unwind-safe the test runner actually is; I don't think there are going to be memory-safety problems, but it seems quite possible the runner could end up with broken assumptions that cause flaky tests.  It seems roughly equivalent to the cancellation issue.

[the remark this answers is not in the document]
kmahar: I agree with Abraham, deferring test runner refactoring to use errors to a follow up ticket SGTM

kmahar (commented):
took another pass through - everything here LGTM, but I'll hold off on approving in case there end up being any more changes

isabelatkinson (commented):
Moving this out of draft/WIP status as the planned maintenance tests are finally all passing. Going to hold off on fixing the merge conflicts until this is ready to merge so I can rebase down to one commit.

abr-egn (approved):

patrickfreed (approved):
LGTM! Awesome job with this.

Just looks like there are some merge conflicts and possibly an MSRV issue that need to be fixed up.

PR #672, opened 2022-06-03
Fix Evergreen task for MSRV compilation with async-std

kmahar:
Pins the version of async-global-executor we use to test async-std with Rust 1.53.


.evergreen/compile-only.sh
```diff
@@ -6,9 +6,25 @@ source ./.evergreen/configure-rust.sh
 rustup update $RUST_VERSION
```
kmahar, line 7: since the runtime-specific compile scripts were only used from this file, I took the opportunity to cut out two scripts and just move their logic into here.

abr-egn (approved):

isabelatkinson (approved):

PR #673, opened 2022-06-03
RUST-1193 follow-up: minor adjustment to ConnectionString -> ClientOptions API

kmahar:
Splits the sync and async APIs each into two methods, one which accepts a resolver config and one which does not.

### DIFF
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -1100,18 +1100,29 @@ impl ClientOptions {
         uri: impl AsRef<str>,
         resolver_config: Option<ResolverConfig>,
     ) -> Result<Self> {
-        Self::parse_connection_string(ConnectionString::parse(uri)?, resolver_config).await
+        Self::parse_connection_string_internal(ConnectionString::parse(uri)?, resolver_config).await
     }
 
     /// Creates a `ClientOptions` from the given `ConnectionString`.
     ///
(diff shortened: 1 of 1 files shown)


kmahar:
(Evergreen still pending, but it seems unlikely the remaining tasks turn up any issues with this and figured we should get the ball rolling on review)

abr-egn (approved):

patrickfreed (approved):

src/client/options/mod.rs
```diff
@@ -1172,16 +1183,25 @@ impl ClientOptions {
         Ok(options)
     }
 
+    /// Creates a `ClientOptions` from the given `ConnectionString`.
+    #[cfg(any(feature = "sync", feature = "tokio-sync"))]
+    pub fn parse_connection_string_sync(conn_str: ConnectionString) -> Result<Self> {
```
isabelatkinson, line 1188: i might've missed a discussion on this on the original PR but does this method need "sync" in the name (and ditto below)? the [sync version](https://docs.rs/mongodb/latest/mongodb/options/struct.ClientOptions.html#method.parse-1) of `ClientOptions::parse` is named the same as the async one because the feature flags for each are mutually exclusive

isabelatkinson: gotcha, thanks for the explanation!

isabelatkinson (approved):

PR #679, opened 2022-06-09
RUST-1328 Only emit SDAM changed events when servers change

patrickfreed:
This fixes a bug where ServerDescriptionChanged and
TopologyDescriptionChanged events were emitted despite no changes to
the topology. This would happen in cases where a ServerDescription was
Unknown due to an error.

This also adds a test for the fix for RUST-1274.

### DIFF
--- a/src/sdam/description/server.rs
+++ b/src/sdam/description/server.rs
@@ -122,6 +122,7 @@ impl PartialEq for ServerDescription {
 
                 self_response == other_response
             }
+            (Err(self_err), Err(other_err)) => self_err == other_err,
             _ => false,
         }
     }
--- a/src/sdam/description/topology/test/sdam.rs
+++ b/src/sdam/description/topology/test/sdam.rs
@@ -646,7 +646,7 @@ async fn heartbeat_events() {
         .await
         .expect("should see server heartbeat succeeded event");
 
-    if !client.supports_fail_command_appname() {
+    if !client.supports_fail_command_appname_initial_handshake() {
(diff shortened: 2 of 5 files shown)


src/sdam/topology.rs
```diff
@@ -477,23 +477,25 @@ impl TopologyWorker {
...
-                };
-                // Mark server as unknown.
-                sd = ServerDescription::new(
-                    sd.address,
-                    Some(Err(format!(
-                        "Connection string replicaSet name {:?} does not match actual name {}",
-                        expected_name, got_display,
-                    ))),
-                );
+            if sd.is_available() {
```
patrickfreed, line 480: I updated this block to only run when the server being updated is available, since it was running any time a server description with an error was being used to update (error => no set name).

src/sdam/topology.rs
```diff
@@ -522,8 +524,8 @@ impl TopologyWorker {
                     s.pool.mark_as_ready().await;
                 }
             }
-            self.publisher.publish_new_state(latest_state)
```
patrickfreed, line 525: we always publish a new state, regardless of whether there were material changes made. This is required so that servers blocked in server selection will continue to wake up every minHeartbeatFrequency and requesting checks instead of just blocking until they time out. A similar change was made in the 2.2.x branch, though that was done at the monitor level rather than topology.

kmahar (approved):

abr-egn (approved):

PR #680, opened 2022-06-09
RUST-1215/RUST-1145/RUST-1070 Add comment option to the estimatedDocumentCountOptions struct, let option to CRUD

sanav33:


src/coll/options.rs
```diff
@@ -574,6 +574,13 @@ pub struct EstimatedDocumentCountOptions {
     /// The level of the read concern.
     #[serde(skip_serializing)]
     pub read_concern: Option<ReadConcern>,
+
+
+    /// Tags the query with an arbitrary BSON object to help trace the operation through the database
+    /// profiler, currentOp and logs.
+    /// Only available for 4.4+
```
isabelatkinson, line 581:
```suggestion
    ///
    /// This option is only available on server versions 4.4+.
```

src/test/spec/crud.rs
```diff
@@ -18,6 +18,11 @@ fn test_predicate(test: &TestCase) -> bool {
     // The Rust driver doesn't support unacknowledged writes.
     let lower = test.description.to_lowercase();
 
-    // TODO: RUST-1071, RUST-1215: unskip comment tests
-    !lower.contains("unacknowledged") && !lower.contains("comment")
+    // TODO: RUST-1071: unskip comment tests
+    // RUST-1215: unskipped comment tests for estimatedDocumentCount
+<<<<<<< HEAD
```
isabelatkinson, line 23: looks like there were some git issues here -- let me know if you need any help fixing these!

src/coll/options.rs
```diff
@@ -574,6 +574,13 @@ pub struct EstimatedDocumentCountOptions {
     /// The level of the read concern.
     #[serde(skip_serializing)]
     pub read_concern: Option<ReadConcern>,
+
+    /// Tags the query with an arbitrary BSON object to help trace the operation through the
+    /// database profiler, currentOp and logs.
+    ///
+    /// This option is only available on server versions 4.4+.
+    #[serde(skip_serializing)]
```
isabelatkinson, line 582: We'll want to omit `#[serde(skip_serializing)]` here. We derive `Serialize` on these options structs so that we can serialize the options specified by the user into bson to send the operation to the server. A few of the other options in this struct have `#[serde(skip_serializing)]` because we serialize/use those options manually, but this one can be serialized by the default implementation.

src/coll/options.rs
```diff
@@ -211,6 +211,14 @@ pub struct UpdateOptions {
 
     /// The write concern for the operation.
     pub write_concern: Option<WriteConcern>,
+
+    /// Map of parameter names and values. Values must be constant or closed
+    /// expressions that do not reference document fields. Parameters can then be
+    /// accessed as varaibles in an aggregate expression context (e.g. "$$var").
```
isabelatkinson, line 217:
```suggestion
    /// accessed as variables in an aggregate expression context (e.g. "$$var").
```

(and ditto below)

src/coll/options.rs
```diff
@@ -641,7 +696,7 @@ pub struct FindOptions {
 
     /// Tags the query with an arbitrary string to help trace the operation through the database
     /// profiler, currentOp and logs.
-    pub comment: Option<String>,
+    pub comment: Option<Bson>,
```
isabelatkinson, line 699: Because this option is already in our public API, we need to leave it as a string. Otherwise, users who are already using the option will get compilation errors because they're no longer specifying the correct type. Was this change needed for the sake of test file deserialization?

(From looking at the specification for this option in the CRUD spec, the server does accept arbitrary BSON on 4.4+, but it doesn't seem like we've made that change yet in Rust.)

src/coll/options.rs
```diff
@@ -783,7 +847,7 @@ pub struct FindOneOptions {
...
```
isabelatkinson, line 850: ditto the above comment here

src/sdam/description/server.rs
```diff
@@ -122,6 +122,7 @@ impl PartialEq for ServerDescription {
 
                 self_response == other_response
             }
+            (Err(self_err), Err(other_err)) => self_err == other_err,
```
isabelatkinson, line 125: It looks like changes from some recent commits got included in the diff for this PR. Can you rebase your branch on main to fix? (let me know if you need any help with this)

src/test/spec/crud.rs
```diff
@@ -18,6 +18,12 @@ fn test_predicate(test: &TestCase) -> bool {
...
-    // TODO: RUST-1071, RUST-1215: unskip comment tests
-    !lower.contains("unacknowledged") && !lower.contains("comment")
+    // TODO: RUST-1071, RUST-663: unskip comment tests
+    // and aggregate $out and $merge tests
+    !lower.contains("unacknowledged")
+        && (!lower.contains("comment")
+            || (lower.contains("estimated")
+                && lower.contains("document")
+                && lower.contains("count")))
+        && !(lower.contains("aggregate") && lower.contains("preference"))
```
isabelatkinson, line 28: It's possible that there are unrelated CRUD tests that contain both "aggregate" and "preference" in their descriptions, so I think we should list out the exact descriptions for the tests we want to skip here to ensure we're not losing any existing coverage.

src/test/spec/crud.rs
```diff
@@ -18,6 +18,12 @@ fn test_predicate(test: &TestCase) -> bool {
     // The Rust driver doesn't support unacknowledged writes.
     let lower = test.description.to_lowercase();
 
-    // TODO: RUST-1071, RUST-1215: unskip comment tests
-    !lower.contains("unacknowledged") && !lower.contains("comment")
+    // TODO: RUST-1071, RUST-663: unskip comment tests
+    // and aggregate $out and $merge tests
+    !lower.contains("unacknowledged")
+        && (!lower.contains("comment")
+            || (lower.contains("estimated")
```
isabelatkinson, line 25: Would this work as `(!lower.contains("comment") || lower.contains("estimateddocumentcount"))`?

src/test/spec/crud.rs
```diff
@@ -18,6 +18,12 @@ fn test_predicate(test: &TestCase) -> bool {
     // The Rust driver doesn't support unacknowledged writes.
     let lower = test.description.to_lowercase();
 
-    // TODO: RUST-1071, RUST-1215: unskip comment tests
-    !lower.contains("unacknowledged") && !lower.contains("comment")
+    // TODO: RUST-1071, RUST-663: unskip comment tests
```
isabelatkinson, line 21: Can we split these TODO comments to be above the relevant boolean checks? That'll make it a little easier for whoever does those tickets to know what to delete.

src/test/spec/unified_runner/operation.rs
```diff
@@ -285,6 +287,8 @@ pub(super) struct DeleteOne {
     session: Option<String>,
     #[serde(flatten)]
     options: DeleteOptions,
+    #[serde(rename = "comment")]
+    _comment: Option<Bson>,
```
isabelatkinson, line 291: Can we add TODO comments to this field (and the other `_comment` fields below) to delete these workarounds when RUST-1071 is done? You can just copy the one you deleted from `CountDocuments`.

isabelatkinson (commented):
looking good! just a few style/documentation comments

isabelatkinson (approved):
lgtm! going to tag in the rest of the team for review

abr-egn (approved):
LGTM!

patrickfreed (approved):
LGTM!

PR #681, opened 2022-06-14
RUST-1357 Bump `clippy` version to `1.61.0`

isabelatkinson:


.evergreen/check-clippy.sh
```diff
@@ -4,7 +4,7 @@ set -o errexit
 
 source ./.evergreen/configure-rust.sh
 # Pin clippy to the lastest version. This should be updated when new versions of Rust are released.
-rustup default 1.59.0
+rustup default 1.61.0
```
patrickfreed, line 7: instead of updating the default toolchain, can we instead do `cargo +1.61.0 clippy <args>`? That way if we run it locally it won't modify our toolchain preferences.

kmahar (approved):

patrickfreed (approved):

PR #682, opened 2022-06-16
RUST-910 Expose server connectionId in command monitoring events

kmahar:
RUST-910

We'll want to expose this in command log messages, so I went ahead with adding this to the monitoring events.

### DIFF
--- a/src/cmap/conn/mod.rs
+++ b/src/cmap/conn/mod.rs
@@ -39,9 +39,14 @@ pub(crate) use wire::next_request_id;
 
 /// User-facing information about a connection to the database.
 #[derive(Clone, Debug, Serialize)]
+#[non_exhaustive]
 pub struct ConnectionInfo {
     /// A driver-generated identifier that uniquely identifies the connection.
     pub id: u32,
+
+    /// A server-generated identifier that uniquely identifies the connection. Available on server
+    /// versions 4.2+. This may be used to correlate driver connections with server logs.
+    pub server_id: Option<i32>,
 
     /// The address that the connection is connected to.
     pub address: ServerAddress,
@@ -51,7 +56,11 @@ pub struct ConnectionInfo {
 #[derive(Derivative)]
 #[derivative(Debug)]
 pub(crate) struct Connection {
+    /// Driver-generated ID for the connection.
     pub(super) id: u32,
+    /// Server-generated ID for the connection.
+    pub(crate) server_id: Option<i32>,
+
     pub(crate) address: ServerAddress,
     pub(crate) generation: ConnectionGeneration,
 
@@ -109,6 +118,7 @@ impl Connection {
 
         let conn = Self {
             id,
+            server_id: None,
             generation: ConnectionGeneration::Normal(generation),
             pool_manager: None,
             command_executing: false,
@@ -174,6 +184,7 @@ impl Connection {
     pub(crate) fn info(&self) -> ConnectionInfo {
         ConnectionInfo {
             id: self.id,
+            server_id: self.server_id,
             address: self.address.clone(),
         }
     }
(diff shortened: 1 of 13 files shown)


src/cmap/conn/mod.rs
```diff
@@ -38,10 +38,15 @@ pub(crate) use wire::next_request_id;
 
 /// User-facing information about a connection to the database.
 #[derive(Clone, Debug)]
+#[non_exhaustive]
```
kmahar, line 41: it seems we did not mark this type non-exhaustive before, and I'm not sure what our current thinking is on retroactively adding it and adding more properties to the type, but I figured I'd start with this and get your opinions. 
if we feel like we should avoid doing this, I could add a new field `server_connection_id` directly to the monitoring events rather than sticking it on `ConnectionInfo`, and we could update the API in 3.0.

src/test/spec/unified_runner/mod.rs
```diff
@@ -91,15 +75,20 @@ pub async fn run_unified_format_test_filtered(
             }
         }
         if !can_run_on {
-            log_uncaptured("Client topology not compatible with test");
+            log_uncaptured(format!(
```
kmahar, line 78: the various logging changes from here down are not strictly related, but I was having a hard time reading the output to figure out if the new test files were being run properly or not so I decided to clarify the log statements a bit

src/test/spec/unified_runner/mod.rs
```diff
@@ -65,18 +59,8 @@ pub async fn run_unified_format_test_filtered(
     test_file: TestFile,
     pred: impl Fn(&TestCase) -> bool,
 ) {
-    let version_matches = SPEC_VERSIONS.iter().any(|req| {
```
kmahar, line 68: as written, this logic would consider a schema version supported if there was any version in the array >= to it. I noticed this when I brought in the new files using schema version 1.6 and didn't need to change the array here to get them running. 

given that, it seemed simpler and more efficient to switch to just storing our max supported version. 

we could take the approach of checking for exact minor version matches instead (which I think may have been the intended idea here), but I think that also doesn't really solve the problem of determining if a file uses unsupported features -- it's possible that a file could e.g. use features from both schema versions 1.6 and 1.7, and therefore report its schema version as 1.7, and we'd consider it supported despite lacking 1.6 support.

src/hello.rs
```diff
@@ -237,6 +237,10 @@ pub(crate) struct HelloCommandResponse {
 
     /// The maximum permitted size of a BSON wire protocol message.
     pub max_message_size_bytes: i32,
+
+    /// The server-generated ID for the connection the "hello" command was run on.
+    /// Present on server versions 4.2+.
+    pub connection_id: Option<i32>,
 }
 
 impl PartialEq for HelloCommandResponse {
```
kmahar, line 246: see my slack message in #rust-team for discussion on this method; I wasn't sure whether to include the connection_id here, or if the current implementation is working as intended

kmahar:
MSRV compile failures are unrelated, have #684 open to fix them

src/cmap/conn/mod.rs
isabelatkinson: It looks like we don't actually expose the `ConnectionInfo` struct in our public API -- it's not exported in lib.rs, so it doesn't show up in documentation and users can't import the type. For that reason I'm not sure that this would actually be a breaking change? `#[non_exhuastive]` is needed to prevent constructing/destructuring, but I don't think that's possible to begin with if you're not able to import the type.

kmahar: oh interesting. well, it seems like we probably do want to expose it in our API so users can see the available fields in the docs. would we want to expose it via the `cmap` module (which currently has nothing else public) or just re-export it via `pub use` from the `event` module?

isabelatkinson: I'd be in favor of `event`, we've done that for other similar types related to SDAM/CMAP monitoring.

abr-egn (approved):
LGTM!

src/cmap/conn/mod.rs
kmahar: added a `pub use` to make it visible in `event::command`

patrickfreed (approved):

isabelatkinson (approved):

PR #688, opened 2022-07-01
RUST-1392 Add GridFS support: Implement public API with placeholder code

sanav33:
WIP: Adding GridFS support to the Rust Driver.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,6 +76,7 @@ bson = { git = "https://github.com/mongodb/bson-rust", branch = "main" }
 chrono = "0.4.7"
 derivative = "2.1.1"
 flate2 = { version = "1.0", optional = true }
+futures = "0.3.14"
 futures-core = "0.3.14"
 futures-util = { version = "0.3.14", features = ["io"] }
 futures-executor = "0.3.14"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,6 @@ bson = { git = "https://github.com/mongodb/bson-rust", branch = "main" }
 chrono = "0.4.7"
 derivative = "2.1.1"
 flate2 = { version = "1.0", optional = true }
-futures = "0.3.14"
 futures-core = "0.3.14"
 futures-util = { version = "0.3.14", features = ["io"] }
 futures-executor = "0.3.14"
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -19,7 +19,7 @@ use crate::{
     concern::{ReadConcern, WriteConcern},
     cursor::Cursor,
     error::{Error, ErrorKind, Result},
-    gridfs::{GridFSBucket, GridFSBucketOptions},
+    gridfs::{GridFsBucket, GridFsBucketOptions},
     operation::{Aggregate, AggregateTarget, Create, DropDatabase, ListCollections, RunCommand},
     options::{
         AggregateOptions,
@@ -566,11 +566,12 @@ impl Database {
             .await
     }
 
+    /// Creates a new [`GridFsBucket`] in the database with the given options.
     pub fn new_gridfs_bucket(
         &self,
-        options: impl Into<Option<GridFSBucketOptions>>,
-    ) -> GridFSBucket {
-        GridFSBucket {
+        options: impl Into<Option<GridFsBucketOptions>>,
+    ) -> GridFsBucket {
+        GridFsBucket {
             db: self.clone(),
             options: options.into(),
         }
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -19,7 +19,7 @@ use crate::{
     concern::{ReadConcern, WriteConcern},
     cursor::Cursor,
     error::{Error, ErrorKind, Result},
-    gridfs::{GridFsBucket, GridFsBucketOptions},
+    gridfs::{GridFsBucket, options::GridFsBucketOptions},
     operation::{Aggregate, AggregateTarget, Create, DropDatabase, ListCollections, RunCommand},
     options::{
         AggregateOptions,
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -571,7 +571,7 @@ impl Database {
             .await
     }
 
-    /// Creates a new [`GridFsBucket`] in the database with the given options.
+    /// Creates a new GridFsBucket in the database with the given options.
     pub fn gridfs_bucket(&self, options: impl Into<Option<GridFsBucketOptions>>) -> GridFsBucket {
         let mut options = options.into().unwrap_or_default();
         options.read_concern = options
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -571,9 +571,25 @@ impl Database {
         &self,
         options: impl Into<Option<GridFsBucketOptions>>,
     ) -> GridFsBucket {
+        let options: GridFsBucketOptions = options.into().unwrap_or_default();
+        let read_concern = options
+            .read_concern
+            .or_else(|| self.read_concern().cloned());
+        let write_concern = options
+            .write_concern
+            .or_else(|| self.write_concern().cloned());
+        let read_preference = options
+            .read_preference
+            .or_else(|| self.selection_criteria().cloned());
+        let bucket_name = options.bucket_name.unwrap_or_else(|| "fs".to_string());
+        let chunk_size_bytes = options.chunk_size_bytes.unwrap_or(255 * 1024);
         GridFsBucket {
             db: self.clone(),
-            options: options.into(),
+            bucket_name,
+            chunk_size_bytes,
+            read_concern,
+            write_concern,
+            read_preference,
         }
     }
 }
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -19,7 +19,7 @@ use crate::{
     concern::{ReadConcern, WriteConcern},
     cursor::Cursor,
     error::{Error, ErrorKind, Result},
-    gridfs::{GridFsBucket, options::GridFsBucketOptions},
+    gridfs::{options::GridFsBucketOptions, GridFsBucket},
     operation::{Aggregate, AggregateTarget, Create, DropDatabase, ListCollections, RunCommand},
     options::{
         AggregateOptions,
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -19,7 +19,12 @@ use crate::{
     concern::{ReadConcern, WriteConcern},
     cursor::Cursor,
     error::{Error, ErrorKind, Result},
-    gridfs::{options::GridFsBucketOptions, GridFsBucket, DEFAULT_BUCKET_NAME, DEFAULT_CHUNK_SIZE_BYTES},
+    gridfs::{
+        options::GridFsBucketOptions,
+        GridFsBucket,
+        DEFAULT_BUCKET_NAME,
+        DEFAULT_CHUNK_SIZE_BYTES,
+    },
     operation::{Aggregate, AggregateTarget, Create, DropDatabase, ListCollections, RunCommand},
     options::{
         AggregateOptions,
@@ -581,7 +586,9 @@ impl Database {
         let selection_criteria = options
             .selection_criteria
             .or_else(|| self.selection_criteria().cloned());
-        let bucket_name = options.bucket_name.unwrap_or_else(|| DEFAULT_BUCKET_NAME.to_string());
+        let bucket_name = options
+            .bucket_name
+            .unwrap_or_else(|| DEFAULT_BUCKET_NAME.to_string());
         let chunk_size_bytes = options.chunk_size_bytes.unwrap_or(DEFAULT_CHUNK_SIZE_BYTES);
         GridFsBucket {
             db: self.clone(),
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -19,7 +19,7 @@ use crate::{
     concern::{ReadConcern, WriteConcern},
     cursor::Cursor,
     error::{Error, ErrorKind, Result},
-    gridfs::{options::GridFsBucketOptions, GridFsBucket},
+    gridfs::{options::GridFsBucketOptions, GridFsBucket, DEFAULT_BUCKET_NAME, DEFAULT_CHUNK_SIZE_BYTES},
     operation::{Aggregate, AggregateTarget, Create, DropDatabase, ListCollections, RunCommand},
     options::{
         AggregateOptions,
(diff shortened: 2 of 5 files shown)


src/gridfs.rs
```diff
@@ -0,0 +1,168 @@
...
+#[non_exhaustive]
+pub struct GridFSBucketOptions {
+    bucket_name: Option<String>,
+    chunk_size_bytes: Option<i32>,
+    write_concern: Option<WriteConcern>,
+    read_concern: Option<ReadConcern>,
+    read_preference: Option<ReadPreference>,
+}
+
+pub struct GridFSUploadOptions {
```
isabelatkinson, line 32: We'll need to add the same annotations as are included on `GridFSBucketOptions` here, and ditto for all other options structs.

src/gridfs.rs
```diff
@@ -0,0 +1,168 @@
...
+use typed_builder::TypedBuilder;
+
+use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
+
+#[derive(Clone, Debug, Default, Deserialize, TypedBuilder)]
+#[builder(field_defaults(setter(into)))]
+#[non_exhaustive]
+pub struct GridFSBucketOptions {
+    bucket_name: Option<String>,
+    chunk_size_bytes: Option<i32>,
```
isabelatkinson, line 26: Can we make all of the options fields `pub`? That way users can reassign fields as needed.

src/gridfs.rs
```diff
@@ -0,0 +1,168 @@
...
+    length: i64,
+    chunk_size: i32,
+    upload_date: DateTime,
+    filename: String,
+    metadata: Document,
+}
+
+pub struct GridFSBucket {
+    // Contains a "chunks" collection
+    pub db: Database,
```
isabelatkinson, line 76: As per the last line in [this section](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#configurable-gridfsbucket-class) of the spec, bucket instances should be immutable. For that reason we'll want to make these fields non-`pub` so that users can't declare a `mut` bucket and change them. We may want to add getter methods for the read concern/write concern/selection criteria (similar to those of [`Collection`](https://docs.rs/mongodb/latest/mongodb/struct.Collection.html#)) so that users can see what those were resolved to.

isabelatkinson (commented):
looking good! i'm also going to work on adding some design comments for the other reviewers once we tag them in

isabelatkinson (commented):
looking good! a few more style/design comments.

src/gridfs.rs
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+
+    /// Opens a [`GridFsStream`] that the application can write the contents of the file to.
+    /// The application provides a custom file id.
+    ///
+    /// Returns a [`GridFsStream`] to which the application will write the contents.
+    pub fn open_upload_stream_with_id<T>(
+        &self,
+        id: T,
```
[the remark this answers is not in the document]
sanav33: I agree with letting the `TFileId` just be a `Bson` type. Introducing generics for IDs doesn't seem like it adds anything significant in the way of usability and it just adds unnecessary complexity to the code.

I was thinking about this while working on the unified test runner support. If we use `ObjectId`s for `id`, does it make sense to even have the `open_upload_stream_with_id`/`upload_from_stream_with_id` in the public API? If the `id`s have to be unique, wouldn't the `id` argument always be the result of an `ObjectId::new()` call?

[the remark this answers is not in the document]
isabelatkinson: Technically the `_id` field in a document just needs to be a unique BSON value, so a user could theoretically provide arbitrary BSON for the `id`, and as long as it doesn't conflict with any already-existing `id`s then it's valid. I think the vast majority of users won't want/need to do this, but we should still include the `_with_id` methods for those who do.

src/gridfs.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    /// Given an `id`, deletes the stored file's files collection document and
+    /// associated chunks from a [`GridFsBucket`].
+    pub fn delete<T>(&self, id: T) {
+        todo!()
+    }
+
+    /// Finds and returns the files collection documents that match the filter.
+    pub fn find<T>(&self, filter: Document, options: GridFsBucketOptions) -> Result<Cursor<T>> {
```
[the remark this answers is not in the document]
isabelatkinson: The nice thing about using the `FilesCollectionDocument` struct rather than a plain `Document` is that the validation of the data returned is done ahead of time by the cursor, so a user wanting to inspect a field just needs to access it on the struct. With a `Document`, a user wanting to inspect the `length` field would need to do something like:
```
doc.get_i64("length").unwrap();
```
which requires knowing what the type of that field should be. Even though the unwrap should never fail it's nice to have the guarantee of the type system rather than needing to assume an invariant/write extra error handling.

That said, I'd be interested in what other members of the team think about using a defined struct here; we can leave this thread open for their input once I tag them in for review.

src/gridfs.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+/// Struct for storing GridFS managed files within a [`Database`].
+pub struct GridFsBucket {
+    // Contains a "chunks" collection
+    pub(crate) db: Database,
+    pub(crate) options: Option<GridFsBucketOptions>,
+}
+
+pub struct GridFsStream {
```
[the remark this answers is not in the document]
sanav33: Another potential implementation could be to make `GridFsStream` a wrapper around a generic with `AsyncRead` / `AsyncWrite` as trait bounds so users can use [any of the types](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWrite.html#implementors) in `tokio` that already implement those traits.

isabelatkinson (commented):
looking good! just one more comment

src/db/mod.rs
```diff
@@ -492,4 +493,15 @@ impl Database {
...
     }
+
+    /// Creates a new [`GridFsBucket`] in the database with the given options.
+    pub fn new_gridfs_bucket(
+        &self,
+        options: impl Into<Option<GridFsBucketOptions>>,
+    ) -> GridFsBucket {
+        GridFsBucket {
+            db: self.clone(),
+            options: options.into(),
```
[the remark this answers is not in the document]
sanav33: I think it would be a good idea to include to store the fields specified in the `GridFsBucketOptions` struct directly within `GridFsBucket`, especially since `chunk_size_bytes` and `bucket_name` are used frequently in the upload and download methods and it makes the code harder to read to figure out those values from all the layers of `Option` around those fields.

src/gridfs.rs
```diff
@@ -0,0 +1,285 @@
...
+        todo!()
+    }
+
+    fn poll_shutdown(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<tokio::io::Result<()>> {
+        todo!()
+    }
+}
+
+pub struct GridFsDownloadStream {
+    pub id: Bson,
```
patrickfreed, line 87: rather than exposing this field publicly, I think it might be preferable to use a helper. That way it's impossible for users to mutate it. Ditto for the upload stream.

src/gridfs.rs
```diff
@@ -0,0 +1,285 @@
...
+        filename: String,
+        source: impl tokio::io::AsyncRead,
+        options: impl Into<Option<GridFsUploadOptions>>,
+    ) {
+        todo!()
+    }
+
+    /// Uploads a user file to a GridFS bucket. The application supplies a custom file id. Uses the
+    /// `futures` crate.
+    pub async fn upload_from_stream_with_id_futures(
```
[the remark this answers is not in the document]
sanav33: > Given this, I think it might be more accurate within the Rust ecosystem to say something like upload_from_reader_with_id, to match up with stuff like [serde_json::from_reader](https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html), epsecially since Stream has its own meaning within async Rust.
I think this is a good idea, and it would also make the difference between `upload_from_stream` and `open_upload_stream` clearer, since they're entirely different ways to upload data but the wording makes that somewhat unclear, in my opinion.

> Also, it might read a little more natural if we used the futures/tokio part to describe the reader rather than just be at the end. e.g. upload_from_futures_reader_with_id. 
I don't have a strong preference either way for this but this change sounds okay to me.

src/gridfs.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub async fn delete(&self, id: Bson) {
+        todo!()
+    }
+
+    /// Finds and returns the files collection documents that match the filter.
+    pub async fn find(
+        &self,
+        filter: Document,
+        options: impl Into<Option<GridFsBucketOptions>>,
+    ) -> io::Result<Cursor<FilesCollectionDocument>> {
```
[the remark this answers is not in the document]
sanav33: Fixed.

src/gridfs/options.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub batch_size: Option<i32>,
+
+    /// The maximum number of documents to return.
+    pub limit: Option<i32>,
+
+    /// The maximum amount of time to allow the query to run.
+    pub max_time_ms: Option<i64>,
+
+    /// The server normally times out idle cursors after an inactivity period
+    /// (10 minutes) to prevent excess memory use. Set this option to prevent that.
```
kmahar, line 84: the timeout value is actually configurable on a per server basis: https://jira.mongodb.org/browse/SERVER-8188 
so I think it would be best to just remove the (10 minutes) part here. the spec pre-dates that work being done. it might be nice to make a PR to remove that from the spec as well it if you have time!

src/gridfs/options.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub no_cursor_timeout: Option<bool>,
+
+    /// The number of documents to skip before returning.
+    pub skip: i32,
```
kmahar, line 88: I know the spec doesn't say it but I think this should probably be optional? also, this is a `u64` for `FindOptions` so I think it would make sense to do the same here

src/gridfs/options.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// of the file to retrieve. Defaults to -1 (the most recent revision).
+    ///
+    /// Revision numbers are defined as follows:
+    /// 0 = the original stored file
+    /// 1 = the first revision
+    /// 2 = the second revision
+    /// etc...
+    /// -2 = the second most recent revision
+    /// -1 = the most recent revision
+    pub revision: Option<i32>,
```
kmahar, line 61: it might be a bit more user-friendly to model this with an enum since there are a set of known values corresponding to constants

src/gridfs.rs
```diff
@@ -0,0 +1,285 @@
...
+        self.read_concern.as_ref()
+    }
+
+    /// Gets the write concern of the [`GridFsBucket`].
+    pub fn write_concern(&self) -> Option<&WriteConcern> {
+        self.write_concern.as_ref()
+    }
+
+    /// Gets the read preference of the [`GridFsBucket`].
+    pub fn read_preference(&self) -> Option<&SelectionCriteria> {
```
kmahar, line 112: if we rename the option as I suggested I think we'd want to rename this `selection_criteria` as well

kmahar (changes requested):
looks good! just various nits around the types used

src/gridfs/options.rs
sanav33: I'm not sure I understand what you mean. Say if someone wanted to get the eighth revision for a file, would we create enum variants for every single revision?
e.g.
```
enum FileRevision {
    OriginalStoredFile,
    First,
    Second,
    // ... and so on
    SecondMostRecent,
    MostRecent,
}

File: src/gridfs/options.rs (hunk 11 above)
Comment (sanav33) [reply src/gridfs/options.rs]: Done and PR created: https://github.com/mongodb/specifications/pull/1289.

File: src/gridfs.rs [hunk 15]
```diff
@@ -0,0 +1,299 @@
...
+}
+
+impl GridFsDownloadStream {
+    /// Gets the file `id` for the stream.
+    pub fn files_id(&self) -> &Bson {
+        &self.files_id
+    }
+}
+
+impl tokio::io::AsyncRead for GridFsDownloadStream {
```
patrickfreed, line 102: we'll also want to implement `futures` version of `AsyncRead` here. Users can decide which trait to import, so there won't be any clashing issues.

src/gridfs.rs
```diff
@@ -0,0 +1,299 @@
...
+        todo!()
+    }
+
+    /// Aborts the upload and discards the upload stream.
+    pub async fn abort(self) {
+        todo!()
+    }
+}
+
+impl tokio::io::AsyncWrite for GridFsUploadStream {
```
patrickfreed, line 73: ditto implementing `futures` version.

src/gridfs.rs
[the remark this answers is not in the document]
kmahar: it is ugly, but I agree with Patrick we need to include the `0_3` part somehow  🤔 I don't have a strong preference between `upload_from_futures_0_3_reader_with_id` or `upload_from_reader_with_id_futures_0_3` though

src/gridfs/options.rs
kmahar: per our conversation offline I misunderstood the possible values of this enum and didn't realize it could be any number, not just one of a few, so I retract this suggestion

src/gridfs.rs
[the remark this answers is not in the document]
kmahar: I like the suggestion to have a `FilesCollectionDocument` type, for the reasons Isabel mentioned; it's a well-defined format and when possible we generally try to model those types

Cargo.toml
```diff
@@ -59,6 +59,7 @@ bson = { git = "https://github.com/mongodb/bson-rust", branch = "main" }
 chrono = "0.4.7"
 derivative = "2.1.1"
 flate2 = { version = "1.0", optional = true }
+futures = "0.3.14"
```
abr-egn, line 62: I think we don't need to explicitly pull in `futures-core`, `-util`, or `-executor` if we're using `futures`.

sanav33: We could do away with those but that would require changing every import of `futures-core`, `-util` and `-executor` in the driver. Is that something we want to do?

patrickfreed: The various individual `futures-x` crates have differing stability guarantees, which is why we depend on them directly. See [this comment](https://github.com/mongodb/mongo-rust-driver/pull/337#discussion_r627881130) for more context / reasoning.

Given that, I think we'll want to continue to depend on the various subcrates directly.

sanav33: Ah I see, that makes sense. Fixed!

kmahar (approved):

patrickfreed (approved):
LGTM!

abr-egn (approved):

isabelatkinson (approved):

PR #689, opened 2022-07-01
RUST-1191 Ensure change stream is in proper state in mid_batch_resume_token test

patrickfreed:
RUST-1191

This PR makes another attempt at reducing the flakiness of the `change_stream::mid_batch_resume_token` test.

### DIFF
--- a/src/test/change_stream.rs
+++ b/src/test/change_stream.rs
@@ -455,7 +455,7 @@ async fn batch_mid_resume_token() -> Result<()> {
         None => return Ok(()),
     };
 
-    // This loop gets the stream to a point where it has been iterated up to but nt including
+    // This loop gets the stream to a point where it has been iterated up to but not including
     // the last event in its batch.
     let mut event_id = None;
     loop {
--- a/src/test/change_stream.rs
+++ b/src/test/change_stream.rs
@@ -455,11 +455,28 @@ async fn batch_mid_resume_token() -> Result<()> {
         None => return Ok(()),
     };
 
-    coll.insert_many((0..2).map(|i| doc! { "_id": i as i32 }), None)
-        .await?;
+    // This loop gets the stream to a point where it has been iterated up to but nt including
+    // the last event in its batch.
+    let mut event_id = None;
+    loop {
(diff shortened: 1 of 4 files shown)


src/test/change_stream.rs
```diff
@@ -455,11 +455,28 @@ async fn batch_mid_resume_token() -> Result<()> {
         None => return Ok(()),
     };
 
-    coll.insert_many((0..2).map(|i| doc! { "_id": i as i32 }), None)
-        .await?;
+    // This loop gets the stream to a point where it has been iterated up to but nt including
+    // the last event in its batch.
+    let mut event_id = None;
```
patrickfreed, line 460: the changes here ensure that the stream exactly meets the preconditions set forth in the prose test. I'm not exactly sure what was wrong last time, but it appears that the insert events may have been appearing in separate batches, making the change stream use the postBatchResumeToken instead of the `_id` of the last event.

kmahar: it's interesting how flaky this test has been in rust.I don't think we've ever had a problem with the Swift one which inserts 5 docs, iterates 3 times, and then checks that the resume token matches the _id of the third doc. maybe having more docs makes it less likely that you end up with a doc returned in its own batch?  🤷‍♀️ 
anyway, fix looks reasonable to me

kmahar (approved):

src/test/change_stream.rs
patrickfreed: I wonder if its because the Swift test uses a single mongos (this problem only happens on sharded clusters). As a side note, shouldn't the Swift one iterate 4 times in order to be up to but not including the last event?

abr-egn (approved):

src/test/change_stream.rs
kmahar: ah yeah the single mongos thing might do it. and yeah you're right, it should be iterating one more time 🤔 I filed [SWIFT-1597](https://jira.mongodb.org/browse/SWIFT-1597) about that

isabelatkinson (approved):

PR #690, opened 2022-07-07
RUST-1382 FLE-related `Client` initialization

abr-egn:
RUST-1382

This allows creating a `Client` with auto-encryption enabled, and spins up the various bits of internal state that are needed for that.


src/client/fle/options.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub bypass_query_analysis: Option<bool>,
+}
+
+/// Options specific to each KMS provider.
+pub type KmsProviders = HashMap<KmsProvider, Document>;
+/// TLS options for connections to each KMS provider.
+pub type KmsProvidersTlsOptions = HashMap<KmsProvider, TlsOptions>;
+
+impl AutoEncryptionOpts {
+    pub(crate) fn extra_option<'a, Opt: ExtraOption<'a, Bson>>(
```
abr-egn, line 63: This machinery incurs a bit of extra boilerplate here but saves it at the usage sites; it also ensures that code using `extra_options` didn't accidentally get the field type wrong.

src/client/mod.rs
```diff
@@ -101,6 +103,8 @@ struct ClientInner {
     topology: Topology,
     options: ClientOptions,
     session_pool: ServerSessionPool,
+    #[cfg(feature = "fle")]
+    fle: std::sync::Mutex<Option<fle::ClientState>>,
```
abr-egn, line 107: It's a little awkward to have to wrap this in a `Mutex` since it's not mutated after client construction, but since making the state for FLE needs the main `Client` as part of its own initialization parameters, I didn't see a way around this.

src/client/mod.rs
```diff
@@ -122,12 +126,25 @@ impl Client {
...
             options,
         });
-
         Ok(Self { inner })
     }
 
+    /// Creates a new `Client` connected to the cluster specified by `options` with auto-encryption
+    /// enabled.
+    #[cfg(feature = "fle")]
+    pub async fn with_fle_options(
```
abr-egn, line 139: Ideally `AutoEncryptionOpts` could have just been another field in `ClientOptions`.  Unfortunately, initializing FLE state requires `async` because it has to construct a client to the mongocryptd from an arbitrary URI.  Since there needs to be a distinct creation method to accommodate that, it seemed worth keeping the options distinct to make it clear that if you're setting those options you have to use this method.

src/client/mod.rs
patrickfreed: We could potentially get around this by creating a `WeakClient` type or some such, but I think we're going to need some form of lock anyways in order to invoke the mutable mongocrypt methods, so it's not a big deal. Since `ClientState` stores `Client`s within it, we need to be really careful not to use them while holding the `std` lock guard, as we can easily deadlock (maybe the compiler will enforce this?). To avoid this, we could introduce a wrapper type or something that manages handling the lock, or alternatively just use an async one.

Also, we may want a `RwLock`, where the read portion can be used to access the clients and the write portion for mutating the state. Though maybe its best to defer a decision on these questions until the implementation moves a little further and the use cases are clearer.

Cargo.toml
```diff
@@ -62,6 +62,9 @@ zstd-compression = ["zstd"]
 zlib-compression = ["flate2"]
 snappy-compression = ["snap"]
 
+# DO NOT USE; see https://jira.mongodb.org/browse/RUST-569 for the status of FLE support in the Rust driver.
+fle = ["mongocrypt", "which"]
```
patrickfreed, line 66: Are we now referring to all of FLE as "Queryable Encryption"? Or is that just FLE 2? If not, should we call this `csfle`? Or have separate features for fle 1 and fle 2?

src/client/mod.rs
patrickfreed, line 139: related to my question above, maybe this would be better named `with_encryption_opts`?

src/error.rs
```diff
@@ -483,6 +490,11 @@ pub enum ErrorKind {
     /// No resume token was present in a change stream document.
     #[error("Cannot provide resume functionality when the resume token is missing")]
     MissingResumeToken,
+
+    /// An error occurred during encryption or decryption.
+    #[cfg(feature = "fle")]
+    #[error("An error occurred during client-side encryption: {0}")]
+    Crypt(mongocrypt::error::Error),
```
patrickfreed, line 497: I think it might be a bit clearer if this errorkind `Fle` or `QueryableEncryption`, depending on what we call things these days. Alternatively, if this is to encompass only errors from `mongocrypt` and not more general, driver-side encryption errors, maybe we name this one `Mongocrypt` and have another kind for regular fle ones? Also happy to defer this question until further in the implementation when we know more about the kinds of errors we'll be returning.

src/client/fle/options.rs
```diff
@@ -0,0 +1,132 @@
...
+/// Automatic encryption is an enterprise only feature that only applies to operations on a
+/// collection. Automatic encryption is not supported for operations on a database or view, and
+/// operations that are not bypassed will result in error (see [libmongocrypt: Auto Encryption
+/// Allow-List](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#libmongocrypt-auto-encryption-allow-list)). To bypass automatic encryption for all operations, set bypassAutoEncryption=true in
+/// AutoEncryptionOpts.
+#[derive(Debug, Default, TypedBuilder, Clone)]
+#[non_exhaustive]
+pub struct AutoEncryptionOpts {
+    /// Used for data key queries.  Will default to an internal client if not set.
+    pub key_vault_client: Option<crate::Client>,
```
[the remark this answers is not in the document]
abr-egn: Done.

src/client/fle/options.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub extra_options: Option<Document>,
+    /// Configure TLS for connections to KMS providers.
+    pub tls_options: Option<KmsProvidersTlsOptions>,
+    /// Maps namespace to encrypted fields.
+    ///
+    /// Supplying an `encrypted_fields_map` provides more security than relying on an
+    /// encryptedFields obtained from the server. It protects against a malicious server
+    /// advertising a false encryptedFields.
+    pub encrypted_fields_map: Option<HashMap<String, Document>>,
+    /// Disable serverside processing of encrypte indexed fields, allowing use of explicit
```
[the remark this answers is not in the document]
abr-egn: Fixed.

src/client/fle/options.rs
```diff
@@ -0,0 +1,132 @@
...
+                None => Err(Error::invalid_argument(format!(
+                    "unexpected type for extra option {:?}: {:?}",
+                    key, b
+                ))),
+            },
+        }
+    }
+}
+
+// `Input` will always be `Bson`; it's present as a parameter to bring the lifetime into scope so
```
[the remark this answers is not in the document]
abr-egn: ... derp, yes.  I came to this from a different direction and that simply didn't occur to me.

Cargo.toml
abr-egn: "Queryable Encryption" is just FLE 2.  I don't think it's worth having distinct features; there's no cost to having FLE2 present if it's not used except for extra compilation, and even that's going to be nearly trivial.

I've renamed this to "csfle" because that seems to be what's consistently used for the overall feature set in public materials.

src/client/mod.rs
abr-egn: Done.

src/error.rs
abr-egn: Renamed this to `Csfle` to match elsewhere.

src/client/fle/options.rs
```diff
@@ -0,0 +1,132 @@
...
+};
+
+/// Options related to automatic encryption.
+///
+/// Automatic encryption is an enterprise only feature that only applies to operations on a
+/// collection. Automatic encryption is not supported for operations on a database or view, and
+/// operations that are not bypassed will result in error (see [libmongocrypt: Auto Encryption
+/// Allow-List](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#libmongocrypt-auto-encryption-allow-list)). To bypass automatic encryption for all operations, set bypassAutoEncryption=true in
+/// AutoEncryptionOpts.
+#[derive(Debug, Default, TypedBuilder, Clone)]
```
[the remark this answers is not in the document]
abr-egn: Done.

src/client/mod.rs
abr-egn: I realized this needed to be a weak ref in any case, since otherwise there's a potential for creating a reference loop with the main `Client` holding a ref to itself.  I also updated it to a `RwLock` because even very early further implementation points to that being the way to go.

patrickfreed (approved):
One more naming nitpick, but otherwise LGTM!

Cargo.toml
```diff
@@ -77,6 +80,7 @@ hex = "0.4.0"
 hmac = "0.12.1"
 lazy_static = "1.4.0"
 md-5 = "0.10.1"
+mongocrypt = { git = "https://github.com/mongodb/libmongocrypt-rust.git", branch = "main", optional = true }
```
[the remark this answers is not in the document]
abr-egn: Yup, fixed.

src/client/mod.rs
```diff
@@ -122,12 +126,25 @@ impl Client {
...
             options,
         });
-
         Ok(Self { inner })
     }
 
+    /// Creates a new `Client` connected to the cluster specified by `options` with auto-encryption
+    /// enabled.
+    #[cfg(feature = "csfle")]
+    pub async fn with_encryption_opts(
```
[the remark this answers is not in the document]
abr-egn: Done.

src/client/csfle.rs
```diff
@@ -0,0 +1,166 @@
...
+    #[allow(dead_code)]
+    key_vault_client: WeakClient,
+    #[allow(dead_code)]
+    metadata_client: Option<WeakClient>,
+    #[allow(dead_code)]
+    internal_client: Option<Client>,
+}
+
+impl ClientState {
+    pub(super) async fn new(client: &Client, opts: AutoEncryptionOpts) -> Result<Option<Self>> {
```
isabelatkinson, line 50: Why does this need to return a `Result<Option<Self>>`? Is there any case in which `Ok(None)` should be returned?

src/client/csfle/options.rs
```diff
@@ -0,0 +1,140 @@
...
+#[derive(Debug, Default, TypedBuilder, Clone)]
+#[non_exhaustive]
+#[builder(field_defaults(setter(into)))]
+pub struct AutoEncryptionOpts {
+    /// Used for data key queries.  Will default to an internal client if not set.
+    #[builder(default)]
+    pub key_vault_client: Option<crate::Client>,
+    /// A collection that contains all data keys used for encryption and decryption (aka the key
+    /// vault collection).
+    pub key_vault_namespace: String,
```
kmahar, line 31: we have a `Namespace` type, should we use that rather than just a string?

src/client/csfle/options.rs
```diff
@@ -0,0 +1,140 @@
...
+/// collection. Automatic encryption is not supported for operations on a database or view, and
+/// operations that are not bypassed will result in error (see [libmongocrypt: Auto Encryption
+/// Allow-List](
+/// https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#libmongocrypt-auto-encryption-allow-list
+/// )). To bypass automatic encryption for all operations, set bypassAutoEncryption=true in
+/// AutoEncryptionOpts.
+#[derive(Debug, Default, TypedBuilder, Clone)]
+#[non_exhaustive]
+#[builder(field_defaults(setter(into)))]
+pub struct AutoEncryptionOpts {
```
kmahar, line 25: nit: seems more consistent with our other options types to use the full `Options` rather than just `Opts`?

kmahar (commented):
looking good! a few minor comments.

src/client/csfle.rs
```diff
@@ -0,0 +1,166 @@
...
+            && crypt.shared_lib_version().is_none()
+        {
+            return Err(crate::error::Error::invalid_argument(
+                "cryptSharedRequired is set but crypt_shared is not available",
+            ));
+        }
+        Ok(crypt)
+    }
+
+    async fn spawn_mongocryptd(opts: &AutoEncryptionOpts, crypt: &Crypt) -> Result<Option<Client>> {
```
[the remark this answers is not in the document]
abr-egn: This is likely to be refactored, but more clarity is good - added.

src/client/csfle.rs
```diff
@@ -0,0 +1,166 @@
...
+        Command::new(bin_path)
+            .args(&args)
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .spawn()?;
+
+        let uri = opts
+            .extra_option(&EO_MONGOCRYPTD_URI)?
+            .unwrap_or("mongodb://localhost:27020");
+        Ok(Some(Client::with_uri_str(uri).await?))
```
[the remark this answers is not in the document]
abr-egn: Missed that, thank you!  Added.

Cargo.toml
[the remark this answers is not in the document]
abr-egn: I'm inclined to go ahead with the MSRV bump - I'll send that in its own PR.

src/client/csfle/options.rs
abr-egn: Yup, done.

Cargo.toml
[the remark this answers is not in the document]
kmahar: SGTM

kmahar (approved):
LGTM!

isabelatkinson (approved):

PR #691, opened 2022-07-09
RUST-1403 Remove `oldtime` feature from chrono dependency

oskgo:
Remove `oldtime` feature from chrono dependency to eliminate a transitive dependency on a deprecated version of the time crate which has [a vulnerability](https://rustsec.org/advisories/RUSTSEC-2020-0071).

The issue was found using [cargo-audit](https://github.com/RustSec/rustsec/tree/main/cargo-audit), which you might want to add to CI.


patrickfreed (approved):

patrickfreed:
On an unrelated note: there looks to be a bunch of failed tasks where all the tests passed. Should keep an eye on that and see if it comes up on more recent PRs (this is currently based off a pretty old branch).

kmahar (approved):

patrickfreed:
Thanks for the contribution!

PR #693, opened 2022-07-11
pin all dependencies for MSRV compilation check

kmahar:
This makes a similar fix to that we made for the BSON library where all dependencies are pinned for our MSRV compilation check to avoid having to pin new deps explicitly anytime they drop support for our MSRV.

successful patch: https://spruce.mongodb.com/version/62cc6fc532f4171e4e2804e4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


abr-egn (approved):

patrickfreed (approved):
LGTM, thanks for taking care of this!

isabelatkinson (approved):

PR #705, opened 2022-07-21
RUST-1071 Add support for the comment field to all helpers

sanav33:
Added a new comment option to crud, change streams, index management, list databases and collection helper methods and added corresponding unified spec tests from [mongodb/specifications@962fb5f](https://github.com/mongodb/specifications/commit/962fb5f20e9e31b528319fe245147928a3100411).


src/coll/options.rs
```diff
@@ -486,9 +542,11 @@ pub struct AggregateOptions {
     /// information on how to use this option.
     pub collation: Option<Collation>,
 
-    /// Tags the query with an arbitrary string to help trace the operation through the database
-    /// profiler, currentOp and logs.
-    pub comment: Option<String>,
```
isabelatkinson, line 491: I think this same change came up in the last `comment` PR: https://github.com/mongodb/mongo-rust-driver/pull/680#discussion_r896959591

There are a few things we can do here (and in other places where this change was made):
- We can wait until we do a 3.0 release of the driver to support `Bson` comments for these commands. At that point we'll be able to make the breaking change and either return an error if a user specifies a non-`Bson::String` pre-4.4 or just propagate the server error that would occur.
- We can introduce separate `comment_bson` options in these structs that accept `Bson` and leave the current `comment` options as-is. At the point of 3.0 we'd probably want to remove them and change the `comment` fields to `Option<Bson>`.

I think I'd prefer the first option given that users haven't been asking for `Bson` comment support here. We can always implement the second option later on if it does turn out to be necessary. In the meantime, users can also use the string representation of their BSON if they do want to use a `Bson` value. (Either way we do this, we'll need to do some workarounds when deserializing the spec tests.)

src/test/spec/unified_runner/test_file.rs
```diff
@@ -268,6 +268,7 @@ pub struct CollectionOrDatabaseOptions {
     #[serde(rename = "readPreference")]
     pub selection_criteria: Option<SelectionCriteria>,
     pub write_concern: Option<WriteConcern>,
+    pub comment: Option<Bson>,
```
isabelatkinson, line 271: Why was this change needed? The `comment` field shouldn't be specified at the db/coll level

src/coll/options.rs
sanav33: Do you think it would be okay then to revert it to an `Option<String>` and leave a TODO comment to change it to `Bson` before the 3.0 release?

src/test/spec/unified_runner/test_file.rs
sanav33: That was an oversight on my part. I confused the `CreateCollectionOptions` with `CollectionOptions` and ended up propagating that error up. I'll fix this and push again.

src/coll/options.rs
isabelatkinson: Looks like I actually filed a ticket about this when it came up originally, thought I had remembered discussing this a bit in triage: https://jira.mongodb.org/browse/RUST-1364

Yeah, whatever we end up doing we'll need to revert this to an `Option<String>`. Also going to leave this thread open for the rest of the reviewers to weigh in on.

src/operation/aggregate/mod.rs
```diff
@@ -95,11 +100,18 @@ impl Operation for Aggregate {
             wc_error_info.validate()?;
         };
 
+        let comment = if description.max_wire_version.unwrap_or(0) < SERVER_4_4_0_WIRE_VERSION {
```
kmahar, line 103: why do we need this check here and for the `find` operation? it seems like outside of cursors we're just relying on the server to error if comment is unsupported, what makes these cases different?

sanav33: The `AggregateOptions` (specification)[https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read] (and also the FindOptions] requires that the comment for an `aggregate` / `find` operation be propagated in subsequent `getMore` operations for server version 4.4+. This was needed to account for that.

kmahar: ahh I see. thanks! I think it could be helpful to add a brief comment in each of these places explaining that on < 4.4 `find` / `aggregate` support comments but `getMore` doesn't so we can't propagate it.

src/operation/list_collections/mod.rs
```diff
@@ -76,6 +76,7 @@ impl Operation for ListCollections {
             description.server_address.clone(),
             self.options.as_ref().and_then(|opts| opts.batch_size),
             None,
+            None,
```
abr-egn, line 79: Should the comment be propagated from somewhere in the server response?

isabelatkinson: the bottom of [this section](https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst#post-mongodb-280-rc3-versions) indicates that the comment should not be attached to `getMore`s (which what propagating the comment here would do)

PR #683, opened 2022-06-17
RUST-999 Link web framework examples/documentation in README.md

isabelatkinson:

PR #709, opened 2022-07-25
RUST-1305 Use uuid 1.x internally and add support for bson uuid 1.x

abr-egn:


isabelatkinson (approved):

docs/manual/web_framework_examples.html
```diff
@@ -0,0 +1,194 @@
+<!DOCTYPE HTML>
```
patrickfreed, line 1: This file seems entirely new although not related to the changes made here, any idea why it's getting generated like this?

kmahar: I think this is because @isabelatkinson added the info to the manual in #683 but we did not actually regenerate the manual files at that time?

kmahar (approved):

patrickfreed (approved):

PR #710, opened 2022-07-25
RUST-1383 Create or drop auxiliary collections when doing so for a collection with encrypted fields.

abr-egn:
RUST-1383

Per the [spec](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#queryable-encryption-create-and-drop-collection-helpers), `create` and `drop` need additional operations when acting on collections with encrypted fields.  The actual operations here are pretty straightforward; following the right defaulting logic for the associated `encrypted_fields` map is the majority of the complexity here.

### DIFF
--- a/src/client/csfle.rs
+++ b/src/client/csfle.rs
@@ -10,7 +10,8 @@ use mongocrypt::Crypt;
 
 use crate::{
     error::{Error, Result},
-    Client, Namespace,
+    Client,
+    Namespace,
 };
 
 use options::{
@@ -175,14 +176,20 @@ impl ClientState {
     }
 }
 
-pub(crate) fn aux_collections(base_name: &str, enc_fields: &bson::Document) -> Result<Vec<Namespace>> {
+pub(crate) fn aux_collections(
+    base_name: &str,
+    enc_fields: &bson::Document,
+) -> Result<Vec<Namespace>> {
     let mut out = vec![];
     for &key in &["esc", "ecc", "ecoc"] {
         let name = match enc_fields.get_str(format!("{}Collection", key)) {
             Ok(s) => s.to_string(),
             Err(_) => format!("enxcol_.{}.{}", base_name, key),
         };
-        out.push(Namespace::from_str(&name).ok_or_else(|| Error::invalid_argument(format!("invalid namespace {:?}", name)))?);
+        out.push(
+            Namespace::from_str(&name)
+                .ok_or_else(|| Error::invalid_argument(format!("invalid namespace {:?}", name)))?,
+        );
     }
     Ok(out)
 }
--- a/src/client/csfle.rs
+++ b/src/client/csfle.rs
@@ -10,7 +10,7 @@ use mongocrypt::Crypt;
 
 use crate::{
     error::{Error, Result},
-    Client,
+    Client, Namespace,
 };
 
 use options::{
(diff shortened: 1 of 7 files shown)


src/client/csfle.rs
```diff
@@ -177,19 +177,19 @@ impl ClientState {
...
-            Err(_) => format!("enxcol_.{}.{}", base_name, key),
+            Err(_) => format!("enxcol_.{}.{}", base_ns.coll, key),
         };
-        out.push(
-            Namespace::from_str(&name)
-                .ok_or_else(|| Error::invalid_argument(format!("invalid namespace {:?}", name)))?,
-        );
+        out.push(Namespace {
+            coll,
+            ..base_ns.clone()
```
kmahar, line 191: what does this syntax do? also, I think? the db name is the first argument and then the collection name?

abr-egn: That's [struct update syntax](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax) - it's creating a new `Namespace` with the `coll` field updated and all other fields (in this case, just `db`) copied from `base_ns`.

kmahar: ah, cool! hadn't seen that before, thanks for the explanation and link.

kmahar (approved):

src/coll/mod.rs
```diff
@@ -246,6 +253,68 @@ impl<T> Collection<T> {
...
+                None => {
+                    self.inner
+                        .db
+                        .list_collections(filter, None)
+                        .await?
+                        .try_collect()
+                        .await?
+                }
+            };
+            for spec in specs {
```
patrickfreed, line 297: since we pass a filter of `{ "name": self.name() }` which is an exact match, this cursor will only ever contain a single result or none. I think it might be easier to grok if we rewrite this for loop as an `if let` instead to better reflect that.

patrickfreed (commented):
Logic looks good, just have one style nit.

Side question: will there be spec tests for this when the rest of FLE is done? If not, we may want to add a few one-off ones.

abr-egn:
> Side question: will there be spec tests for this when the rest of FLE is done? If not, we may want to add a few one-off ones.

Yup, this is covered by [fle2-CreateCollection.yml](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/legacy/fle2-CreateCollection.yml).

src/coll/mod.rs
abr-egn: Done.

src/operation/drop_collection/test.rs
```diff
@@ -15,6 +15,7 @@ fn build() {
             w: Some(Acknowledgment::Custom("abc".to_string())),
             ..Default::default()
         }),
+        ..Default::default()
```
isabelatkinson, line 18: I think we've talked in the past about removing these tests when they break rather than continuing to maintain them -- should we do that here?

src/coll/mod.rs
```diff
@@ -217,18 +217,25 @@ impl<T> Collection<T> {
...
     async fn drop_common(
         &self,
         options: impl Into<Option<DropCollectionOptions>>,
         session: impl Into<Option<&mut ClientSession>>,
     ) -> Result<()> {
-        let session = session.into();
+        let mut session = session.into();
 
-        let mut options = options.into();
+        let mut options: Option<DropCollectionOptions> = options.into();
```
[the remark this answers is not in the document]
abr-egn: It's not strictly needed, but it made autocomplete work :)

src/operation/drop_collection/test.rs
abr-egn: Whoops, that had dropped out of my head - removed the test.

isabelatkinson (approved):

patrickfreed (approved):

PR #711, opened 2022-07-25
minor: remove unnecessary tests

isabelatkinson:
This PR removes some unnecessary tests that I found while poking around our test runner code during Skunkworks.

### DIFF
--- a/src/test/spec/transactions.rs
+++ b/src/test/spec/transactions.rs
@@ -2,8 +2,7 @@ use serde::{Deserialize, Serialize};
 use tokio::sync::{RwLockReadGuard, RwLockWriteGuard};
 
 use crate::{
-    bson::{doc, serde_helpers::serialize_u64_as_i32, Document},
-    client::session::TransactionState,
+    bson::{doc, Document},
     test::{log_uncaptured, run_spec_test, TestClient, LOCK},
     Collection,
 };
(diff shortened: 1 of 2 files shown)


src/test/spec/transactions.rs
```diff
@@ -27,66 +26,6 @@ async fn run_unified() {
     run_spec_test(&["transactions", "unified"], run_unified_format_test).await;
 }
 
-#[cfg_attr(feature = "tokio-runtime", tokio::test(flavor = "multi_thread"))]
-#[cfg_attr(feature = "async-std-runtime", async_std::test)]
-#[function_name::named]
-/// This test replicates the test cases in errors-client.json as we do not have the document
-/// validation required to trigger client-side errors in those tests.
```
isabelatkinson, line 34: This issue is no longer relevant as of [this commit](https://github.com/mongodb/specifications/commit/a124e21be89cbfc78606725c6ed1fa00dc992777); we currently have `errors-client.json` synced and running, so we don't need this additional test.

src/test/spec/versioned_api.rs
```diff
@@ -1,100 +1,12 @@
...
-async fn run_non_transaction_handling_test(path: PathBuf, file: TestFile) {
-    if path.ends_with("transaction-handling.json") {
-        log_uncaptured("skipping run_non_transaction_handling_test");
-        return;
-    }
-    run_unified_format_test(file).await
-}
-
-// TODO RUST-817 Remove this test in favor of transaction-handling.json versioned API spec test when
-// transactions are implemented in the unified runner
```
isabelatkinson, line 55: This test was a replication of the test in `transaction-handling.json` that I wrote before we had transaction support implemented in the unified test runner. The ticket linked here implies that we would need all of the transactions spec tests ported over to the unified test format to remove this, but we really just needed transactions support in our own runner (which we now have).

src/test/spec/versioned_api.rs
```diff
@@ -1,100 +1,12 @@
...
 #[cfg_attr(feature = "tokio-runtime", tokio::test)]
 #[cfg_attr(feature = "async-std-runtime", async_std::test)]
 async fn run() {
     let _guard: RwLockWriteGuard<_> = LOCK.run_exclusively().await;
-    run_spec_test_with_path(&["versioned-api"], run_non_transaction_handling_test).await;
-}
-
-#[cfg_attr(feature = "tokio-runtime", tokio::test)]
-#[cfg_attr(feature = "async-std-runtime", async_std::test)]
-async fn run_transaction_handling_spec_test() {
```
isabelatkinson, line 30: IIRC this test was originally split out from the rest so we could skip it on certain topologies, but it looks like that conditional skip was removed. Now that we're running this unconditionally there's no need to separate it from the rest of the versioned API tests.

abr-egn (approved):

kmahar (approved):

patrickfreed (approved):

PR #712, opened 2022-07-25
RUST-1373 Update unified test format runner to support SDAM integration tests

patrickfreed:
RUST-1373

This PR updates the test runner to preemptively implement the changes proposed in https://github.com/mongodb/specifications/pull/1274. It also makes some general test runner changes and improvements.


Cargo.toml
```diff
@@ -7,7 +7,7 @@ authors = [
     "Kaitlin Mahar <kaitlin.mahar@mongodb.com>",
 ]
 description = "The official MongoDB driver for Rust"
-edition = "2018"
+edition = "2021"
```
patrickfreed, line 10: The [Disjoint capture in closures](https://doc.rust-lang.org/edition-guide/rust-2021/disjoint-capture-in-closures.html) feature of the 2021 edition helped make using lifetimes in the runner a lot easier, so I bumped it here. Now that the MSRV is updated to the minimum for this edition, this shouldn't be a problem.

src/sdam/description/server.rs
```diff
@@ -106,7 +107,36 @@ pub(crate) struct ServerDescription {
     // allows us to ensure that only valid states are possible (e.g. preventing that both an error
     // and a reply are present) while still making it easy to define helper methods on
     // ServerDescription for information we need from the hello reply by propagating with `?`.
-    pub(crate) reply: Result<Option<HelloReply>, String>,
+    pub(crate) reply: Result<Option<HelloReply>>,
+}
+
+impl Serialize for ServerDescription {
```
patrickfreed, line 113: This manual implementation here just replicates what the old generated one was before the error change.

src/sdam/description/server.rs
```diff
@@ -106,7 +107,36 @@ pub(crate) struct ServerDescription {
     // allows us to ensure that only valid states are possible (e.g. preventing that both an error
     // and a reply are present) while still making it easy to define helper methods on
     // ServerDescription for information we need from the hello reply by propagating with `?`.
-    pub(crate) reply: Result<Option<HelloReply>, String>,
+    pub(crate) reply: Result<Option<HelloReply>>,
```
patrickfreed, line 110: The fact that we store `String` here instead of the actual error is kind of a historical quirk at this point. In order to implement the full error matching for the test runner, I needed to make this a regular `Result`. IMO this is an improvement anyways.

src/test/atlas_planned_maintenance_testing/mod.rs
```diff
@@ -120,7 +121,20 @@ fn write_json(test_runner: &mut TestRunner, mut errors: Vec<Bson>) {
     // The events key is expected to be present regardless of whether storeEventsAsEntities was
     // defined.
     write!(&mut writer, ",\"events\":[").unwrap();
-    test_runner.write_events_list_to_file("events", &mut writer);
+    let event_list_entity = match entities.get("events") {
```
patrickfreed, line 124: I just moved the logic from the function that was previously called here inline.

src/test/spec/mod.rs
```diff
@@ -97,10 +97,8 @@ where
     G: Future<Output = ()>,
     T: DeserializeOwned,
 {
-    let json: Value = serde_json::from_reader(File::open(path.as_path()).unwrap()).unwrap();
-
-    // Printing the name of the test file makes it easier to debug deserialization errors.
-    println!("Running tests from {}", path.display());
+    let json: Value = serde_json::from_reader(File::open(path.as_path()).unwrap())
```
patrickfreed, line 100: since we print the path when deserialization fails, I removed the extra print here since it clashed with the new logging changes I made in the runner itself.

src/test/spec/retryable_writes/test_file.rs
```diff
@@ -8,46 +8,46 @@ use crate::{
 
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct TestFile {
-    pub run_on: Option<Vec<RunOn>>,
-    pub data: Vec<Document>,
-    pub tests: Vec<TestCase>,
+pub(crate) struct TestFile {
```
patrickfreed, line 11: these being `pub` actually ran into some issues when trying to use `pub(crate)` stuff in the driver, and this happens because `cargo` can't really tell the difference between regular `pub` and `cfg(test)`'d `pub`. Since the tests are always crate private, I just converted everything to `pub(crate)`. I started with only a few things, but it has kind of a viral effect, leading to a lot of things needing to change, including stuff outside of the unified runner like this.

src/test/spec/unified_runner/matcher.rs
```diff
@@ -154,6 +163,57 @@ fn cmap_events_match(actual: &CmapEvent, expected: &ExpectedCmapEvent) -> Result
     }
 }
 
+fn sdam_events_match(
+    actual: &SdamEvent,
+    expected: &ExpectedSdamEvent,
+    description: impl AsRef<str>,
```
patrickfreed, line 169: taking a description here is a step towards having more context aware failures, but I limited the changes to mostly just here. We'll want to make a more comprehensive effort for that in later PRs.

src/test/spec/unified_runner/mod.rs
```diff
@@ -31,14 +32,21 @@ pub use self::{
     test_runner::{EntityMap, TestRunner},
 };
 
+use super::run_spec_test_with_path;
+
 static MIN_SPEC_VERSION: Version = Version::new(1, 0, 0);
-static MAX_SPEC_VERSION: Version = Version::new(1, 7, 0);
+static MAX_SPEC_VERSION: Version = Version::new(1, 10, 0);
+
+fn file_level_log(message: impl AsRef<str>) {
```
patrickfreed, line 40: This function is used to log stuff in a way that separates it from other files. I took the format from some stuff I wrote in the Swift runner a long time ago, let me know if you think it looks good.

Here's a sample:
```
cargo test sessions::run_unified
   Compiling mongodb v2.4.0 (/home/patrick/mongo-rust-driver)
    Finished test [unoptimized + debuginfo] target(s) in 53.69s
     Running unittests (target/debug/deps/mongodb-dc9092005b6c81b2)

running 1 test

------------
Running tests from /home/patrick/mongo-rust-driver/src/test/spec/json/sessions/snapshot-sessions-unsupported-ops.json

Executing "Server returns an error on insertOne with snapshot"
Executing "Server returns an error on insertMany with snapshot"
Executing "Server returns an error on deleteOne with snapshot"
Executing "Server returns an error on updateOne with snapshot"
Executing "Server returns an error on findOneAndUpdate with snapshot"
Executing "Server returns an error on listDatabases with snapshot"
Executing "Server returns an error on listCollections with snapshot"
Executing "Server returns an error on listIndexes with snapshot"
Executing "Server returns an error on runCommand with snapshot"

------------
Skipping file /home/patrick/mongo-rust-driver/src/test/spec/json/sessions/snapshot-sessions-not-supported-client-error.json: client topology not compatible with test


------------
Running tests from /home/patrick/mongo-rust-driver/src/test/spec/json/sessions/driver-sessions-dirty-session-errors.json

Executing "Dirty explicit session is discarded (insert)"
Executing "Dirty explicit session is discarded (findAndModify)"
Executing "Dirty implicit session is discarded (insert)"
Executing "Dirty implicit session is discarded (findAndModify)"
Executing "Dirty implicit session is discarded (read returning cursor)"
Executing "Dirty implicit session is discarded (read not returning cursor)"

------------
Skipping file /home/patrick/mongo-rust-driver/src/test/spec/json/sessions/snapshot-sessions-not-supported-server-error.json: client topology not compatible with test


------------
Running tests from /home/patrick/mongo-rust-driver/src/test/spec/json/sessions/driver-sessions-server-support.json

Executing "Server supports explicit sessions"
Executing "Server supports implicit sessions"

------------
Running tests from /home/patrick/mongo-rust-driver/src/test/spec/json/sessions/snapshot-sessions.json

Executing "Find operation with snapshot"
Executing "Distinct operation with snapshot"
Executing "Aggregate operation with snapshot"
Executing "countDocuments operation with snapshot"
Executing "Mixed operation with snapshot"
Executing "Write commands with snapshot session do not affect snapshot reads"
Executing "First snapshot read does not send atClusterTime"
Executing "StartTransaction fails in snapshot session"
test test::spec::sessions::run_unified ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 391 filtered out; finished in 10.90s
```

src/test/spec/unified_runner/observer.rs
```diff
@@ -0,0 +1,103 @@
...
+    runtime,
+    test::Event,
+};
+
+use super::{events_match, EntityMap, ExpectedEvent};
+
+/// Observer used to cache all the seen events for a given client in a unified test.
+/// Used to implement assertEventCount and waitForEvent operations.
+#[derive(Debug)]
+pub(crate) struct EventObserver {
```
patrickfreed, line 22: The existing `EventHandler` "consumed" events as they were observed, which doesn't work for the functionality we need in the unified runner. Down the road, we may want to convert all the usages of `EventHandler` to use `EventObserver` so we can ditch the other type, but I figured that was out of scope for this work.

src/test/spec/unified_runner/operation.rs
```diff
@@ -93,16 +103,104 @@ pub trait TestOperation: Debug + Send + Sync {
     }
 }
 
+macro_rules! with_mut_session {
```
patrickfreed, line 106: to facilitate working with sessions through the lock, this macro pops it out of the entity map, "passes" it to the provided block, and then returns it to the entity map. It does it this way so that we can continue to borrow the entity map in other ways even when we're using a session, which we'd have to borrow mutably from the map.

src/test/spec/unified_runner/operation.rs
```diff
@@ -93,16 +103,104 @@ pub trait TestOperation: Debug + Send + Sync {
...
-    pub name: String,
-    pub object: OperationObject,
-    pub expectation: Expectation,
+    pub(crate) name: String,
+    pub(crate) object: OperationObject,
+    pub(crate) expectation: Expectation,
+}
+
+impl Operation {
+    pub(crate) async fn execute<'a>(&self, test_runner: TestRunner, description: &str) {
```
patrickfreed, line 138: this was just copy/pasted here so that we could call `operation.execute` from test runner threads. Previously this was all inline in the runner.

src/test/spec/unified_runner/test_runner.rs
```diff
@@ -60,58 +64,74 @@ const SKIPPED_OPERATIONS: &[&str] = &[
...
+pub(crate) type EntityMap = HashMap<String, Entity>;
 
-pub struct TestRunner {
-    pub internal_client: TestClient,
-    pub entities: EntityMap,
-    pub fail_point_guards: Vec<FailPointGuard>,
+#[derive(Clone)]
+pub(crate) struct TestRunner {
+    pub(crate) internal_client: TestClient,
+    pub(crate) entities: Arc<RwLock<EntityMap>>,
```
patrickfreed, line 72: the entity map needed to be arc'd + locked so that it could be accessed from all the different test runner tasks. ditto for the failpoint guards.

src/test/spec/unified_runner/observer.rs
abr-egn: I'm having trouble following the distinction between this and `EventHandler` - can you explain more about what was needed that couldn't be done using that?

src/test/spec/unified_runner/test_file.rs
```diff
@@ -25,15 +26,14 @@ use crate::{
...
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
-pub struct TestFile {
-    pub description: String,
+pub(crate) struct TestFile {
+    pub(crate) description: String,
     #[serde(deserialize_with = "deserialize_schema_version")]
-    pub schema_version: Version,
-    pub run_on_requirements: Option<Vec<RunOnRequirement>>,
-    pub allow_multiple_mongoses: Option<bool>,
```
abr-egn, line 33: Was this dropped intentionally?

src/test/spec/unified_runner/test_runner.rs
```diff
@@ -60,58 +64,74 @@ const SKIPPED_OPERATIONS: &[&str] = &[
...
+        log_uncaptured(format!(
+            "\n------------\nRunning tests from {}\n",
+            file_title
+        ));
 
         for test_case in test_file.tests {
             if let Some(skip_reason) = test_case.skip_reason {
                 log_uncaptured(format!(
-                    "Skipping test case {}: {}",
+                    "Skipping test case \"{}\": {}",
```
abr-egn, line 134: Nit: when logging strings that need to be quoted, I tend to use `{:?}` to catch special characters as well.

abr-egn (commented):
Overall LG!

src/test/spec/unified_runner/test_runner.rs
```diff
@@ -60,58 +64,74 @@ const SKIPPED_OPERATIONS: &[&str] = &[
...
-            entities: HashMap::new(),
-            fail_point_guards: Vec::new(),
+            entities: Arc::new(RwLock::new(EntityMap::new())),
+            fail_point_guards: Arc::new(RwLock::new(Vec::new())),
         }
     }
 
-    pub async fn run_test(&mut self, test_file: TestFile, pred: impl Fn(&TestCase) -> bool) {
+    pub(crate) async fn run_test(
+        &self,
```
kmahar, line 98: it seems like we are always creating two test runners now? `run_unified_format_test_filtered` and the workload executor both create them and call this method on them, but then this method creates its own

src/test/spec/unified_runner/operation.rs
kmahar: nice solution. might be worth a comment explaining why we need it?

src/test/spec/unified_runner/test_runner.rs
```diff
@@ -462,62 +436,113 @@ impl TestRunner {
...
+                    let d = description.as_ref().to_string();
+                    let id = thread.id.clone();
+                    runtime::execute(async move {
+                        while let Some(msg) = receiver.recv().await {
+                            match msg {
+                                ThreadMessage::ExecuteOperation(op) => {
+                                    op.execute(runner.clone(), d.as_str()).await;
+                                }
+                                ThreadMessage::Stop(sender) => {
+                                    sender.send(Ok(())).unwrap_or_else(|_| {
```
kmahar, line 451: idk if I follow the error case here that we're panicking on here... if I understand correctly, we only receive `ThreadMessage::Stop` here once `wait()` has been called on a `ThreadEntity`, which looks to only occur via a `waitForThread` operation being executed. 

from the docs it looks `oneshot::Sender::send` only fails if the receiver has already been dropped. maybe this can happen if we for some reason kick off but don't finish awaiting the `waitForThread` operation, so `ThreadEntity.wait()` doesn't actually finish executing and the receiver is dropped? or how were you imagining we end up in this state?

src/test/spec/unified_runner/observer.rs
```diff
@@ -0,0 +1,103 @@
...
+            Ok(e) => {
+                self.seen_events.push(e.clone());
+                Some(e)
+            }
+            Err(TryRecvError::Lagged(_)) => panic!("event receiver lagged"),
+            Err(TryRecvError::Closed | TryRecvError::Empty) => None,
+        }
+    }
+
+    pub(crate) async fn event_count(
```
kmahar, line 57: nit: can we make the names here and for `wait_for_event` a little more descriptive like `matching_event_count` and `wait_for_matching_events` or something?

kmahar (commented):
looks good! a few minor comments and questions

patrickfreed:
In order to preserve the PR comment history, I've started using merge commits, hopefully the diff isn't too bad. Will still squash it all back to a single one at the end as usual though.

src/sdam/description/server.rs
isabelatkinson: at least for the workload executor it doesn't really matter what the format is. all of the serialized stuff just gets dumped into JSON logs

src/test/spec/unified_runner/observer.rs
abr-egn: Got it, thank you!  Can you leave a TODO in the code pointing to a new ticket so we don't lose track of this?

abr-egn (approved):
LGTM!  (modulo others' comments)

src/test/spec/unified_runner/observer.rs
patrickfreed: Filed RUST-1425, added comment.

isabelatkinson (approved):

src/sdam/public.rs
```diff
@@ -100,6 +101,11 @@ impl<'a> ServerInfo<'a> {
     pub fn tags(&self) -> Option<&TagSet> {
         self.command_response_getter(|r| r.tags.as_ref())
     }
+
+    /// Gets the error this server encountered, if any.
+    pub fn error(&self) -> Option<&Error> {
```
patrickfreed, line 106: Filed RUST-1432 for this change.

patrickfreed (commented):
After the spec PR went through review, we removed the requirement to make assertions on the error that a ServerDescription contains, since none of the existing tests do that. This means the changes related to having `ServerDescription` store a `mongodb::Result` instead of `std::Result<T, String>` aren't necessary, though I think given that they're already written we should keep them in. They'll make implementing RUST-1422 easier.

kmahar (approved):
lgtm mod a nit but don't need to re-review

src/sdam/public.rs
```diff
@@ -100,6 +101,11 @@ impl<'a> ServerInfo<'a> {
     pub fn tags(&self) -> Option<&TagSet> {
         self.command_response_getter(|r| r.tags.as_ref())
     }
+
+    /// Gets the error this server encountered, if any.
```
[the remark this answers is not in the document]
patrickfreed: Done

PR #714, opened 2022-07-29
RUST-1253 Add AWS Lambda Examples

isabelatkinson:
Adds examples for using the driver with AWS Lambda. The examples are based off of [this blog post](https://aws.amazon.com/blogs/opensource/rust-runtime-for-aws-lambda/) and [this example](https://github.com/pkdone/mongo-rust-lambda-demo) (thank you @pkdone!).

I've verified that these examples work by running them manually in Lambda.


src/test/lambda_examples.rs
```diff
@@ -0,0 +1,5 @@
+#![allow(dead_code)]
+
+#[cfg(feature = "aws-auth")]
+mod auth;
+mod no_auth;
```
isabelatkinson, line 5: These are split into two separate modules to avoid including conditional compilation statements within the examples. Although we've talked about containing our examples in the top-level `tests` directory, I needed to include these ones here instead because the `tests` directory does not follow the module structure and therefore does not permit conditional compilation of an entire file.

src/test/lambda_examples/auth.rs
```diff
@@ -0,0 +1,71 @@
...
+async fn main() -> Result<(), lambda_runtime::Error> {
+    let service = service_fn(handler);
+    lambda_runtime::run(service).await?;
+    Ok(())
+}
+// end lambda connection example 2
+
+// Runs a ping operation on the "db" database and returns the response.
+async fn handler_create_client(_: LambdaEvent<Value>) -> Result<Value, lambda_runtime::Error> {
+    let uri = std::env::var("MONGODB_URI").unwrap();
```
isabelatkinson, line 56: our URI for the AWS auth tests is configured a little differently than what is done for `MONGODB_CLIENT` so this uses the URI set for the tests directly.

src/test/lambda_examples/no_auth.rs
```diff
@@ -0,0 +1,51 @@
...
+    let service = service_fn(handler);
+    lambda_runtime::run(service).await?;
+    Ok(())
+}
+// end lambda connection example 1
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test(flavor = "multi_thread"))]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn test_handler() {
+    if std::env::var("MONGODB_API_VERSION").is_ok() {
```
isabelatkinson, line 46: A similar URI configuration issue -- it didn't seem super important to get versioned API coverage here so I just skipped the test.

abr-egn (approved):

kmahar (approved):

patrickfreed (approved):
LGTM!

PR #717, opened 2022-08-04
RUST-1384 Implement auto encryption

abr-egn:
RUST-1384

This implements support for auto encryption and decryption; the large majority of the complexity is in `state_machine.rs`; the changes needed to the main executor were pretty minor.

This is the companion PR to https://github.com/mongodb/libmongocrypt-rust/pull/11 (and its ill-fated predecessor https://github.com/mongodb/libmongocrypt-rust/pull/9).

### DIFF
--- a/src/client/csfle/state_machine.rs
+++ b/src/client/csfle/state_machine.rs
@@ -2,7 +2,7 @@ use std::convert::TryInto;
 use std::sync::Arc;
 
 use bson::{RawDocumentBuf, Document, RawDocument};
-use futures_util::TryStreamExt;
+use futures_util::{StreamExt, TryStreamExt, stream};
 use mongocrypt::ctx::{Ctx, State};
 
 use crate::{Client};
@@ -22,7 +22,28 @@ impl Client {
         };
         let mut result = None;
         loop {
-            match ctx.state()? {
+            let state = ctx.state()?;
+            match state {
+                State::NeedMongoCollinfo => {
+                    let filter = raw_to_doc(ctx.mongo_op()?)?;
+                    let db = self.database(db.as_ref().ok_or_else(|| Error::internal("db required for NeedMongoCollinfo state"))?);
+                    let mut cursor = db.list_collections(filter, None).await?;
+                    if cursor.advance().await? {
+                        ctx.mongo_feed(cursor.current())?;
+                    }
+                    ctx.mongo_done()?;
+                }
+                State::NeedKms => {
+                    let scope = ctx.kms_scope();
+                    let mut kms_ctxen: Vec<Result<_>> = vec![];
+                    while let Some(kms_ctx) = scope.next_kms_ctx()? {
+                        kms_ctxen.push(Ok(kms_ctx));
+                    }
+                    let f = stream::iter(kms_ctxen).try_for_each_concurrent(None, |kms_ctx| async move {
+                        Ok(())
+                    });
+                    f.await;
+                }
                 State::Ready => result = Some(ctx.finalize()?.to_owned()),
                 State::Done => break,
                 _ => todo!(),
--- a/src/client/csfle/state_machine.rs
+++ b/src/client/csfle/state_machine.rs
@@ -1,16 +1,28 @@
-use bson::RawDocumentBuf;
+use std::convert::TryInto;
+
+use bson::{RawDocumentBuf, Document, RawDocument};
 use mongocrypt::ctx::{Ctx, State};
 
-use crate::Client;
-use crate::error::Result;
+use crate::{Client, Database};
+use crate::error::{Error, Result};
+use crate::operation::{ListCollections, RawOutput};
+
+fn raw_to_doc(raw: &RawDocument) -> Result<Document> {
+    raw.try_into().map_err(|e| Error::internal("???"))
+}
 
 impl Client {
-    pub(crate) async fn run_mongocrypt_ctx(&self, ctx: &mut Ctx) -> Result<RawDocumentBuf> {
+    pub(crate) async fn run_mongocrypt_ctx(&self, ctx: &mut Ctx, db: Option<Database>) -> Result<RawDocumentBuf> {
         let mut result = RawDocumentBuf::new();
         loop {
             match ctx.state()? {
                 State::NeedMongoCollinfo => {
-                    let filter = ctx.mongo_op()?;
+                    let filter: Document = raw_to_doc(ctx.mongo_op()?)?;
+                    let db = db.as_ref().ok_or_else(|| Error::internal("db required for NeedMongoCollinfo state"))?;
+                    let mut cursor = db.list_collections(filter, None).await?;
+                    if cursor.advance().await? {
+                        ctx.mongo_feed(cursor.current())?;
+                    }
                 }
                 _ => todo!(),
             }
--- a/src/client/csfle/state_machine.rs
+++ b/src/client/csfle/state_machine.rs
@@ -4,6 +4,7 @@ use std::sync::mpsc as sync_mpsc;
 use std::thread;
 
 use bson::{RawDocumentBuf, Document, RawDocument};
+use futures_util::TryStreamExt;
 use mongocrypt::ctx::{Ctx, State, CtxBuilder};
 use tokio::sync::mpsc::UnboundedSender;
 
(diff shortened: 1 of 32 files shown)


src/operation/mod.rs
```diff
@@ -67,12 +68,17 @@ pub(crate) use insert::Insert;
...
 pub(crate) use run_command::RunCommand;
 pub(crate) use update::Update;
 
 const SERVER_4_2_0_WIRE_VERSION: i32 = 8;
 
 /// A trait modeling the behavior of a server side operation.
+///
+/// No methods in this trait should have default behaviors to ensure that wrapper operations
+/// replicate all behavior.  Default behavior is provided by the `OperationDefault` trait.
 pub(crate) trait Operation {
```
abr-egn, line 82: The `Trait` / `TraitWithDefaults` pattern introduced here solves a gripe I've had for a while, which is that when implementing something that delegates most trait logic (e.g. `raw_output::RawOutput`) you won't get any help from the compiler if new trait methods are added that have defaults, which in turn could cause nasty logic bugs downstream.  It's definitely a little awkward and not work using for most traits, but `Operation` is critical enough that it seems worth using here.

kmahar (commented):
looks good overall! just a couple comments re: spec details

src/operation/insert/mod.rs
```diff
@@ -49,7 +54,7 @@ impl<'a, T> Insert<'a, T> {
     }
 }
 
-impl<'a, T: Serialize> Operation for Insert<'a, T> {
+impl<'a, T: Serialize> OperationWithDefaults for Insert<'a, T> {
```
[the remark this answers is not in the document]
abr-egn: Ah, you're right - I'd initially read that as applying only to the bulk write API, but it also applies here.  Updated.

src/client/csfle/state_machine.rs
```diff
@@ -0,0 +1,155 @@
...
+        loop {
+            let state = result_ref(&ctx)?.state()?;
+            match state {
+                State::NeedMongoCollinfo => {
+                    let ctx = result_mut(&mut ctx)?;
+                    let filter = raw_to_doc(ctx.mongo_op()?)?;
+                    let db = self.database(db.as_ref().ok_or_else(|| {
+                        Error::internal("db required for NeedMongoCollinfo state")
+                    })?);
+                    let mut cursor = db.list_collections(filter, None).await?;
```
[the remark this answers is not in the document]
abr-egn: Good catch, fixed.  I'll send a PR for the integration guide.

src/operation/insert/mod.rs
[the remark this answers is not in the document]
kmahar: I filed RUST-1438 so we remember to do the same for bulk writes when the time comes

kmahar (approved):

src/client/csfle/state_machine.rs
```diff
@@ -0,0 +1,171 @@
...
+                    let kv_client = csfle
+                        .aux_clients
+                        .key_vault_client
+                        .upgrade()
+                        .ok_or_else(|| Error::internal("key vault client dropped"))?;
+                    let kv_coll = kv_client
+                        .database(&kv_ns.db)
+                        .collection::<RawDocumentBuf>(&kv_ns.coll);
+                    let mut cursor = kv_coll.find(filter, None).await?;
+                    while let Some(result) = cursor.try_next().await? {
```
patrickfreed, line 87: It's a little bit faster to use the `cursor.advance` and `cursor.current` API here to avoid having to trip through `serde`.

src/client/csfle/state_machine.rs
```diff
@@ -0,0 +1,160 @@
...
+                    ctx.mongo_done()?;
+                }
+                State::NeedKms => {
+                    let ctx = result_mut(&mut ctx)?;
+                    let scope = ctx.kms_scope();
+                    let mut kms_ctxen: Vec<Result<_>> = vec![];
+                    while let Some(kms_ctx) = scope.next_kms_ctx() {
+                        kms_ctxen.push(Ok(kms_ctx));
+                    }
+                    stream::iter(kms_ctxen)
```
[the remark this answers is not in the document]
patrickfreed: I think the stream versoin works basically the same as `FuturesUnordered`: the current thread polls all the futures, rather than having potentially many threads doing it in parallel.

src/client/csfle/state_machine.rs
```diff
@@ -0,0 +1,171 @@
...
+                        &mut ctx,
+                        Err(Error::internal("crypto context not present")),
+                    )?;
+                    csfle.crypto_threads.spawn(move || {
+                        let result = thread_ctx.finalize().map(|doc| doc.to_owned());
+                        let _ = tx.send((thread_ctx, result));
+                    });
+                    let (ctx_again, output) = rx
+                        .await
+                        .map_err(|_| Error::internal("crypto thread dropped"))?;
```
patrickfreed, line 145: what are the consequences of this happening? also are you aware of any ways this could reasonably occur?

src/client/executor.rs
```diff
@@ -765,6 +799,34 @@ impl Client {
...
+        csfle: &'a super::csfle::ClientState,
+        command: &'a RawDocument,
+        target_db: &'a str,
+    ) -> BoxFuture<'a, Result<RawDocumentBuf>> {
+        Box::pin(async move {
+            let ctx = csfle
+                .crypt
+                .ctx_builder()
+                .build_encrypt(target_db, command)?;
+            self.run_mongocrypt_ctx(ctx, Some(target_db)).await
```
patrickfreed, line 814: WDYT about passing the `&ClientState` down into this method too? That way we won't need to acquire the lock again, and it also makes it explicit that this function is called when the read lock is held and can therefore not acquire the write lock? Ditto for `auto_decrypt`.

src/operation/raw_output.rs
```diff
@@ -0,0 +1,77 @@
+use crate::{
+    cmap::{Command, RawCommandResponse, StreamDescription},
+    error::Result,
+};
+
+use super::Operation;
+
+pub(crate) struct RawOutput<Op>(pub(crate) Op);
```
patrickfreed, line 8: Can we add a docstring indicating this type's purpose? If I'm understanding it right, it forwards all implementation to the underlying op except that it performs no validation or parsing of the response, is that right?

src/client/csfle/state_machine.rs
```diff
@@ -0,0 +1,171 @@
...
+            let state = result_ref(&ctx)?.state()?;
+            match state {
+                State::NeedMongoCollinfo => {
+                    let ctx = result_mut(&mut ctx)?;
+                    let filter = raw_to_doc(ctx.mongo_op()?)?;
+                    let metadata_client = csfle
+                        .aux_clients
+                        .metadata_client
+                        .as_ref()
+                        .and_then(|w| w.upgrade())
```
patrickfreed, line 46: I know this was a prior PR, but where do the strong references for these live again? Is the user in charge of ensuring they're alive?

patrickfreed (commented):
Looks good! I just have some clarifying questions

src/operation/raw_output.rs
abr-egn: Yup, that's correct - added a comment.

patrickfreed (approved):
LGTM! (bar a commit from a previous comment, no need for me to re-review)

patrickfreed:
Oh and it looks like the clippy task failed.

src/operation/raw_output.rs
abr-egn: Whoops, fixed.

abr-egn:
> Oh and it looks like the clippy task failed.

Looks like those failures are in `src/gridfs.rs`, unrelated to this PR (@sanav33).

src/operation/mod.rs
```diff
@@ -398,3 +391,140 @@ macro_rules! remove_empty_write_concern {
 }
 
 pub(crate) use remove_empty_write_concern;
+
+// A mirror of the `Operation` trait, with default behavior where appropriate.  Should only be
+// implemented by leaf operation types.
```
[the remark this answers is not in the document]
abr-egn: Derp, my inner terminology leaking out.  Updated the comment to clarify.

src/operation/run_command/test.rs
```diff
@@ -20,7 +22,8 @@ fn build() {
         command
             .body
             .get("hello")
-            .and_then(crate::bson_util::get_int),
+            .unwrap()
```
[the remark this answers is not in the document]
abr-egn: Good call, done.  Sorry that I keep forgetting that!

isabelatkinson (approved):
looks like the format task failed but lgtm otherwise!

PR #721, opened 2022-08-11
RUST-360 Streaming monitoring protocol

patrickfreed:
RUST-360

This PR updates the driver to make use of the streaming monitoring protocol (described [here](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#streaming-protocol)), which allows the driver to accept topology updates that are pushed to it rather than polling the server regularly. This should allow the driver to more quickly recover from failover events.


src/cmap/establish/handshake/mod.rs
```diff
@@ -214,17 +211,12 @@ impl Handshaker {
     }
 
     /// Handshakes a connection.
-    pub(crate) async fn handshake(
-        &self,
-        conn: &mut Connection,
-        topology: Option<&Topology>,
-        handler: &Option<Arc<dyn SdamEventHandler>>,
```
patrickfreed, line 221: The monitors now handle emitting these events, so we don't need to pipe these event handlers down into the handshaker anymore.

src/cmap/manager.rs
```diff
@@ -38,17 +38,11 @@ impl PoolManager {
...
-            .sender
-            .send(PoolManagementRequest::MarkAsReady {
-                _completion_handler: message,
-            })
-            .is_ok()
-        {
-            acknowledgment_receiver.wait_for_acknowledgment().await;
-        }
+    ///
+    /// Since management requests are treated with the highest priority by the pool and will be
```
patrickfreed, line 42: I updated the pool worker task to use a biased `select!`, which means that it will poll in a certain order. Specifically for this, I updated it to check for management requests (clearing, marking as ready, etc) before checking for check-out requests. This means that we don't actually have to wait for acknowledgment that the pool is ready before returning here, since we know that no check-out requests will be processed before this `MarkAsReady` is.

src/hello.rs
```diff
@@ -53,91 +51,36 @@ pub(crate) fn hello_command(
...
-    let end_time = Instant::now();
-
-    let round_trip_time = end_time.duration_since(start_time);
-
-    match response_result.and_then(|raw_response| {
-        let hello_reply = raw_response.into_hello_reply(round_trip_time)?;
-        Ok(hello_reply)
-    }) {
-        Ok(hello_reply) => {
-            emit_event(topology, handler, |handler| {
```
patrickfreed, line 92: this logic was moved into the monitor.

src/lib.rs
```diff
@@ -297,7 +297,8 @@
     allow(
         clippy::unreadable_literal,
         clippy::cognitive_complexity,
-        clippy::float_cmp
+        clippy::float_cmp,
+        clippy::match_like_matches_macro
```
patrickfreed, line 301: this macro forces you to write something like
```rust
match a {
    SomeEnum::B => false,
    _ => true
}
```
as 
```rust
!matches!(a, SomeEnum::B)
```
which I personally find less readable (super easy to miss that first `!`). What do you all think?

src/runtime/worker_handle.rs
```diff
@@ -18,24 +18,31 @@ impl WorkerHandle {
 /// Listener used to determine when all handles have been dropped.
 #[derive(Debug)]
 pub(crate) struct WorkerHandleListener {
-    receiver: mpsc::Receiver<()>,
+    sender: watch::Sender<()>,
```
patrickfreed, line 21: This was rewritten to use a `watch` channel which is more appropriate for this type and also allowed us to use `&self` everywhere instead of `&mut self`. The functionality is otherwise unchanged.

src/sdam/description/server.rs
```diff
@@ -82,6 +83,30 @@ impl Default for ServerType {
...
+    pub(crate) counter: i64,
+}
+
+impl TopologyVersion {
+    pub(crate) fn is_more_recent_than(&self, existing_tv: TopologyVersion) -> bool {
+        self.process_id != existing_tv.process_id || self.counter > existing_tv.counter
+    }
+}
+
+impl From<TopologyVersion> for Bson {
```
patrickfreed, line 101: This is used instead of `serde` so we can convert to BSON without having to handle errors (this is called in `hello_command`, which currently is infallible).

src/sdam/description/topology/mod.rs
```diff
@@ -305,20 +304,6 @@ impl TopologyDescription {
         self.compatibility_error.as_ref()
     }
 
-    /// Update the ServerDescription's round trip time based on the rolling average.
-    fn update_round_trip_time(&self, server_description: &mut ServerDescription) {
```
patrickfreed, line 309: The RTT monitor handles this now

src/sdam/description/topology/mod.rs
```diff
@@ -462,14 +464,48 @@ impl TopologyDescription {
...
+                        }
+                    }
+                }
+            }
         }
 
-        // Update the round trip time on the server description to the weighted average as described
-        // by the spec.
-        self.update_round_trip_time(&mut server_description);
+        if let Some(expected_name) = &self.set_name {
```
patrickfreed, line 483: This used to be checked in the Topology worker task, but it makes more sense to do it at the `TopologyDescription` level.

src/sdam/monitor.rs
```diff
@@ -67,109 +98,180 @@ impl Monitor {
...
-                .await?;
-
-                let start = Instant::now();
-                let res = self
-                    .handshaker
-                    .handshake(
-                        &mut connection,
-                        None,
-                        &self.client_options.sdam_event_handler,
+        let timeout = if self.connect_timeout().as_millis() == 0 {
```
patrickfreed, line 183: The way the tests are implemented, this timeout is supposed to apply to the entire handshake, including connection establishment, authentication, etc. Previously, we only applied it to establishment.

src/sdam/topology.rs
```diff
@@ -53,146 +52,56 @@ use super::{
...
             runtime::execute(async move {
                 while let Some(event) = rx.recv().await {
                     let (event, ack) = event.into_parts();
-
-                    let is_closed = matches!(event, SdamEvent::TopologyClosed(_));
                     handle_sdam_event(handler.as_ref(), event);
-
                     ack.acknowledge(());
-
-                    // no more events should be emitted after a TopologyClosedEvent
```
patrickfreed, line 80: This was kind of a hack to get this to behave right. Instead, I updated the worker to wait until all monitors have shut down before emitting the `TopologyClosedEvent`. This helps catch the bug I mentioned at standup where monitors weren't ever exiting.

src/sdam/topology.rs
```diff
@@ -53,146 +52,56 @@ use super::{
...
-                        break;
-                    };
                 }
             });
             SdamEventEmitter { sender: tx }
         });
 
         let (updater, update_receiver) = TopologyUpdater::channel();
         let (worker_handle, handle_listener) = WorkerHandleListener::channel();
-
```
patrickfreed, line 91: We used to do a lot of ad-hoc initialization to match the events and outcomes that were expected in the SDAM tests, but it was pretty hard to follow / maintain. I've updated this to use the standard topology update process that all other updates go through, which cuts down on a lot of this hard-coded stuff.

src/sdam/topology.rs
```diff
@@ -415,173 +393,139 @@ impl TopologyWorker {
...
-                    self.event_emitter.clone(),
-                    self.update_requester.subscribe(),
-                    self.options.clone(),
-                );
-            }
-        }
+    fn sync_hosts(&mut self, hosts: HashSet<ServerAddress>) -> bool {
+        let mut new_description = self.topology_description.clone();
+        new_description.sync_hosts(&hosts);
+        self.update_topology(new_description)
```
patrickfreed, line 417: Everything that modifies the topology now goes through the following steps:
1. clone the existing topology description
2. modify it in some way
3. pass it to `update_topology`

This cuts down on the one-off logic in these various methods and hopefully makes the worker easier to read / reason about.

src/test/spec/unified_runner/test_file.rs
```diff
@@ -94,7 +94,13 @@ impl RunOnRequirement {
             }
         }
         if let Some(ref topologies) = self.topologies {
-            if !topologies.contains(&client.topology().await) {
+            let client_topology = client.topology().await;
+            if !topologies.iter().any(|expected_topology| {
```
patrickfreed, line 98: Per the spec, a "sharded" runOnRequirement should match both "sharded" and "sharded-replicaset" topologies. Since we were using direct equality, we may have been skipping some tests accidentally.

src/test/util/event.rs
```diff
@@ -505,9 +505,12 @@ impl EventSubscriber<'_> {
         F: FnMut(&Event) -> bool,
     {
         let mut events = Vec::new();
-        while let Some(event) = self.wait_for_event(timeout, &mut filter).await {
```
patrickfreed, line 508: This changes how this method works at a fundamental level, but I think in a way that makes much more sense. Before, it would collect events until an event was not received for the given timeout, so if there was a constant stream of events it would go on forever. The new version collects events for the given time and returns what it observed.

src/sdam/topology.rs
```diff
@@ -319,7 +228,21 @@ impl Topology {
 #[derive(Debug, Clone)]
 pub(crate) struct TopologyState {
     pub(crate) description: TopologyDescription,
-    pub(crate) servers: HashMap<ServerAddress, Arc<Server>>,
+    servers: HashMap<ServerAddress, Weak<Server>>,
```
patrickfreed, line 231: Storing `Arc<Server>` here actually clones them for every receiver (aka `TopologyWatcher`), which led to a reference cycle wherein the `Server`s were never actually getting dropped because the monitors held watchers, but the monitors never exited because the `Server`s never dropped. Storing `Weak` pointers here fixes this.

One future area to explore would be to ditch `Arc` / `Weak` here altogether and just pass clones of `Server` around, but I didn't want to go too deep down the refactor rabbit hole.

2.3.0 (which has the Topology refactor) is not affected by this issue, but that's because the monitors exit once they detect the topology is dropped, which then causes the `Server`s to drop. They have no need for the `Server` to be alive in order to be dropped. This does highlight a separate issue, which is that monitors on 2.3.0 (and probably earlier) don't exit if a server is removed from the Topology. This is fixed now, however.

src/sdam/topology.rs
```diff
@@ -1,7 +1,7 @@
 use std::{
```
patrickfreed, line 1: This file has gotten kind of long and complicated, so once this PR is merged I think I'll go in and break it down into a few smaller ones. For the purposes of preserving the diff I didn't do that in this PR.

src/cmap/establish/test.rs
```diff
@@ -1,179 +0,0 @@
-use tokio::sync::RwLockWriteGuard;
```
patrickfreed, line 1: This file used to contain unit tests of the speculative authentication behavior, but I had to remove them now that authentication was moved into the `Handshaker` layer, and I couldn't really think of a way to test this without introducing specific logic to make unit testing possible. These being unit tests, they don't use the actual driver and so don't exercise the same code paths a real client would, so it didn't seem like they were providing enough value to be worth preserving. If we do want to unit test this behavior, we can introduce some new functionality to connections for tracking which messages were sent and what not, but I figured we could defer that to later. What do you all think?

src/cmap/conn/mod.rs
```diff
@@ -490,49 +529,63 @@ impl PinnedConnectionHandle {
...
+/// monitoring case and the Option<> wrapper can be dropped from this.
+#[derive(Debug, Clone, Copy)]
 pub(crate) enum ConnectionGeneration {
+    Monitoring,
     Normal(u32),
-    LoadBalanced {
-        generation: u32,
-        service_id: ObjectId,
-    },
+    LoadBalanced(Option<LoadBalancedGeneration>),
```
patrickfreed, line 544: This was updated to include an `Option` to model the situation where we have a load balanced connection that has not yet received its service id. In RUST-1454 we can make this more robust and remove the need for this per the comment above.

src/cmap/establish/handshake/mod.rs
```diff
@@ -116,19 +112,7 @@ lazy_static! {
                 version: None,
             },
             platform: format!("{} with {}", rustc_version_runtime::version_meta().short_version_string, RUNTIME_NAME),
-        };
-
-        let info = os_info::get();
```
patrickfreed, line 121: This actually performs an `lsb_release` command (if present) and if that fails does some other stuff, so it can take a somewhat significant amount of time to run (~50ms or so in my testing). Given that we only currently use it for the distro version but don't report the distro itself, it seemed worth it to drop for now unti lwe do RUST-1285. (I did this while investigating those timeouts that ultimately ended up being TLS related, but this contributed to it too).

@kmahar are we currently using this version field in a significant way that would inhibit removing it for now?

patrickfreed (commented):
The changes in this PR are kinda all over the place, so here's a high level summary of the major changes in this PR that might serve as good starting points for review:
- Monitors now can read from a socket repeatedly rather than issue a new hello and read the responce 
- The RTT for each server is now measured by a separate RTT monitor
- Monitor checks can now be cancelled by the topology. Also the way monitor checks are requested was tweaked slightly
- Monitors and connection pools now create connections the same way (via `ConnectionEstablisher`)
- Updates to the topology are now processed uniformly / more simply by the topology worker

kmahar (commented):
so far this looks good! mostly pretty minor comments and questions. 

I will say that I don't have a great mental model of how all of the components in the driver interact so I think the parts of the code that involve coordination/communication between different components have been difficult for me to review and provide meaningful input on, especially in cases where things have been refactored. 

I'm going to try to spend some time reasoning through how it all fits together before doing another round of review, but in the meantime figured I would go ahead and give these comments.

src/test/client.rs
```diff
@@ -731,6 +731,11 @@ async fn retry_commit_txn_check_out() {
         return;
     }
 
+    if setup_client.server_version_gte(4, 4) {
```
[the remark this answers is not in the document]
patrickfreed: Ah yeah, I hadn't considered that. Definitely a good idea, and really easy to implement on `TestClient` as well. Updated to use that.

src/test/client.rs
```diff
@@ -790,6 +795,10 @@ async fn retry_commit_txn_check_out() {
         .await
         .expect("should see marked unknown event");
 
+    // If this test were run when using the streaming protocol, this assertion would never succeed.
```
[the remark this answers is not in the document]
patrickfreed: Updated, lmk if its clearer now.

src/test/spec/sdam.rs
```diff
@@ -0,0 +1,232 @@
...
+        .collect_events(Duration::from_secs(2), |e| {
+            if let Event::Sdam(SdamEvent::ServerDescriptionChanged(e)) = e {
+                assert!(
+                    e.new_description.average_round_trip_time().unwrap() > Duration::from_millis(0)
+                );
+            };
+            true
+        })
+        .await;
+    assert!(!events.is_empty());
```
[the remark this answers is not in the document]
patrickfreed: Updated

src/test/spec/unified_runner/observer.rs
```diff
@@ -55,19 +55,20 @@ impl EventObserver {
...
+    ) -> Vec<Event> {
         // first retrieve all the events buffered in the channel
         while self.try_recv().is_some() {}
         let es = entities.read().await;
         // then count
         self.seen_events
             .iter()
             .filter(|e| events_match(e, event, Some(&es)).is_ok())
-            .count()
+            .cloned()
```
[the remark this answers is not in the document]
patrickfreed: good catch, fixed.

src/cmap/establish/test.rs
patrickfreed: Not specific to speculative authentication, but our auth test suite should exercise the code paths. We don't have any actual verification that they do, though.

src/test/spec/retryable_reads.rs
```diff
@@ -65,7 +65,7 @@ async fn retry_releases_connection() {
         .collection("retry_releases_connection");
     collection.insert_one(doc! { "x": 1 }, None).await.unwrap();
 
-    let options = FailCommandOptions::builder().error_code(91).build();
+    let options = FailCommandOptions::builder().close_connection(true).build();
```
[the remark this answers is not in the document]
patrickfreed: done

src/test/spec/retryable_writes/mod.rs
```diff
@@ -45,19 +46,20 @@ async fn run_unified() {
...
             }
-            let hosts = &CLIENT_OPTIONS.get().await.hosts;
-            let options = test_case.client_options.map(|mut opts| {
-                opts.hosts = hosts.clone();
-                opts
-            });
+            let mut options = test_case.client_options.unwrap_or_default();
+            options.hosts = CLIENT_OPTIONS.get().await.hosts.clone();
+            if options.heartbeat_freq.is_none() {
+                options.heartbeat_freq = Some(MIN_HEARTBEAT_FREQUENCY);
```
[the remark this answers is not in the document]
patrickfreed: So this is actually the same problem that made us disable that other test: only the operation receives the error via failpoint. As far as the monitor that is streaming is concerned, nothing has happened, and so it continues to wait for the server to push a new response (heartbeatFrequencyMS / maxAwaitTimeMS) all the while ignoring check requests. Setting this really low tells the server to push updates more often so that the monitor can more quickly rediscover the server. In actual scenarios, if an operation is seeing a state change error, the monitor will likely have already seen it.

src/test/spec/retryable_writes/mod.rs
```diff
@@ -85,17 +87,23 @@ async fn run_legacy() {
...
             }
 
-            if let Some(ref mut fail_point) = test_case.fail_point {
-                client
-                    .database("admin")
-                    .run_command(fail_point.clone(), None)
-                    .await
-                    .unwrap();
-            }
+            let _fp_guard = if let Some(ref mut fail_point) = test_case.fail_point {
```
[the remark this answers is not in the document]
patrickfreed: Yep, the latter. If a test failed, it would panic before reaching the disable fail point logic and cause a bunch of other tests to fail too.

patrickfreed:
>I will say that I don't have a great mental model of how all of the components in the driver interact so I think the parts of the code that involve coordination/communication between different components have been difficult for me to review and provide meaningful input on, especially in cases where things have been refactored.
>
>I'm going to try to spend some time reasoning through how it all fits together before doing another round of review, but in the meantime figured I would go ahead and give these comments.

Sounds good. And yeah, the various refactors + cross component nature is part of what made this difficult from the implementation side too. Would be happy to do a walkthrough of some of the changes over zoom if that would be helpful to reviewers. If not, no worries too.

src/runtime/worker_handle.rs
kmahar: ahh that all makes sense. thanks!

src/cmap/conn/mod.rs
kmahar: oh right, I found that code later on but forgot to come back and get rid of this comment :) thanks

src/hello.rs
```diff
@@ -53,91 +51,32 @@ pub(crate) fn hello_command(
...
+    if let Some(opts) = awaitable_options {
+        command.insert("topologyVersion", opts.topology_version);
+        command.insert("maxAwaitTimeMS", opts.max_await_time.as_millis() as i64);
+    }
+
     let mut command = Command::new(command_name.into(), "admin".into(), command);
     if let Some(server_api) = server_api {
         command.set_server_api(server_api);
     }
+    command.exhaust_allowed = awaitable_options.is_some();
```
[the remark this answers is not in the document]
kmahar: > Note how command is reassigned from a Document to a Command above

whoops, missed that! 

a builder API seems nice 👍

src/sdam/monitor.rs
```diff
@@ -208,12 +326,330 @@ impl Monitor {
...
+            };
+
+            if check_succeded {
+                self.sender
+                    .send_modify(|rtt_info| rtt_info.add_sample(start.elapsed()));
+            } else {
+                // From the SDAM spec: "Errors encountered when running a hello or legacy hello
+                // command MUST NOT update the topology."
+                self.connection.take();
+
```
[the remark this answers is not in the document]
kmahar: ohhhh that makes more sense. thanks

src/cmap/worker.rs
```diff
@@ -254,15 +247,21 @@ impl ConnectionPoolWorker {
 
         loop {
             let task = tokio::select! {
-                Some(request) = self.request_receiver.recv() => {
-                    PoolTask::CheckOut(request)
-                },
+                // This marker indicates that the futures will be polled in order from top to
```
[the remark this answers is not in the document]
kmahar: thanks! that all makes sense.

src/runtime/tls_openssl.rs
```diff
@@ -23,11 +23,38 @@ pub(crate) struct AsyncTlsStream {
...
+pub(crate) struct TlsConfig {
+    connector: SslConnector,
+    verify_hostname: bool,
+}
+
+impl TlsConfig {
+    /// Create a new `TlsConfig` from the provided options from the user.
+    /// This operation is expensive, so the resultant `TlsConfig` should be cached.
+    pub(crate) fn new(options: TlsOptions) -> Result<TlsConfig> {
+        let verify_hostname = options.allow_invalid_hostnames.unwrap_or(true);
```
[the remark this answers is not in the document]
patrickfreed: Oh good catch, yeah it is. Updated to use a `match` to make the logic clearer here. As a side note, I'm a little concerned this didn't trigger any test failures. Filed RUST-1467 for introducing test coverage for this.

src/runtime/stream.rs
```diff
@@ -33,6 +32,26 @@ pub(crate) enum AsyncStream {
...
 }
 
+impl AsyncStream {
+    pub(crate) async fn connect(
+        address: ServerAddress,
+        tls_cfg: Option<&TlsConfig>,
+    ) -> Result<Self> {
+        let inner = AsyncTcpStream::connect(&address).await?;
+
+        // If there are TLS options, wrap the inner stream with rustls.
```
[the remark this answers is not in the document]
patrickfreed: fixed

kmahar (approved):
LGTM! great job with this 👏

PR #724, opened 2022-08-15
RUST-1295 Sync tests for pre and post image options in collection helpers

abr-egn:
RUST-1295

This is mostly just a test sync; however, one of the test files is exercising the `modifyCollection` helper, which the Rust driver doesn't have, so it needed a code change to skip that.  I couldn't find a bug to implement that, so I left it with a non-TODO comment.

### DIFF
--- a/src/test/spec/collection_management.rs
+++ b/src/test/spec/collection_management.rs
@@ -1,4 +1,4 @@
-use crate::test::LOCK;
+use crate::test::{log_uncaptured, LOCK};
 
 use super::{run_spec_test_with_path, run_unified_format_test};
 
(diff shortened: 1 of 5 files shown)


kmahar (approved):
LGTM!

> I couldn't find a bug to implement that, so I left it with a non-TODO comment.

AFAIK only the PHP driver actually has a helper for the collMod command; since it's more of an administrative command than one an app would run I think we're unlikely to ever need to add a helper. so no ticket seems fine.

isabelatkinson (approved):

patrickfreed (approved):

PR #726, opened 2022-08-17
RUST-1385 Implement explicit encryption

abr-egn:
RUST-1385

This implements the `ClientEncryption` struct which provides the explicit encryption API.  This required a bit of refactoring to allow a `Ctx` to be executed independently of a `Client`, but is otherwise mostly just data plumbing; the API is just a wrapper around CRUD operations plus `Ctx` operations.

### DIFF
--- a/src/client/csfle/client_encryption.rs
+++ b/src/client/csfle/client_encryption.rs
@@ -1,11 +1,14 @@
 #![allow(dead_code)]
 
+use crate::coll::options::CollectionOptions;
+use crate::options::{WriteConcern, ReadConcern};
 use crate::results::DeleteResult;
-use crate::{Cursor, Client, Namespace};
+use crate::{Cursor, Client, Namespace, Collection};
 use crate::bson::{Binary, Document};
-use crate::error::Result;
+use crate::error::{Result, Error};
+use bson::{doc, RawDocumentBuf};
 use mongocrypt::Crypt;
-use mongocrypt::ctx::KmsProvider;
+use mongocrypt::ctx::{KmsProvider, Ctx};
 
 use super::options::{KmsProviders, KmsProvidersTlsOptions};
 use super::state_machine::CryptExecutor;
@@ -14,6 +17,7 @@ pub struct ClientEncryption {
     crypt: Crypt,
     exec: CryptExecutor,
     opts: ClientEncryptionOptions,
+    key_vault: Collection<RawDocumentBuf>,
 }
 
 impl ClientEncryption {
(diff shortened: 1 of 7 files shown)


src/client/csfle/client_encryption.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// Removes the key document with the given UUID (BSON binary subtype 0x04) from the key vault
+    /// collection. Returns the result of the internal deleteOne() operation on the key vault
+    /// collection.
+    pub async fn delete_key(&self, id: &Binary) -> Result<DeleteResult> {
+        self.key_vault.delete_one(doc! { "_id": id }, None).await
+    }
+
+    /// Finds a single key document with the given UUID (BSON binary subtype 0x04).
+    /// Returns the result of the internal find() operation on the key vault collection.
+    pub async fn get_key(&self, id: &Binary) -> Result<Option<RawDocumentBuf>> {
```
abr-egn, line 108: This API uses raw types because it seemed likely that a user performing encryption/decryption would want to avoid the additional overhead of parsing.  This was a handwavy judgement call that I'd be happy to revisit if people feel strongly the other way.

src/client/csfle/client_encryption.rs
```diff
@@ -0,0 +1,294 @@
...
+        }
+        if let Some(qtype) = opts.query_type {
+            builder = builder.query_type(&qtype)?;
+        }
+        Ok(builder.build_explicit_encrypt(value)?)
+    }
+
+    /// Decrypts an encrypted value (BSON binary of subtype 6).
+    /// Returns the original BSON value.
+    pub async fn decrypt<'a>(&self, value: RawBinaryRef<'a>) -> Result<bson::RawBson> {
```
abr-egn, line 204: This could just accept `&[u8]` rather than `RawBinaryRef`, but the subtype check seems like a useful guardrail.  `Binary::as_raw_binary` from https://github.com/mongodb/bson-rust/pull/370 will make this reasonably convenient to use with non-raw data as well.

src/concern/mod.rs
```diff
@@ -29,6 +29,29 @@ pub struct ReadConcern {
     pub level: ReadConcernLevel,
 }
 
+impl ReadConcern {
```
abr-egn, line 32: I added these utility constants (and the equivalent for `WriteConcern`) because it's _way_ more readable to construct a collection that uses majority read and write concern using them than having to nest construction of the concern structs (e.g. in `ClientEncryption::new`).

src/gridfs.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(dead_code, unused_variables)]
```
abr-egn, line 1: This is unrelated, but removes the noise from `cargo check` and evergreen runs :)

isabelatkinson: can you add a TODO RUST-1395 comment so we make sure to remove these?

src/concern/mod.rs
isabelatkinson: nice, yeah the construction of these can be pretty verbose

src/client/csfle/state_machine.rs
```diff
@@ -3,31 +3,58 @@ use std::convert::TryInto;
...
-            Some(csfle) => csfle,
-            None => return Err(Error::internal("no csfle state for mongocrypt ctx")),
-        };
+use super::options::KmsProvidersTlsOptions;
+
+#[derive(Debug)]
+pub(crate) struct CryptExecutor {
+    key_vault_client: WeakClient,
+    key_vault_namespace: Namespace,
+    mongocryptd_client: Option<Client>,
```
isabelatkinson, line 28: why is this one a non-weak client whereas the others are weak?

src/client/csfle/client_encryption.rs
```diff
@@ -0,0 +1,294 @@
...
+        Ok(result
+            .get("v")?
+            .ok_or_else(|| Error::internal("invalid decryption result"))?
+            .to_raw_bson())
+    }
+}
+
+/// Options for initializing a new `ClientEncryption`.
+#[non_exhaustive]
+pub struct ClientEncryptionOptions {
```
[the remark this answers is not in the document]
isabelatkinson: oh i think we should also add the defaults:
`#[builder(field_defaults(setter(into)))]`

isabelatkinson (approved):
lgtm, just one more small comment (but no need for re-review)

kmahar (approved):
LGTM mod typo

src/client/csfle/client_encryption.rs
```diff
@@ -0,0 +1,354 @@
...
+                }
+            }
+        };
+        self.key_vault
+            .find_one_and_update(doc! { "_id": id }, vec![update], None)
+            .await
+    }
+
+    /// Returns a key document in the key vault collection with the given keyAltName.
+    pub async fn get_key_by_alt_name(&self, key_alt_name: &str) -> Result<Option<RawDocumentBuf>> {
```
patrickfreed, line 168: This `&str` can be `impl AsRef<str>`

src/client/csfle/client_encryption.rs
```diff
@@ -0,0 +1,354 @@
...
+                builder = builder.key_alt_name(&name)?;
+            }
+        }
+        if let Some(material) = opts.key_material {
+            builder = builder.key_material(&material)?;
+        }
+        Ok(builder.build_datakey()?)
+    }
+
+    // pub async fn rewrap_many_data_key(&self, _filter: Document, _opts: RewrapManyDataKeyOptions)
```
patrickfreed, line 98: Are these commented sections here and below intentionally left in?

src/client/csfle/client_encryption.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// The contention factor.
+    #[builder(default)]
+    pub contention_factor: Option<i64>,
+    /// The query type.
+    #[builder(default)]
+    pub query_type: Option<String>,
+}
+
+/// An encryption key reference.
+pub enum EncryptKey {
```
patrickfreed, line 341: Do we need `#[non_exhaustive]` here? Also, we'll probably want `Debug` and `Clone`.

src/client/csfle/client_encryption.rs
```diff
@@ -0,0 +1,354 @@
...
+        let result = self.exec.run_ctx(ctx, None).await?;
+        Ok(result
+            .get("v")?
+            .ok_or_else(|| Error::internal("invalid decryption result"))?
+            .to_raw_bson())
+    }
+}
+
+/// Options for initializing a new `ClientEncryption`.
+#[derive(TypedBuilder)]
```
patrickfreed, line 228: I think it might be worth going over these various structs to see which ones should be `Clone` (unless they're intentionally not so already).

src/client/csfle/client_encryption.rs
```diff
@@ -0,0 +1,354 @@
+//! Support for explicit encryption.
```
patrickfreed, line 1: This is perhaps more appropriate for a later PR, but I think it would be beneficial to have more extended documentation of what this module is for, how it differs from auto encryption, and some doc-examples to demonstrate basic usage. Ditto for adding examples to any of the new types as you think is worth it.

src/client/csfle/client_encryption.rs
abr-egn: Good thought, went through and added `Clone` and `Debug` derives.

src/client/csfle/client_encryption.rs
abr-egn: Done.

src/client/csfle/client_encryption.rs
patrickfreed: Leaving them sound good, figured that was the case but just wanted to make sure.

patrickfreed (approved):
LGTM!

PR #730, opened 2022-08-24
RUST-1396 Respawn mongocryptd on server selection failure

abr-egn:
RUST-1396

[Spec](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#spawning-mongocryptd); this is the "If spawning is necessary..." clause.  Control of mongocryptd spawning moved into the `CryptExecutor`, which IMO mostly made things a bit cleaner.  The split between `CryptExecutor::new_implicit` and `::new_explicit` is a bit clunky, but because one needs async and the other doesn't it seemed necessary.  Sadly, we're probably a ways off from dependently typed function keywords :)

### DIFF
--- a/src/client/csfle/state_machine.rs
+++ b/src/client/csfle/state_machine.rs
@@ -238,7 +238,13 @@ impl Mongocryptd {
                 return unit_err(&*self.child.lock().await);
             }
         };
-        *child = Self::spawn(&self.opts);
+        let new_child = Self::spawn(&self.opts);
+        if new_child.is_ok() {
+            if let Ok(old_child) = child.as_mut() {
+                let _ = old_child.wait().await;
+            }
+        }
+        *child = new_child;
         unit_err(&*child)
     }
 
--- a/src/client/csfle/state_machine.rs
+++ b/src/client/csfle/state_machine.rs
@@ -1,4 +1,4 @@
-use std::convert::TryInto;
+use std::{convert::TryInto, path::PathBuf};
 
 use bson::{Document, RawDocument, RawDocumentBuf};
 use futures_util::{stream, TryStreamExt};
@@ -25,20 +25,20 @@ use super::options::KmsProvidersTlsOptions;
 pub(crate) struct CryptExecutor {
     key_vault_client: WeakClient,
     key_vault_namespace: Namespace,
-    mongocryptd_client: Option<Client>,
-    metadata_client: Option<WeakClient>,
     tls_options: Option<KmsProvidersTlsOptions>,
     crypto_threads: ThreadPool,
+    mongocryptd_opts: Option<MongocryptdOptions>,
+    mongocryptd_client: Option<Client>,
+    metadata_client: Option<WeakClient>,
 }
 
 impl CryptExecutor {
-    pub(crate) fn new(
+    pub(crate) fn new_explicit(
         key_vault_client: WeakClient,
         key_vault_namespace: Namespace,
-        mongocryptd_client: Option<Client>,
-        metadata_client: Option<WeakClient>,
         tls_options: Option<KmsProvidersTlsOptions>,
-    ) -> Result<Self> {
+    ) -> Result<Self>
+    {
         let num_cpus = std::thread::available_parallelism()?.get();
         let crypto_threads = rayon::ThreadPoolBuilder::new()
             .num_threads(num_cpus)
@@ -47,11 +47,30 @@ impl CryptExecutor {
         Ok(Self {
             key_vault_client,
             key_vault_namespace,
-            mongocryptd_client,
-            metadata_client,
             tls_options,
             crypto_threads,
+            mongocryptd_opts: None,
+            mongocryptd_client: None,
+            metadata_client: None,
         })
+    }
+
+    pub(crate) async fn new_implicit(
+        key_vault_client: WeakClient,
+        key_vault_namespace: Namespace,
+        tls_options: Option<KmsProvidersTlsOptions>,
+        mongocryptd_opts: Option<MongocryptdOptions>,
+        metadata_client: Option<WeakClient>,
+    ) -> Result<Self> {
+        let mongocryptd_client = match &mongocryptd_opts {
+            Some(opts) => opts.spawn().await?,
+            None => None,
+        };
+        let mut exec = Self::new_explicit(key_vault_client, key_vault_namespace, tls_options)?;
+        exec.mongocryptd_opts = mongocryptd_opts;
+        exec.mongocryptd_client = mongocryptd_client;
+        exec.metadata_client = metadata_client;
+        Ok(exec)
     }
 
     pub(crate) async fn run_ctx(&self, ctx: Ctx, db: Option<&str>) -> Result<RawDocumentBuf> {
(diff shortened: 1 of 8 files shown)


src/client/csfle/state_machine.rs
```diff
@@ -181,6 +216,48 @@ impl CryptExecutor {
...
+        let mut args: Vec<&str> = vec![];
+        let mut has_idle = false;
+        for arg in &self.spawn_args {
+            has_idle |= arg.starts_with("--idleShutdownTimeoutSecs");
+            args.push(arg);
+        }
+        if !has_idle {
+            args.push("--idleShutdownTimeoutSecs=60");
+        }
+        Command::new(bin_path)
```
patrickfreed, line 248: Spawning a process is potentially blocking, so I think we'll want to use the async runtimes versions of this. Unfortunately, I don't think they're cross-platform, so we'll need to add a new wrapper layer to our `runtime` module. Fortunately, I think the API is the same between the two runtimes:
- [tokio](https://docs.rs/tokio/latest/tokio/process/index.html)
- [async-std](https://docs.rs/async-std/latest/async_std/process/index.html)

Alternatively, we could just use `std`'s API and throw it in a `spawn_blocking` task, but I think using the runtime native APIs would probably be preferable.

src/client/csfle/state_machine.rs
```diff
@@ -89,11 +113,22 @@ impl CryptExecutor {
...
+                    let mongocryptd = self
+                        .mongocryptd
                         .as_ref()
                         .ok_or_else(|| Error::internal("mongocryptd client not found"))?;
-                    let response = mongocryptd_client.execute_operation(op, None).await?;
+                    let mut mongocryptd_client = mongocryptd.client.lock().await;
+                    let mut result = mongocryptd_client.execute_operation(op.clone(), None).await;
+                    if let Err(e) = &result {
+                        if matches!(&*e.kind, ErrorKind::ServerSelection { .. }) {
+                            *mongocryptd_client = mongocryptd.opts.spawn().await?;
```
patrickfreed, line 124: I don't think we need to recreate the `Client` each time we spawn mongocryptd, we just need to restart the process and the monitor in the client should rediscover the new process. This will also allow us to drop the `Mutex`, which is currently limiting us to one operation at a time using this client.

We will need a lock around something to ensure we don't try to spawn `mongocryptd` multiple times at once though, maybe around the opts?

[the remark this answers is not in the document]
abr-egn: Done.

abr-egn: Updated - this meshed nicely with the switch to the runtime process APIs, since it turns out keeping a `Child` handle around is needed to have the opportunity to reap dead processes.

src/client/csfle/state_machine.rs
```diff
@@ -181,6 +209,71 @@ impl CryptExecutor {
...
+
+    async fn respawn(&self) -> Result<()> {
+        let mut child = match self.child.try_lock() {
+            Ok(l) => l,
+            _ => {
+                // Another respawn is in progress.  Lock to wait for it.
+                return unit_err(&*self.child.lock().await);
+            }
+        };
+        *child = Self::spawn(&self.opts);
```
patrickfreed, line 241: I think we may want to do an explicit `wait` here on the prior child process (if the result is `Ok`) to ensure its resources were reaped properly before starting a new one. I think the `try_wait` in the `Drop` of process isn't necessarily sufficient to ensure reaping happened.

src/client/csfle/state_machine.rs
```diff
@@ -181,6 +209,71 @@ impl CryptExecutor {
...
+        options.server_selection_timeout = Some(std::time::Duration::from_millis(10_000));
+        let client = Client::with_options(options)?;
+        Ok(Self {
+            opts,
+            client,
+            child,
+        })
+    }
+
+    async fn respawn(&self) -> Result<()> {
```
patrickfreed, line 233: The spec is somewhat quiet about the specifics of this, so I'm going off of DRIVERS-1861 here:
- In the event that `mongocryptd` has already been spawned by another client instance, we'll still want this to return `Ok`. I'm not exactly sure how to detect that though.
- It looks like we'd want to report any other errors related to spawning a process here though.

src/client/csfle/state_machine.rs
```diff
@@ -89,11 +107,21 @@ impl CryptExecutor {
...
-                        .mongocryptd_client
+                    let mongocryptd = self
+                        .mongocryptd
                         .as_ref()
                         .ok_or_else(|| Error::internal("mongocryptd client not found"))?;
-                    let response = mongocryptd_client.execute_operation(op, None).await?;
+                    let mut result = mongocryptd.client.execute_operation(op.clone(), None).await;
+                    if let Err(e) = &result {
+                        if matches!(&*e.kind, ErrorKind::ServerSelection { .. }) {
+                            mongocryptd.respawn().await?;
```
patrickfreed, line 117: DRIVERS-1861 also mentions that we're supposed to be doing this spawning at MongoClient construction rather than on first operation, but given that it could technically involve SRV resolution (and therefore be async), I think we'll have to ignore that and keep it as we have it here. It's not encoded in the specification anyways, so we're not technically violating anything by doing so.

src/runtime/process.rs
```diff
@@ -0,0 +1,40 @@
...
+#[cfg(feature = "async-std-runtime")]
+use async_std::process::{Child, Command};
+#[cfg(feature = "tokio-runtime")]
+use tokio::process::{Child, Command};
+
+use crate::error::Result;
+
+#[derive(Debug)]
+pub(crate) struct Process {
+    child: Child,
```
patrickfreed, line 12: DRIVERS-1861 mentions that ideally the mongocryptd process would not be a child process of the driver, since it would want to keep executing for other driver instances even after the first one exited. I looked briefly, but I'm not sure of a way to do this cleanly with `tokio`'s API. I did see some solutions that use unsafe + libc, but I assume they are not very portable. Might be worth looking into some more, but I am not super hopeful we'll find a way around this worth pursuing.

abr-egn: This is complicated and messy.  My understanding of the situation:
* On Linux, a child process will continue to run if the parent process terminated normally; however, if the parent process receives `SIGINT`/`SIGTERM` that will be propagated to the children (killing them unless they have special handlers).  The normal way to avoid this is via the double-fork dance, which is nicely captured by the [`fork` crate](https://crates.io/crates/fork), but that just provides a raw `pid_t` to the parent, so we'd need to implement our own waiting/reaping logic.
* On Windows, you just need to set the [`CREATE_NEW_PROCESS_GROUP` flag](https://docs.microsoft.com/en-us/windows/win32/procthread/process-creation-flags) - this can be done in std via `CommandExt::creation_flags`, but neither Tokio nor async-std expose that in their respective `Command` APIs, and the `fork` crate doesn't support Windows.
* On OS X, it seems like pre-10.5 it was the same as Linux, but in 10.5+ it's its own thing (not entirely clear exactly what).  It looks like the `fork` crate handles this.

My inclination is to punt on this; if the user needs a persistent mongocryptd they can spawn one themselves and tell the driver to use it and not autospawn.

src/client/csfle/state_machine.rs
abr-egn: Done.

src/client/csfle/state_machine.rs
patrickfreed: Ah right, I forgot we had made a separate method for that.

src/client/csfle/state_machine.rs
patrickfreed: Yeah I don't think we can go that route. I actually just checked it and `spawn` will return `Ok` just fine, so there isn't anything to be done. If we wait for the process to exit and assert its status is how we'd run into issues, but we're not doing that here.

patrickfreed (approved):
LGTM!

src/client/csfle/state_machine.rs
```diff
@@ -89,11 +107,21 @@ impl CryptExecutor {
...
+                            let new_result = mongocryptd.client.execute_operation(op, None).await;
```
kmahar, line 118: so the spec says:
> If the MongoClient fails to connect after spawning, the server selection error is propagated to the user.

is it possible that we successfully connect to the respawned mongocryptd here, but then `execute_operation` fails with a non-server-selection error? if so, should we be propagating that rather than the original server selection error?

src/client/csfle/state_machine.rs
```diff
@@ -181,6 +209,77 @@ impl CryptExecutor {
...
+struct Mongocryptd {
+    opts: MongocryptdOptions,
+    client: Client,
+    child: Mutex<Result<Process>>,
+}
+
+impl Mongocryptd {
+    async fn new(opts: MongocryptdOptions) -> Result<Self> {
+        let child = Mutex::new(Ok(Self::spawn(&opts)?));
+        let uri = opts.uri.as_deref().unwrap_or("mongodb://localhost:27020");
```
kmahar, line 222: nit: could we make this and the server selection timeout named constants on this type

src/client/csfle/state_machine.rs
```diff
@@ -181,6 +209,77 @@ impl CryptExecutor {
...
+        let client = Client::with_options(options)?;
+        Ok(Self {
+            opts,
+            client,
+            child,
+        })
+    }
+
+    async fn respawn(&self) -> Result<()> {
+        let mut child = match self.child.try_lock() {
```
kmahar, line 234: I'm probably missing something here, but what is the benefit of calling `try_lock` first and then `lock` if it fails? 
my assumption (which could be wrong) is that if the lock is free, `try_lock` and `lock` would both immediately return the lock. and that if it isn't free, we are going to end up calling `lock` regardless.

src/client/csfle/state_machine.rs
abr-egn: Good thought, changed to that.  Also reorganized this code block a little to (hopefully) be easier to read.

src/client/csfle/state_machine.rs
abr-egn: Done.

src/client/csfle/state_machine.rs
abr-egn: If `try_lock` fails that tells us another task is attempting a respawn, so the failure branch doesn't need to call `Self::spawn`, it just needs to wait for the task holding the lock to finish and then return the result of that.  Just calling `lock` rather than `try_lock` would mean we can't distinguish between lock-immediately-acquired and lock-acquired-after-waiting.

kmahar: oh! I missed that we were just going ahead and returning here. that makes sense, thanks for the explanation

kmahar (approved):

PR #733, opened 2022-08-30
RUST-1443 Stop executing server monitors after the server has been closed (2.3 backport)

patrickfreed:
RUST-1443

This PR fixes a bug where server monitors would continue executing even after a server had been removed from the topology.

### DIFF
--- a/src/sdam/description/topology/mod.rs
+++ b/src/sdam/description/topology/mod.rs
@@ -126,23 +126,6 @@ impl PartialEq for TopologyDescription {
 }
 
 impl TopologyDescription {
-    pub(crate) fn new_empty() -> Self {
-        Self {
-            single_seed: false,
-            topology_type: TopologyType::Unknown,
-            set_name: None,
-            max_set_version: None,
-            max_election_id: None,
-            compatibility_error: None,
-            session_support_status: SessionSupportStatus::Undetermined,
-            transaction_support_status: TransactionSupportStatus::Undetermined,
-            cluster_time: None,
-            local_threshold: None,
-            heartbeat_freq: None,
-            servers: HashMap::new(),
-        }
-    }
-
     pub(crate) fn initialized(options: &ClientOptions) -> crate::error::Result<Self> {
         verify_max_staleness(
             options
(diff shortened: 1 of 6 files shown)


src/sdam/test.rs
```diff
@@ -485,3 +490,86 @@ async fn repl_set_name_mismatch() -> crate::error::Result<()> {
 
     Ok(())
 }
+
+/// Test verifying that a server's monitor stops after the server has been removed from the
+/// topology.
+#[cfg_attr(feature = "tokio-runtime", tokio::test(flavor = "multi_thread"))]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn removed_server_monitor_stops() -> crate::error::Result<()> {
```
patrickfreed, line 498: This test will need to be forward-ported to the main branch once RUST-360 lands.

src/sdam/description/topology/mod.rs
```diff
@@ -849,10 +871,12 @@ impl Default for TransactionSupportStatus {
 /// Returned from `TopologyDescription::diff`.
 #[derive(Debug)]
 pub(crate) struct TopologyDescriptionDiff<'a> {
-    pub(crate) removed_addresses: HashSet<&'a ServerAddress>,
-    pub(crate) added_addresses: HashSet<&'a ServerAddress>,
-    pub(crate) changed_servers:
-        HashMap<&'a ServerAddress, (&'a ServerDescription, &'a ServerDescription)>,
+    pub(crate) removed_addresses: Vec<&'a ServerAddress>,
```
abr-egn, line 874: Is the change to `Vec` purely to allow sorting in tests?  It looks like they're still "morally" sets, i.e. there won't be any duplicate keys.  If so, it might be clearer to do the sorting when emitting the events in `TopologyWorker::process_topology_diff` and leave these as `HashSet/Map`.

abr-egn (commented):
SDAM stuff can be subtle, so to make sure I'm following what's going on:
* `Monitor`'s loop condition changed from topology liveness to server liveness (this is the core fix).
* To enable `is_alive` for polling in the above, `WorkerHandle[Listener]` changed to use `watch::Receiver`.
* `TopologyWorker` pairs `WorkerHandle` with `Server` in a map of `MonitoredServer`s; dropping those values is what will terminate the worker.
* The server map is maintained in `sync_hosts`;
  * `sync_hosts` used to take a `&mut TopologyState` to update, but that state is now rolled directly into `TopologyWorker`.

Is that a reasonable summary?

src/sdam/description/topology/mod.rs
patrickfreed: Good idea, done.

patrickfreed:
Yep, that's all exactly right. The change from using the `TopologyState` that was stored in the `watch::channel` to using one stored on the worker was to avoid a reference cycle: receivers of the channel (which the monitor has) would keep the `MonitoredServer` instance alive, leading to the monitor never exiting.

abr-egn (approved):

kmahar (approved):

isabelatkinson (commented):
one nit but otherwise looks good!

isabelatkinson (approved):

PR #738, opened 2022-09-08
RUST-1071 Add `comment` field to all options structs

isabelatkinson:


src/coll/options.rs
```diff
@@ -95,6 +95,7 @@ pub enum CursorType {
 
 /// Specifies the options to a
 /// [`Collection::insert_one`](../struct.Collection.html#method.insert_one) operation.
+#[skip_serializing_none]
```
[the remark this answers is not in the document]
isabelatkinson: I can't find any specific language about this in the specs, but it doesn't appear that `Bson::Null` (which is what `None` values here would serialize to) is a significant value vs. not being specified for these fields. We also have `#[skip_serializing_none)]` on a lot of our options structs that have the same fields already.

abr-egn (approved):
LGTM!

src/coll/options.rs
```diff
@@ -486,10 +542,21 @@ pub struct AggregateOptions {
...
     pub collation: Option<Collation>,
 
-    /// Tags the query with an arbitrary string to help trace the operation through the database
-    /// profiler, currentOp and logs.
+    /// Tags the query with an arbitrary string to help trace the operation through the
+    /// database profiler, currentOp and logs.
+    ///
+    /// If both this option and `comment_bson` are specified, `comment_bson` will take precedence.
+    #[serde(skip_serializing)]
     pub comment: Option<String>,
```
kmahar, line 550: is the plan in the long run to consolidate these into a single option in 3.0? if so, should we file a ticket about that / add a TODO?

this also made me think about whether it makes sense to deprecate `comment` here. on the one hand it would hopefully prevent people from trying to specify both `comment` and `comment_bson` and would get people to migrate to the Bson form. but on the other hand, it's kind of funny to deprecate the one with the property we would use in 3.0, since that means anyone using the string option now would rename it, and then rename it back. 🤔

isabelatkinson: Yes, in 3.0 there should be one `comment` field of type `Option<Bson>`. I filed RUST-1364 for this a while back. I think our original consensus when that ticket was filed was to wait for 3.0 for any bson comment support, so I'll update the description to reflect what was done here. I'll also add TODOs pointing to that ticket to this/the other place we have this workaround.

I was thinking about deprecation when I was originally doing this work but I didn't think it made much sense; both of them are effectively deprecated right now because neither will exist in their current form in 3.0, so I think users can use whatever currently suits their needs and then migrate when they upgrade. Both options will require migration anyway.

kmahar: SGTM!

kmahar (approved):

patrickfreed (approved):

PR #739, opened 2022-09-08
Collect only deserializable documents

sgasse:
The [documentation](https://docs.rs/mongodb/latest/mongodb/struct.Cursor.html) mentiones several ways to collect documents from a typed collection:
```rust
// regular Stream uses collect() and collects into a Vec<Result<T>>
let v: Vec<Result<_>> = cursor.collect().await;
```

```rust
// TryStream uses try_collect() and collects into a Result<Vec<T>>
let v: Vec<_> = cursor.try_collect().await?;
```

In the project I work on, we have heterogeneous collections and often only want the documents that can be deserialized successfully. What I currently do is:

```rust
let valid_documents: Vec<LongDocumentTypeName> = cursor
    .filter_map(|e| async move { e.ok() })
    .collect()
    .await;
```

An alternative with collecting the results first before mapping them looks like this:

```rust
let valid_documents: Vec<LongDocumentTypeName> = cursor
    .collect::<Vec<Result<LongDocumentTypeName, _>>>()
    .await
    .filter_map(|e| { e.ok() })
    .collect();
```

While I wonder what is the idiomatic way, I want to be able to retrieve the valid, deserializable documents with a one-liner. For this, I added a helper function which I tried to integrate into the MongoDB Rust driver in this PR.

Do you think it is helpful? Am I missing some existing functionality that would allow me to get just the valid documents?

### DIFF
--- a/src/cursor/mod.rs
+++ b/src/cursor/mod.rs
@@ -13,6 +13,7 @@ use bson::RawDocument;
 #[cfg(test)]
 use bson::RawDocumentBuf;
 use futures_core::{future::BoxFuture, Stream};
+use futures_util::StreamExt;
 use serde::{de::DeserializeOwned, Deserialize};
 #[cfg(test)]
 use tokio::sync::oneshot;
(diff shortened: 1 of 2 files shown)


sgasse:
(I saw in the diff that my `rustfmt` version probably does not match so I marked this a draft for now)

sgasse:
I fixed my toolchain, ran all checks and this is now ready for review.

sgasse:
Ah alright @abr-egn , thank you for explaining. Great suggestion to filter already before - is there a convenient way to create a filter checking for the existence/type of a Rust `struct`?

abr-egn:
We don't have one ready-made, but you could potentially use [`serde_aux::serde_introspection::serde_introspect`](https://docs.rs/serde-aux/4.0.0/serde_aux/serde_introspection/fn.serde_introspect.html) to get the list of field names and construct a filter from that.

PR #740, opened 2022-09-09
RUST-828 Reduce flakiness of `cluster_time_in_commands` test

patrickfreed:
RUST-828


src/client/session/test/mod.rs
```diff
@@ -282,24 +310,73 @@ async fn cluster_time_in_commands() {
...
-    cluster_time_test("ping", |client| async move {
+    let handler = Arc::new(EventHandler::new());
+    let mut options = CLIENT_OPTIONS.get().await.clone();
+    options.heartbeat_freq = Some(Duration::from_secs(1000));
+    options.command_event_handler = Some(handler.clone());
+    options.sdam_event_handler = Some(handler.clone());
+
+    // Ensure we only connect to one server so the monitor checks from other servers
+    // don't affect the TopologyDescription's clusterTime value between commands.
+    if options.load_balanced != Some(true) {
```
patrickfreed, line 321: I believe this test was failing occasionally due to two different possible causes:
- When connected to topologies with multiple servers, there's a possibility that a monitor check to another server completes between the first command in this test and the second, thereby updating the cluster time and causing a failure
- The first command begins executing before the server in question has been discovered, causing it to request an immediate check which is then serviced as soon as the initial check completes. This second check could succeed between commands and mess up the client's cluster time too.
  - With the changes in the streamable hello PR, this won't be a concern now that we properly ignore check requests while one is currently in progress.

Both of these are fixed, and the test didn't fail in this patch, so hopefully that solves it.

src/test/util/event.rs
```diff
@@ -500,15 +529,52 @@ impl EventSubscriber<'_> {
...
-    {
-        let mut events = Vec::new();
-        while let Some(event) = self.wait_for_event(timeout, &mut filter).await {
-            events.push(event);
-        }
-        events
+    /// Waits for the next CommandStartedEvent/CommandFailedEvent pair.
+    /// If the next CommandStartedEvent is associated with a CommandFailedEvent, this method will
+    /// panic.
+    pub(crate) async fn wait_for_successful_command_execution(
```
patrickfreed, line 535: This functionality existed on `EventClient`, but I wanted to use a regular `Client` with a handler, so I reimplemented it here too.

abr-egn (approved):
LGTM!

kmahar (approved):

isabelatkinson (approved):

PR #743, opened 2022-09-15
RUST-360 Guard connection establishment by connectTimeoutMS

patrickfreed:
In #721, I moved around a lot of the logic for enforcing `connectTimeoutMS` and inadvertently removed it from the operation connection establishment code path. This PR adds it back in.

### DIFF
--- a/src/cmap/establish/mod.rs
+++ b/src/cmap/establish/mod.rs
@@ -65,10 +65,20 @@ impl ConnectionEstablisher {
             None
         };
 
+        let connect_timeout = if let Some(timeout) = options.connect_timeout {
+            if timeout.is_zero() {
+                Duration::MAX
+            } else {
+                timeout
+            }
+        } else {
+            DEFAULT_CONNECT_TIMEOUT
+        };
+
         Ok(Self {
             handshaker,
             tls_config,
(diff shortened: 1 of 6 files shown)


src/sdam/monitor.rs
```diff
@@ -198,16 +197,6 @@ impl Monitor {
             // Otherwise, just use connectTimeoutMS.
             self.connect_timeout()
         };
-        let timeout_future = async {
```
patrickfreed, line 201: This was moved into the `runtime` module.

src/cmap/establish/mod.rs
```diff
@@ -59,14 +65,29 @@ impl ConnectionEstablisher {
             None
         };
 
+        let connect_timeout = if let Some(timeout) = options.connect_timeout {
```
abr-egn, line 68: Nit: I think this might be easier to read as a match against `Some(0)`, `Some(n)`, and `None`.

abr-egn (approved):
LG!  One small comment only.

kmahar (approved):

isabelatkinson (approved):

src/cmap/establish/mod.rs
patrickfreed: Good idea, done. I had to do it a little bit different since we're matching on a `Duration`, but I still think it's more readable than before.

PR #744, opened 2022-09-15
RUST-1443 Ensure monitors close after server is removed from topology

patrickfreed:
RUST-1443

This PR forward-ports the test added in #733 that verifies that monitors stop executing after their servers are removed from the topology. Running the test yielded a panic in the monitoring task due to a subtraction overflow (heartbeatFrequencyMS < minHeartbeatFrequencyMS), which made me notice we weren't validating heartbeatFrequencyMS when constructing a client via `ClientOptions` directly. I also fixed that here (RUST-1476).

### DIFF
--- a/src/client/options/mod.rs
+++ b/src/client/options/mod.rs
@@ -1325,6 +1325,19 @@ impl ClientOptions {
             }
         }
 
+        if let Some(heartbeat_frequency) = self.heartbeat_freq {
+            if heartbeat_frequency < self.min_heartbeat_frequency() {
+                return Err(ErrorKind::InvalidArgument {
+                    message: format!(
+                        "'heartbeat_freq' must be at least {}ms, but {}ms was given",
+                        self.min_heartbeat_frequency().as_millis(),
+                        heartbeat_frequency.as_millis()
+                    ),
+                }
+                .into());
+            }
+        }
+
         Ok(())
     }
 
@@ -1372,6 +1385,21 @@ impl ClientOptions {
     #[cfg(test)]
     pub(crate) fn test_options_mut(&mut self) -> &mut TestOptions {
         self.test_options.get_or_insert_with(Default::default)
+    }
+
+    pub(crate) fn min_heartbeat_frequency(&self) -> Duration {
+        #[cfg(test)]
+        {
+            self.test_options
+                .as_ref()
+                .and_then(|to| to.min_heartbeat_freq)
+                .unwrap_or(MIN_HEARTBEAT_FREQUENCY)
+        }
+
+        #[cfg(not(test))]
+        {
+            MIN_HEARTBEAT_FREQUENCY
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/client/options/mod.rs
```diff
@@ -1325,6 +1325,19 @@ impl ClientOptions {
             }
         }
 
+        if let Some(heartbeat_frequency) = self.heartbeat_freq {
```
patrickfreed, line 1328: This validation existed, but it was in the URI parsing stage. I've moved it to `ClientOptions::validate` to ensure it gets hit even when the options are constructed directly.

src/sdam/monitor.rs
```diff
@@ -101,25 +101,11 @@ impl Monitor {
...
-                    .client_options
-                    .test_options
-                    .as_ref()
-                    .and_then(|to| to.min_heartbeat_freq)
-                    .unwrap_or(MIN_HEARTBEAT_FREQUENCY);
-
-                #[cfg(not(test))]
-                let min_frequency = MIN_HEARTBEAT_FREQUENCY;
-
-                tokio::select! {
```
patrickfreed, line 115: I moved the minimum delay into `wait_for_check_request` to consolidate the checks for the server closing.

src/sdam/monitor.rs
```diff
@@ -607,17 +593,28 @@ impl MonitorRequestReceiver {
...
+    /// It is covered by the provided timeout.
+    async fn wait_for_check_request(&mut self, delay: Duration, timeout: Duration) {
         let _ = runtime::timeout(timeout, async {
+            let wait_for_check_request = async {
+                runtime::delay_for(delay).await;
+                self.topology_check_request_receiver
+                    .wait_for_check_request()
+                    .await;
+            };
+            tokio::pin!(wait_for_check_request);
```
patrickfreed, line 610: we need the `pin` here since we're polling the same future repeatedly in a loop.

src/sdam/test.rs
```diff
@@ -269,3 +274,91 @@ async fn repl_set_name_mismatch() -> crate::error::Result<()> {
 
     Ok(())
 }
+
+/// Test verifying that a server's monitor stops after the server has been removed from the
+/// topology.
+#[cfg_attr(feature = "tokio-runtime", tokio::test(flavor = "multi_thread"))]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn removed_server_monitor_stops() -> crate::error::Result<()> {
```
patrickfreed, line 282: this test was copied pretty much as-is from #733

src/sdam/topology.rs
```diff
@@ -1081,10 +1081,11 @@ pub(crate) struct TopologyCheckRequestReceiver {
 
 impl TopologyCheckRequestReceiver {
```
patrickfreed, line 1082: This was a minor change I made to mitigate the case where the value hasn't changed but there are still operations requesting an update. I don't believe this case actually could actually happen with the existing implementation, but I think it's best to make that clear here.

abr-egn (approved):
LGTM!

src/sdam/monitor.rs
```diff
@@ -607,17 +593,28 @@ impl MonitorRequestReceiver {
...
 
-    /// Wait for a request to immediately check the server to come in, guarded by the provided
-    /// timeout. If a cancellation request is received indicating the topology has closed, this
-    /// method will return. All other cancellation requests will be ignored.
-    async fn wait_for_check_request(&mut self, timeout: Duration) {
+    /// Wait for a request to immediately check the server to be received, guarded by the provided
+    /// timeout. If the server associated with this monitor is removed from the topology, this
+    /// method will return.
+    ///
+    /// The `delay` parameter indicates how long this method should wait before listing to requests.
```
kmahar, line 600: listing -> listening?

ee

src/sdam/monitor.rs
```diff
@@ -607,17 +593,28 @@ impl MonitorRequestReceiver {
...
+    /// It is covered by the provided timeout.
```
kmahar, line 601: does "covered by" mean "counts toward"? like if `delay` is 1s and timeout is 5s, the method delays for 1s then times out 4s if no requests come in? (I think yes from reading the logic)

src/sdam/monitor.rs
patrickfreed: Fixed

src/sdam/monitor.rs
patrickfreed: Yep, updated the docstring to more be clearer about this.

kmahar (approved):

isabelatkinson (approved):

PR #747, opened 2022-09-16
RUST-1400 GridFS download methods

isabelatkinson:
Implements the `download_to_stream` method described [here](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#file-download) and the `download_to_stream_by_name` method described [here](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#file-download-by-filename)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -457,6 +457,11 @@ pub enum ErrorKind {
     #[non_exhaustive]
     DnsResolve { message: String },
 
+    /// A GridFS error occurred.
+    #[error("{0:?}")]
+    #[non_exhaustive]
+    GridFs(GridFsErrorKind),
+
     #[error("Internal error: {message}")]
     #[non_exhaustive]
     Internal { message: String },
@@ -691,6 +696,49 @@ impl WriteFailure {
             .into())
         }
     }
+}
+
+/// An error that occurred during a GridFS operation.
+#[derive(Clone, Debug)]
+#[allow(missing_docs)]
+#[non_exhaustive]
+pub enum GridFsErrorKind {
+    /// The file with the given identifier was not found.
+    #[non_exhaustive]
+    FileNotFound { identifier: GridFsFileIdentifier },
+
+    /// The file with the given revision was not found.
+    #[non_exhaustive]
+    RevisionNotFound { revision: i32 },
+
+    /// The chunk at index 'n' was missing.
+    #[non_exhaustive]
+    MissingChunk { n: u32 },
+
+    /// The chunk was the incorrect size.
+    #[non_exhaustive]
+    WrongSizeChunk {
+        actual_size: u32,
+        expected_size: u32,
+    },
+
+    /// An incorrect number of chunks was present for the file.
+    #[non_exhaustive]
+    WrongNumberOfChunks {
+        actual_number: u32,
+        expected_number: u32,
+    },
+}
+
+/// An identifier for a file stored in a GridFS bucket.
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub enum GridFsFileIdentifier {
+    /// The name of the file. Not guaranteed to be unique.
+    Filename(String),
+
+    /// The file's unique [`Bson`] ID.
+    Id(Bson),
 }
 
 /// Translates ErrorKind::BulkWriteError cases to ErrorKind::WriteErrors, leaving all other errors
--- a/src/gridfs.rs
+++ b/src/gridfs.rs
@@ -1,52 +1,72 @@
-#![allow(dead_code, unused_variables)]
 // TODO(RUST-1395) Remove these allows.
+#![allow(dead_code, unused_variables)]
 
+mod download;
 pub mod options;
 
 use core::task::{Context, Poll};
-use std::pin::Pin;
+use std::{
+    pin::Pin,
+    sync::{atomic::AtomicBool, Arc},
+};
+
+use serde::{Deserialize, Serialize};
+use tokio::io::ReadBuf;
 
 use crate::{
+    bson::{doc, oid::ObjectId, Bson, DateTime, Document, RawBinaryRef},
     concern::{ReadConcern, WriteConcern},
     cursor::Cursor,
     error::Result,
-    selection_criteria::SelectionCriteria,
+    options::SelectionCriteria,
+    Collection,
     Database,
 };
-use bson::{oid::ObjectId, Bson, DateTime, Document};
+
 use options::*;
-use serde::{Deserialize, Serialize};
-use tokio::io::ReadBuf;
 
 pub const DEFAULT_BUCKET_NAME: &str = "fs";
 pub const DEFAULT_CHUNK_SIZE_BYTES: u32 = 255 * 1024;
 
 // Contained in a "chunks" collection for each user file
-struct Chunk {
+#[derive(Debug, Deserialize, Serialize)]
+struct Chunk<'a> {
+    #[serde(rename = "_id")]
     id: ObjectId,
     files_id: Bson,
     n: u32,
-    // default size is 255 KiB
-    data: Vec<u8>,
+    #[serde(borrow)]
+    data: RawBinaryRef<'a>,
 }
 
 /// A collection in which information about stored files is stored. There will be one files
 /// collection document per stored file.
-#[derive(Serialize, Deserialize)]
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
 pub struct FilesCollectionDocument {
-    id: Bson,
-    length: i64,
-    chunk_size: u32,
-    upload_date: DateTime,
-    filename: String,
-    metadata: Document,
+    #[serde(rename = "_id")]
+    pub id: Bson,
+    pub length: u64,
+    pub chunk_size: u32,
+    pub upload_date: DateTime,
+    pub filename: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub metadata: Option<Document>,
+}
+
+#[derive(Debug)]
+struct GridFsBucketInner {
+    options: GridFsBucketOptions,
+    files: Collection<FilesCollectionDocument>,
+    chunks: Collection<Chunk<'static>>,
+    created_indexes: AtomicBool,
 }
 
 /// Struct for storing GridFS managed files within a [`Database`].
+#[derive(Debug, Clone)]
 pub struct GridFsBucket {
-    // Contains a "chunks" collection
-    pub(crate) db: Database,
-    pub(crate) options: GridFsBucketOptions,
+    inner: Arc<GridFsBucketInner>,
 }
 
 // TODO: RUST-1395 Add documentation and example code for this struct.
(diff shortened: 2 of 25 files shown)


src/test/spec/unified_runner/test_runner.rs
```diff
@@ -128,6 +129,14 @@ impl TestRunner {
         ));
 
         for test_case in test_file.tests {
+            // to run a single test, set the TEST_DESCRIPTION environment variable to the test's
+            // description
+            if let Ok(description) = std::env::var("TEST_DESCRIPTION") {
+                if description != test_case.description {
+                    continue;
+                }
+            }
```
isabelatkinson, line 138: I think we all probably write something to this effect when trying to debug an individual test, so I thought it would be useful to have this in the code permanently

src/gridfs.rs
```diff
@@ -4,49 +4,66 @@
...
 pub const DEFAULT_BUCKET_NAME: &str = "fs";
 pub const DEFAULT_CHUNK_SIZE_BYTES: u32 = 255 * 1024;
 
 // Contained in a "chunks" collection for each user file
-struct Chunk {
-    id: ObjectId,
-    files_id: Bson,
-    n: u32,
+#[derive(Debug, Deserialize, Serialize)]
+pub struct Chunk {
```
isabelatkinson, line 31: going to disable the lint suppression for missing docs in places like these and add them in when I do RUST-1395

isabelatkinson:
converting to draft as I pull in some improvements from #751, feel free to hold off on reviewing for now

isabelatkinson:
ready for review again

src/gridfs.rs
```diff
@@ -1,52 +1,71 @@
...
-    data: Vec<u8>,
+    data: Binary,
 }
 
 /// A collection in which information about stored files is stored. There will be one files
 /// collection document per stored file.
-#[derive(Serialize, Deserialize)]
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
 pub struct FilesCollectionDocument {
```
patrickfreed, line 46: Think we'll want `non_exhaustive` here.

src/gridfs.rs
```diff
@@ -1,52 +1,71 @@
...
+    #[serde(rename = "_id")]
     id: ObjectId,
     files_id: Bson,
     n: u32,
     // default size is 255 KiB
-    data: Vec<u8>,
+    data: Binary,
 }
 
 /// A collection in which information about stored files is stored. There will be one files
```
patrickfreed, line 42: I know this is existing code, but I think this docstring could be a tad clearer. For instance, this struct models the documents in the files collection, rather than being the collection itself. Also, it could be better to say something like "metadata" rather than information, and also could be useful to mention the collection in which the files are actually stored (the "chunks" collection, right?).

src/gridfs.rs
```diff
@@ -1,52 +1,71 @@
...
-    filename: String,
-    metadata: Document,
+    #[serde(rename = "_id")]
+    pub id: Bson,
+    pub length: u64,
+    pub chunk_size: u32,
+    pub upload_date: DateTime,
+    pub filename: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub metadata: Option<Document>,
```
patrickfreed, line 54: Would be good to document each of these fields, similar to what we do for `ChangeStreamEvent`.

src/gridfs/download.rs
```diff
@@ -0,0 +1,181 @@
...
+use crate::{
+    bson::{doc, Bson},
+    error::{ErrorKind, Result},
+    options::{FindOneOptions, FindOptions},
+};
+
+impl GridFsBucket {
+    /// Downloads the contents of the stored file specified by `id` and writes
+    /// the contents to the `destination`.
+    pub async fn download_to_tokio_writer<T>(&self, id: Bson, destination: &mut T) -> Result<()>
```
patrickfreed, line 14: We can actually just take in `T` here instead of `&mut T`. See the relevant API guideline: [C-RW-VALUE](https://rust-lang.github.io/api-guidelines/interoperability.html#c-rw-value). Ditto for the other methods

src/gridfs/download.rs
```diff
@@ -0,0 +1,181 @@
...
+use tokio_util::compat::FuturesAsyncWriteCompatExt;
+
+use super::{options::GridFsDownloadByNameOptions, FilesCollectionDocument, GridFsBucket};
+use crate::{
+    bson::{doc, Bson},
+    error::{ErrorKind, Result},
+    options::{FindOneOptions, FindOptions},
+};
+
+impl GridFsBucket {
```
patrickfreed, line 11: Could we provide doctest examples for each of these? (sorry i know it will be a lot of similar examples, but I think it's something users really appreciate)

src/gridfs/download.rs
```diff
@@ -0,0 +1,181 @@
...
+        let file = match self
+            .inner
+            .files
+            .find_one(doc! { "filename": filename.as_ref() }, options)
+            .await?
+        {
+            Some(fcd) => fcd,
+            None => {
+                return Err(ErrorKind::InvalidArgument {
+                    message: format!("couldn't find file with name {}", &filename.as_ref()),
```
patrickfreed, line 90: This error case might be one that a user would want to handle programatically, so I think it could be worth introducing an enum case of some kind for `FileNotFound`. We could do this at the `ErrorKind` level (e.g. `ErrorKind::GridFsFileNotFound`), but given that we have a `GridFs` case, I think there's merit in folding them up all under that.

e.g.
```rust
enum ErrorKind {
    GridFs(GridFsErrorKind),
    // rest of cases
}

enum GridFsErrorKind {
    #[non_exhaustive]
    FileNotFound { filename: String },
    Other { message: String },
}
```
WDYT?

src/gridfs/download.rs
```diff
@@ -0,0 +1,181 @@
...
+        T: futures_util::io::AsyncWrite + std::marker::Unpin,
+    {
+        self.download_to_tokio_writer(id, &mut destination.compat_write())
+            .await
+    }
+
+    /// Downloads the contents of the stored file specified by `filename` and writes the contents to
+    /// the `destination`. If there are multiple files with the same filename, the `revision` in the
+    /// options provided is used to determine which one to download. If no `revision` is specified,
+    /// the most recent file with the given filename is chosen.
```
patrickfreed, line 58: Can we document in these methods that an error will be returned if the file is not found?

src/gridfs/download.rs
```diff
@@ -0,0 +1,181 @@
...
+            .chunks
+            .find(doc! { "files_id": &file.id }, options)
+            .await?;
+
+        let mut n = 0;
+        while cursor.advance().await? {
+            let chunk = cursor.deserialize_current()?;
+            if chunk.n != n {
+                return Err(ErrorKind::GridFS {
+                    message: format!("missing chunk {} in file", n),
```
patrickfreed, line 148: Ditto might want to have cases for the ChunkIsMissing, ChunkIsWrongSize, and RevisionNotFound errors (basically anything the spec tests refer to in the yml comments).

Having them all enumerated helps give the user an idea of what might go wrong, even if they don't programmatically handle each case imo.

src/gridfs/download.rs
```diff
@@ -0,0 +1,181 @@
...
+                .into());
+            }
+
+            destination.write_all(&chunk.data.bytes).await?;
+            n += 1;
+        }
+
+        let expected_n =
+            total_bytes_expected / chunk_size + u64::from(total_bytes_expected % chunk_size != 0);
+        if (n as u64) < expected_n {
```
patrickfreed, line 172: Would we want to make this more general and check for `expected_n != n`? Then we could say "number of chunks read (n) did not match expected number of chunks (expected_n)".

src/test/spec/unified_runner/test_runner.rs
patrickfreed: Could we make this `contains` instead of strict equality? I normally add something like that in case I want to run a few tests at once. Great idea to add this btw; I think it'll save a ton of time for all of us going forward.

src/gridfs/download.rs
```diff
@@ -0,0 +1,181 @@
...
+    async fn download_to_tokio_writer_common<T>(
+        &self,
+        file: FilesCollectionDocument,
+        destination: &mut T,
+    ) -> Result<()>
+    where
+        T: tokio::io::AsyncWrite + std::marker::Unpin,
+    {
+        let total_bytes_expected = file.length;
+        let chunk_size = file.chunk_size as u64;
```
patrickfreed, line 126: Since this value is coming from the DB, we may want to validate that it is within the range of `u64`, returning an error if it isn't.

src/gridfs/download.rs
```diff
@@ -0,0 +1,181 @@
...
+                }
+                .into());
+            }
+        };
+        self.download_to_tokio_writer_common(file, destination)
+            .await
+    }
+
+    /// Downloads the contents of the stored file specified by `id` and writes
+    /// the contents to the `destination`.
```
patrickfreed, line 42: The method name implies this of course, but can we explicitly mention that the destination here implements the `AsyncWrite` trait from the `futures 0.3` crate, and link to it? Ditto for the tokio ones. Might also be worth mentioning that once `std` stabilizes `AsyncWrite`, this method will be deprecated (ditto for tokio ones on this too).

patrickfreed (commented):
Sorry for the delay on the review!

src/error.rs
```diff
@@ -693,6 +698,61 @@ impl WriteFailure {
...
 }
 
+/// An error that occurred during a GridFS operation.
+#[derive(Clone, Debug)]
+#[allow(missing_docs)]
+#[non_exhaustive]
+pub enum GridFsErrorKind {
+    /// The file with the given identifier was not found.
+    #[non_exhaustive]
+    FileNotFound { identifier: GridFsFileIdentifier },
```
isabelatkinson, line 708: This is a little convoluted but the `FileNotFound` error can be returned from either `download` (which has an id) or `download_by_name` (which has a filename), and it didn't seem much better to have separate error cases for those.

src/gridfs/download.rs
isabelatkinson: Good idea, added an enum to model this. I ended up including a variant for each of the errors returned, but that approach might be worth revisiting if the list of possible errors becomes too bloated as the rest of the API is implemented.

src/error.rs
```diff
@@ -693,6 +698,61 @@ impl WriteFailure {
...
+    fn from(filename: &str) -> Self {
+        Self::Filename(filename.into())
+    }
+}
+
+impl From<Bson> for GridFsFileIdentifier {
+    fn from(id: Bson) -> Self {
+        Self::Id(id)
+    }
+}
```
patrickfreed, line 754: I think it might actually be better to omit these since a string can be used to represent either a filename or an ID. Having to explicitly indicate which variant helps avoid this ambiguity.

src/gridfs.rs
patrickfreed: Follow up PR sounds good to me! Sorry, yeah I missed your previous comment.

src/gridfs/download.rs
patrickfreed: Yeah, the concern is that the data gets messed up in some other way by a different driver or application (i.e. it's untrusted). Basically, don't want to accidentally use a junk value in the logic that follows. Seems mostly harmless, but I think it would be good to explicitly verify that.

src/gridfs.rs
```diff
@@ -1,52 +1,72 @@
...
 // Contained in a "chunks" collection for each user file
+#[derive(Debug, Deserialize, Serialize)]
 struct Chunk {
+    #[serde(rename = "_id")]
     id: ObjectId,
     files_id: Bson,
     n: u32,
     // default size is 255 KiB
-    data: Vec<u8>,
+    data: Binary,
```
patrickfreed, line 39: This would be a good opportunity to leverage borrowed deserialization via `RawBinaryRef` to save on some potentially big copies. You'll need to throw a `#[serde(borrow)]` annotation on too.

isabelatkinson: good idea, updated

src/gridfs/download.rs
patrickfreed: Ah ok my bad, I thought it was an `i32` or `i64` being cast to a `u64`.

patrickfreed (approved):
LGTM! tagging in the other reviews.

Going to leave my docs related commends unresolved so you can go back to them when doing the follow on PR.

src/test/spec/unified_runner/test_runner.rs
kmahar: nice 😎

src/gridfs/download.rs
abr-egn: Since the `to_futures_0_3` methods are just using `compat_write` under the hood and are temporary, what would people think about leaving them out entirely and recommending users do this directly if needed in documentation?

abr-egn (approved):
LGTM!

src/test/spec/unified_runner/operation.rs
```diff
@@ -2640,6 +2643,62 @@ impl TestOperation for CreateEntities {
...
+    fn execute_entity_operation<'a>(
+        &'a self,
+        id: &'a str,
+        test_runner: &'a TestRunner,
+    ) -> BoxFuture<'a, Result<Option<Entity>>> {
+        async move {
+            let bucket = test_runner.get_bucket(id).await;
+            let mut buf: Vec<u8> = vec![];
+            bucket
+                .download_to_futures_0_3_writer(self.id.clone(), &mut buf)
```
kmahar, line 2662: I know we talked at some point about figuring out a way to get coverage of the various upload/download methods via the unified tests. but maybe using the futures methods makes sense here since (I think) they end up leading us to call through to the other versions of the method anyway?
if that's the case I think it might be worth explaining in a comment here why we chose this variant of `download` (likewise for `DownloadByName`)

src/error.rs
```diff
@@ -693,6 +698,49 @@ impl WriteFailure {
...
+    WrongNumberOfChunks {
+        actual_number: u32,
+        expected_number: u32,
+    },
+}
+
+/// An identifier for a file stored in a GridFS bucket.
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub enum GridFsFileIdentifier {
```
kmahar, line 736: note: I thought about whether it would make sense to accept this type in the API methods (I guess`String` and `Bson` could implement `Into<GridFsFileIdentifier>` so users wouldn't have to construct identifiers?) and avoid having separate methods for download by filename and ID. 
but I think the separate named methods is probably a little nicer in terms of discovering the allowed functionality, and it allows us to separately document the behavior of querying by filename vs ID. so I think the current API is fine.

src/gridfs/download.rs
kmahar: @abr-egn I'd be fine with that

kmahar (commented):
a few minor questions/comments but LGTM overall!

src/gridfs/download.rs
patrickfreed: It's a little inconsistent with our various cursor APIs that implement the `futures` version of `Stream`, but having seen what the implementation actually entails here, I think I agree that maybe these separate methods probably aren't worth it. Instead, if we want to have consistency, maybe we should just also implement tokio's `Stream` on our cursor types. Then we can omit the `futures` one here.

src/gridfs.rs
```diff
@@ -1,52 +1,72 @@
...
+    created_indexes: AtomicBool,
 }
 
 /// Struct for storing GridFS managed files within a [`Database`].
+#[derive(Debug, Clone)]
 pub struct GridFsBucket {
-    // Contains a "chunks" collection
-    pub(crate) db: Database,
-    pub(crate) options: GridFsBucketOptions,
+    inner: Arc<GridFsBucketInner>,
```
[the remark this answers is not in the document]
isabelatkinson: This pattern is in line with some of our other top-level types -- it allows users to clone them cheaply and share them across threads/tasks.

src/gridfs/download.rs
isabelatkinson: Would it make more sense to have only the `futures` implementations then? That way we're runtime-agnostic and in line with what we already support for `Cursor`. The tokio compat utils go both ways, so I think the user conversion code would be basically the same regardless of which library we choose to support in our API.

src/test/spec/unified_runner/operation.rs
isabelatkinson: Yeah, I chose to use the `futures` API because it runs through the `tokio` API, so we're getting coverage for both here. It seems like we're leaning towards only having one API but I'll add a comment here explaining why `futures` is used if we decide to keep both.

isabelatkinson:
Updated to remove the tokio methods. I initially wrote a test to verify that this works but removed it in favor of adding a doc example when I refresh all of the documentation which will give the same coverage.

patrickfreed (approved):

kmahar (approved):

PR #751, opened 2022-09-26
RUST-1478 GridFS upload methods

isabelatkinson:
Implements the `upload_from_stream` and `upload_from_stream_with_id` methods described in this section of the spec: https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#file-upload

### DIFF
--- a/src/gridfs/upload.rs
+++ b/src/gridfs/upload.rs
@@ -1,0 +1,162 @@
+use std::{marker::Unpin, sync::atomic::Ordering};
+
+use futures_util::{
+    io::{AsyncRead, AsyncReadExt},
+    stream::TryStreamExt,
+};
+
+use super::{options::GridFsUploadOptions, Chunk, FilesCollectionDocument, GridFsBucket};
+use crate::{
+    bson::{doc, oid::ObjectId, spec::BinarySubtype, Bson, DateTime, Document, RawBinaryRef},
+    bson_util::get_int,
+    error::{ErrorKind, Result},
+    index::IndexModel,
+    options::{FindOneOptions, ReadPreference, SelectionCriteria},
+    Collection,
+};
+
+impl GridFsBucket {
+    /// Uploads a user file to a GridFS bucket. Bytes are read from `source` and stored in chunks in
+    /// the bucket's chunks collection. After all the chunks have been uploaded, a corresponding
+    /// [`FilesCollectionDocument`] is stored in the bucket's files collection.
+    ///
+    /// This method generates an [`ObjectId`] for the `files_id` field of the
+    /// [`FilesCollectionDocument`] and returns it.
+    pub async fn upload_from_futures_0_3_reader<T>(
+        &self,
+        filename: impl AsRef<str>,
+        source: T,
+        options: impl Into<Option<GridFsUploadOptions>>,
+    ) -> Result<ObjectId>
+    where
+        T: AsyncRead + Unpin,
+    {
+        let id = ObjectId::new();
+        self.upload_from_futures_0_3_reader_with_id(id.into(), filename, source, options)
+            .await?;
+        Ok(id)
+    }
+
+    /// Uploads a user file to a GridFS bucket with the given `files_id`. Bytes are read from
+    /// `source` and stored in chunks in the bucket's chunks collection. After all the chunks have
+    /// been uploaded, a corresponding [`FilesCollectionDocument`] is stored in the bucket's files
+    /// collection.
+    pub async fn upload_from_futures_0_3_reader_with_id<T>(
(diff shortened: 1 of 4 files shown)


kmahar (commented):
looks good! just a couple of small comments re: things from the spec.

src/gridfs/upload.rs
```diff
@@ -0,0 +1,147 @@
...
+            .unwrap_or_else(|| self.chunk_size_bytes());
+        let mut length = 0u64;
+        let mut n = 0;
+
+        loop {
+            let mut buf = vec![0u8; chunk_size as usize];
+            let bytes_read = match source.read(&mut buf).await {
+                Ok(0) => break,
+                Ok(n) => n,
+                Err(error) => {
```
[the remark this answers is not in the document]
isabelatkinson: The mechanism to abort an upload is more relevant for `open_download_stream`, when users are incrementally writing their chunks to the stream and closing it out themselves. I think a user could possibly somehow close the source while this method is ongoing, but that would be addressed here, as the call to `read` would fail.

src/gridfs/upload.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        Ok(())
+    }
+
+    async fn create_indexes(&self) -> Result<()> {
+        if !self.inner.created_indexes.load(Ordering::SeqCst) {
+            let find_options = FindOneOptions::builder()
+                .selection_criteria(SelectionCriteria::ReadPreference(ReadPreference::Primary))
+                .build();
+            if self.files().find_one(None, find_options).await?.is_none() {
```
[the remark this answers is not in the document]
isabelatkinson: Good catch, updated. The projection will cause `find_one` to fail on `self.files()` if a doc gets returned because it won't match the `FilesCollectionDocument` schema, so I also added a `clone_with_type::<Document>` call to avoid this.

src/gridfs/upload.rs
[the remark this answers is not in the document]
kmahar: ah that makes sense. thanks 👍

kmahar (approved):

src/gridfs/upload.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        let chunk_size = options
+            .as_ref()
+            .and_then(|opts| opts.chunk_size_bytes)
+            .unwrap_or_else(|| self.chunk_size_bytes());
+        let mut length = 0u64;
+        let mut n = 0;
+
+        loop {
+            let mut buf = vec![0u8; chunk_size as usize];
```
abr-egn, line 66: Micro-optimization nit: since `chunk_size` is static during iteration, it might be more efficient to construct `buf` outside the loop.

src/test/spec/unified_runner/operation.rs
```diff
@@ -2699,6 +2700,49 @@ impl TestOperation for DownloadByName {
...
+impl TestOperation for Upload {
+    fn execute_entity_operation<'a>(
+        &'a self,
+        id: &'a str,
+        test_runner: &'a TestRunner,
+    ) -> BoxFuture<'a, Result<Option<Entity>>> {
+        async move {
+            let bucket = test_runner.get_bucket(id).await;
+            let hex_bytes = self.source.get("$$hexBytes").unwrap().as_str().unwrap();
+            // Iterates over two characters of the hex string at a time, parses each pair into a
```
abr-egn, line 2727: Could this use `hex::decode`?

src/gridfs/upload.rs
patrickfreed: As part of this, rather than truncating each time, we'll probably need to just slice the buffer.

src/gridfs/upload.rs
```diff
@@ -0,0 +1,154 @@
...
+        // create them.
+        let mut indexes = coll.list_indexes(None).await?;
+        'outer: while let Some(index_model) = indexes.try_next().await? {
+            if index_model.keys.len() != keys.len() {
+                continue;
+            }
+            // Indexes should be considered equivalent regardless of numeric value type.
+            // e.g. { "filename": 1, "uploadDate": 1 } is equivalent to
+            // { "filename": 1.0, "uploadDate": 1.0 }
+            let number_matches = |key: &str| index_model.keys.get(key).and_then(get_int) == Some(1);
```
patrickfreed, line 140: I know the indexes we create all currently use 1 for the value, but I think it would be good to avoid future pitfalls by checking against `get_int` of the input document too instead of `Some(1)`.

patrickfreed (approved):

abr-egn (approved):

PR #756, opened 2022-10-03
minor: remove expansion.yml after load

abr-egn:
I'm not sure how to test this short of pushing another release, but this _should_ fix the problem we had with 2.3.1.


kmahar (approved):
lgtm, let's hope this is the final fix 🤞

PR #757, opened 2022-10-05
RUST-580 Add tracing integration and command tracing events

kmahar:
This is the Rust implementation of the changes detailed in mongodb/specifications#1303. Various comments inline.


.evergreen/run-tests.sh
```diff
@@ -10,7 +10,7 @@ if [ "$SINGLE_THREAD" = true ]; then
 	OPTIONS="$OPTIONS --test-threads=1"
 fi
 
-FEATURE_FLAGS="zstd-compression,snappy-compression,zlib-compression,${TLS_FEATURE}"
+FEATURE_FLAGS="zstd-compression,snappy-compression,zlib-compression,tracing-unstable,${TLS_FEATURE}"
```
kmahar, line 13: I initially had planned to introduce separate tasks for testing tracing to avoid introducing any potential extra overhead from having a tracing subscriber registered in every test. however, I tried out just running the tracing tests alongside the other ones, and the total runtime for a full patch seemed to fall within the usual bounds; for the last 10 commits to main patches took anywhere from 14 days, 7 hours to 16 days, 2 hours computing time, and both of mine took around 15 days, 8 hours. 

So there does not seem to be a ton of overhead. also, if we split out the tracing tasks we would need to separately install rustup, cargo, and driver deps for even more tasks which would probably cancel out any speedup in the other tests.

Cargo.toml
```diff
@@ -70,6 +70,11 @@ snappy-compression = ["snap"]
 # DO NOT USE; see https://jira.mongodb.org/browse/RUST-569 for the status of CSFLE support in the Rust driver.
 csfle = ["mongocrypt", "rayon"]
 
+# DO NOT USE; see https://jira.mongodb.org/browse/RUST-580 for the status of tracing/logging support in the Rust driver.
+# Enables support for emitting tracing events.
+# TODO: pending https://github.com/tokio-rs/tracing/issues/2036 stop depending directly on log.
```
kmahar, line 75: if this doesn't end up getting in before we release I don't think it is a huge deal because we already transitively depend on log via a bunch of our dependencies so we're not adding anything extra to compile here.

src/error.rs
```diff
@@ -40,7 +40,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// [`ErrorKind`](enum.ErrorKind.html) is wrapped in an `Arc` to allow the errors to be
 /// cloned.
 #[derive(Clone, Debug, Error)]
-#[error("{kind}")]
+#[error("Kind: {kind}, labels: {labels:?}")]
```
kmahar, line 43: currently we omitted the labels from the Display implementations so I added that. I currently use the Display impls to generate the string representations of errors to attach to "command failed" events. (tracing events only support attaching a few primitive types like strings and various numbers.)

otherwise, I wanted to avoid going down a rabbit hole of hashing out the ideal string representations for every error type in the driver so I just left those as-is. since they are just string representations I don't think there are any semver concerns, and if we find issues with them in the future we could improve them as we go.

src/error.rs
```diff
@@ -551,7 +561,11 @@ pub struct CommandError {
 
 impl fmt::Display for CommandError {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        write!(fmt, "({}): {})", self.code_name, self.message)
+        write!(
```
kmahar, line 564: we weren't actually including the error code here before.

src/test/spec/command_monitoring/mod.rs
```diff
@@ -11,8 +11,9 @@ async fn command_monitoring_unified() {
         // This test relies on old OP_QUERY behavior that many drivers still use for < 4.4, but we do not use, due to never implementing OP_QUERY.
         tc.description != "A successful find event with a getmore and the server kills the cursor (<= 4.4)";
 
-    run_spec_test_with_path(&["command-monitoring", "unified"], |path, f| {
-        run_unified_format_test_filtered(path, f, pred)
-    })
+    run_spec_test_with_path(
```
kmahar, line 14: the "command-monitoring" directly has been renamed to "command-logging-and-monitoring" (CLAM!). monitoring tests now live in a "monitoring" subdirectory, and logging tests in a "logging" subdirectory.

src/test/spec/trace.rs
```diff
@@ -0,0 +1,215 @@
...
+    assert_eq!(reply.len(), 8); // 5 + 3 for trailing "..."
+
+    // TODO RUST-1405: when we expose the full server reply for command errors, we should confirm
+    // that gets correctly truncated in command failed events here as well.
+}
+
+/// Prose test 3: mid-codepoint truncation
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn command_logging_truncation_mid_codepoint() {
```
kmahar, line 143: this test is unfortunately really brittle to changes in the server response, but I don't think there's really a way around that

src/error.rs
```diff
@@ -370,6 +370,16 @@ impl Error {
             _ => None,
         }
     }
+
+    /// Per the CLAM spec, for sensitive commands we must redact everything besides the
+    /// error labels, error code, and error code name from errors received in response to
+    /// sensitive commands. Currently, the only field besides those that we expose is the
+    /// error message.
+    pub(crate) fn redact(&mut self) {
```
kmahar, line 378: you can find the ongoing thread about error redaction [here](https://github.com/mongodb/specifications/pull/1303#discussion_r973265187). 
I think we have consensus on what the redaction behavior should be (only redact command errors, and only redact fields that are not code, codename, and labels) but not necessarily consensus yet on the testing approach for it.

src/test/spec/unified_runner/matcher.rs
```diff
@@ -39,6 +46,72 @@ pub(crate) fn events_match(
...
+        ));
+    }
+    if actual.level != expected.level {
+        return Err(format!(
+            "Expected and actual tracing event levels do not match: expected {}. got {}",
+            expected.level, actual.level
+        ));
+    }
+
+    if let Some(has_failure) = expected.has_failure {
```
kmahar, line 67: this logic is likely going to change based on the thread I link above: re error redaction. 

potentially, we'll end up with both `has_failure` (which would simply check whether there is a key named "failure") and then `failure_is_redacted` which would assert that the error has been redacted. I wrote a sort of quick and dirty implementation of that here which checks for "REDACTED" but if this ends up being the path forward I think we could maybe match the error string against a regex too to confirm that the message is the redacted part.

src/test/util/event.rs
```diff
@@ -464,74 +458,7 @@ impl CommandEventHandler for EventHandler {
...
-#[derive(Debug)]
-pub(crate) struct EventSubscriber<'a> {
-    /// A reference to the handler this subscriber is receiving events from.
-    /// Stored here to ensure this subscriber cannot outlive the handler that is generating its
-    /// events.
-    _handler: &'a EventHandler,
-    receiver: tokio::sync::broadcast::Receiver<Event>,
-}
-
-impl EventSubscriber<'_> {
```
kmahar, line 476: this type has been made generic over the handler and event types and moved into `src/test/util/subscriber.rs` so we can reuse it for tracing events as well. 
`wait_for_successful_command_execution` is only meaningful for an `EventSubscriber<'_, EventHandler, Event>` though so I left that method here.

src/trace/command.rs
```diff
@@ -0,0 +1,128 @@
...
+fn serialize_command_or_reply(doc: bson::Document, max_length_bytes: usize) -> String {
+    let mut ext_json = doc.tracing_representation();
+    truncate_on_char_boundary(&mut ext_json, max_length_bytes);
+    ext_json
+}
+
+// TODO: subject to change based on what exact version of truncation we decide to go with.
+/// Truncates the given string at the closest UTF-8 character boundary >= the provided length.
+/// If the new length is >= the current length, does nothing.
+pub(crate) fn truncate_on_char_boundary(s: &mut String, new_len: usize) {
```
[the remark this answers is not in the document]
patrickfreed: Sounds good to me to keep this for now. Can we file a RUST ticket to track the `ceil_char_boundary` portion?

Also, if we did roll our own extJSON serializer that did support graceful truncation, would the spec allow for that? I think it would require a `"truncated": true` field or something to be added to the JSON instead of an ellipsis to ensure the JSON is valid.

patrickfreed (commented):
Looks great! I only have some minor suggestions about docstrings / comments and a few questions

src/test/spec/unified_runner/matcher.rs
kmahar: this logic has been updated. note that we redact a few different error types now and not just command errors. before I go down the path of writing regexes matching the Display impl for each possible type of error that gets redacted, do we think that is worthwhile or have any other ideas for how to make these assertions a bit more robust?

src/test/spec/unified_runner/matcher.rs
```diff
@@ -39,6 +46,74 @@ pub(crate) fn events_match(
...
+        ));
+    }
+    if actual.level != expected.level {
+        return Err(format!(
+            "Expected and actual tracing event levels do not match: expected {}. got {}",
+            expected.level, actual.level
+        ));
+    }
+
+    if let Some(failure_is_redacted) = expected.failure_is_redacted {
```
kmahar, line 67: copying from a now outdated thread for visibility: this logic has been updated. note that we redact a few different error types now and not just command errors. before I go down the path of writing regexes matching the Display impl for each possible type of error that gets redacted, do we think that is worthwhile or have any other ideas for how to make these assertions a bit more robust?

src/error.rs
patrickfreed: Makes sense, thanks for the explanation!

src/trace/command.rs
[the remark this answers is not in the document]
patrickfreed: Sounds good, thanks for another clear + detailed explanation. Java not having any complaints seems like pretty good signal that the "dumb" approach should be good for us :+1:

src/client/options/mod.rs
```diff
@@ -575,6 +588,12 @@ pub(crate) struct TestOptions {
 
     /// Mock response for `SrvPollingMonitor::lookup_hosts`.
     pub(crate) mock_lookup_hosts: Option<Result<LookupHosts>>,
+
+    /// Optionally stores the unified test runner entity ID for this client.
+    /// Used to include client IDs in tracing events which allows correlating
+    /// the events with specific clients.
+    #[cfg(feature = "tracing-unstable")]
+    pub(crate) client_id: Option<String>,
```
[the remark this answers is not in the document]
patrickfreed: >So this property is actually on TestOptions, which is only defined under #[cfg(test)]. so I'm not sure it makes sense to add that comment here given this definition is not included in the driver. (the way GH collapses the intermediate code does make it look like this is on ClientOptions though lol)

Ah ok, I missed that. I thought this was on `ClientOptions`. No need for a comment here.

>that said, is there any nice way to avoid the unconditional definition there?

Hmm I don't think so, or at least I don't know of any way around it. You could look into the generated code of the macro and possibly use it there, but I doubt that will end up being cleaner overall.

src/trace.rs
```diff
@@ -0,0 +1,81 @@
...
+            TracingOrLogLevel::Info => tracing::Level::INFO,
+            TracingOrLogLevel::Debug => tracing::Level::DEBUG,
+            TracingOrLogLevel::Trace => tracing::Level::TRACE,
+        }
+    }
+}
+
+/// Emits a tracing event at debug level. In tests, the provided client ID will be added as field to
+/// each event to enable distinguishing which client produced the event. Outside of tests, the ID
+/// will be ignored.
```
patrickfreed, line 63: I wonder if including client ID in non-tests scenarios would actually be useful too, since it would allow a user to distinguish between clients if they happen to use multiple in their applications. I feel like we may have discussed this before (and I apologize if that is the case!), but is there a reason why we wouldn't want this in the actual driver?

src/client/mod.rs
```diff
@@ -145,9 +153,30 @@ impl Client {
         Ok(client)
     }
 
-    pub(crate) fn emit_command_event(&self, emit: impl FnOnce(&Arc<dyn CommandEventHandler>)) {
+    pub(crate) fn emit_command_event(&self, generate_event: impl Fn() -> CommandEvent) {
```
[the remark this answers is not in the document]
patrickfreed: Hah yeah, this case is particularly tricky in Rust's type system. When I originally reviewed it I couldn't think of a great way to do it hence my suggestion about just leaving a comment, but I think the solution you ended up with looks great!

patrickfreed (commented):
Looks great! I only have a few really minor comments / questions, so going to tag in other reviews so they can get started too.

src/test/spec/unified_runner/matcher.rs
kmahar: > Alternatively, we could just replace the entire Display implementation of redacted errors with "REDACTED" and then perform a strict equality, though we'd lose some information that way. Given that this redaction is pretty rare (it's just for when the user uses run_command to create users and what not, right?), I don't think it would be the worst thing in the world.

so we decided in the spec that drivers should not redact error code, error code name, and error labels, so I don't think we'd want to take this approach as we'd strip out everything.

what I ended up doing: 
* writing regexes here for the two types of errors that actually show up and have redaction asserted on them in spec tests; using those to extract components of the errors and assert redaction/lack thereof
* adding unit tests in test/spec/trace.rs that construct all the various types of server errors we redact, call `redact()` on them, then assert that those properties have been redacted as expected

src/test/spec/unified_runner/operation.rs
```diff
@@ -101,6 +101,12 @@ pub(crate) trait TestOperation: Debug + Send + Sync {
     fn returns_root_documents(&self) -> bool {
         false
     }
+
+    /// If this operation specifies entities to create, returns those entities. Otherwise,
+    /// returns None.
+    fn as_test_file_entities(&self) -> Option<&Vec<TestFileEntity>> {
```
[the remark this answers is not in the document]
kmahar: oh yeah, I think that name was a carry over from a previous version of the method which returned the `CreateEntities` operation itself rather than just its entities. renamed.

src/test/spec/unified_runner/matcher.rs
```diff
@@ -39,6 +48,145 @@ pub(crate) fn events_match(
...
+    }
+
+    if let Some(failure_should_be_redacted) = expected.failure_is_redacted {
+        match actual.fields.get("failure") {
+            Some(failure) => {
+                match failure {
+                    TracingEventValue::String(failure_str) => {
+                        // lazy_static saves us having to recompile this regex every time this
+                        // function is called.
+                        lazy_static! {
```
kmahar, line 76: as an aside, I kind of assumed that a nested `lazy_static` like this would get re-evaluated on every function call, but this was explicitly recommended in the regex crate docs as a way to avoid re-compiling within helper methods that use regexes: https://docs.rs/regex/latest/regex/index.html#example-avoid-compiling-the-same-regex-in-a-loop

src/trace.rs
patrickfreed: 1. I don't think its strictly necessary to include in the public API right now, though it could definitely be useful. I'd lean towards leaving it out for now in case something like this gets stabilized across drivers.
2. An ObjectID / UUID would also work fine and not require synchronization across client instances. As an aside, I wonder if we could use the Topology ID for this? We already publish it for TopologyOpening events, and AFAIK there is only one generated per client's lifetime.

src/test/spec/unified_runner/matcher.rs
patrickfreed: LGTM

src/trace.rs
kmahar: good idea re: the topology ID. it does appear it is generated just once when the `Topology` is initialized. 
I refactored to use that instead, and to add it to events unconditionally (i.e. not based on whether we are in tests). 

I named it `topology_id`, because I figured that's the name we'll go with when including it in SDAM tracing events, and this will make it more obvious that it's the same type of ID and can be used to correlate SDAM events with other events.

abr-egn (commented):
Looks good, just a few smallish comments.

src/event/command.rs
```diff
@@ -144,3 +144,19 @@ pub trait CommandEventHandler: Send + Sync {
     /// whenever a database command fails to complete successfully.
     fn handle_command_failed_event(&self, _event: CommandFailedEvent) {}
 }
+
+#[derive(Clone)]
+pub(crate) enum CommandEvent {
```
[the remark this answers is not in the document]
kmahar: good point. I've combined the test `CommandEvent` type into this one. I will tackle combining the SDAM ones when adding SDAM logging

src/test/spec/unified_runner/matcher.rs
```diff
@@ -39,6 +46,148 @@ pub(crate) fn events_match(
...
+                                r"^Kind: I/O error: (?P<message>.+), labels: (?P<labels>.+)$"
+                            ).unwrap();
+                        }
+
+                        // We redact all server-returned errors, however at this time the only types
+                        // of errors that show up in tracing redaction tests
+                        // are command errors (which should be redacted) and
+                        // I/O errors (which should not redacted). redaction of other errors is
+                        // unit-tested in src/test/spec/trace.rs.
+                        if COMMAND_FAILED_REGEX.is_match(failure_str) {
```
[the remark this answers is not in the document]
kmahar: nice; updated here and for the similar case below with IO_ERROR_REGEX

src/test/spec/unified_runner/matcher.rs
kmahar: added comments in the relevant places for command errors and I/O errors 👍

src/trace.rs
patrickfreed: Given that we're using topology ID now (which every driver should have), would that be something to consider upstreaming into the spec as a requirement?

patrickfreed (approved):
LGTM!

src/trace.rs
kmahar: good question; I started a thread in the spec channel about this.

abr-egn (approved):

src/test/spec/unified_runner/test_file.rs
```diff
@@ -164,6 +168,9 @@ pub(crate) struct Client {
     #[serde(default, deserialize_with = "deserialize_server_api_test_format")]
     pub(crate) server_api: Option<ServerApi>,
     pub(crate) store_events_as_entities: Option<Vec<StoreEventsAsEntity>>,
+    #[cfg(feature = "tracing-unstable")]
+    #[serde(default, deserialize_with = "deserialize_tracing_level_map")]
+    pub observe_log_messages: Option<HashMap<String, tracing::Level>>,
```
isabelatkinson, line 173:
```suggestion
    pub(crate) observe_log_messages: Option<HashMap<String, tracing::Level>>,
```
(small nit, i think we decided upon using `pub(crate)` for internal test stuff)

isabelatkinson (commented):
looks great! just a few very small comments

src/client/options/mod.rs
```diff
@@ -530,6 +530,20 @@ pub struct ClientOptions {
     #[builder(default)]
     pub tls: Option<Tls>,
 
+    /// The maximum number of bytes that the driver should include in a tracing event
+    /// or log message's extended JSON string representation of a BSON document, e.g. a
+    /// command or reply from the server.
+    /// If truncation of a document at the exact specified length would occur in the middle
+    /// of a Unicode codepoint, the document will be truncated at the closest larger length
+    /// which falls on a boundary between codepoints.
+    /// Note that in cases where truncation occurs the output will not be valid JSON.
```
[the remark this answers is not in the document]
isabelatkinson: actually never mind, I got mixed up here. "will" is the right word to use

src/test/spec/unified_runner/test_file.rs
kmahar: oh yeah - updated

src/test/mod.rs
```diff
@@ -87,6 +88,26 @@ lazy_static! {
...
+// needs to be defined separately.
+#[cfg(feature = "tracing-unstable")]
+lazy_static! {
+    /// A global default tracing handler that will be installed the first time this
+    /// value is accessed. A global handler must be used anytime the multi-threaded
+    /// test runtime is in use, as non-global handlers only apply to the thread
+    /// they are registered in.
+    /// By default this handler will collect no tracing events.
+    /// Its minimum severity levels can be configured on a per-component basis using
+    /// [`TracingHandler:set_levels`]. The test lock MUST be acquired in any test
```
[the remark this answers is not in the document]
kmahar: yes; updated

isabelatkinson (approved):

PR #758, opened 2022-10-05
RUST-1386 Implement the first batch of FLE prose tests

abr-egn:
RUST-1386

This implements a bunch of the FLE prose tests, the evergreen config to support running them, and the fixes for the various bugs the tests showed up (all pretty minor).  Evergreen is configured to only run the csfle tests on a single version/os here; I expect that getting this running on other configurations will require minor tweaking which didn't seem worth including in this initial round.


.evergreen/config.yml
```diff
@@ -460,6 +473,52 @@ functions:
...
+      working_dir: "src"
+      background: true
+      script: |
+        ${PREPARE_SHELL}
+        cd ${DRIVERS_TOOLS}/.evergreen/csfle
+        . ./activate_venv.sh
+        # TMPDIR is required to avoid "AF_UNIX path too long" errors.
+        TMPDIR="$(dirname ${DRIVERS_TOOLS})" python3 kms_kmip_server.py
+
+  "run csfle tests":
```
abr-egn, line 501: I was initially intending to have the csfle tests just run as part of the main batch, but given the combination of extra setup needed and extra restrictions on server versions, it seemed better to leave them out and run as their own task.

.evergreen/find-crypt_shared.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
abr-egn, line 1: The crypt_shared library is named different things on different OSs, and the exact name needs to be passed into libmongocrypt.  Fortunately, the fetcher puts it as the sole file in the `lib/` directory, so this script just globs and validates that there's a single result.

src/client/csfle/client_encryption.rs
```diff
@@ -62,8 +64,8 @@ impl ClientEncryption {
     /// Returns the _id of the created document as a UUID (BSON binary subtype 0x04).
     pub async fn create_data_key(
         &self,
-        kms_provider: KmsProvider,
-        opts: impl Into<Option<DataKeyOptions>>,
+        kms_provider: &KmsProvider,
```
abr-egn, line 67: Not strictly a bugfix, but I realized from usage that there's no reason this (and other similar methods) need to consume the args, and it's often convenient to be able to reuse options.

src/test/csfle.rs
```diff
@@ -0,0 +1,1184 @@
...
+async fn corpus_local_schema() -> Result<()> {
+    if !check_env("corpus_local_schema", true) {
+        return Ok(());
+    }
+    let _guard = LOCK.run_exclusively().await;
+    run_corpus_test(true).await?;
+    Ok(())
+}
+
+async fn run_corpus_test(local_schema: bool) -> Result<()> {
```
isabelatkinson, line 629: I'm curious if it would be any easier/more readable to model the corpus docs here with structs and use serde rather than parse them out directly. Was this something you tried/considered? There might be a lot of deserialization customization that would make it overkill...

src/test/csfle.rs
```diff
@@ -0,0 +1,1184 @@
...
+                .algorithm(Algorithm::AeadAes256CbcHmacSha512Deterministic)
+                .build(),
+        )
+        .await?;
+    let decrypted = client_encryption.decrypt(encrypted.as_raw_binary()).await?;
+    assert_eq!(value, decrypted);
+    Ok(())
+}
+
+async fn custom_endpoint_aws_ok(endpoint: Option<String>) -> Result<()> {
```
isabelatkinson, line 875: nit: can we add which test number these tests correspond to in comments?

src/test/csfle.rs
```diff
@@ -0,0 +1,1184 @@
...
+            key_version: None,
+            endpoint: Some("doesnotexist.invalid:443".to_string()),
+        })
+        .build();
+
+    let client_encryption = custom_endpoint_setup(true).await?;
+    let result = client_encryption
+        .create_data_key(&KmsProvider::Gcp, &key_options)
+        .await;
+    assert!(result.unwrap_err().is_csfle_error());
```
isabelatkinson, line 1105: I think we need a string check here as well, from the spec:

> Expect this to fail with an exception with a message containing the string: "Invalid KMS response".

src/client/csfle/client_encryption.rs
```diff
@@ -271,42 +273,52 @@ pub struct DataKeyOptions {
...
 #[allow(missing_docs)]
 pub enum MasterKey {
+    #[serde(rename_all = "camelCase")]
     Aws {
         region: String,
         /// The Amazon Resource Name (ARN) to the AWS customer master key (CMK).
         key: String,
         /// An alternate host identifier to send KMS requests to. May include port number. Defaults
         /// to "kms.<region>.amazonaws.com"
+        #[serde(skip_serializing_if = "Option::is_none")]
```
isabelatkinson, line 287: Can these be replaced with a `#[serde_with::skip_serializing_none]` tag on the entire enum? Or does that not propagate to the nested fields?

isabelatkinson (commented):
going to tag in the rest of the team for review so i'm not blocking; i'll do a final pass when i'm back

src/test/csfle.rs
```diff
@@ -0,0 +1,1184 @@
...
+    doc.insert("_id", "encryption_exceeds_16mib");
+    let mut value = String::new();
+    for _ in 0..(16777216 - 2000) {
+        value.push('a');
+    }
+    doc.insert("unencrypted", value);
+    let result = coll.insert_one(doc, None).await;
+    assert!(result.is_err());
+
+    Ok(())
```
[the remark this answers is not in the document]
abr-egn: Ah, good catch!  Added.

src/test/csfle.rs
```diff
@@ -0,0 +1,1184 @@
...
+    Ok(bson::Binary {
+        subtype: BinarySubtype::Uuid,
+        bytes: base64::decode(bytes.as_ref())?,
+    })
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn external_key_vault() -> Result<()> {
+    if !check_env("external_key_vault", true) {
```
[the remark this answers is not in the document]
patrickfreed: Yeah I also stopped using it due to rust-analyzer problems and also sometimes I think it would mess with backtraces in failed tests.

src/client/csfle/client_encryption.rs
patrickfreed: I'm a little bit concerned about the inconsistency between this and our other methods that take options, since in all those other cases we accept them by value. Based on your usage thus far, is accepting a reference here (for `DataKeyOptions` at least) a big benefit? If not, then it might be better to keep it by value to prevent user compilation errors in the future. If it saves a ton of copying in a hot path or otherwise makes the users life a lot easier, then I think the inconsistency would be okay.

patrickfreed (commented):
Looks good! Have some minor comments from a first pass, will doe another one tomorrow

src/test/csfle.rs
```diff
@@ -0,0 +1,1192 @@
...
+            )
+            .await?;
+        assert_eq!(other_encrypted.subtype, BinarySubtype::Encrypted);
+        assert_eq!(other_encrypted.bytes, encrypted.bytes);
+
+        // Attempt to auto-encrypt an already encrypted field.
+        let result = coll
+            .insert_one(doc! { "encrypted_placeholder": encrypted }, None)
+            .await;
+        assert!(result.is_err());
```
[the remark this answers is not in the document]
abr-egn: This isn't an auth error (and the test spec doesn't expect it to be), it's a csfle error - added an assert to verify.

src/test/csfle.rs
```diff
@@ -0,0 +1,1184 @@
...
+            log_uncaptured(format!("skipping {}: KMIP requires openssl", name));
+            return false;
+        }
+    }
+    true
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn custom_key_material() -> Result<()> {
```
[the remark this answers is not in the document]
abr-egn: Happily, we get to have both - I switched to `anyhow::Error`, which includes a backtrace.

src/test/csfle.rs
```diff
@@ -0,0 +1,1192 @@
...
+        {
+            // rustls is incompatible with the driver-tools kmip server.
+            log_uncaptured(format!("skipping {}: KMIP requires openssl", name));
+            return false;
+        }
+    }
+    true
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
```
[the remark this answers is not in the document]
abr-egn: Done.

src/client/csfle/client_encryption.rs
abr-egn: It's not likely to be part of a hot path, but it is convenient when creating multiple data keys from a given provider (see the various `custom_endpoint_*` tests).  The ergonomics win isn't huge, it's just saving a `.clone()`, but the inconsistency seems like a very minor concern in this case.

src/test/csfle.rs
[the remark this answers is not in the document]
abr-egn: Filed RUST-1511 for this.

src/test/csfle.rs
```diff
@@ -0,0 +1,1261 @@
...
+        let client_encryption = ClientEncryption::new(enc_opts)?;
+
+        // Test: encrypted client.
+        let result = client_encrypted
+            .database("db")
+            .collection::<Document>("coll")
+            .insert_one(doc! { "encrypted": "test" }, None)
+            .await;
+        if with_external_key_vault {
+            assert!(result.is_err());
```
kmahar, line 422: > expect an authentication exception to be thrown

can we assert it's an auth error?

src/test/csfle.rs
```diff
@@ -0,0 +1,1261 @@
...
+            .encrypt(
+                RawBson::String("test".to_string()),
+                &EncryptOptions::builder()
+                    .key(EncryptKey::Id(base64_uuid("LOCALAAAAAAAAAAAAAAAAA==")?))
+                    .algorithm(Algorithm::AeadAes256CbcHmacSha512Deterministic)
+                    .build(),
+            )
+            .await;
+        if with_external_key_vault {
+            assert!(result.is_err());
```
kmahar, line 441: same comment re asserting its an auth error

src/test/csfle.rs
```diff
@@ -0,0 +1,1261 @@
...
+    datakeys
+        .insert_one(load_testdata("limits-key.json")?, None)
+        .await?;
+
+    // Setup: encrypted client.
+    let mut opts = CLIENT_OPTIONS.get().await.clone();
+    let handler = Arc::new(EventHandler::new());
+    let mut events = handler.subscribe();
+    opts.command_event_handler = Some(handler.clone());
+    // TODO: register an event handler
```
kmahar, line 498: it looks like we register a handler just before this. should this TODO still be here?

src/test/csfle.rs
```diff
@@ -0,0 +1,1261 @@
...
+    .iter()
+    .collect();
+    Ok(std::fs::read_to_string(path)?)
+}
+
+fn load_testdata(name: &str) -> Result<Document> {
+    Ok(serde_json::from_str(&load_testdata_raw(name)?)?)
+}
+
+// Prose test 4. BSON Size Limits and Batch Splitting
```
kmahar, line 469: nit: for future readability it might be helpful to number the steps in this test since all of them look kind of similar at first read due to doing inserts

src/test/csfle.rs
```diff
@@ -0,0 +1,1261 @@
...
+        "_id": "under_16mib",
+        "unencrypted": repeat_a(16_777_216 - 2_000),
+    };
+    coll.insert_one(doc, None).await?;
+
+    let mut doc: Document = load_testdata("limits-doc.json")?;
+    doc.insert("_id", "encryption_exceeds_16mib");
+    doc.insert("unencrypted", repeat_a(16_777_216 - 2_000));
+    let result = coll.insert_one(doc, None).await;
+    assert!(result.is_err());
```
kmahar, line 584: is there a specific error type we throw here / can assert we throw?

kmahar (commented):
looks good! some very minor comments and questions

src/client/csfle/client_encryption.rs
patrickfreed: In practice, I agree it's not a huge concern. That said, at an abstract level, strict API consistency is somewhat important, even if in an intangible way. On a more practical level, my concerns would be:
- if in the future we want to add new options to `DataKeyOptions` that we want to take by value, we can't
- users will need to refer to the documentation for each method or their compiler output to know whether they need to pass options in by reference or not, which might be a poor ergonomic experience.

If performance is a motivating factor here, I think I'd personally lean towards making the options struct lighter (e.g. by `Arc`-ing constant and/or big fields) rather than taking it in by reference. If performance isn't a concern, then I lean towards strict consistency.

I think it's totally fine to just take `KmsProvider` in by reference though. `EncryptOptions` I think I lean towards taking by value also for similar reasons, though given that it is required and not an `Option`, I think consistency is less of a factor there.

All this being said, happy to leave this as-is the team doesn't really see this as being a concern.

kmahar (approved):

src/test/csfle.rs
```diff
@@ -0,0 +1,1192 @@
...
+    // Setup: db initialization.
+    let (client, _) = init_client().await?;
+    client
+        .database("db")
+        .collection::<Document>("view")
+        .drop(None)
+        .await?;
+    client
+        .database("db")
+        .run_command(doc! { "create": "view", "viewOn": "coll" }, None)
```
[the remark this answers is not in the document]
patrickfreed: Yeah, it might be more ergonomic to have separate helpers for creating an actual collection and creating a view. Unfortunately, it looks like there's only tests for collection management but not actually a spec (DRIVERS-2118). Could you file a RUST ticket for this, and maybe when 3.0 comes around it might be enough inertia for one of us to just write the drivers spec for it.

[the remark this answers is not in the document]
abr-egn: Filed RUST-1512 for this.

isabelatkinson (approved):

patrickfreed (approved):
LGTM!

PR #759, opened 2022-10-06
RUST-1490 Fix build errors with csfle enabled

abr-egn:
RUST-1490

This switches to using the `num_cpus` crate instead of the stdlib function that's not available in our MSRV; it also updates the lock file to use a more recent git revision of `bson`, which had some knock-on changes.

### DIFF
--- a/src/client/csfle/state_machine.rs
+++ b/src/client/csfle/state_machine.rs
@@ -40,6 +40,7 @@ impl CryptExecutor {
         key_vault_namespace: Namespace,
         tls_options: Option<KmsProvidersTlsOptions>,
     ) -> Result<Self> {
+        // TODO(RUST-1492): Replace num_cpus with std::thread::available_parallelism.
         let crypto_threads = rayon::ThreadPoolBuilder::new()
             .num_threads(num_cpus::get())
             .build()
--- a/src/client/csfle/state_machine.rs
+++ b/src/client/csfle/state_machine.rs
@@ -40,7 +40,7 @@ impl CryptExecutor {
         key_vault_namespace: Namespace,
         tls_options: Option<KmsProvidersTlsOptions>,
     ) -> Result<Self> {
-        // TODO(RUST-1492): Replace num_cpus with std::thread::available_parallelism.
+        // TODO RUST-1492: Replace num_cpus with std::thread::available_parallelism.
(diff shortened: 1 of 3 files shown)


src/client/csfle/state_machine.rs
```diff
@@ -40,9 +40,8 @@ impl CryptExecutor {
         key_vault_namespace: Namespace,
         tls_options: Option<KmsProvidersTlsOptions>,
     ) -> Result<Self> {
-        let num_cpus = std::thread::available_parallelism()?.get();
```
kmahar, line 43: do we want to file a ticket / add a TODO about switching back to this once our MSRV is 1.59+?

kmahar (approved):
one minor comment but LGTM otherwise

src/client/csfle/state_machine.rs
abr-egn: Good thought, done.

src/client/csfle/state_machine.rs
```diff
@@ -40,9 +40,9 @@ impl CryptExecutor {
...
+        // TODO(RUST-1492): Replace num_cpus with std::thread::available_parallelism.
```
patrickfreed, line 43:
```suggestion
        // TODO RUST-1492: Replace num_cpus with std::thread::available_parallelism.
```
slight nit to help with grep-ability, since our existing TODOs don't have the parens.

src/client/csfle/state_machine.rs
```diff
@@ -40,9 +40,9 @@ impl CryptExecutor {
...
         let crypto_threads = rayon::ThreadPoolBuilder::new()
-            .num_threads(num_cpus)
+            .num_threads(num_cpus::get())
```
patrickfreed, line 45: On Linux, this appears to [do file I/O](https://github.com/seanmonstar/num_cpus/blob/master/src/linux.rs#L60) to read `/proc/cpuinfo`. I doubt this is a huge problem, but just to confirm: this isn't generally called in a hot path or per operation, right? Typically this is done once per client?

abr-egn: Correct, once per `Client` that has encryption enabled (not run at all for clients that don't), and once per `ClientEncryption`.  Neither of those are in a hot path.

isabelatkinson (approved):

PR #761, opened 2022-10-06
RUST-1488 Populate inserted_ids when an insert_many fails

### LINKED ISSUE #748: RUST-1488 BulkWriteFailure does not contain inserted_ids when insert_many fails
## Versions/Environment
1. What version of Rust are you using? cargo 1.60.0
2. What operating system are you using? MS Windows 10
3. What versions of the driver and its dependencies are you using? mongodb:2.3.0 & bson:2.4.0
4. What version of MongoDB are you using? AtlasDB 5.0.12
5. What is your MongoDB topology (standalone, replica set, sharded cluster, serverless)? M5 (General) Replica Set - 3 nodes

## Describe the bug
The documentation states:

> ordered: 
If true, when an insert fails, return without performing the remaining writes. If false, when a write fails, continue with the remaining writes, if any.

So when inserting documents, with some violating an index, I expect that the insert does not panic but inserts the non-violating documents.
The actual behaviour is: The driver panics, and no document is inserted:

    let options = mongodb::options::InsertManyOptions::builder().ordered(false).build();
    let res = col.insert_many(document, options).await.unwrap();

The panic is

    thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: 
    Error { kind: BulkWrite(BulkWriteFailure { write_errors: 
      Some([BulkWriteError {
       index: 0, code: 11000, code_name: None, message: "E11000 duplicate key error collection: col_name index: 
       index_name dup key: { field_name: \"X1\" }", details: None }]), write_concern_error: None, inserted_ids: {} }
     ), labels: {}, wire_version: None, source: None }'

The use case is optimistic locking: Bulk insert those documents which are not yet inserted.

kmahar:
fixes #748 

The fix ended up being a little more complicated than I initially expected in order to get it work properly for multi-batch bulk writes but was still relatively straightforward; most of the diff here is just adding testing. @isabelatkinson we should consider adding automated tests around this type of thing as part of the new bulk spec

### DIFF
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -1235,6 +1235,10 @@ where
 
                             let failure_ref =
                                 cumulative_failure.get_or_insert_with(BulkWriteFailure::new);
+                            for (index, id) in bw.inserted_ids {
+                                failure_ref.inserted_ids.insert(index + n_attempted, id);
+                            }
+
                             if let Some(write_errors) = bw.write_errors {
                                 for err in write_errors {
                                     let index = n_attempted + err.index;
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -1,6 +1,12 @@
 pub mod options;
 
-use std::{borrow::Borrow, collections::HashSet, fmt, fmt::Debug, sync::Arc};
+use std::{
+    borrow::Borrow,
+    collections::{HashMap, HashSet},
+    fmt,
+    fmt::Debug,
+    sync::Arc,
+};
 
 use futures_util::{
     future,
@@ -1196,7 +1202,7 @@ where
 
         let mut cumulative_failure: Option<BulkWriteFailure> = None;
         let mut error_labels: HashSet<String> = Default::default();
-        let mut cumulative_result: Option<InsertManyResult> = None;
+        let mut cumulative_inserted_ids = HashMap::new();
 
         let mut n_attempted = 0;
 
@@ -1211,13 +1217,8 @@ where
             {
                 Ok(result) => {
                     let current_batch_size = result.inserted_ids.len();
-
-                    let cumulative_result =
-                        cumulative_result.get_or_insert_with(InsertManyResult::new);
                     for (index, id) in result.inserted_ids {
-                        cumulative_result
-                            .inserted_ids
-                            .insert(index + n_attempted, id);
+                        cumulative_inserted_ids.insert(index + n_attempted, id);
                     }
 
                     n_attempted += current_batch_size;
@@ -1236,7 +1237,7 @@ where
                             let failure_ref =
                                 cumulative_failure.get_or_insert_with(BulkWriteFailure::new);
                             for (index, id) in bw.inserted_ids {
-                                failure_ref.inserted_ids.insert(index + n_attempted, id);
+                                cumulative_inserted_ids.insert(index + n_attempted, id);
                             }
 
                             if let Some(write_errors) = bw.write_errors {
@@ -1259,7 +1260,8 @@ where
                             if ordered {
                                 // this will always be true since we invoked get_or_insert_with
                                 // above.
-                                if let Some(failure) = cumulative_failure {
+                                if let Some(mut failure) = cumulative_failure {
+                                    failure.inserted_ids = cumulative_inserted_ids;
                                     return Err(Error::new(
                                         ErrorKind::BulkWrite(failure),
                                         Some(error_labels),
@@ -1275,11 +1277,14 @@ where
         }
 
         match cumulative_failure {
-            Some(failure) => Err(Error::new(
-                ErrorKind::BulkWrite(failure),
-                Some(error_labels),
-            )),
-            None => Ok(cumulative_result.unwrap_or_else(InsertManyResult::new)),
+            Some(mut failure) => {
+                failure.inserted_ids = cumulative_inserted_ids;
+                Err(Error::new(
+                    ErrorKind::BulkWrite(failure),
+                    Some(error_labels),
+                ))
+            }
+            None => Ok(InsertManyResult::new(cumulative_inserted_ids)),
         }
     }
 
--- a/src/test/coll.rs
+++ b/src/test/coll.rs
@@ -27,7 +27,7 @@ use crate::{
         UpdateOptions,
         WriteConcern,
     },
-    results::DeleteResult,
+    results::{DeleteResult, InsertManyResult},
     runtime,
     test::{
         log_uncaptured,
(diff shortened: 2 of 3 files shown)


src/coll/mod.rs
```diff
@@ -1196,7 +1202,7 @@ where
 
         let mut cumulative_failure: Option<BulkWriteFailure> = None;
         let mut error_labels: HashSet<String> = Default::default();
-        let mut cumulative_result: Option<InsertManyResult> = None;
+        let mut cumulative_inserted_ids = HashMap::new();
```
kmahar, line 1205: I initially had just added logic to set the `inserted_ids` on the `cumulative_failure` for failed batches, but I realized that would not correctly handle cases where there are a mix of successful and unsuccessful batches -- for example, if the first batch succeeds and the second batch fails, we would put the `inserted_ids` from batch 1 on the `cumulative_result` but the `inserted_ids` from batch 2 on the `cumulative_failure`, and then only return the failure. 
given this, it seemed simplest to switch to tracking the `inserted_ids` in one place and then just setting them on the `InsertManyResult` / `BulkWriteFailure` at the end.

src/test/coll.rs
```diff
@@ -1201,3 +1201,216 @@ fn assert_duplicate_key_error_with_utf8_replacement(error: &ErrorKind) {
...
+                &Bson::Int32(2),
+                "inserted document should have _id 2"
+            );
+        }
+        _ => panic!("Expected BulkWrite error, but got: {:?}", err),
+    }
+
+    // an ordered, 2-batch bulk write where a document in the first batch generates write error:
+    // nothing should be inserted
+    // note: these numbers were chosen because maxWriteBatchSize is 100,000
```
kmahar, line 1293: @isabelatkinson for the purpose of trying to write spec tests around batching behavior for bulk, we might consider requiring drivers have some way to internally override `maxWriteBatchSize`. otherwise we'd need  massive yaml/json files to fit in enough data to generate multiple batches (the limit is either 100k writes or 16 MB, whichever is smaller). 
alternatively, drivers can write prose tests like this one, but I think it would be much easier if we could artificially use bulk batches with a smaller size.
alternatively/additionally we could consider asking the server to make the parameter somehow configurable so we could temporarily decrease it for the bulk tests

kmahar:
the evg failures don't seem related.
also, I suppose we should probably backport this to the 2.3 branch and tag 2.3.2

abr-egn (approved):
LGTM!

src/test/coll.rs
```diff
@@ -1201,3 +1201,213 @@ fn assert_duplicate_key_error_with_utf8_replacement(error: &ErrorKind) {
...
+                    None => panic!("Document at index {} should have been inserted", i),
+                }
+            }
+        }
+        _ => panic!("Expected BulkWrite error, but got: {:?}", err),
+    }
+
+    // an unordered, 2-batch bulk write where the second-to-last document in the second batch
+    // generates a write error: everything besides that document should be inserted
+    let res = run_inserted_ids_test(&client, &docs, false);
```
patrickfreed, line 1380: Can we add one more case where we have a 2 batch write with a write error in both batches?

src/coll/mod.rs
```diff
@@ -1271,11 +1277,14 @@ where
         }
 
         match cumulative_failure {
-            Some(failure) => Err(Error::new(
-                ErrorKind::BulkWrite(failure),
-                Some(error_labels),
-            )),
-            None => Ok(cumulative_result.unwrap_or_else(InsertManyResult::new)),
+            Some(mut failure) => {
+                failure.inserted_ids = cumulative_inserted_ids;
```
patrickfreed, line 1281: This field is actually currently `pub(crate)`, and the only reason the user was able to observe it was the derived `Debug` implementation. I did some digging and found that RUST-260 was filed to eventually expose this field properly and to stop discarding the intermediate results, and I believe the field originally was introduced for counting writes attempted rather than actually tracking inserted ids (per https://github.com/mongodb/mongo-rust-driver/pull/377#discussion_r657342471). 

Given that we've done the work for populating this field, I think we probably ought to do RUST-260 also and mark the field as `pub`. 

Regarding the Jira logistics of this, I wonder if it makes sense to rename RUST-1488 to cover the bug of the `Debug` implementation leaking the unset, `pub(crate)` `inserted_ids` map, and instead have this PR be titled RUST-260. Wdyt?

src/coll/mod.rs
```diff
@@ -1235,6 +1236,10 @@ where
```
patrickfreed, line 1236: In the event that one of the batches fails with a non-BulkWriteError, we just return early without continuing it looks like. I think this is probably inconsistent with the docstring of unordered writes ("If false, when a write fails, continue with the remaining writes, if any), but besides that, it also means we discard our cumulative idea of what has been inserted thus far. I wonder if we would instead want to add a `Vec<Error>` to `BulkWriteFailure` to account for the other errors that might occur during a multi-batch write in order to preserve the `inserted_ids` map. We have something similar in Swift, though I think it's only a single error instead of an array of them.

This may get kind of tricky when deciding which errors we want to continue on with and which constitute a "terminal" error. For example, I imagine an auth error would not be something we'd want to continue attempting to insert after, but a network error would be (e.g. so that we could get a different mongos for the next batch). Maybe its worth deferring to another ticket? Or maybe we keep the existing behavior?

qm3ster: For prior art of how to decide, consider https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.error_for_status

qm3ster:
Being as ignorant and disorganized as I am, I have reopened this in https://github.com/mongodb/mongo-rust-driver/issues/1157 and attempted an implementation in https://github.com/mongodb/mongo-rust-driver/pull/1158.
I am currently using that from my private fork where `BulkWriteFailure.inserted_ids` is `pub`.
If anyone would like to look over the code and give it a general go-ahead I'd be happy to write (steal from this PR) some tests.

But if this is a whole process, (I thought it may not be since it's not a breaking change) I believe it would be a lot more ergonomic to start breaking application "errors" out of the general error.
Examples would range from `Result<InsertManyResult{ids: Vec<Result<Id, Err>>}, mongodb::Error>`
to even `Result<Result<InsertOneResult, InsertOneFailure>, mongodb::Error>`

But that sounds like `4.0`.

isabelatkinson:
@qm3ster yeah that would be quite a major overhaul of our error API. 3.0 is very fresh so breaking changes will likely not happen again for a while.

Also worth noting that the new bulk write API does report inserted IDs in its error type; `BulkWriteError` contains a partial result which, if `verbose_results` was configured on the call to `bulk_write`, will record an `InsertResult` for any successful inserts that happened before the error occurred. e.g.:

```
let write_models = <InsertOneModel for each insert document>;
let bulk_write_result = client.bulk_write(write_models).verbose_results().await;
let insert_results = match bulk_write_result {
    Ok(result) => Some(result.insert_results),
    Err(error) => {
        if let ErrorKind::BulkWrite(bulk_write_error) = *error.kind {
            match bulk_write_error.partial_result {
                Some(PartialBulkWriteResult::Verbose(result)) => Some(result.insert_results),
                _ => None,
            }
        } else {
            None
        }
    }
}; // returns Option<HashMap<usize, InsertOneResult>>
```

(The major caveat here is that bulk write will not be available for use in production until MongoDB 8.0 is released, but once that happens this will be a way to inspected inserted IDs upon error without requiring changes to the driver.)

PR #764, opened 2022-10-12
RUST-1399 Implement `AsyncRead` for `GridFsDownloadStream`

isabelatkinson:


src/gridfs/download.rs
```diff
@@ -139,16 +140,196 @@ impl GridFsBucket {
...
+        if chunk.data.bytes.len() != (expected_len as usize) {
+            return Err(ErrorKind::GridFs(GridFsErrorKind::WrongSizeChunk {
+                actual_size: chunk.data.bytes.len(),
+                expected_size: expected_len,
+            })
+            .into());
+        }
+
+        let bytes_to_write = std::cmp::min(chunk.data.bytes.len(), buf_size) as usize;
+        bytes.extend_from_slice(&chunk.data.bytes[..bytes_to_write]);
```
isabelatkinson, line 268: I was trying to figure out a way not to need to copy these bytes into the intermediary `Vec<u8>` that gets returned from this method, but I don't think it's possible. The two ways to do this would be:

- Pass a mutable reference to the `buf` provided in `poll_read` into this method, and copy the bytes directly in here. This won't work, however, because the lifetime of `buf` is contained to one call to `poll_read`, whereas a reference here would require it to be for the lifetime of the `GridFsDownloadStream`.
- Return something like a `Vec<&[u8]>` that holds pointers to each chunk we deserialize in this method. This isn't feasible because the lifetime of `chunk` (returned from the call to `deserialize_current`) is contained within the loop.

src/gridfs/download.rs
```diff
@@ -139,16 +140,196 @@ impl GridFsBucket {
...
+    if buf_size > 0
+        && (current_n as u64) < FilesCollectionDocument::n_from_vals(expected_file_len, chunk_size)
+    {
+        return Err(ErrorKind::GridFs(GridFsErrorKind::MissingChunk { n: current_n }).into());
+    }
+
+    Ok(bytes)
+}
+
+impl From<Error> for futures_util::io::Error {
```
patrickfreed, line 287: Is the only usage of this impl on line 224? If so (or if there are only a few usages), I think it might be better to keep this as an internal function rather than a public trait impl to avoid bleeding into our public API.

patrickfreed (commented):
Looks great! Good job digging into tokio for some prior art on implementing something like this. I just have a few minor suggestions

src/gridfs/download.rs
patrickfreed: If I'm following the logic right, the buffer being allocated each time will always have capacity equal to the chunk size from the files collection document except in the following cases:

- the final chunk is slightly smaller than a full chunk
- the output buffer is full
 
So in either case, a reusable buffer of capacity equal to the chunk size should be sufficient since the required capacity will be at most chunk size, just in some cases the entire buffer won't be used.

isabelatkinson: Just to clarify -- I'm referencing the buffer being created on line 251. This method reads however many chunks are necessary to fill up the buffer the user provided (note the loop over the cursor), so the `bytes` vec isn't of a fixed size relative to the file/stream. Were you thinking of an intermediary buffer for the chunks being read?

patrickfreed: Ohh sorry, I misunderstood how the cached future works. I see that it takes in the length of the input buffer and reads up to that amount of bytes. 

>Were you thinking of an intermediary buffer for the chunks being read?

Yeah, I was thinking that instead of one future for the whole file, we could create a new future for each chunk. That way, we can use a chunk-size intermediary buffer in each future instead of one the size of the user's buffer / the file.

isabelatkinson: Yeah, I actually had an implementation like that originally. My concern with that design, however, is the additional overhead that would be required to read through an entire file, particularly when the file is large and the chunk size is small. Whereas the current design only requires a single future to be created upon a single call to one of the `AsyncWriteExt` methods (e.g. `read`, `read_exact` or `read_to_end`), creating a new future for each chunk could possibly require a ton of loops through the state machine when a user calls `read_exact` or `read_to_end` to read in the requested number of bytes or the entirety of the file, respectively. It could also require lots of allocations in the way of creating all of those intermediary futures, although that could be mitigated by using tokio's [`ReusableBoxFuture`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.ReusableBoxFuture.html).

That said, `AsyncReadExt::read` (which calls down directly to `poll_read`) only says it returns "some bytes", and the presence of `AsyncReadExt::read_exact` and `AsyncReadExt::read_to_end` implies that `read` doesn't necessarily do either of those things. So perhaps it makes sense to be more conservative with how much we're actually reading in this method and allow the `AsyncWriteExt` methods to handle satisfying the quantity of bytes the user is requesting. I'm not sure how other libraries/implementations choose to define what "some" means; I can do some investigation into this. This would also be technically configurable by the user in that they could adjust their chunk size if we switch over to your proposal, although that seems like more of an implementation detail than a user knob and would require foresight upon inserting the data.

patrickfreed: >That said, read (which calls down directly to poll_read) only says it returns ["some bytes"](https://docs.rs/futures/latest/futures/io/trait.AsyncReadExt.html#method.read), and the presence of read_exact and read_to_end implies that read doesn't necessarily do either of those things. 

Yeah, I think the "some" here basically means "whatever is available". For example, in the case of a TCP socket, this might entail blocking until bytes have been received and just returning those bytes. In our case, it might mean a single chunk, but it could just as well mean "up to the number of bytes in the provided buffer". 

> This value would also be technically configurable by the user if we switch over to your proposal in that they could adjust their chunk size, although that seems like more of an implementation detail than a user knob and would require foresight upon inserting the data.

This is a good point. I think you're right that we're okay to buffer up to the size of the user's input (or a little more to fit the whole chunk in our case), so we probably don't need to do the multiple future route. This is what `tokio` does for `File` it looks like.

That said, it has made me realize that we _do_ need to update the future logic to support partial reads without discarding the unread portions of the chunks. I think that right now, it currently does discard any bytes in a chunk that exceed the users input buffer size. This is a problem when doing repeated reads (such as in `read_to_end`) with a buffer size < file size. Instead, we'll want to cache the leftover portion of the chunk and pick it up next time `poll_read` is called.

patrickfreed: Hm, I would think the spec tests for this would catch this error though. Could you write a test that calls `read` with a buffer half the chunk size repeatedly until it returns 0? The collected bytes from the `read` invocations _should_ be missing half of each chunk if my understanding of the various `AsyncRead` traits is correct. If the test passes then I might be missing something.

src/gridfs/download.rs
```diff
@@ -139,16 +142,214 @@ impl GridFsBucket {
...
+
+                stream.state = if cursor.has_next() {
+                    State::Idle(Box::new(Some(cursor)))
+                } else {
+                    State::Done
+                };
+
+                Poll::Ready(Ok(bytes_to_write))
+            }
+            Err(error) => {
```
isabelatkinson, line 277: I originally allowed errors to be recoverable (i.e. not set the stream's state to `Done`), but after our discussion about losing partial chunks when the buffer size doesn't exactly divide the chunk size, I realized the same thing would happen if an error occurred in the middle of iterating the cursor. While we could cache what has been read so far, there is a question around what should be done with the data in malformed chunks -- should it be included in the cache, or should it be ignored? Additionally, all data in chunks that couldn't deserialize properly would be lost. Because of this I think it makes more sense to allow these errors to invalidate the stream. Our retryability logic will retry recoverable errors, so it's likely that the errors returned here will be unrecoverable anyway.

src/gridfs/upload.rs
```diff
@@ -116,16 +116,29 @@ impl GridFsBucket {
...
+        // listIndexes returns an error if the collection has not yet been created.
+        let options = CreateCollectionOptions::builder()
+            .write_concern(self.write_concern().cloned())
+            .build();
+        // Ignore NamespaceExists errors if the collection has already been created.
+        if let Err(error) = self.inner.db.create_collection(coll.name(), options).await {
+            if error.code() != Some(48) {
+                return Err(error);
+            }
+        }
```
isabelatkinson, line 140: I ran into some `NamespaceNotFound` errors when running the new test which necessitated this addition. I considered creating the collections upon instantiation of the bucket, but that seemed out of line with our other methods for retrieving handles to top-level types, which don't do any I/O. Checking against the error code here isn't ideal, but it seemed like a better solution than running a `listCollections` to see if the collection exists first.

src/gridfs.rs
```diff
@@ -52,12 +52,39 @@ pub struct FilesCollectionDocument {
...
+impl FilesCollectionDocument {
+    /// Returns the total number of chunks expected to be in the file.
+    fn n(&self) -> u32 {
+        Self::n_from_vals(self.length, self.chunk_size)
+    }
+
+    fn n_from_vals(length: u64, chunk_size: u32) -> u32 {
+        let chunk_size = chunk_size as u64;
+        let n = length / chunk_size + u64::from(length % chunk_size != 0);
+        n as u32
```
isabelatkinson, line 67: This number is much easier to use in calculations against chunks' `n` fields when it's a `u32`, and needed to be cast to `u32` in a few places anyway. A file would need to be ~1M GB for its total number of chunks to exceed `u32::MAX` with the default chunk size, so it's extremely unlikely that this cast will ever fail.

isabelatkinson:
this is ready for another round! sorry for the delay

src/gridfs/download.rs
```diff
@@ -139,16 +142,214 @@ impl GridFsBucket {
...
+                stream.state = State::Done;
+                let error = futures_io::Error::new(futures_io::ErrorKind::Other, error);
+                Poll::Ready(Err(error))
+            }
+        }
+    }
+}
+
+async fn get_bytes(
```
patrickfreed, line 286: Why is the separate helper function necessary? If it isn't can we consolidate?

src/gridfs/download.rs
```diff
@@ -139,16 +142,214 @@ impl GridFsBucket {
...
+
+                let new_future = get_bytes(
+                    cursor.take().unwrap(),
+                    n_range,
+                    stream.file.chunk_size,
+                    stream.file.length,
+                )
+                .boxed();
+
+                stream.current_n += total_chunks;
```
patrickfreed, line 246: I think this logic might be a little clearer if we just assign `current_n` to `final_n` directly.

src/gridfs/download.rs
```diff
@@ -139,16 +142,214 @@ impl GridFsBucket {
...
+
+impl AsyncRead for GridFsDownloadStream {
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::result::Result<usize, futures_util::io::Error>> {
+        let stream = self.get_mut();
+
+        if let Some(cache) = &mut stream.cached_bytes {
```
patrickfreed, line 219: Following up on my earlier comments, we can save a lot of allocations by passing this buffer to and from the future and reusing it both as the cache for leftover bytes and also as the temporary cache of bytes we're reading from chunks. This works because `Vec` is [guaranteed to never shrink its allocated capacity](https://doc.rust-lang.org/std/vec/struct.Vec.html#guarantees) even if we remove elements from it. I think I was barking up the wrong tree by suggesting limiting the cache to one chunk--we can just keep a (possibly ever-growing) cache around of arbitrary size.

patrickfreed (commented):
Logic looks good to me! I just have a few nits and one follow up perf suggestion. Going to tag the other reviewers in now since I think it's getting close to ready

src/gridfs/download.rs
```diff
@@ -139,16 +142,210 @@ impl GridFsBucket {
...
+        let future = match &mut stream.state {
+            State::Idle { buffer, cursor } => {
+                let mut taken_buffer = buffer.take().unwrap();
+                let taken_cursor = cursor.take().unwrap();
+
+                if !taken_buffer.is_empty() {
+                    let bytes_to_write = std::cmp::min(taken_buffer.len(), buf.len());
+                    buf[..bytes_to_write]
+                        .copy_from_slice(taken_buffer.drain(..bytes_to_write).as_slice());
+                    if taken_cursor.has_next() {
```
patrickfreed, line 232: Think we also need to verify that the buffer is empty here before transitioning to done (e.g. if the input buffer is less than half the chunk size). This would actually be a good test case to cover I think.

isabelatkinson: Good catch, this is fixed by the refactor you suggested above. I updated the test to include this scenario.

patrickfreed (approved):
LGTM!

abr-egn (approved):
LGTM!

In particular, I appreciate the clarity and commenting in the `GridFsDownloadStream` implementation - it's easy for manual future implementations to get gnarly but I found yours to be quite readable.

src/gridfs/download.rs
```diff
@@ -100,18 +111,11 @@ impl GridFsBucket {
...
-        let chunk_size = file.chunk_size as u64;
-
-        if total_bytes_expected == 0 {
+        if file.length == 0 {
             return Ok(());
         }
 
-        let options = FindOptions::builder()
-            .sort(doc! { "n": 1 })
-            .read_concern(self.read_concern().cloned())
```
kmahar, line 112: were we not supposed to be using the read concern and read preference here?

src/gridfs/download.rs
```diff
@@ -139,16 +142,206 @@ impl GridFsBucket {
...
+                            n_range,
+                            stream.file.chunk_size,
+                            stream.file.length,
+                        )
+                        .boxed(),
+                    );
+
+                    match new_future.poll_unpin(cx) {
+                        Poll::Ready(result) => result,
+                        Poll::Pending => return Poll::Pending,
```
kmahar, line 256: reading over the docs for `Poll::Pending` I see:

> When a function returns Pending, the function must also ensure that the current task is scheduled to be awoken when progress can be made.

Is that requirement covered somehow here?

src/gridfs/download.rs
```diff
@@ -139,16 +142,206 @@ impl GridFsBucket {
...
     }
 }
+
+pub struct GridFsDownloadStream {
+    state: State,
+    current_n: u32,
+    file: FilesCollectionDocument,
+}
+
+type GetBytesFuture = BoxFuture<'static, Result<(Vec<u8>, Box<Cursor<Chunk<'static>>>)>>;
```
kmahar, line 163: could you explain the usage of `'static`? my prior understanding was that static lifetimes were for things that would live for the lifetime of a program, but it sounds wrong to me that every `Chunk` we encounter would live for the lifetime of the program so I think I'm missing something 🙂

kmahar (commented):
looks great! I just have a few questions mostly for my own understanding.

src/gridfs/download.rs
isabelatkinson: Hm, the reference implementations I looked at when writing this don't do anything with the waker. I believe this is a lower-level detail handled by the future we pass `cx` into -- this function itself wouldn't be able to know when that future is ready to make progress.

src/gridfs/download.rs
kmahar: ah right, I remember that now. thanks!

src/gridfs/download.rs
kmahar: ahhh this post is very helpful! thanks for the explanation. I think I at least sort of get it now 🙂

src/gridfs/download.rs
kmahar: 👍 cool, yeah I wasn't sure how we would be expected to know when to wake either. sgtm

kmahar (approved):
awesome job with this!!! 👏

PR #766, opened 2022-10-20
RUST-1510 Implement connection pool tracing messages

kmahar:
This is the Rust PR corresponding to the spec changes in https://github.com/mongodb/specifications/pull/1324:

Summary:
* Sync new logging tests and reorganize yml/JSON CMAP files
* Emit tracing events corresponding to all existing CMAP events 
    * Refactor connection pool code to pass around a new `CmapEventEmitter` type that emits both tracing events and the existing events
* Support running `close` operations on client entities in the unified test runner (previously only supported cursors)


src/event/cmap.rs
```diff
@@ -327,3 +335,100 @@ pub trait CmapEventHandler: Send + Sync {
     /// whenever a connection is checked back into a connection pool.
     fn handle_connection_checked_in_event(&self, _event: ConnectionCheckedInEvent) {}
 }
+
+#[derive(Clone)]
+pub(crate) enum CmapEvent {
```
kmahar, line 340: I wasn't sure if it was worth trying to combine this with the `cmap::test::event::Event` type in the tests because that has a decent amount very Astrolabe-specific serialization logic associated with it which I think we'd have to pull up into the driver itself. Let me know if you think that refactor makes sense though.

src/test/spec/unified_runner/entity.rs
```diff
@@ -267,7 +284,13 @@ impl Deref for ClientEntity {
     type Target = Client;
 
     fn deref(&self) -> &Self::Target {
-        &self.client
+        match &self.client {
+            ClientEntityState::Client(c) => c,
+            ClientEntityState::Dropped => panic!(
```
kmahar, line 289: it didn't seem great to panic here (and the docs say implementations of this method should never fail), but this can only happen in a very specific case where a test tries to use a client that was closed via a `close` operation, which is documented as something that tests should not do in my spec PR. so it seemed worth the ergonomic tradeoff of implementing deref to leave this.

src/test/spec/unified_runner/entity.rs
```diff
@@ -48,9 +48,19 @@ pub(crate) enum Entity {
...
+    /// Indicates the underlying client has been dropped via a `close` operation.
+    Dropped,
+}
+
 #[derive(Clone, Debug)]
 pub(crate) struct ClientEntity {
-    client: Client,
+    pub(crate) client: ClientEntityState,
+    #[cfg(feature = "tracing-unstable")]
+    pub(crate) topology_id: bson::oid::ObjectId,
```
kmahar, line 63: this needs to be stored since if we close the underlying client we can't access it directly via that anymore.

kmahar:
also, just to make sure my refactoring and changing of serialization logic didn't mess anything up for the astrolabe integration I ran a [patch](https://spruce.mongodb.com/version/6359bb13d6d80a51ad53909d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) there and it all passed.

src/event/cmap.rs
isabelatkinson: From looking at these [docs](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute), I think that it would be possible to do something like `#[cfg_attr(test, derive(Deserialize)]` on this enum to define a `Deserialize` implementation conditionally, and then do the same for the custom deserialization attributes. Then we could avoid doing a custom implementation that requires the duplicate struct.

What was your concern about having the `Deserialize` impl on this enum in the regular driver code, though? Because it's `pub(crate)` it shouldn't leak into the public API, but maybe I'm missing another issue at hand.

isabelatkinson (commented):
looks good, just one more suggestion/question. gonna tag in the rest of the team

src/event/cmap.rs
kmahar: I think I'm maybe trying to follow more of a Swift driver precendent here where we tended to, when possible, put any Codable implementations that were custom and specific to test file structure in the tests rather than in the driver itself. (was not always possible if the only change was custom `CodingKeys`, but can be done for anything with custom `init(from: Decoder)` and encode(to: Encoder)` methods). 

in this case I think we would need to pull up all of this: https://github.com/mongodb/mongo-rust-driver/blob/5ad1c6127138ecd26b95256ac98ac937fab7f220/src/cmap/test/event.rs#L214-L301 

but if that's not a concern and/or we have existing precedent for this type of thing being in the driver itself then I don't mind moving the logic

patrickfreed: If we have to pull in all that test code, then I think it makes sense to go with separate implementations. The risk is low, but less code in the release driver -> fewer bug opportunities.

>I think I'm maybe trying to follow more of a Swift driver precendent here where we tended to, when possible, put any Codable implementations that were custom and specific to test file structure in the tests rather than in the driver itself.

We do have `Serialize` and `Deserialize` implementations that are derived on public types but used only for tests. I think we treat that kind of thing as a "SHOULD" for convenience's sake. In this case, there's a lot more custom logic than a simple derive and you've already done the "SHOULD" part, so it seems fine to keep it separate imo.

src/test/spec/unified_runner/operation.rs
```diff
@@ -2120,12 +2121,25 @@ impl TestOperation for Close {
...
-            let rx = cursor.make_kill_watcher().await;
-            *cursor = TestCursor::Closed;
-            drop(entities);
-            let _ = rx.await;
-            Ok(None)
+            let target_entity = entities.get_mut(id).unwrap();
+            match target_entity {
+                Entity::Client(client) => {
+                    client.client = ClientEntityState::Dropped;
+                    drop(entities);
```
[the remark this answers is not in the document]
patrickfreed: Hmm, I can't think of a reason either. This might have just been a mistake. If the tests pass without it, I'd say leave it out for cursors.

For clients, I think we do need it (or something similar) in order to drop all descendant collections / databases to ensure the client actually gets dropped.

src/trace/connection.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+impl TracingRepresentation for ConnectionCheckoutFailedReason {
+    type Representation = &'static str;
+
+    fn tracing_representation(self) -> &'static str {
+        match self {
+            ConnectionCheckoutFailedReason::Timeout => {
+                "Wait queue timeout elapsed without a connection becoming available"
```
patrickfreed, line 181: We don't actually use this reason anywhere in the driver because we don't have a wait queue timeout, but I think once DRIVERS-2347 is done in Rust, we should start using this for connection establishment timeout. To prepare for that, we might want to update this string accordingly.

src/test/spec/unified_runner/operation.rs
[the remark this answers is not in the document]
kmahar: you raise a good point regarding the descendant databases and collections keeping a client alive. 

that said, re-reading this code, all `drop(entities)` does is drop the `RwLockWriteGuard` we acquired which gave exclusive write access to the entities. I'm not sure why that would have been needed for cursors , and I don't see why we would need it for clients, still but I'll try running a full patch with it removed for cursors to see if the test failures (or lack thereof) give any indication here.

in any case, I think we need to update the logic here to actually remove all descendant objects from the entity map so the client can be dropped, and I think I should update my spec PR to state that it is an error for a test to use any descendant object for an operation after its parent client is closed. 
there is already a sentence to this effect re using the client which I can expand on: 
> Test files SHOULD NOT
specify any operations for a client entity following a `close` operation on it, as
driver behavior when an operation is attempted on a closed client is not consistent.

src/trace/connection.rs
kmahar: Just to check my understanding from reading over that ticket, would this specifically mean that we hit `connectTimeoutMS`? e.g. should the message be something like "Failed to establish a new connection within connectTimeoutMS"?

patrickfreed: Yep that sounds perfect.

src/test/spec/unified_runner/operation.rs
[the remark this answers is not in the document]
patrickfreed: Ohh, I see now. I think it was dropped to avoid holding it across await point in case it caused a deadlock, since the guard constitutes exclusive access. Probably would have been good if I left a comment about that.

>in any case, I think we need to update the logic here to actually remove all descendant objects from the entity map so the client can be dropped

Sounds good :+1:

src/trace/connection.rs
```diff
@@ -0,0 +1,188 @@
...
+impl ConnectionTracingEventEmitter {
+    pub(crate) fn new(topology_id: ObjectId) -> ConnectionTracingEventEmitter {
+        Self { topology_id }
+    }
+}
+
+impl CmapEventHandler for ConnectionTracingEventEmitter {
+    fn handle_pool_created_event(&self, event: PoolCreatedEvent) {
+        let options_ref = event.options.as_ref();
+        tracing::debug!(
```
patrickfreed, line 37: This probably has already been discussed so sorry for retreading ground, but does `tracing` handle the non-structured format for us (e.g. if someone is just using `log`)? Is there a way to configure that output so it looks like the spec defined format?

kmahar: yes, tracing handles the conversion for us, and unfortunately I don't believe it is possible for us to configure the output format at this time. the output looks something like "a=b, c=d, e=f"; it is generated by the `Display` implementation here: https://github.com/tokio-rs/tracing/blob/3258a2bb2ae5f1b3be755bffbaef7ff851ec3b4f/tracing/src/lib.rs#L1162-L1205 

I think in order to customize it we would have to add logic to emit log messages from scratch ourselves. I'm also not sure if/how we could prevent our tracing events from getting converted to log messages if a user used `tracing` with the `log` feature on. (not sure if it is possible to check whether a feature flag in another crate has been enabled?) if not, we might end up giving them two copies of every message, the one generating from the tracing event and the one we emit directly.

on top of those issues, I considered this acceptable because users of the `log` feature of tracing will be used to receiving logged key-value pairs in this format, and also because the strings defined in the spec are SHOULDs, not MUSTs, which was done both because we aren't actually testing string representations and because there were different schools of thought amongst reviewers as to whether a concise a=b, c=d, .. form or a more sentence-like form was preferable.

abr-egn (commented):
Looks good (modulo the entity map pruning), just a few minor comments.

src/event/cmap.rs
```diff
@@ -144,6 +156,15 @@ pub struct ConnectionClosedEvent {
 
     /// The reason that the connection was closed.
     pub reason: ConnectionClosedReason,
+
+    /// If the `reason` connection checkout failed was `Error`,the associated
+    /// error is contained here. This is attached so we can include it in log messages;
+    /// in future work we may add this to public API on the event itself. TODO: add
```
kmahar, line 162: @patrickfreed was going to file a DRIVERS ticket about drivers adding these errors to their connection monitoring events, so once there is a ticket to link to I'll add that here. I figured for now we could just leave it out from the public API in case the eventual spec work leads to some different proposal to expose the error.  

ideally, we would probably actually add the error to the `ConnectionClosedReason::Error` enum case but that would be a breaking change.

src/event/cmap.rs
```diff
@@ -144,6 +156,15 @@ pub struct ConnectionClosedEvent {
...
+    /// ticket link here.
+    #[cfg(feature = "tracing-unstable")]
+    #[serde(skip)]
+    #[derivative(PartialEq = "ignore")]
```
kmahar, line 166: we rely on this `PartialEq` implementation in the CMAP tests. but since `Error` does not implement `PartialEq` right now and this field is currently just here as an implementation detail for tracing it seemed like we could leave it out for now when comparing events in tests. 

that said, looking more closely, I'm not sure the `PartialEq` implementation is even needed. When I remove it from this type, the only thing that breaks is that the `CmapEvent` enum can't derive `PartialEq`. And if I remove the derived impl from that, the only thing that breaks is this line of code:

https://github.com/mongodb/mongo-rust-driver/blob/6e3d43d824facf8b051e3593d32f0ab378d095ef/src/cmap/test/mod.rs#L224

and IIUC all that line is doing is checking that the length of `subscriber.all(filter)` is 0??? so maybe we should just remove the `PartialEq` stuff altogether.

src/cmap/worker.rs
```diff
@@ -651,20 +636,19 @@ async fn establish_connection(
...
-                let event = ConnectionClosedEvent {
+            event_emitter.emit_event(|| {
+                ConnectionClosedEvent {
                     address,
                     reason: ConnectionClosedReason::Error,
                     connection_id,
-                };
-                handler.handle_connection_closed_event(event);
-            }
+                    error: Some(e.cause.clone()),
```
kmahar, line 644: I just extracted the `cause` here since `EstablishError` is a different type than `Error`. AFAICT the extra info stored on the `EstablishError` is what phase of the handshake the error happened during. I wasn't sure if that info was worth including here or more exists so the driver can decide what to do about the error. LMK what you think.

src/test/spec/unified_runner/operation.rs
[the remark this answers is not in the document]
patrickfreed: Tests can run in parallel using the `runOnThread` operation introduced as part of the SDAM integration test port to the unified format. I'm not sure any test includes operations that would be problematic for this, but it's technically possible I think.

src/trace/connection.rs
patrickfreed: Sounds good, thanks for the explanation!

src/cmap/test/event.rs
```diff
@@ -1,17 +1,14 @@
...
-use crate::{event::cmap::*, options::ServerAddress, runtime};
-use tokio::sync::broadcast::error::{RecvError, SendError};
+use crate::{event::cmap::*, options::ServerAddress, test::util::EventSubscriber};
+use tokio::sync::broadcast::error::SendError;
 
 #[derive(Clone, Debug)]
 pub struct EventHandler {
-    pub events: Arc<RwLock<Vec<Event>>>,
-    channel_sender: tokio::sync::broadcast::Sender<Event>,
+    pub(crate) events: Arc<RwLock<Vec<CmapEvent>>>,
```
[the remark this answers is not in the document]
patrickfreed: Another thing is that `pub` things (even in tests) can't expose `pub(crate)` things from the driver. In practice this doesn't matter for us because our tests don't get compiled in to the actual library, but it will prevent compilation from succeeding.

src/cmap/worker.rs
patrickfreed: Yep, that's just implementation detail stuff for SDAM error handling purposes. Using `cause` here SGTM.

src/event/cmap.rs
patrickfreed: Our `Serialize` / `Deserialize` impls are mostly there for test convenience, but I think something like `PartialEq` _might_ be functionality that a user could rely on, even if inadvertently like we are via a `Derive`. For that reason, I'd be hesitant to remove any of those implementations until 3.0, even if they're largely unnecessary.

src/event/cmap.rs
patrickfreed: Ticket here: https://jira.mongodb.org/browse/DRIVERS-2495. Leaving it out until the RUST ticket for that sgtm

src/test/spec/unified_runner/operation.rs
```diff
@@ -2120,12 +2120,48 @@ impl TestOperation for Close {
...
+                    let mut client = entities.get_mut(id).unwrap().as_mut_client();
+                    let closed_client_topology_id = client.topology_id;
+                    client.client = None;
+
+                    let mut entities_to_remove = vec![];
+                    for (key, value) in entities.iter() {
+                        match value {
+                            // skip clients so that we don't remove the client entity itself from
+                            // the map: we want to preserve it so we can
+                            // access the other data stored on the entity.
```
patrickfreed, line 2135: What data do we need to access on the client entity after its been closed? And is this also why we set the inner client to `None` instead of just removing it?

patrickfreed (commented):
Looks good, just have one small question

src/test/spec/unified_runner/operation.rs
kmahar: the `ClientEntity` stores:
*  the `EventHandler` and corresponding `EventObserver` for the client. no tests I've added here actually need the ability to inspect e.g. command monitoring events produced for a client after it was closed, since these tests mostly just use tracing events, but it seems plausible that would be needed in the future. (one use case: it would be good to have a spec test that checks drivers actually send `endSessions` when a client is closed, and we'd probably assert that via a `close` operation + command monitoring event.) 
* the `topology_id` (used to filter on tracing events and get those for this particular entity): currently used to make the tracing assertions. 

and yes, that's why we only get rid of the inner client and not the whole entity.

patrickfreed (approved):
LGTM!

src/event/cmap.rs
kmahar: oh yeah, that's a good point. I wasn't thinking about this being a public type. I agree it seems best to keep it around then. filed RUST-1538 about considering a removal in 3.0

src/event/cmap.rs
kmahar: added reference to DRIVERS-2495 here and below

isabelatkinson (approved):
lgtm!

abr-egn (approved):
LGTM!

PR #767, opened 2022-10-25
RUST-1386 Implement the remainder of the FLE prose tests

abr-egn:
RUST-1386

This adds the remaining prose tests, as well as expanding the Evergreen configuration to cover more cases.  I went with a sparse matrix for testing rather than every permutation, as there didn't seem to be much value in the latter.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -493,10 +493,7 @@ functions:
       background: true
       script: |
         ${PREPARE_SHELL}
-        cd ${DRIVERS_TOOLS}/.evergreen/csfle
-        . ./activate_venv.sh
-        # TMPDIR is required to avoid "AF_UNIX path too long" errors.
-        TMPDIR="$(dirname ${DRIVERS_TOOLS})" python3 kms_kmip_server.py
+        .evergreen/run-csfle-kmip-servers.sh
 
   "run csfle tests":
     - command: shell.exec
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -506,6 +506,7 @@ functions:
           export ASYNC_RUNTIME=${ASYNC_RUNTIME}
           export CSFLE_TLS_CERT_DIR=$DRIVERS_TOOLS_X509
           export CSFLE_SHARED_LIB_PATH=$(.evergreen/find-crypt_shared.sh "$PROJECT_DIRECTORY/crypt_shared/lib/")
+          export DISABLE_CRYPT_SHARED=${DISABLE_CRYPT_SHARED}
           # Exported without xtrace to avoid leaking credentials
           set +o xtrace
           export KMS_PROVIDERS=$(cat << "EOF"
@@ -1578,6 +1579,7 @@ axes:
         display_name: "5.0"
         variables:
           MONGODB_VERSION: "5.0"
+          DISABLE_CRYPT_SHARED: "true"
       - id: "4.4"
         display_name: "4.4"
         variables:
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1575,11 +1575,6 @@ axes:
         variables:
           MONGODB_VERSION: "6.0"
           CRYPT_SHARED_VERSION: "6.0.0"
-      - id: "5.0"
-        display_name: "5.0"
-        variables:
-          MONGODB_VERSION: "5.0"
-          DISABLE_CRYPT_SHARED: "true"
       - id: "4.4"
         display_name: "4.4"
         variables:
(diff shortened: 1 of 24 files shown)


src/test/csfle.rs
```diff
@@ -1269,3 +1309,1439 @@ async fn custom_endpoint_kmip_invalid_endpoint() -> Result<()> {
...
+    };
+    let enc_coll = testdata
+        .encrypted_client
+        .database("db")
+        .collection::<Document>("explicit_encryption");
+
+    let insert_payload = testdata
+        .client_encryption
+        .encrypt(
+            RawBson::String("encrypted indexed value".to_string()),
```
kmahar, line 2176: I don't think it makes a functional difference, but it looks like this and the other strings in this test are supposed to use "unindexed" rather than "indexed"

src/test/csfle.rs
```diff
@@ -1269,3 +1309,1439 @@ async fn custom_endpoint_kmip_invalid_endpoint() -> Result<()> {
...
+        return Ok(());
+    }
+    let _guard = LOCK.run_exclusively().await;
+
+    let testdata = match explicit_encryption_setup().await? {
+        Some(t) => t,
+        None => return Ok(()),
+    };
+
+    let raw_value = RawBson::String("encrypted indexed value".to_string());
```
kmahar, line 2250: similarly, should be "unindexed"

kmahar (commented):
great job with this!  I just have some minor comments

src/test/csfle.rs
abr-egn: Fixed.

src/test/csfle.rs
abr-egn: Fixed.

src/test/csfle.rs
```diff
@@ -1269,3 +1309,1439 @@ async fn custom_endpoint_kmip_invalid_endpoint() -> Result<()> {
...
+            (KmsProvider::Aws, tls_opts.clone()),
+            (KmsProvider::Azure, tls_opts.clone()),
+            (KmsProvider::Gcp, tls_opts.clone()),
+            (KmsProvider::Kmip, tls_opts.clone()),
+        ]
+        .into_iter()
+        .collect()
+    }
+
+    let client_encryption_no_client_cert = ClientEncryption::new(
```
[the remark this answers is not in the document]
abr-egn: Yup, that comes from the default provider setup (`KMS_PROVIDERS`).

src/cmap/conn/stream_description.rs
```diff
@@ -59,7 +59,10 @@ impl StreamDescription {
                 .logical_session_timeout_minutes
                 .map(|mins| Duration::from_secs(mins as u64 * 60)),
             max_bson_object_size: reply.command_response.max_bson_object_size,
-            max_write_batch_size: reply.command_response.max_write_batch_size,
+            max_write_batch_size: reply
+                .command_response
+                .max_write_batch_size
+                .unwrap_or(100_000),
```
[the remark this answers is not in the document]
abr-egn: Creating the `StreamDescription` happens unconditionally on handshake; to error on usage would require propagating the `Option`ality further, and I'd rather keep the modification needed for this minimal.  I added a comment and a todo pointing at RUST-1532 for this.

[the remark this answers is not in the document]
kmahar: sgtm!

kmahar (approved):

.evergreen/config.yml
```diff
@@ -1868,14 +1893,70 @@ buildvariants:
...
       - ubuntu-20.04
     mongodb-version:
       - "6.0"
-  display_name: "CSFLE on mongodb ${mongodb-version} / ${os}"
+    topology:
+      - "standalone"
+    crypt-shared:
+      - "enabled"
+    tls-feature:
+      - "openssl"
```
patrickfreed, line 1958: Given the weirdness of OpenSSL on different platforms, I think it might be a good idea to expand this to Windows and macOS.

patrickfreed (commented):
Looks great! I have a few minor comments / questions

.evergreen/config.yml
```diff
@@ -1868,14 +1893,70 @@ buildvariants:
   tasks:
     - "serverless_task_group"
 
-# TODO(RUST-1386): Expand the os and version listing
-- matrix_name: "csfle"
+- matrix_name: "csfle-topology"
```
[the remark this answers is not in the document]
patrickfreed: Ah that's unfortunate that it doesn't work quite like the documentation says. Definitely agree that the current set of definitions is preferable to this giant exclusion.

And yeah, moving to `generate.tasks` is something the drivers team has needed to do for some time. I filed RUST-1539 to cover the work for us doing that in Rust at least.

patrickfreed (approved):
LGTM!

isabelatkinson (approved):
lgtm! two small comments but no need to re-review

PR #768, opened 2022-10-26
RUST-1536: tidy the client-side encryption API

abr-egn:
This adds convenience builders for the two entry points (`Client` and `ClientEncryption`), and updates methods on the latter to use the action pattern.

### DIFF
--- a/src/client/csfle/client_encryption.rs
+++ b/src/client/csfle/client_encryption.rs
@@ -41,23 +41,45 @@ impl ClientEncryption {
     /// # let kv_client = todo!();
     /// # let kv_namespace = todo!();
     /// # let local_key = doc! { };
-    /// let enc = ClientEncryption::builder()
-    ///     .key_vault_client(kv_client)
-    ///     .key_vault_namespace(kv_namespace)
-    ///     .kms_providers([
+    /// let enc = ClientEncryption::new(
+    ///     kv_client,
+    ///     kv_namespace,
+    ///     [
     ///         (KmsProvider::Local, doc! { "key": local_key }, None),
     ///         (KmsProvider::Kmip, doc! { "endpoint": "localhost:5698" }, None),
-    ///     ])?
-    ///     .build();
+    ///     ]
+    /// )?;
     /// # Ok(())
     /// # }
     /// ```
-    pub fn builder() -> ClientEncryptionBuilder<BUILDER_UNSET, BUILDER_UNSET, BUILDER_UNSET> {
-        ClientEncryptionBuilder {
-            key_vault_client: None,
-            key_vault_namespace: None,
-            kms_providers: None,
-        }
+    pub fn new(
+        key_vault_client: Client,
+        key_vault_namespace: Namespace,
+        kms_providers: impl IntoIterator<Item = (KmsProvider, bson::Document, Option<TlsOptions>)>,
+    ) -> Result<Self> {
+        let kms_providers = KmsProviders::new(kms_providers)?;
+        let crypt = Crypt::builder()
+            .kms_providers(&kms_providers.credentials()?)?
+            .build()?;
+        let exec = CryptExecutor::new_explicit(
+            key_vault_client.weak(),
+            key_vault_namespace.clone(),
+            kms_providers.tls_options().clone(),
+        )?;
+        let key_vault = key_vault_client
+            .database(&key_vault_namespace.db)
+            .collection_with_options(
+                &key_vault_namespace.coll,
+                CollectionOptions::builder()
+                    .write_concern(WriteConcern::MAJORITY)
+                    .read_concern(ReadConcern::MAJORITY)
+                    .build(),
+            );
+        Ok(ClientEncryption {
+            crypt,
+            exec,
+            key_vault,
+        })
     }
 
     /// Creates a new key document and inserts into the key vault collection.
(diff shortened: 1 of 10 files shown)


src/client/csfle/client_encryption.rs
```diff
@@ -234,8 +281,7 @@ impl ClientEncryption {
 }
 
 /// Options for initializing a new `ClientEncryption`.
-#[derive(Debug, Clone, TypedBuilder)]
```
abr-egn, line 237: With the builder for the primary entrypoint, it seemed like it would be confusing and redundant to have a builder for the options as well (this also goes for `AutoEncryptionOptions`).

src/client/csfle/client_encryption.rs
```diff
@@ -331,24 +483,40 @@ pub enum MasterKey {
...
-    /// `AutoEncryptionOptions`. `AutoEncryptionOptions.bypass_query_analysis may be true.
-    /// `AutoEncryptionOptions.bypass_auto_encryption` must be false.
-    pub algorithm: Algorithm,
-    /// The contention factor.
-    #[builder(default)]
-    pub contention_factor: Option<i64>,
-    /// The query type.
-    #[builder(default)]
-    pub query_type: Option<String>,
+pub(crate) struct EncryptOptions {
```
abr-egn, line 488: With the action pattern, there's no immediate reason for this to be a public type.  If it turns out that there's a need for it adding it back in will be semver-compatible 🙂

src/test/csfle.rs
```diff
@@ -107,25 +96,17 @@ async fn custom_key_material() -> Result<()> {
...
          ZCvpXSyxTICWSXTUYsWYPyu3IoHbuBZdmw2faM3WhcRIgbMReU5").unwrap();
     let id = enc
-        .create_data_key(
-            &KmsProvider::Local,
-            DataKeyOptions::builder()
-                .master_key(MasterKey::Local)
-                .key_material(key)
-                .build(),
-        )
+        .create_data_key(MasterKey::Local)
```
abr-egn, line 107: Examples like this are, I think, a pretty strong argument in favor of going this way.

src/client/csfle/client_encryption.rs
```diff
@@ -250,25 +296,118 @@ pub struct ClientEncryptionOptions {
...
+            key_vault_namespace,
+            kms_providers,
+            tls_options: None,
+        }
+    }
+}
+
+/// Builder for convenient construction of a new `ClientEncryption`.
+#[derive(Debug)]
+pub struct ClientEncryptionBuilder<const KMS_SET: bool> {
```
patrickfreed, line 320: Is it possible to construct a `ClientEncryptionBuilder<false>` via the public API?

src/client/csfle/client_encryption.rs
```diff
@@ -182,7 +207,29 @@ impl ClientEncryption {
 
     /// Encrypts a BsonValue with a given key and algorithm.
     /// Returns an encrypted value (BSON binary of subtype 6).
-    pub async fn encrypt(&self, value: bson::RawBson, opts: EncryptOptions) -> Result<Binary> {
```
patrickfreed, line 185: One slight drawback of this approach is that we no longer see the return value in the method declaration. This isn't totally unprecedented though, since many of tokio's methods return a named future type. Would there be any value in introducing a sealed trait for our various Action types that indicates the output type?

src/client/csfle/client_encryption.rs
```diff
@@ -182,7 +207,29 @@ impl ClientEncryption {
...
+    ///
+    /// To insert or query with an "Indexed" encrypted payload, use a `Client` configured with
+    /// `AutoEncryptionOptions`. `AutoEncryptionOptions.bypass_query_analysis` may be true.
+    /// `AutoEncryptionOptions.bypass_auto_encryption` must be false.
+    pub fn encrypt(
```
patrickfreed, line 214: I think a `#[must_use]` annotation would be useful here so users don't forget to execute their operation. Ditto for any other method that returns an action.

patrickfreed, line 214: I think it would be worth including a doc example for any public API method that uses the pattern so that users can more easily discover how they're supposed to ultimately execute their operation.

src/client/csfle/options.rs
```diff
@@ -20,12 +19,10 @@ use crate::{
 /// https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#libmongocrypt-auto-encryption-allow-list
 /// )). To bypass automatic encryption for all operations, set bypassAutoEncryption=true in
 /// AutoEncryptionOpts.
-#[derive(Debug, TypedBuilder, Clone)]
+#[derive(Debug, Clone)]
 #[non_exhaustive]
-#[builder(field_defaults(setter(into)))]
```
patrickfreed, line 25: Since there's still a public `Client::with_encryption_options`, I think we probably want to retain the builder functionality here.

src/client/mod.rs
```diff
@@ -153,6 +161,21 @@ impl Client {
         Ok(client)
     }
 
+    /// Return an `EncryptedClientBuilder` for constructing a `Client` with auto-encryption enabled.
+    #[cfg(feature = "csfle")]
+    pub fn encrypted_builder(
+        options: ClientOptions,
+        key_vault_namespace: crate::Namespace,
+    ) -> EncryptedClientBuilder<false> {
```
patrickfreed, line 169: Could be nice to introduce aliases for these to better communicate what the const generic means, e.g.
```rust
pub type EncryptedClientBuilderNoKmsProviders = EncrypedClientBuilder<false>;
pub type EncryptedClientBuilderReady = EncrypedClientBuilder<true>;
```
These names aren't great but I imagine there could be good ones we could choose.

src/test/csfle.rs
patrickfreed: It definitely is way less verbose, which is a huge win. I also imagine the integration with rust-analyzer is much better too. So while overall it does seem to be a better user experience, it is a pretty big departure from what the rest of our API currently looks like, which concerns me a bit. @kmahar @isabelatkinson what do you all think?

src/client/mod.rs
```diff
@@ -536,3 +559,100 @@ impl WeakClient {
...
+
+#[cfg(feature = "csfle")]
+pub struct EncryptedClientBuilder<const KMS_SET: bool> {
+    options: ClientOptions,
+    enc_options: AutoEncryptionOptions,
+}
+
+#[cfg(feature = "csfle")]
+impl<const KMS_SET: bool> EncryptedClientBuilder<KMS_SET> {
+    /// Add configuration for a single KMS provider.
```
patrickfreed, line 571: I think it would be good to include the note about already-added configurations in this docstring too, as well as an example showing that setting multiple providers is okay.

src/client/mod.rs
```diff
@@ -536,3 +559,100 @@ impl WeakClient {
...
+    ) -> EncryptedClientBuilder<true> {
+        self.enc_options.kms_providers.insert(provider, config);
+        EncryptedClientBuilder {
+            options: self.options,
+            enc_options: self.enc_options,
+        }
+    }
+
+    /// Add configuration for multiple KMS providers.  Already-added configuration for other
+    /// providers will be preserved.
```
patrickfreed, line 585: I think it would be good to have a doc-example that demonstrates this vs an equalent invocation using `kms_provider`.

src/client/mod.rs
```diff
@@ -536,3 +559,100 @@ impl WeakClient {
...
+        self
+    }
+
+    pub fn extra_options(mut self, extra_opts: impl Into<Option<Document>>) -> Self {
+        self.enc_options.extra_options = extra_opts.into();
+        self
+    }
+
+    /// Add TLS options for a single KMS provider.
+    pub fn tls_options(mut self, provider: KmsProvider, options: TlsOptions) -> Self {
```
patrickfreed, line 618: Would we want to combine these TLS options ones with the methods for setting the KMS provider(s)? That way it wouldn't be possible to specify TLS options for a provider that wasn't specified.

If we don't go that route, then I think we'll want to document what happens in that case in the docstring.

patrickfreed (commented):
The new API looks really interesting! I left a bunch of comments on possible improvements if we go this route, but I'd hold off on implementing them until the rest of the team gets a chance to weigh in on the direction generally.

src/client/mod.rs
abr-egn: Done.

src/client/mod.rs
abr-egn: Done.

src/client/csfle/options.rs
abr-egn: Hmm.  Is it worth keeping `Client::with_encryption_options`, though?

src/client/csfle/client_encryption.rs
abr-egn: The drawback of introducing a trait is that clients would have to import that trait; for now I've gone with adding specific documentation on return types.

src/client/csfle/client_encryption.rs
abr-egn: Good call, done.

src/client/csfle/client_encryption.rs
```diff
@@ -250,25 +330,123 @@ pub struct ClientEncryptionOptions {
...
+impl<'a> CreateDataKeyAction<'a> {
+    /// Execute the pending data key creation.
+    pub async fn run(self) -> Result<Binary> {
+        self.client_enc
+            .create_data_key_final(&self.opts.master_key.provider(), self.opts)
+            .await
+    }
+
+    /// Set an optional list of alternate names that can be used to reference the key.
+    pub fn key_alt_names(mut self, names: impl IntoIterator<Item = String>) -> Self {
```
patrickfreed, line 438: Would we want to do the `key_alt_name` + `extend_key_alt_names` approach here too?

src/client/mod.rs
```diff
@@ -536,3 +561,129 @@ impl WeakClient {
...
+
+    pub fn bypass_query_analysis(mut self, bypass: impl Into<Option<bool>>) -> Self {
+        self.enc_options.bypass_query_analysis = bypass.into();
+        self
+    }
+}
+
+#[cfg(feature = "csfle")]
+impl EncryptedClientBuilder<{ client_encryption::BUILDER_READY }> {
+    pub async fn build(self) -> Result<Client> {
```
patrickfreed, line 686: Think it would be good to note in this docstring what kind of work this method does, since I think it is a little bit uncommon (but not an issue) for builder's `build` methods to be async.

src/client/mod.rs
patrickfreed: If we go the route of making `AutoEncryptionOptions` `pub(crate)`, what do you think about unifying them, since the correspondence with the options would matter less in that case?

In general, I think I would be a bit surprised as a user if the bulk and single version of the same method had different functionality, so I'd be in favor of updating the `extend` ones regardless.

patrickfreed (commented):
Looks good! I mostly have documentation comments at this point. Going to tag the other reviewers in now as this is mostly an API-related change.

Also, could we file a ticket to capture this work given the API ramifications? I think that would also be a good place to capture some of the discussion we had in triage about this.

src/client/mod.rs
```diff
@@ -153,6 +163,21 @@ impl Client {
         Ok(client)
     }
 
+    /// Return an `EncryptedClientBuilder` for constructing a `Client` with auto-encryption enabled.
```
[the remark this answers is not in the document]
patrickfreed: As an alternative to the const generic approach, have we considered having the user supply the KMS providers up front like we do with the key vault namespace?

patrickfreed (commented):
Looks great! Just have one remaining API question, so tagging in the other reviewers.

kmahar (approved):

src/client/mod.rs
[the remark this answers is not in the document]
abr-egn: From the discussion in person, the team had a leaning towards using builders with typestate for the required fields; in practice, this turned out to be _very_ noisy, so I went with option b of using plain arguments for required fields, including this one.  This meant that there was no reason for `ClientEncryption` to have a builder, so it just has a `new` method now; I figure if new optional fields are introduced in the future we can just introduce a `::builder` and leave `new` as "all options with default".

patrickfreed (approved):
LGTM!

isabelatkinson (approved):

PR #769, opened 2022-10-27
minor: bump clippy version to 1.64

kmahar:


abr-egn (approved):
LGTM!  This is small enough that I don't think it needs the full round.

PR #770, opened 2022-11-02
RUST-1387 Execute CSFLE unified spec tests

abr-egn:
RUST-1387

This updates the unified runner to support csfle tests (mostly just new operations).  This PR also pulls in the legacy test files - nothing uses those yet but there didn't seem a reason not to.


Cargo.toml
```diff
@@ -75,6 +75,9 @@ csfle = ["mongocrypt", "rayon", "num_cpus"]
 # TODO: pending https://github.com/tokio-rs/tracing/issues/2036 stop depending directly on log.
 tracing-unstable = ["tracing", "log"]
 
+[patch."https://github.com/mongodb/bson-rust"]
```
abr-egn, line 78: This won't get merged, it's just needed until https://github.com/mongodb/bson-rust/pull/382 goes in.

src/test/csfle.rs
```diff
@@ -459,7 +441,7 @@ async fn external_key_vault() -> Result<()> {
 fn load_testdata_raw(name: &str) -> Result<String> {
     let path: PathBuf = [
         env!("CARGO_MANIFEST_DIR"),
-        "src/test/spec/json/client-side-encryption/testdata",
+        "src/test/spec/json/testdata/client-side-encryption",
```
abr-egn, line 444: The testsdata for the prose tests needed to move so it doesn't conflict with the way `update-spec-tests.sh` works.

src/test/spec/unified_runner/operation.rs
```diff
@@ -170,24 +176,25 @@ impl Operation {
...
-                                            "[{}] result mismatch, expected = {:#?}  actual = \
-                                             {:#?}\nmismatch detail: {}",
-                                            description, expected_bson, actual, e
-                                        );
-                                    }
-                                } else {
-                                    panic!(
+                                let actual = match &entity {
+                                    Entity::Bson(bs) => Some(bs),
+                                    Entity::None => None,
```
abr-egn, line 181: This allows `Entity::None` to successfully match against `Bson::None`; the logic here is otherwise unchanged.

src/test/spec/unified_runner/operation.rs
```diff
@@ -2769,7 +2792,192 @@ impl TestOperation for Upload {
...
+            };
+            Ok(Some(entity))
+        }
+        .boxed()
+    }
+}
+
+#[derive(Debug, Deserialize)]
+pub(super) struct UnimplementedOperation {
+    _name: String,
```
kmahar, line 2980: nice improvement to store this

src/test/spec/unified_runner/operation.rs
kmahar: out of curiosity, where did the need for that come up? also, what actually is `Entity::None` used for and supposed to correspond to? the only usage I see is from the `FindOne` operation, but we already return a `Result<Option<Entity>>` there so I'm not sure why we wouldn't just return `None`

.evergreen/run-csfle-tests.sh
```diff
@@ -26,6 +26,9 @@ cargo_test() {
 
 set +o errexit
 
-cargo_test test::csfle > results.xml
+cargo_test test::csfle > prose.xml
```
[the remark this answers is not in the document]
abr-egn: Ah, well spotted - I've filed RUST-1535 to track this, it'll take a bit of doing since both serverless and csfle have substantial setup.

src/test/spec/unified_runner/test_runner.rs
```diff
@@ -538,6 +544,31 @@ impl TestRunner {
...
+                TestFileEntity::ClientEncryption(client_enc) => {
+                    let id = client_enc.id.clone();
+                    let opts = &client_enc.client_encryption_opts;
+                    let kv_client = self
+                        .get_client(&opts.key_vault_client)
+                        .await
+                        .client()
+                        .clone();
+                    let kv_namespace =
+                        crate::Namespace::from_str(&opts.key_vault_namespace).unwrap();
```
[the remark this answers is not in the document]
kmahar: think this wasn't actually pushed

kmahar:
LGTM mod the outstanding namespace thread so tagging everyone else in!

src/test/spec/unified_runner/test_runner.rs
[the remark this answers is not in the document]
abr-egn: Whoops, yup, fixed.

kmahar (approved):

src/test/spec/unified_runner/test_file.rs
```diff
@@ -133,6 +134,11 @@ impl RunOnRequirement {
                 return false;
             }
         }
+        if let Some(csfle) = &self.csfle {
+            if *csfle && std::env::var("KMS_PROVIDERS").is_err() {
```
isabelatkinson, line 138: if `csfle` is set in the test file, does that mean that `KMS_PROVIDERS` should be set? or is it valid for one of these values to be set but not the other? if it's the former, perhaps we should panic or log some kind of message here to avoid skipping CSFLE tests silently

src/client/csfle/client_encryption.rs
```diff
@@ -32,6 +32,12 @@ pub struct ClientEncryption {
     key_vault: Collection<RawDocumentBuf>,
 }
 
+impl std::fmt::Debug for ClientEncryption {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ClientEncryption").finish()
```
patrickfreed, line 37: Is this for redaction purposes? If so, I think it could be good to follow the existing format for `Credential` (which uses `debug_tuple` and "REDACTED" as the body) as well as introduce a test ensuring it is in fact redacted.

Also, are there any other `Debug` implementations that we might want to redact? Maybe in `Client`?

patrickfreed (commented):
Looks good mod one minor suggestion and Isabel's comments

patrickfreed (approved):
LGTM!

PR #773, opened 2022-11-08
RUST-1477 Add 1.x EOL notice

patrickfreed:
RUST-1477

This PR adds a notice to the README that the 1.x version of the driver has reached EOL status and won't receive any more updates. I ended up not adding any compiler warnings or anything since the `compile_warning` macro never got introduced to the language.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ For more detailed documentation, see [the manual](https://mongodb.github.io/mong
 - [Contributing](#contributing)
 - [Running the tests](#running-the-tests)
 - [Continuous Integration](#continuous-integration)
+- [Minimum supported Rust version (MSRV)](#minimum-supported-rust-version-msrv)
+- [`mongodb` 1.x EOL notice](#mongodb-1x-eol-notice)
 - [License](#license)
 
 ## Installation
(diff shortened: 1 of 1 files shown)


isabelatkinson (approved):

abr-egn (approved):

README.md
```diff
@@ -28,6 +28,8 @@ For more detailed documentation, see [the manual](https://mongodb.github.io/mong
 - [Contributing](#contributing)
 - [Running the tests](#running-the-tests)
 - [Continuous Integration](#continuous-integration)
+- [Minimum supported Rust version (MSRV)](#minimum-supported-rust-version-msrv)
```
kmahar, line 31: would we want to move these sections, maybe along with the one about which platforms we support higher up in the README, maybe with the installation info? these things seem more relevant and important to users off the bat than a lot of the info that precedes this, like around contributing to the driver

patrickfreed: Good idea, moved both the EOL notice and the platform support into the installation section (as a side note, the platform support could be more explicit. Filed RUST-1543 about updating it). I also updated the wording of the EOL notice to fit the installation section better.

The MSRV is actually already in the installation section under "Requirements". I added a link to the MSRV policy from that point, but kept the policy section here.

kmahar (approved):

PR #774, opened 2022-11-09
RUST-1507 Implement `AsyncWrite` for `GridFsUploadStream`

isabelatkinson:


isabelatkinson:
Note: there are some rustdoc failures caused by making various types public. These will be resolved in a future PR that updates all the GridFS docs.

src/gridfs/upload.rs
```diff
@@ -173,3 +179,283 @@ impl GridFsBucket {
...
+                    chunk_size: stream.chunk_size,
+                    upload_date: DateTime::now(),
+                    filename: stream.filename.take(),
+                    metadata: stream.metadata.take().unwrap(),
+                };
+
+                let new_future = close(stream.bucket.clone(), buffer, file).boxed();
+                stream.state.set_closing(new_future)
+            }
+            State::Writing(_) => return Poll::Pending,
```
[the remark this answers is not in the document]
kmahar: this seems reasonable to me. can we capture what you said about the unreachability in a comment here?

src/gridfs/upload.rs
```diff
@@ -173,3 +179,283 @@ impl GridFsBucket {
...
+        stream.state = State::Closed;
+        match result {
+            Ok(()) => Poll::Ready(Ok(())),
+            Err(error) => Poll::Ready(Err(error.into_futures_io_error())),
+        }
+    }
+}
+
+impl Drop for GridFsUploadStream {
+    fn drop(&mut self) {
```
kmahar, line 343: can we add a TODO here referencing RUST-1439? (I updated the ticket to include this type as one to cover there)

kmahar (changes requested):
some really minor comments, but LGTM besides those! tagging in the others

src/gridfs/options.rs
```diff
@@ -33,11 +33,12 @@ pub struct GridFsBucketOptions {
 /// [`GridFsBucket`].
 #[derive(Clone, Debug, Default, Deserialize, TypedBuilder)]
 #[serde(rename_all = "camelCase")]
-#[builder(field_defaults(setter(into)))]
+#[builder(field_defaults(default, setter(into)))]
 #[non_exhaustive]
 pub struct GridFsUploadOptions {
     /// The number of bytes per chunk of this file. Defaults to the `chunk_size_bytes` specified
     /// in the [`GridFsBucketOptions`].
+    #[serde(rename = "chunkSizeBytes")]
```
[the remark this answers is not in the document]
isabelatkinson: oh I think this was pulled in from a rebase, removed

kmahar (approved):

src/gridfs.rs
```diff
@@ -322,3 +230,9 @@ impl GridFsBucket {
         Ok(())
     }
 }
+
+impl Error {
+    fn into_futures_io_error(self) -> futures_io::Error {
+        futures_io::Error::new(futures_io::ErrorKind::Other, self)
```
patrickfreed, line 236: A `futures_io` error is just a `std::io::Error`, which we wrap in our `ErrorKind` too. In the event that our `Error` is that variant, we should just return it straight without mapping to the `Other` case.

src/gridfs/upload.rs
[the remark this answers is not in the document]
patrickfreed: I think it _might_ be possible to reach this if a user used something like `tokio::time::timeout` and stopped polling the future before it was done and then tried to close it, either explicitly or via the `Drop` implementation.

Given that this is effectively a "cancellation" case, we may want to use the `abort` behavior here. Would be worth writing a test to verify if this is actually reachable in that case.

src/gridfs/upload.rs
```diff
@@ -173,3 +179,287 @@ impl GridFsBucket {
...
+
+        stream.state = State::Closed;
+        match result {
+            Ok(()) => Poll::Ready(Ok(())),
+            Err(error) => Poll::Ready(Err(error.into_futures_io_error())),
+        }
+    }
+}
+
+async fn write_bytes(
```
patrickfreed, line 358: I think it would be worth documenting here that the `u32` returned is number of chunks rather than number of bytes, which someone might assume without reading the implementation.

src/gridfs/upload.rs
```diff
@@ -173,3 +179,287 @@ impl GridFsBucket {
...
+    bucket.chunks().insert_many(chunks, None).await?;
+    buffer.drain(..(n * chunk_size) as usize);
+
+    Ok((n, buffer))
+}
+
+async fn close(bucket: GridFsBucket, buffer: Vec<u8>, file: FilesCollectionDocument) -> Result<()> {
+    let insert_result: Result<()> = async {
+        if !buffer.is_empty() {
+            let final_chunk = Chunk {
```
patrickfreed, line 395: might be worth adding a debug assert here that `buf.len() < chunk_size` for testing purposes.

src/gridfs/upload.rs
```diff
@@ -173,3 +179,287 @@ impl GridFsBucket {
...
+    }
+    .await;
+
+    match insert_result {
+        Ok(()) => Ok(()),
+        Err(error) => {
+            bucket
+                .chunks()
+                .delete_many(doc! { "files_id": file.id }, None)
+                .await?;
```
patrickfreed, line 417: Hm this is kind of a weird case here if we run into an error executing the `delete_many`. I wonder if it would be better to issue a `GridFs` error that was something like `"failed inserting files document: {}, then failed removing orphaned chunks: {}"`. WDYT?

src/gridfs/upload.rs
```diff
@@ -173,3 +179,287 @@ impl GridFsBucket {
...
+            }
+            Err(error) => {
+                stream.state = State::Closed;
+                Poll::Ready(Err(error.into_futures_io_error()))
+            }
+        }
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<tokio::io::Result<()>> {
+        // The buffer only contains leftover bytes that couldn't fill an entire chunk, so there's
```
patrickfreed, line 309: Wouldn't writing such leftover bytes to the db be the correct usage of `flush`?

abr-egn (approved):
LGTM modulo Patrick's comments!

src/gridfs.rs
patrickfreed: Ohh right, forgot about that issue. We could try to deal with this via [`Arc::try_unwrap`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap), but it's probably not worth it. Seems fine to leave this as-is.

src/gridfs/upload.rs
patrickfreed: Ah yeah I was imagining just having an irregularly sized chunk, but I remember now that we can't do this. Agreed that doing partial chunk writes doesn't seem worth it, so this seems fine as-is. In the documentation PR, it might be worth calling out that `flush` doesn't do anything due to this requirement.

src/gridfs/upload.rs
```diff
@@ -173,3 +181,318 @@ impl GridFsBucket {
...
+                    filename: stream.filename.take(),
+                    metadata: stream.metadata.take().unwrap(),
+                };
+
+                let new_future = close(stream.bucket.clone(), buffer, file).boxed();
+                stream.state.set_closing(new_future)
+            }
+            // This case is effectively unreachable, as the AsyncWriteExt methods take &mut self and
+            // poll the futures to completion. If a user were to call these polling methods directly
+            // and intersperse futures, we should just pend here until the writing future resolves.
```
patrickfreed, line 337: This comment can be removed now I think.

src/gridfs/upload.rs
```diff
@@ -173,3 +181,318 @@ impl GridFsBucket {
...
+            None => Ok(()),
+        },
+        Err(error) => {
+            let mut message = String::new();
+            if let Some(original_error) = original_error {
+                message.push_str(&format!("failed operation: {}, then ", original_error));
+            };
+            message.push_str("failed to delete orphaned chunks: {}");
+            Err(ErrorKind::GridFs(GridFsErrorKind::AbortError { message }).into())
+        }
```
patrickfreed, line 452: Sorry for the back-and-forth on this, but it seems like it might be nice to have both of these as `Error` in the `AbortError` variant, and then have the `Display` implementation join them into a string like this. WDYT?

patrickfreed (commented):
Looks good! just some minor comments left

src/gridfs/upload.rs
isabelatkinson: updated. I think the naming/docs make it clear enough what's going on but I can also add a custom `Display` impl if you think that would be better

patrickfreed (approved):
LGTM!

PR #781, opened 2022-11-21
RUST-1387 Run csfle legacy spec tests

abr-egn:
RUST-1387

Along with the additions to the v2 test runner to support the tests, this also has various minor bugfixes that the tests turned up and some updates to the test code to help debugging and make test client initialization a little less messy.


.evergreen/config.yml
```diff
@@ -511,12 +511,19 @@ functions:
...
+          export PYTHON=${PYTHON}
+          ${PYTHON} -m pip install boto3
+
           # Exported without xtrace to avoid leaking credentials
           set +o xtrace
-          export KMS_PROVIDERS=$(cat << "EOF"
-          ${kms_providers}
-          EOF
-          )
+          export AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
```
abr-egn, line 519: These need to be individual rather than a single json blob because the aws temporary credentials script needs the aws parts of this.  I also realized it's better to do it this way so that individual credentials can be easily rotated.

src/test/spec/v2_runner/test_file.rs
```diff
@@ -23,6 +31,12 @@ pub(crate) struct TestFile {
     #[allow(unused)]
     pub(crate) bucket_name: Option<String>,
     pub(crate) data: Option<TestData>,
+    #[cfg(feature = "csfle")]
```
abr-egn, line 34: I would have liked to use `#[serde(flatten)]` with a struct defined elsewhere for clarity, but that continues to be incompatible with `deny_unknown_fields` :(

src/test/util/mod.rs
```diff
@@ -67,43 +68,129 @@ impl std::ops::Deref for TestClient {
...
+            min_heartbeat_freq: None,
+            #[cfg(feature = "csfle")]
+            encrypted: None,
+        }
     }
+}
 
-    pub(crate) async fn with_options(options: impl Into<Option<ClientOptions>>) -> Self {
-        Self::with_handler(None, options).await
+pub(crate) struct TestClientBuilder {
```
abr-egn, line 83: I realize I'm at risk of the "when the only tool you have is a hammer, every problem looks like a nail" pitfall here, but I think the builder pattern works well for this - it means that the previously-duplicated constructors for accepting permutations of options/additional options/event handler/etc. can be folded into a single implementation.  My motivation for this was being able to construct both `TestClient` and `EventClient` with auto-encryption enabled without having to copy-paste the code across them.

src/test/spec/mod.rs
```diff
@@ -103,9 +103,14 @@ where
     G: Future<Output = ()>,
     T: DeserializeOwned,
 {
-    let json: Value = serde_json::from_reader(File::open(path.as_path()).unwrap())
+    let mut json: Value = serde_json::from_reader(File::open(path.as_path()).unwrap())
         .unwrap_or_else(|err| panic!("{}: {}", path.display(), err));
 
+    // TODO RUST-36 Remove this when decimal128 support is implemented
+    if path.ends_with("client-side-encryption/legacy/types.json") {
```
abr-egn, line 110: It's a very ugly hack to have this here, but I couldn't think of a cleaner way - the stripping has to happen before the json is parsed as bson, so test runner predicates are too late.

src/test/spec/v2_runner/operation.rs
```diff
@@ -1212,11 +1215,12 @@ impl TestOperation for AssertCollectionExists {
     ) -> BoxFuture<'a, Result<Option<Bson>>> {
         async move {
             let collections = client
-                .database(&self.database)
+                .database_with_options(&self.database, DatabaseOptions::builder().read_concern(ReadConcern::MAJORITY).build())
+                //.database(&self.database)
```
isabelatkinson, line 1219:
```suggestion

```

src/test/spec/v2_runner/mod.rs
```diff
@@ -39,9 +43,15 @@ const SKIPPED_OPERATIONS: &[&str] = &[
...
 
-pub(crate) async fn run_v2_test(test_file: TestFile) {
+pub(crate) async fn run_v2_test(path: std::path::PathBuf, test_file: TestFile) {
     let internal_client = TestClient::new().await;
 
+    file_level_log(format!("Running tests from {}", path.display(),));
+    #[cfg(not(feature = "csfle"))]
+    let is_csfle_test = false;
+    #[cfg(feature = "csfle")]
+    let is_csfle_test = path.to_string_lossy().contains("client-side-encryption");
```
isabelatkinson, line 53: why is the feature check needed here? shouldn't the check on the path always evaluate to false if the feature is off?

abr-egn: Yup, you're right.  Fixed.

src/test/util/event.rs
```diff
@@ -537,6 +538,32 @@ impl std::ops::DerefMut for EventClient {
...
+        if inner.handler.is_none() {
+            inner = inner.event_handler(Arc::new(EventHandler::new()));
+        }
+        let handler = inner.handler().unwrap().clone();
+        let client = inner.build().await;
+
+        // clear events from commands used to set up client.
+        handler.command_events.write().unwrap().clear();
+
+        EventClient { client, handler }
```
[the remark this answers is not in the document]
abr-egn: Yup, this makes sense.  I'd prefer to keep that work distinct, so I've updated RUST-1449.

src/test/util/mod.rs
```diff
@@ -67,43 +68,129 @@ impl std::ops::Deref for TestClient {
...
+    pub(crate) fn options(mut self, options: impl Into<Option<ClientOptions>>) -> Self {
+        let options = options.into();
+        assert!(self.options.is_none() || options.is_none());
+        self.options = options;
+        self
     }
 
-    async fn build_test_options(
-        event_handler: Option<Arc<EventHandler>>,
+    pub(crate) async fn additional_options(
```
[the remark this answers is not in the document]
abr-egn: I left a somewhat minimal comment and a TODO - the actual effects of `options_for_multiple_mongoses` are super messy and unclear, and a prime candidate for dismantling in RUST-1449 (already called out there).

src/test/spec/mod.rs
abr-egn: I think this is independent of that - the field that has the issue is `clientOptions.autoEncryptOpts.schemaMap`, which is itself a `bson::Document`; the test file has some fields in that doc that are decimal128 which currently can't be represented in the `bson` crate datastructures.

src/test/util/mod.rs
abr-egn: Yup, makes sense.  Updated that ticket and left a TODO here.

isabelatkinson (approved):
code changes lgtm, although it looks like there are some evg failures. tagging in the rest of the team

src/test/spec/client_side_encryption.rs
```diff
@@ -28,3 +33,29 @@ fn spec_predicate(test: &TestCase) -> bool {
...
     true
 }
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test(flavor = "multi_thread"))]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn run_legacy() {
+    let _guard: RwLockWriteGuard<()> = LOCK.run_exclusively().await;
+
+    run_spec_test_with_path(&["client-side-encryption", "legacy"], |path, test| async {
+        if path.ends_with("timeoutMS.json") {
```
patrickfreed, line 43: can we add a TODO for RUST-582 to unskip this?

patrickfreed (approved):
Looks great! Just have one minor comment about adding a TODO, otherwise lgtm

src/test/spec/client_side_encryption.rs
abr-egn: Done.

kmahar (approved):

PR #782, opened 2022-11-22
RUST-1531 Improve test coverage for retryable handshake errors

isabelatkinson:


src/coll/options.rs
```diff
@@ -975,7 +975,10 @@ pub struct FindOneOptions {
     ///
     /// This options maps to the `maxTimeMS` MongoDB query option, so the duration will be sent
     /// across the wire as an integer number of milliseconds.
-    #[serde(deserialize_with = "bson_util::deserialize_duration_option_from_u64_millis")]
+    #[serde(
+        default,
```
isabelatkinson, line 979: adding `default` here was the only code change needed; just requesting review as a sanity check

kmahar (approved):

kmahar:
LGTM, I don't think others need to review since it's so small.

PR #787, opened 2022-11-28
RUST-1535 Run csfle tests against serverless

abr-egn:
RUST-1535

This updates evergreen configuration to run csfle tests against a serverless setup.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -483,6 +483,39 @@ functions:
         export TLS_FEATURE=${TLS_FEATURE}
         .evergreen/run-csfle-kmip-servers.sh
 
+  "build csfle expansions":
+    - command: shell.exec
+      params:
+        shell: bash
+        working_dir: "src"
+        script: |
+          ${PREPARE_SHELL}
+          set +o xtrace
+          cat <<EOT > csfle-expansions.yml
+          PREPARE_CSFLE: |
+            export ASYNC_RUNTIME=${ASYNC_RUNTIME}
+            export CSFLE_TLS_CERT_DIR=$DRIVERS_TOOLS_X509
+            export DISABLE_CRYPT_SHARED=${DISABLE_CRYPT_SHARED}
+            export TLS_FEATURE=${TLS_FEATURE}
+            export CRYPT_SHARED_LIB_PATH=${CRYPT_SHARED_LIB_PATH}
+            export PYTHON=${PYTHON}
+            ${PYTHON} -m pip install boto3
+            # Exported without xtrace to avoid leaking credentials
+            set +o xtrace
+            export AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -1180,8 +1223,24 @@ tasks:
       - func: "fetch crypt_shared"
       - func: "bootstrap mongo-orchestration"
       - func: "run kmip server"
+      - func: "build csfle expansions"
       - func: "run csfle tests"
 
+  - name: "test-csfle-serverless"
+    tags: ["csfle", "serverless"]
+    commands:
+      - command: expansions.update
```
abr-egn, line 1232: I have _no dang idea_ why evergreen doesn't have expansions set when it runs this task.  If I had to guess, it would be that there's some internal state reset between `task_group` tasks that clears them, so they persist from setup into the `test-serverless` task but not in this one.

.evergreen/run-csfle-tests.sh
```diff
@@ -26,7 +26,7 @@ CARGO_RESULT=0
 
 cargo_test() {
     RUST_BACKTRACE=1 \
-        cargo test --features ${FEATURE_FLAGS} $1 ${OPTIONS} | cargo2junit
+        cargo test --features ${FEATURE_FLAGS} $1 ${OPTIONS} | grep -v '{"t":' | cargo2junit
```
abr-egn, line 29: The other thing I have no dang idea about - occasionally some output from the spawned `mongocryptd` shows up in the stdout of the test, causing `cargo2junit` to barf.  This shouldn't be happening - stdout and stderr are closed before spawning it, and I've validated that behavior with a test binary - but there it is.

kmahar: weird 🤷‍♀️ filtering out the test lines seems reasonable to me

.evergreen/config.yml
```diff
@@ -1889,6 +1948,10 @@ buildvariants:
     os:
       - ubuntu-18.04
     async-runtime: "*"
+    crypt-shared:
+      - "enabled"
+    tls-feature:
```
kmahar, line 1953: does this mean we now only test serverless with openssl? seems probably fine (I guess we were doing the opposite before, only rustls)

.evergreen/config.yml
kmahar: weird. I don't understand evergreen expansions very well at all so I'm not sure I can help answer this either 🙂

kmahar (approved):

isabelatkinson (approved):

patrickfreed (approved):
lgtm, but am curious about @kmahar's question re: testing only on OpenSSL

.evergreen/config.yml
abr-egn: I don't actually recall why I was only testing against openssl; I've updated this to test against both.  Currently serverless instance creation is timing out on evergreen, but initial runs showed passes for all of it.

Issue #777, opened 2022-11-13
Build Error: the name `RUNTIME_NAME` is defined multiple times


Y0h4n3s:
## Versions/Environment
1. Rust version: 1.65.0-nightly (4ed54cecc 2022-08-27)
2. Os: Ubuntu 22, 6.0.2-76060002-generic x86_64
3. bson@2.4.0, mongodb@2.3.1
4. MongoDb version: 5.0.8 Standalone

I'm getting this build error
```
error[E0428]: the name `RUNTIME_NAME` is defined multiple times
  --> /.../mongodb-2.3.1/src/cmap/establish/handshake/mod.rs:28:1
   |
22 | const RUNTIME_NAME: &str = "tokio";
   | ----------------------------------- previous definition of the value `RUNTIME_NAME` here
...
28 | const RUNTIME_NAME: &str = "sync (with async-std)";
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `RUNTIME_NAME` redefined here
   |
   = note: `RUNTIME_NAME` must be defined only once in the value namespace of this module

error[E0252]: the name `Interval` is defined multiple times
  --> /.../mongodb-2.3.1/src/runtime/mod.rs:34:5
   |
28 | use interval::Interval;
   |     ------------------ previous import of the type `Interval` here
...
34 | use tokio::time::Interval;
   |     ^^^^^^^^^^^^^^^^^^^^^ `Interval` reimported here
   |
   = note: `Interval` must be defined only once in the type namespace of this module
help: you can use `as` to change the binding name of the import
   |
34 | use tokio::time::Interval as OtherInterval;
   |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

error: `tokio-runtime` and `async-std-runtime` can't both be enabled; either disable `async-std-runtime` or set `default-features = false` in your Cargo.toml
   --> /.../mongodb-2.3.1/src/lib.rs:357:1
    |
357 | / compile_error!(
358 | |     "`tokio-runtime` and `async-std-runtime` can't both be enabled; either disable \
359 | |      `async-std-runtime` or set `default-features = false` in your Cargo.toml"
360 | | );
    | |_^

error[E0433]: failed to resolve: use of undeclared crate or module `futures`
  --> /.../mongodb-2.3.1/src/runtime/mod.rs:83:40
   |
[...]

PR #788, opened 2022-11-29
minor: improve compile error when `sync` and `tokio-runtime` are enabled

patrickfreed:
Related to #777


abr-egn (approved):
LGTM!  I don't think this needs full review.

PR #793, opened 2022-12-05
RUST-1310 FLE 2.0 example

abr-egn:
RUST-1310

This provides the requested FLE 2.0 example for the docs team; this is a pretty direct transliteration of the Go example.


src/test/csfle.rs
```diff
@@ -2839,3 +2839,120 @@ async fn bind(addr: &str) -> Result<TcpListener> {
...
+    let ce = ClientEncryption::new(
+        test_client.clone().into_client(),
+        KV_NAMESPACE.clone(),
+        LOCAL_KMS.clone(),
+    )?;
+    let key1_id = ce.create_data_key(MasterKey::Local).run().await?;
+    let key2_id = ce.create_data_key(MasterKey::Local).run().await?;
+
+    // Create an encryptedFieldsMap.
+    let encrypted_fields_map = vec![(
```
kmahar, line 2881: minor: I notice Go accepts the encrypted fields map as a map rather than a vec. I can't really tell if the spec cares which way we do this - is it specified anywhere? I don't remember if we talked about this - out of curiosity, was there a reason we took the `impl IntoIterator<Item = (String, Document)>` approach vs a map?

kmahar (approved):
I have one question, but it's more about our API than this example. LGTM!

src/test/csfle.rs
patrickfreed: I personally like the types we use for all the different configuration options. It can make it hard to reason about things when everything is just a document.

To cut back on the `to_string()` spam at least, we could accept `impl IntoIterator<Item = (impl Into<String>, Document)>` instead I believe.

abr-egn: Huh, I thought I'd tried using nested `impl Foo` for something else and it wasn't a thing that worked, but it works fine here.  That removes that friction point and I has me agreeing on the types.

abr-egn: ... and I just realized I applied that change to #794, not this PR.  Want me to back it out and reapply here, or is it okay where it is?

patrickfreed: If it affects the examples here (which I think it does), then it probably makes sense to reapply here, or whichever one gets merged second. If it doesn't have any bearing on the examples, then either one works imo.

abr-egn: It affects the examples in the manual much more than the example here, so I've left it in that PR.  I'll merge that one first and then update this one before merging.

patrickfreed (approved):

isabelatkinson (approved):

Issue #796, opened 2022-12-17
RUST-1559 Impl Serialize for WriteFailure, WriteConcernError, WriteError, and BulkWriteFailure


borngraced:
This will very useful for returning detailed errors in JSON format in actix web and co.

Currently we can't do something like this with `KhWriteFailure` etc unless `Serialize` is also implemented for the structures either by using `derive_macro!` or manually!
```
#[derive(Debug, Deserialize, Serialize)]
pub enum KhoomiResponseError {
    WriteError(WriteFailure),
}
```

```rust
impl ResponseError for KhoomiResponseError {
    fn status_code(&self) -> StatusCode {
        match self {
              ...,
            _ => StatusCode::INTERNAL_SERVER_ERROR,
        }
    }

    fn error_response(&self) -> HttpResponse {
        HttpResponse::build(self.status_code())
            .insert_header(ContentType::html())
            .json(self)
    }
}
```

I was able to fix this by creating a custom `KhWriteFailure` that implements Serialize and then `impl From<WriteFailure> for KhWriteFailure`

PR #797, opened 2022-12-21
extend error types to impl serialization

borngraced:
Fixes: #796 

Also, fixed cargo clippy and remove unnecessary lifetime annotations

### DIFF
--- a/src/bson_util/mod.rs
+++ b/src/bson_util/mod.rs
@@ -177,7 +177,7 @@ where
 {
     let bson = Bson::deserialize(deserializer)?;
     get_u64(&bson)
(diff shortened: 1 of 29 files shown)


abr-egn:
Hi!  Thanks for the contribution.

Unfortunately, variable interpolation in format strings isn't supported in our MSRV of 1.57 - can you revert that portion of your change?

borngraced:
> Hi! Thanks for the contribution.
> 
> Unfortunately, variable interpolation in format strings isn't supported in our MSRV of 1.57 - can you revert that portion of your change?

all done

abr-egn:
This will go out with the next release (which coincidentally will be in the next week or so) - thanks again!

PR #801, opened 2023-01-04
minor: refactor transactions example

isabelatkinson:
This PR updates our transactions example to align more closely with [other drivers' examples](https://www.mongodb.com/docs/manual/core/transactions-in-applications/#core-api), specifically by including the call to `start_transaction` within the retry. There are still some minor discrepancies due to Rust's limited support of async closures.


tests/transactions_example.rs
```diff
@@ -0,0 +1,91 @@
...
+use mongodb::{
+    bson::{doc, Document},
+    error::{Result, TRANSIENT_TRANSACTION_ERROR, UNKNOWN_TRANSACTION_COMMIT_RESULT},
+    options::{Acknowledgment, ReadConcern, TransactionOptions, WriteConcern},
+    Client,
+    ClientSession,
+};
+
+#[tokio::main]
+async fn main() -> Result<()> {
```
abr-egn, line 13: The previous example didn't include a `main` and it doesn't look like other languages' do either - maybe move this outside the `// START` comment?

abr-egn (approved):

tests/transactions_example.rs
```diff
@@ -0,0 +1,91 @@
...
+                    session.abort_transaction().await?;
+                    return Err(error);
+                }
+            }
+        }
+    }
+}
+
+async fn execute_transaction(session: &mut ClientSession) -> Result<()> {
+    let transaction_options = TransactionOptions::builder()
```
kmahar, line 42: the other lang examples explicitly set read preference primary here; maybe we should too?

tests/transactions_example.rs
```diff
@@ -0,0 +1,91 @@
...
+        .read_concern(ReadConcern::snapshot())
+        .write_concern(WriteConcern::builder().w(Acknowledgment::Majority).build())
+        .build();
+    session.start_transaction(transaction_options).await?;
+
+    let client = session.client();
+    let employees = client.database("hr").collection::<Document>("employees");
+    let events = client.database("reporting").collection("events");
+
+    employees
```
kmahar, line 52: as an aside, this is a perfect use-case for a one-shot transaction in the future 🙂

tests/transactions_example.rs
isabelatkinson: The pseudocode this example is based off of (transactions retry example 3 in DRIVERS-488) does include construction of the client/session and some of the scripting languages' examples (e.g. JS and Python) show those steps. I think it's helpful to demonstrate how to start a session and emphasize that `run_transaction_with_retry` is the entry point here.

kmahar (approved):

patrickfreed (approved):
LGTM, example looks really clear and easy to follow

PR #805, opened 2023-01-12
RUST-1587 Implement server selection tracing events

kmahar:
This implements the server selection tracing/log events/messages as defined in mongodb/specifications#1341. 

There is a workaround in here for the fact that we still currently perform server selection to detect sessions support; logging messages in those cases makes these tests flaky. That flakiness is why the spec PR hasn't been merged yet, though I expect for any driver that doesn't log on server selection attempts for sessions support the tests will pass.

We could merge this and remove the workaround as part of RUST-1499, or alternatively we can wait to merge until that work is in; I don't have a strong opinion either way.

Will leave some more comments inline.


src/client/mod.rs
```diff
@@ -497,31 +504,149 @@ impl Client {
...
-                && watcher
-                    .wait_for_update(timeout - start_time.elapsed())
-                    .await;
-            if !change_occurred {
-                return Err(ErrorKind::ServerSelection {
-                    message: state
-                        .description
-                        .server_selection_timeout_error_message(criteria),
+            );
+            match result {
```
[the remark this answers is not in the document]
kmahar: great idea, done. I ended up adding a helper type to store the shared info and avoid having to pass it with each message. this matches nicely with our existing `XEventEmitter` types for CMAP and command tracing.

src/client/mod.rs
```diff
@@ -480,13 +481,18 @@ impl Client {
         &self,
         criteria: Option<&SelectionCriteria>,
     ) -> Result<ServerAddress> {
-        let server = self.select_server(criteria).await?;
+        let server = self.select_server(criteria, "test_select_server").await?;
```
abr-egn, line 484: The names given elsewhere are short phrases rather than a snake_case tag, might be better to be consistent one way or the other.

src/client/mod.rs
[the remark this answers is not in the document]
abr-egn: Very nice!

src/trace/server_selection.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    pub(crate) fn emit_failed_event(
+        &self,
+        topology_description: &TopologyDescription,
+        error: &Error,
+    ) {
+        if trace_or_log_enabled!(target: SERVER_SELECTION_TRACING_EVENT_TARGET, TracingOrLogLevel::Debug)
+        // TODO: RUST-1585 Remove this condition.
+        && self.operation_name != "Check sessions support status"
```
abr-egn, line 80: Since these values are being matched on, they should be named constants.

abr-egn (approved):
LGTM!  Just a few comments that don't require re-review.

src/trace/server_selection.rs
kmahar: good point. since the constant needs to be used unconditional of the tracing feature (in `get_session_support_status`), this couldn't be hidden away in the trace module. I ended up putting it in `client/mod.rs` for now.

src/client/mod.rs
kmahar: renamed to "Test select server" for consistency.

kmahar:
pushed one more small change: I noticed that the other event emitter types were named `XTracingEventEmitter` but this one was missing the `Tracing` bit so I've renamed the type.

isabelatkinson (approved):
lgtm!

PR #810, opened 2023-01-18
chore: fix clippy `1.67`

attila-lin:


kmahar:
Hi @attila-lin, thanks for the contribution! 

A lot of these changes appear to be applying a new nightly-only rule `uninlined_format_args`. Unfortunately, that rule forces inlined format arguments which were not introduced until Rust 1.58, and our MSRV is 1.57. 

However, there are some other changes in this PR that we can certainly apply now.

It seems what we should do is add a [clippy.toml](https://doc.rust-lang.org/clippy/configuration.html#specifying-the-minimum-supported-rust-version) file specifying our MSRV. Would you be up for adding that and removing the `uninlined_format_args` changes, and also editing the file `.evergreen/check-clippy.sh` to set `CLIPPY_VERSION` to 1.66?

attila-lin:
Yes！ Thanks for your response.

kmahar (approved):
LGTM! Tagging in the the rest of the team for review.

abr-egn (approved):
LGTM, thank you!

isabelatkinson (approved):

kmahar:
thank you for the contribution @attila-lin!

PR #811, opened 2023-01-18
RUST-1069 Defer implicit session check out until after connection check out

abr-egn:
RUST-1069

Previously, retries were handled by a distinct `execute_retry` method; this was an issue for this change because there are two retry cases (connection check out failure and operation failure), and a retry should create an implicit session in the first case but not the second.  This PR changes it to have a loop inside `execute_operation_with_retry`; this makes the handling there a bit more complicated, but removes the `execute_retry` method entirely, and also collapses `ExecutionOutput` into `ExecutionDetails`, so I think it's a net wash on complexity.  Having the loop will also make executing multiple retries easier if/when we get to CSOT.

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -118,8 +118,8 @@ impl Client {
             if let Some(session) = &session {
                 if !Arc::ptr_eq(&self.inner, &session.client().inner) {
                     return Err(ErrorKind::InvalidArgument {
-                        message: "the session provided to an operation must be created from \
-                                  the same client as the collection/database"
+                        message: "the session provided to an operation must be created from the \
+                                  same client as the collection/database"
                             .into(),
                     }
                     .into());
@@ -150,10 +150,8 @@ impl Client {
     {
         Box::pin(async {
             let mut details = self.execute_operation_with_details(op, None).await?;
-            let pinned = self.pin_connection_for_cursor(
-                &details.output,
-                &mut details.connection,
-            )?;
+            let pinned =
+                self.pin_connection_for_cursor(&details.output, &mut details.connection)?;
             Ok(Cursor::new(
                 self.clone(),
                 details.output,
@@ -176,16 +174,9 @@ impl Client {
             .execute_operation_with_details(op, &mut *session)
             .await?;
 
-        let pinned = self.pin_connection_for_session(
-            &details.output,
-            &mut details.connection,
-            session,
-        )?;
-        Ok(SessionCursor::new(
-            self.clone(),
-            details.output,
-            pinned,
-        ))
+        let pinned =
+            self.pin_connection_for_session(&details.output, &mut details.connection, session)?;
+        Ok(SessionCursor::new(self.clone(), details.output, pinned))
     }
 
     fn is_load_balanced(&self) -> bool {
@@ -247,8 +238,7 @@ impl Client {
                 details.implicit_session = Some(session);
             }
             let (cursor_spec, cs_data) = details.output;
-            let pinned =
-                self.pin_connection_for_cursor(&cursor_spec, &mut details.connection)?;
+            let pinned = self.pin_connection_for_cursor(&cursor_spec, &mut details.connection)?;
             let cursor = Cursor::new(self.clone(), cursor_spec, details.implicit_session, pinned);
 
             Ok(ChangeStream::new(cursor, args, cs_data))
@@ -280,11 +270,8 @@ impl Client {
                 .execute_operation_with_details(op, &mut *session)
                 .await?;
             let (cursor_spec, cs_data) = details.output;
-            let pinned = self.pin_connection_for_session(
-                &cursor_spec,
-                &mut details.connection,
-                session,
-            )?;
+            let pinned =
+                self.pin_connection_for_session(&cursor_spec, &mut details.connection, session)?;
             let cursor = SessionCursor::new(self.clone(), cursor_spec, pinned);
 
             Ok(SessionChangeStream::new(cursor, args, cs_data))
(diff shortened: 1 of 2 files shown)


src/client/executor.rs
```diff
@@ -309,167 +285,152 @@ impl Client {
...
-        prior_txn_number: Option<i64>,
-        first_error: Error,
-    ) -> Result<ExecutionOutput<T>> {
-        op.update_for_retry();
-
-        let server = match self.select_server(op.selection_criteria()).await {
-            Ok(server) => server,
-            Err(_) => {
-                return Err(first_error);
+            let retryability = self.get_retryability(&conn, &op, &session)?;
```
isabelatkinson, line 357: Is there any chance this could fail during a retry? If so, I think we should return the original error rather than the one that occurs in this call.

src/client/executor.rs
```diff
@@ -309,167 +285,152 @@ impl Client {
...
-            .or_else(|| op.selection_criteria());
 
-        let server = match self.select_server(selection_criteria).await {
-            Ok(server) => server,
-            Err(mut err) => {
-                err.add_labels_and_update_pin(None, &mut session, None)?;
-                return Err(err);
+            // If the current transaction has been committed/aborted and it is not being
+            // re-committed/re-aborted, reset the transaction's state to TransactionState::None.
+            if let Some(ref mut session) = session {
```
isabelatkinson, line 298: Can this be moved outside of the loop? I don't think these conditions would change on a retry: the session and operation should each stay the same.

src/client/executor.rs
```diff
@@ -309,167 +285,152 @@ impl Client {
...
-                    || retryability == Retryability::Write && err.is_write_retryable()
-                {
-                    self.execute_retry(&mut op, &mut session, txn_number, err)
-                        .await
-                } else {
-                    Err(err)
-                }
+            if session.is_none() {
+                implicit_session = self.start_implicit_session(&op).await?;
+                session = implicit_session.as_mut();
```
isabelatkinson, line 354: IIUC here an implicit session will be started only on the first attempt and reused in subsequent retries. Is this the right behavior? I'm not sure if we should be dropping/re-creating the implicit session after checking out a connection each time around.

src/client/executor.rs
```diff
@@ -309,167 +285,152 @@ impl Client {
...
-            Ok(c) => c,
-            Err(mut err) => {
-                err.add_labels_and_update_pin(None, &mut session, None)?;
-                if err.is_read_retryable() && self.inner.options.retry_writes != Some(false) {
-                    err.add_label(RETRYABLE_WRITE_ERROR);
-                }
+            let selection_criteria = session
+                .as_ref()
+                .and_then(|s| s.transaction.pinned_mongos())
+                .or_else(|| op.selection_criteria());
```
isabelatkinson, line 313: ditto above, does this need to be done in the loop?

isabelatkinson (commented):
Just a few questions. As I reviewed this I noticed some areas in this file that could use refactoring, but I don't think they're relevant to this PR so I filed RUST-1575 to address them in the future.

src/client/executor.rs
abr-egn: Yup, done.

src/client/executor.rs
abr-egn: Unfortunately not - this (potentially) borrows from `op`, so moving it outside the loop would block the call to `op.update_for_retry()` (which requires a &mut).

src/client/executor.rs
abr-egn: I believe this is correct - the goal of the upstream change was to minimize the number of sessions created serverside.

src/client/executor.rs
abr-egn: The only error this can return is an internal one indicating an impossible/invalid connection state, which I think would be valuable to expose if it does happen :)

isabelatkinson (approved):

isabelatkinson:
not sure if we want another review on this -- I'll leave it up to you, Abraham, if you want to wait until Kaitlin is back

abr-egn:
> not sure if we want another review on this -- I'll leave it up to you, Abraham, if you want to wait until Kaitlin is back

No rush. and this is core enough that more eyes will be good :)

src/client/executor.rs
kmahar: the retryable writes spec actually _requires_ that the retry attempt use the same session as the initial operation, so this is definitely right, and I assume there are some spec tests that would start failing if we didn't preserve this behavior.

per [spec](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#executing-retryable-write-commands):

>  When retrying a write command, drivers MUST resend the command with the same transaction ID.

where "transaction ID" is [defined](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#terms) as:

> the top-level lsid and txnNumber fields.

and per the sessions [spec](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#why-do-automatic-retry-attempts-re-use-a-dirty-implicit-session):

> The retryable writes spec requires that both the original and retry attempt use the same server session. The server will block the retry attempt until the initial attempt completes at which point the retry attempt will continue executing.
For retryable reads that use an implicit session, drivers could choose to use a new server session for the retry attempt however this would lose the information that these two reads are related.

kmahar (approved):
one comment that I don't need to re-review for, LGTM otherwise.

src/test/spec/sessions.rs
```diff
@@ -62,3 +69,107 @@ async fn explicit_session_created_on_same_client() {
...
+                }
+            }
+            unique.push(lsid);
+        }
+
+        min_lsids = std::cmp::min(min_lsids, unique.len());
+        max_lsids = std::cmp::max(max_lsids, unique.len());
+        lsids.clear();
+    }
+    assert!(
```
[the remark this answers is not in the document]
abr-egn: Good call, done.

PR #814, opened 2023-01-24
RUST-1576 Allow "timeseries"-typed collections in list_collections output

abr-egn:


isabelatkinson (approved):
lgtm, don't think we need extra reviewers on this one

PR #820, opened 2023-02-02
Improvements to ListCollections

main--:
This PR includes two changes to the ListCollections functionality. The first one is just a small bugfix: `nameOnly` allows filtering on the `type`, which previously caused the code to disable `nameOnly`. The second change is adding support for the `authorizedCollections` flag, which enables listing collections when access control is enforced and the user is only allowed to access a limited subset of the collections in the database.

### DIFF
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -48,7 +48,7 @@ use crate::{
 /// so it can safely be shared across threads or async tasks. For example:
 ///
 /// ```rust
-/// 
+///
 /// # #[cfg(all(not(feature = "sync"), not(feature = "tokio-sync")))]
 /// # use mongodb::{bson::Document, Client, error::Result};
 /// # #[cfg(feature = "async-std-runtime")]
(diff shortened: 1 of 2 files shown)


abr-egn:
Hi!

Thanks for the contribution!  We'd like to avoid adding a toplevel method to support `authorizedCollections` - can you update this PR to include it in the `ListCollectionsOptions` instead?  You'll need to tag it so serde doesn't try to serialize the field because it's not actually part of the options expected by the server.

main--:
>can you update this PR to include it in the `ListCollectionsOptions` instead

I originally wanted to do something like that, but  [`list_collection_names`](https://docs.rs/mongodb/latest/mongodb/struct.Database.html#method.list_collection_names) doesn't take `ListCollectionsOptions` and to my understanding there is no other way to set the `names_only` flag.

abr-egn:
Ah, I see, that's unfortunate.  We'd really prefer to avoid adding a new method just for this option; we'll update the API in 3.0 to allow this (RUST-1603), and in the meanwhile there's work this quarter to add a helper to directly run commands that return a cursor (RUST-1588).

PR #821, opened 2023-02-02
RUST-1566 Sync (and skip) command logging unacknowledged write test

kmahar:
This test was added after our implementation. We can't run it since we don't support uancknowledged writes but syncing the file nonetheless to keep us up-to-date.


abr-egn (approved):
LGTM!  I don't think this needs additional reviews.

Issue #812, opened 2023-01-18
RUST-1582 Would It Be Possible To Expose The Session Creation Functionality From a `Collection<T>` Object?


LuisOsta:
Currently the only way to create a session is through the client object. But from my investigations it seems that the collection object has the client object stored in its struct.

Would it be possible to either expose session creation functionality or allow the consumer to get the client from the collection object?

The reason for this is that we separate the code based on collections so each domain only has access to the relevant pieces and I'd prefer to not have to pass the whole `client` object to all areas that need to create sessions or handle transactions.

LuisOsta:
The ideal usage I'd love would be something like:


```rust
collection.start_session(None);
```

Or

```rust
collection.get_client();
```

isabelatkinson:
Hey @LuisOsta, thanks for filing this issue! While we would prefer not to add the `start_session` method directly to the `Collection` object, we do think it would be a good idea to have a method on `Collection` that returns its parent client. I've filed RUST-1582 to cover this work. Feel free to make a pull request implementing this if you'd like to; otherwise, we'll pick it up when we have time to do so.

One small thing to note is that we probably want to name the method `client` (i.e. `collection.cllient()`) for consistency with the [corresponding method](https://docs.rs/mongodb/latest/mongodb/struct.ClientSession.html#method.client) on our `ClientSession`type.

LuisOsta:
Alright sounds good. I'll see if I can make the PR that makes the appropriate changes, thanks for the quick reply!

LuisOsta:
@kmahar I've opened the PR that allows external consumers to access the underlying client object - https://github.com/mongodb/mongo-rust-driver/pull/822

isabelatkinson:
Closing this out as we've just merged #822.

PR #822, opened 2023-02-02
RUST-1582 Expose collection.client() as public

LuisOsta:
This aims to address https://jira.mongodb.org/browse/RUST-1582 as originally filed as an issue in https://github.com/mongodb/mongo-rust-driver/issues/812.

There was already a member function in the collection object which allowed client access but it was only accessible from within the driver crate. So the only changes needed are to make it externally accessible.


isabelatkinson (approved):
LGTM - thanks for your contribution!

abr-egn (approved):
LGTM!

PR #823, opened 2023-02-03
minor: fix documentation of new features

isabelatkinson:
As far as I can tell there's not a great way to test what this is going to look like on docs.rs short of doing a release, but the FLE docs should show up in the same way that the sync docs currently do.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,11 +67,12 @@ zstd-compression = ["zstd"]
 zlib-compression = ["flate2"]
 snappy-compression = ["snap"]
 
+# Enables support for client-side field level encryption and queryable encryption.
 # The In Use Encryption API is unstable and may have backwards-incompatible changes in minor version updates.
 in-use-encryption-unstable = ["mongocrypt", "rayon", "num_cpus"]
 
-# DO NOT USE; see https://jira.mongodb.org/browse/RUST-580 for the status of tracing/logging support in the Rust driver.
 # Enables support for emitting tracing events.
+# The tracing API is unstable and may have backwards-incompatible changes in minor version updates.
 # TODO: pending https://github.com/tokio-rs/tracing/issues/2036 stop depending directly on log.
 tracing-unstable = ["tracing", "log"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,8 @@ exclude = [
     "tests/**",
 ]
 
+# NOTE: any new features added to this list should also be added to the features
+# list in the [package.metadata.docs.rs] section below.
 [features]
 default = ["tokio-runtime"]
 tokio-runtime = [
@@ -186,3 +188,16 @@ regex = "1.6.0"
 
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
+# async-std-runtime and sync are excluded here because they conflict with the default features.
+# Neither feature has any unique documentation associated with it, so we do not need to
+# include them in our documentation build.
+features = [
+    "tokio-sync",
+    "zstd-compression",
+    "snappy-compression",
+    "zlib-compression",
+    "openssl-tls",
+    "aws-auth",
+    "tracing-unstable",
+    "in-use-encryption-unstable"
+]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -343,7 +343,8 @@ mod test;
 #[cfg(feature = "tracing-unstable")]
 mod trace;
 
-#[cfg(feature = "in-use-encryption-unstable")]
+#[cfg(any(feature = "in-use-encryption-unstable", docsrs))]
+#[cfg_attr(docsrs, doc(cfg(any(feature = "sync", feature = "tokio-sync"))))]
 pub use crate::client::csfle::client_encryption;
 pub use crate::{
     client::{session::ClientSession, Client},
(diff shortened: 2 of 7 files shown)


src/lib.rs
```diff
@@ -49,21 +49,23 @@
...
+//! | `zstd-compression`           | Enable support for compressing messages with [`zstd`](http://facebook.github.io/zstd/).                                                                                 | no      |
+//! | `snappy-compression`         | Enable support for compressing messages with [`snappy`](http://google.github.io/snappy/).                                                                               | no      |
+//! | `openssl-tls`                | Switch TLS connection handling to use [`openssl`](https://docs.rs/openssl/0.10.38/).                                                                                    | no      |
+//! | `in-use-encryption-unstable` | Enable support for client-side field level encryption and queryable encryption. This API is unstable and may be subject to breaking changes in minor releases.          | no      |
+//! | `tracing-unstable`           | Enable support for emitting [`tracing`](https://docs.rs/tracing/latest/tracing/) events. This API is unstable and may be subject to breaking changes in minor releases. | no      |
```
isabelatkinson, line 68: These are the two new additions to the table; I realized they were missing. I also removed the "Extra dependencies" column as a lot of the information was outdated/incorrect. If we think this information is valuable then I can go through and fix it, but I'm not sure how useful it is given that users can just look at our `Cargo.toml`. There's also a risk that it will become out-of-date again in the future if we add/change dependencies and forget to update here. Thoughts?

src/lib.rs
```diff
@@ -343,7 +345,8 @@ mod test;
 #[cfg(feature = "tracing-unstable")]
 mod trace;
```
isabelatkinson, line 346: I don't think there's anything here that needs to be exported publicly/documented but let me know if that's wrong, @kmahar.

src/lib.rs
abr-egn: Agreed this doesn't need to be in the table.

src/lib.rs
```diff
@@ -343,7 +345,8 @@ mod test;
...
 
-#[cfg(feature = "in-use-encryption-unstable")]
+#[cfg(any(feature = "in-use-encryption-unstable", docsrs))]
+#[cfg_attr(docsrs, doc(cfg(any(feature = "sync", feature = "tokio-sync"))))]
```
abr-egn, line 349: I think the second line should be
```
#[cfg_attr(docsrs, doc(cfg(feature = "in-use-encryption-unstable")))]
```
(and below).

Cargo.toml
```diff
@@ -67,11 +67,12 @@ zstd-compression = ["zstd"]
 zlib-compression = ["flate2"]
 snappy-compression = ["snap"]
 
+# Enables support for client-side field level encryption and queryable encryption.
 # The In Use Encryption API is unstable and may have backwards-incompatible changes in minor version updates.
 in-use-encryption-unstable = ["mongocrypt", "rayon", "num_cpus"]
 
-# DO NOT USE; see https://jira.mongodb.org/browse/RUST-580 for the status of tracing/logging support in the Rust driver.
```
kmahar, line 73: thanks for removing this! I forgot :)

src/lib.rs
kmahar: SGTM

src/lib.rs
kmahar: the only thing explicitly in the API around tracing is the `tracing_max_document_length_bytes` option in `ClientOptions`. that already has the docsrs annotations and appears to show up correctly in the docs: https://docs.rs/mongodb/2.4.0-beta.1/mongodb/options/struct.ClientOptions.html#structfield.tracing_max_document_length_bytes

kmahar (approved):
LGTM mod Abraham's comment.

src/lib.rs
isabelatkinson: whoops, bad copy/paste - fixed

abr-egn (approved):
LGTM, thanks!

isabelatkinson:
The new changes update our docs.rs build approach to use feature flags rather than the `#[cfg(docsrs)]` attribute. Currently, anything behind a feature flag that we want to document publicly has annotations to the effect of:
```
#[cfg(any(feature = "feature", docsrs))]
#[cfg_attr(docsrs, doc(cfg(feature = "feature")))]
```
The first line includes the code in our docs build and the second line adds a note in the docs that the item is gated behind a feature flag. We also have the following in our Cargo.toml:
```
```

This approach was necessary in part because the `doc(cfg(...))` annotation is only available on nightly, so we needed to gate its usage behind `docsrs`. However, a new crate-wide `doc_auto_cfg` annotation was recently added which will automatically add feature-flag notes in the docs. Because this `cfg_attr` is no longer needed, we don't really need to use `docsrs` throughout the code anymore. We do still need it to gate the usage of `doc_auto_cfg` in `lib.rs`, as that is also only available on nightly.

The new approach is to build our documentation with all of the non-default-conflicting features (i.e. everything but `async-std-runtime` and `sync`) specified rather than adding `docsrs` inline to anything public behind a feature flag. We can get away with this because we don't have any public API associated with `async-std-runtime` and all of the API associated with `sync` also builds with the `tokio-sync` feature. (`doc_auto_cfg` will still make a note that our sync API is available with the `sync` feature flag.) I find this new approach to be more maintainable because we no longer need to remember to add a `docsrs` annotation when we introduce new feature-flag-gated API. However, if we do add any API gated behind those conflicting features, we can fall back to the `docsrs` annotation in those cases.

Cargo.toml
```diff
@@ -185,3 +188,16 @@ regex = "1.6.0"
...
+features = [
+    "tokio-sync",
+    "zstd-compression",
+    "snappy-compression",
+    "zlib-compression",
+    "openssl-tls",
+    "aws-auth",
+    "tracing-unstable",
+    "in-use-encryption-unstable"
+]
```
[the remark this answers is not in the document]
abr-egn: Yup, agreed, definitely prefer this way.

abr-egn (approved):
This is much better than how it was set up before - thank you!

PR #824, opened 2023-02-03
RUST-1219, RUST-863: Assert that a ping succeeds in DNS seedlist tests; sync DB name with comma tests

kmahar:
I POCed RUST-1219 when the DRIVERS ticket was done, and reviewed the PR + confirmed the tests worked in Rust for RUST-863, so figured I'd get those changes merged in.

### DIFF
--- a/src/test/spec/initial_dns_seedlist_discovery.rs
+++ b/src/test/spec/initial_dns_seedlist_discovery.rs
@@ -5,7 +5,7 @@ use tokio::sync::RwLockReadGuard;
 
 use crate::{
     bson::doc,
-    client::{auth::AuthMechanism, Client},
+    client::Client,
     options::{ClientOptions, ResolverConfig},
     runtime,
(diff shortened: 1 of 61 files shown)


src/test/spec/initial_dns_seedlist_discovery.rs
```diff
@@ -136,37 +137,19 @@ async fn run_test(mut test_file: TestFile) {
             "skipping initial_dns_seedlist_discovery test case due to TLS requirement mismatch",
         )
     } else {
-        // If the connection URI provides authentication information, manually create the user
```
kmahar, line 139: previously, we unconditionally ran a `ping`, which the spec didn't strictly require, and doing so required creating the user in those cases.
the spec does now require a ping, but only for tests with `ping`: true. the ones with `ping` set to true are ones without user information provided, so this code is no longer needed.

isabelatkinson (approved):

abr-egn (approved):
LGTM!

PR #826, opened 2023-02-06
RUST-1100 Fix find_and_getmore_share_session flakiness

abr-egn:
RUST-1100

This uses a blunt hammer fix: the failure was the cursor immediately ending, so presumably there was still a race between replication and the test logic.  Since the point of the test isn't replication behavior but sessions, this just adds a loop until the cursor has data.

### DIFF
--- a/.evergreen/env.sh
+++ b/.evergreen/env.sh
@@ -19,5 +19,7 @@ if [[ "$OS" == "Windows_NT" ]]; then
     export PATH
     echo "updated path on windows PATH=$PATH"
 else
+    # Turn off tracing for the very-spammy nvm script.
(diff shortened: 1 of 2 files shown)


.evergreen/env.sh
```diff
@@ -19,5 +19,7 @@ if [[ "$OS" == "Windows_NT" ]]; then
     export PATH
     echo "updated path on windows PATH=$PATH"
 else
+    # Turn off tracing for the very-spammy nvm script.
```
abr-egn, line 22: This is unrelated, but the spam was making finding the actual test failures annoying.

isabelatkinson (approved):

PR #827, opened 2023-02-07
RUST-1273 Increase timeout for retry pool cleared tests

isabelatkinson:
This PR bumps the timeout used to wait for a `ConnectionCheckoutFailed` event. I also added some error messages to make debugging this test easier in the future if we run into more failures.


abr-egn (approved):
LGTM!

PR #831, opened 2023-02-15
RUST-1314 Support on-demand AWS credentials for in-use encryption

abr-egn:
RUST-1314

Happily, this turned out to be more straightforward than expected.


src/test/csfle.rs
```diff
@@ -2751,7 +2751,66 @@ impl CommandEventHandler for DecryptionEventsHandler {
...
+    let result = ce
+        .create_data_key(MasterKey::Aws {
+            region: "us-east-1".to_string(),
+            key: "arn:aws:kms:us-east-1:579766882180:key/89fcc2c4-08b0-4bd9-9f25-e30687b580d0"
+                .to_string(),
+            endpoint: None,
+        })
+        .run()
+        .await;
+    assert!(result.is_err());
```
isabelatkinson, line 2783: is there a more specific error case/variant we can assert on here?

abr-egn: Good call, updated this to check that it's an auth error.

isabelatkinson (approved):

PR #841, opened 2023-03-09
Refactor spec test parser and runners

isabelatkinson:
This is my low-stakes skunkworks project; review at your leisure!

This PR refactors the way we parse spec tests and our spec test runners, using the action pattern to improve the latter. Because we're not bound to our MSRV in our test code I also got to try out some new features, including `IntoFuture`.


abr-egn (approved):
This is _way_ more readable, thank you!  One extremely minor style nit, no need for re-review.

PR #844, opened 2023-03-16
RUST-1501 Collect FaaS platform metadata

abr-egn:
RUST-1501

This implements the new data collection from environment variables and the associated fallback logic to keep the handshake within the size limit.


isabelatkinson (commented):
Code changes LGTM. Is it expected that users will be setting these environment variable values? If so, should we document somewhere how they should do so?

PR #849, opened 2023-04-04
RUST-107 Convenient transactions

abr-egn:
RUST-107

This adds the convenient transactions API and associated tests.  Running the v2 spec tests required quite a bit of refactoring in order to allow recursively running operations within the `withTransaction` operation.

### DIFF
--- a/src/client/session/mod.rs
+++ b/src/client/session/mod.rs
@@ -563,7 +563,7 @@ impl ClientSession {
     }
 
     /// TODO
-    pub async fn with_transaction<F, R, C>(&mut self, mut callback: F, options: impl Into<Option<TransactionOptions>>, mut context: C) -> Result<R>
+    pub async fn with_transaction<R, C, F>(&mut self, mut context: C, mut callback: F, options: impl Into<Option<TransactionOptions>>) -> Result<R>
         where F: for<'a> FnMut(&'a mut ClientSession, &'a mut C) -> BoxFuture<'a, Result<R>>,
     {
         let options = options.into();
--- a/src/client/session/mod.rs
+++ b/src/client/session/mod.rs
@@ -567,13 +567,22 @@ impl ClientSession {
     }
 
     /// TODO
-    pub async fn with_transaction<R, C, F>(&mut self, mut context: C, mut callback: F, options: impl Into<Option<TransactionOptions>>) -> Result<R>
-        where F: for<'a> FnMut(&'a mut ClientSession, &'a mut C) -> BoxFuture<'a, Result<R>>,
+    pub async fn with_transaction<R, C, F>(
+        &mut self,
+        mut context: C,
+        mut callback: F,
+        options: impl Into<Option<TransactionOptions>>,
+    ) -> Result<R>
+    where
+        F: for<'a> FnMut(&'a mut ClientSession, &'a mut C) -> BoxFuture<'a, Result<R>>,
     {
         let options = options.into();
         let timeout = Duration::from_secs(120);
         #[cfg(test)]
-        let timeout = self.convenient_transaction_timeout.clone().unwrap_or(timeout);
+        let timeout = self
+            .convenient_transaction_timeout
+            .clone()
+            .unwrap_or(timeout);
         let start = Instant::now();
 
         use crate::error::{TRANSIENT_TRANSACTION_ERROR, UNKNOWN_TRANSACTION_COMMIT_RESULT};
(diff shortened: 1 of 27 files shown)


src/test/spec/v2_runner/mod.rs
```diff
@@ -232,220 +226,282 @@ async fn run_v2_test(path: std::path::PathBuf, test_file: TestFile) {
...
+        mut sessions: OpSessions<'b>,
+    ) -> Option<Result<Option<bson::Bson>, crate::error::Error>> {
+        if operation.name == "withTransaction" {
+            if !matches!(&operation.object, Some(OperationObject::Session0)) {
+                panic!("invalid object for withTransaction: {:?}", operation.object);
+            }
+            return Some(operation.execute_recursive(self, sessions).await);
+        }
+
+        let db = match &operation.database_options {
```
abr-egn, line 320: Despite what github's diff says, the rest of the logic in this method is unchanged from the original loop.

src/test/spec/v2_runner/mod.rs
```diff
@@ -232,220 +226,282 @@ async fn run_v2_test(path: std::path::PathBuf, test_file: TestFile) {
...
 
-            let session = match operation.session.as_deref() {
-                Some("session0") => session0.as_mut(),
-                Some("session1") => session1.as_mut(),
-                Some(other) => panic!("unknown session name: {}", other),
-                None => None,
+    async fn run_operation(
+        &mut self,
+        operation: &Operation,
+    ) -> Option<Result<Option<bson::Bson>, crate::error::Error>> {
```
abr-egn, line 265: The return type here is a little complex because along with succeeding or failing an operation can also be skipped.

src/test/spec/v2_runner/operation.rs
```diff
@@ -87,6 +97,74 @@ pub(crate) struct Operation {
     pub(crate) session: Option<String>,
 }
 
+impl Operation {
+    pub(crate) fn assert_result_matches(&self, result: &Result<Option<Bson>>, description: &str) {
```
abr-egn, line 101: This was also moved unchanged from its original place in `run_v2_test`.

src/test/spec/v2_runner/operation.rs
```diff
@@ -1493,6 +1572,59 @@ impl TestOperation for AssertIndexNotExists {
...
+        async move {
+            let session = sessions.session0.unwrap();
+            session
+                .with_transaction(
+                    (runner, &self.callback.operations, sessions.session1),
+                    |session, (runner, operations, session1)| {
+                        async move {
+                            for op in operations.iter() {
+                                let sessions = OpSessions {
+                                    session0: Some(session),
```
abr-egn, line 1601: This is literally the same pointer as `session0` passed to `execute_recursive`, but to satisfy the borrow checker we need to re-bind it with the lifetime provided to the callback.

src/client/session/mod.rs
```diff
@@ -561,6 +566,107 @@ impl ClientSession {
         }
     }
 
+    /// Runs a callback inside a transaction.  Transient transaction errors will cause the callback
```
isabelatkinson, line 569: Can we add a bit more detail here explaining that the transaction will be started and committed/aborted within this method?

src/client/session/mod.rs
```diff
@@ -561,6 +566,107 @@ impl ClientSession {
...
+    ///     |session, _| async move {
+    ///         coll.insert_one_with_session(doc! { "data": my_data }, None, session).await
+    ///     }.boxed(),
+    ///     None,
+    /// ).await?;
+    /// */
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn with_transaction<R, C, F>(
```
isabelatkinson, line 608: we'll also need to add this method to the sync API

src/client/session/mod.rs
```diff
@@ -561,6 +566,107 @@ impl ClientSession {
...
+        &mut self,
+        mut context: C,
+        mut callback: F,
+        options: impl Into<Option<TransactionOptions>>,
+    ) -> Result<R>
+    where
+        F: for<'a> FnMut(&'a mut ClientSession, &'a mut C) -> BoxFuture<'a, Result<R>>,
```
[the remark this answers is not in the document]
abr-egn: Hmm, that's unfortunate.  The type erasure is necessary - there's no way to express `for<'a> FnMut(&'a mut ClientSession, &'a mut C) -> impl Future<'a, Result<R>>` currently without it.  Since `BoxFuture` is just a type alias, I've duplicated the definition here instead of using the one from the `futures_core` crate.

src/client/session/mod.rs
abr-egn: Good catch.  Unfortunately, I had to duplicate the body of the function rather than just wrapping the async one to avoid nested `block_on` calls from within the callback.

src/client/session/mod.rs
abr-egn: Updated.

abr-egn:
Note that the evergreen failures are unrelated - testing against AWS auth is broken right now due to https://jira.mongodb.org/browse/BF-28319.

PR #853, opened 2023-04-13
RUST-1411 Add `create_encrypted_collection` helper

abr-egn:


src/error.rs
```diff
@@ -259,7 +259,7 @@ impl Error {
 
     /// Gets the code from this error for performing SDAM updates, if applicable.
     /// Any codes contained in WriteErrors are ignored.
-    pub(crate) fn code(&self) -> Option<i32> {
+    pub(crate) fn sdam_code(&self) -> Option<i32> {
```
abr-egn, line 262: Not for the first time, I got a bit confused when implementing the prose tests because this doesn't return a code for write errors, so I renamed this to be more clear it's got sdam-specific logic and added an unrestricted version.

isabelatkinson (approved):

PR #858, opened 2023-04-18
RUST-1411 Move `create_encrypted_collection` helper to `ClientEncryption`

abr-egn:
RUST-1411

While writing up the release notes for 2.5.0, I realize that `create_encrypted_collection` is in the spec as a method on `ClientEncryption`, not (as I'd implemented it) `Database`.  While it makes some sense to be on the latter as a parallel to `create_collection`, since the spec and all other language drivers implement it the other way, I updated us to follow suit.

### DIFF
--- a/src/client/csfle/client_encryption.rs
+++ b/src/client/csfle/client_encryption.rs
@@ -719,7 +719,7 @@ impl<'a> CreateEncryptedCollectionAction<'a> {
         )
     }
 
-    /// Set the encrypted fields for the colllection.
+    /// Set the encrypted fields for the collection.
     pub fn encrypted_fields(mut self, encrypted_fields: Document) -> Self {
         self.options.encrypted_fields = Some(encrypted_fields);
         self
--- a/src/client/csfle/client_encryption.rs
+++ b/src/client/csfle/client_encryption.rs
@@ -720,8 +720,8 @@ impl<'a> CreateEncryptedCollectionAction<'a> {
     }
 
     /// Set the encrypted fields for the colllection.
-    pub fn encrypted_fields(mut self, encrypted_fields: impl Into<Option<Document>>) -> Self {
-        self.options.encrypted_fields = encrypted_fields.into();
+    pub fn encrypted_fields(mut self, encrypted_fields: Document) -> Self {
+        self.options.encrypted_fields = Some(encrypted_fields);
         self
     }
 
--- a/src/client/csfle/client_encryption.rs
+++ b/src/client/csfle/client_encryption.rs
@@ -421,20 +421,47 @@ impl ClientEncryption {
     ///
     /// Does not affect any auto encryption settings on existing MongoClients that are already
     /// configured with auto encryption.
-    #[must_use]
-    pub fn create_encrypted_collection(
+    pub async fn create_encrypted_collection(
         &self,
         db: &Database,
-        name: impl Into<String>,
+        name: impl AsRef<str>,
         master_key: MasterKey,
-    ) -> CreateEncryptedCollectionAction {
-        CreateEncryptedCollectionAction {
-            client_enc: self,
-            db: db.clone(),
-            name: name.into(),
-            options: CreateCollectionOptions::default(),
-            master_key,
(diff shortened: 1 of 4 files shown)


src/client/csfle/client_encryption.rs
```diff
@@ -639,3 +665,70 @@ impl From<String> for EncryptKey {
...
+        let mut opts_prime = self.options.clone();
+        opts_prime.encrypted_fields = Some(ef_prime.clone());
+        (
+            ef_prime,
+            self.db.create_collection(&self.name, opts_prime).await,
+        )
+    }
+
+    /// Set the encrypted fields for the colllection.
+    pub fn encrypted_fields(mut self, encrypted_fields: impl Into<Option<Document>>) -> Self {
```
abr-egn, line 723: Since it's an error (caught elsewhere) to call this method without `encrypted_fields`, it seemed worth giving this option its own action helper.

src/sync/db.rs
```diff
@@ -219,28 +219,6 @@ impl Database {
         ))
     }
 
-    /// Creates a new collection with encrypted fields, automatically creating new data encryption
```
abr-egn, line 222: There's now no sync api for this, but that's consistent with the rest of `ClientEncryption` - we made the decision a while back to defer any work on that until/unless there are user requests for it.

src/client/csfle/client_encryption.rs
isabelatkinson: Can this just accept a `Document` since it's an error for `None` to be passed?

src/client/csfle/client_encryption.rs
```diff
@@ -639,3 +665,70 @@ impl From<String> for EncryptKey {
...
+        self.options.encrypted_fields = encrypted_fields.into();
+        self
+    }
+
+    /// Set the complete creation options for the collection.  This will overwrite prior calls to
+    /// `encrypted_fields`.
+    pub fn create_collection_options(mut self, options: CreateCollectionOptions) -> Self {
```
isabelatkinson, line 730: Should we take the opportunity here to add action methods for each option rather than accept the options struct? I think that's the direction we're going to want to head in for 3.0. anyway. That way we won't have to deal with multiple encrypted fields maps being passed in.

src/client/csfle/client_encryption.rs
abr-egn: Good call, done.

src/client/csfle/client_encryption.rs
abr-egn: I'd prefer not to - I agree for 3.0 something like that is what we'll want, but for now there are a lot of fields in `CreateCollectionOptions` that aren't directly relevant.

isabelatkinson: If that's the case I'd be inclined to add options methods just for the fields that are relevant or to revert to accepting the options struct as an argument to the method. I think users might be confused as to which way of providing the encrypted fields is correct (i.e. through the options struct or via the separate action method).

I'll defer to you on this though; I think we're in a weird in-between situation when it comes to actions vs. options structs in our API right now, so whatever we end up doing can be ironed out when we get to 3.0.

src/client/csfle/client_encryption.rs
```diff
@@ -639,3 +665,70 @@ impl From<String> for EncryptKey {
...
+        }
+        let mut opts_prime = self.options.clone();
+        opts_prime.encrypted_fields = Some(ef_prime.clone());
+        (
+            ef_prime,
+            self.db.create_collection(&self.name, opts_prime).await,
+        )
+    }
+
+    /// Set the encrypted fields for the colllection.
```
isabelatkinson, line 722:
```suggestion
    /// Set the encrypted fields for the collection.
```

isabelatkinson (approved):
lgtm - small typo fix and I'll leave it up to you to decide on API!

src/client/csfle/client_encryption.rs
abr-egn: Yeah, makes sense.  I went back to just accepting options as a parameter.

PR #863, opened 2023-04-21
RUST-1605 Update to use libmongocrypt fle2v2

abr-egn:
RUST-1605 / RUST-1581

The code changes here are pretty minor, it's largely a test sync.  Because the tests had also been updated for RUST-1581 I included that in this PR.


.evergreen/config.yml
```diff
@@ -1983,6 +1983,7 @@ buildvariants:
     os:
       - ubuntu-20.04
     mongodb-version:
+      - "latest"
```
abr-egn, line 1986: The updated tests can only run on 7.0+, which the csfle suite wasn't running on before.  `latest` as of yesterday was `"7.1.0-alpha-220-g77c517c"` so that covers it, and it seems good to be running csfle tests against `latest` anyway.

src/client/mod.rs
```diff
@@ -565,6 +565,28 @@ impl Client {
         &self.inner.topology
     }
 
+    pub(crate) async fn primary_description(&self) -> Option<crate::sdam::ServerDescription> {
```
abr-egn, line 568: The spec includes a clause that creating collections with encrypted fields must be denied if the primary is <7.0.  The reference implementation in the C driver actually does a full connection checkout to test that, but this lets us avoid that.

Unfortunately, simply peeking the latest topology description doesn't work because it's possible to call `create_collection` while the topology is still being discovered, so this waits for data to be populated.

isabelatkinson (commented):
just a small comment!

isabelatkinson (approved):

PR #869, opened 2023-05-01
RUST-1629 Sync testdata updates for csfle

abr-egn:
RUST-1629

This required adding some extra logic to our test syncing script and reorganizing the testadata files to match.


abr-egn:
Just a note, the test-csfle-serverless failures are fixed in #870.

isabelatkinson (approved):

PR #870, opened 2023-05-01
RUST-1621 Remove uses of the soon-to-be-unsupported ubuntu-18.04

abr-egn:
RUST-1621

Some of them were to run old versions, which I switched to use rhel-8.0; others had just defaulted to that version, which I updated to ubuntu-20.04.


isabelatkinson (approved):
changes lgtm!

> Unfortunately, the AWS auth tests seem to depend on compiling the binary specifically on ubuntu-18.04, so those are remaining as is for now.

can we file a ticket to update this in the future?

abr-egn:
> can we file a ticket to update this in the future

I'm going to keep the existing ticket open so the upstream DRIVERS ticket tracks the status properly; I'll update it once I merge this PR.

PR #872, opened 2023-05-09
RUST-1442 On-demand Azure KMS credentials

abr-egn:
RUST-1442

This adds support for fetching on-demand csfle credentials from the Azure IMDS service.  Notably, this requires an HTTP fetch, so this updates the internal `HttpClient` to not be aws auth only; this also means that azure auth has the same restriction as aws auth (i.e. tokio), and is behind a feature flag.

This PR includes the unit tests; integration tests to come in a follow-up PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,9 @@ bson-uuid-1 = ["bson/uuid-1"]
 # This can only be used with the tokio-runtime feature flag.
 aws-auth = ["reqwest"]
 
+# Enable support for on-demand Azure KMS credentials.
+azure-kms = ["reqwest"]
+
 zstd-compression = ["zstd"]
 zlib-compression = ["flate2"]
 snappy-compression = ["snap"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,7 @@ bson-uuid-1 = ["bson/uuid-1"]
 aws-auth = ["reqwest"]
 
 # Enable support for on-demand Azure KMS credentials.
+# This can only be used with the tokio-runtime feature flag.
 azure-kms = ["reqwest"]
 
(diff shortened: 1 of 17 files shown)


src/client/csfle/state_machine.rs
```diff
@@ -346,3 +364,134 @@ fn raw_to_doc(raw: &RawDocument) -> Result<Document> {
...
+                .get_and_deserialize_json(self.make_url()?, &self.make_headers())
+                .await
+                .map_err(|e| Error::authentication_error("azure imds", &format!("{}", e)))?;
+            let expires_in_secs: u64 = server_response.expires_in.parse().map_err(|e| {
+                Error::authentication_error(
+                    "azure imds",
+                    &format!("invalid `expires_in` response field: {}", e),
+                )
+            })?;
+            #[allow(clippy::redundant_clone)]
```
isabelatkinson, line 431: why is a redundant clone needed here?

Cargo.toml
```diff
@@ -65,6 +65,9 @@ bson-uuid-1 = ["bson/uuid-1"]
 # This can only be used with the tokio-runtime feature flag.
 aws-auth = ["reqwest"]
 
+# Enable support for on-demand Azure KMS credentials.
```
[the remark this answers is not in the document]
abr-egn: Good call, done.  And yeah, agreed that it's not a great situation w.r.t. features and dependencies.

src/client/csfle/state_machine.rs
abr-egn: The `#[cfg(test)]` line below needs `server_response`, but without the `clone` this line is a partial move.

isabelatkinson (approved):

PR #874, opened 2023-05-10
Add partial file retrieval when using GridFS stream download

laleksiunas:
Implemented support for partial file retrieval according to GridFS [spec](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#partial-file-retrieval) for `GridFsBucket::open_download_stream` and `GridFsBucket::open_download_stream_by_name`.
**Unfortunately, there is a breaking change in `GridFsBucket::open_download_stream` interface - additional options argument to accept start/end range.**

- [x] start and end values are non negative (ensured by design using u64)
- [x] validate whether start is less than or equal to end
- [x] validate whether start and end are less than or equal to file length
- [x] file chunk size is being used to skip and limit chunks when performing partial reads

### DIFF
--- a/src/cursor/mod.rs
+++ b/src/cursor/mod.rs
@@ -148,6 +148,7 @@ impl<T> Cursor<T> {
     }
 
     /// Whether this cursor has any additional items to return.
+    #[allow(dead_code)]
     pub(crate) fn has_next(&self) -> bool {
         !self.is_exhausted()
             || !self
--- a/src/error.rs
+++ b/src/error.rs
@@ -905,6 +905,15 @@ pub enum GridFsErrorKind {
     /// [`GridFsUploadStream`](crate::gridfs::GridFsUploadStream) while a write was still in
     /// progress.
     WriteInProgress,
+
+    /// Partial file download range is invalid when start is greater then end
+    InvalidPartialDownloadRange { start: u64, end: u64 },
+
+    /// Partial file download range is invalid when start or end are greater then file length
+    PartialDownloadRangeOutOfBounds {
+        out_of_bounds_value: u64,
+        file_length: u64,
+    },
 }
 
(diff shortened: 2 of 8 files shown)


isabelatkinson:
Hi @laleksiunas, thanks for creating this pull request! The team is currently focused on other priorities, but we will take a look at this when we have more bandwidth in the future. One piece of immediate feedback I have is that making a breaking change here means that we would not be able to merge this work until we do a major release, which we don't currently have a time frame for. One way to avoid this breaking change would be to add a `GridFsBucket::open_download_stream_with_range` method rather than change the signature for the existing `open_download_stream` method.

laleksiunas:
@isabelatkinson Thanks for review, I will add `GridFsBucket::open_download_stream_with_options` in the upcoming days

nathanfranke:
I rebased this to `upstream/main`.
Tests aren't passing for me even on `upstream/v3.5.1` or `upstream/main`; perhaps my fault.
Some code was moved between files and I made some members public, notably `actions::gridfs::download::DownloadRange` and those modules.

Otherwise, the async `by_name` API seems to work well; I did not test with IDs or `sync`.

---

> One way to avoid this breaking change would be to add a `GridFsBucket::open_download_stream_with_range` method rather than change the signature for the existing `open_download_stream` method.

This should be resolved now. `open_download_stream` and `open_download_stream_by_name` use the builder pattern (e.g. `.open_download_stream_by_name("test").start(0).end(100)`)

---

Test project:
[rust-mongodb.zip](https://github.com/user-attachments/files/17571978/rust-mongodb.zip)

Note: The `mongodb` dependency path is hardcoded in this project.

```
rust mongodb: 100 x 1MiB
    WRITE: 534ms
    READ FULL: 320ms
    READ PARTIAL: 3ms
```

PR #877, opened 2023-05-16
RUST-1417 Add support for GCP attached service accounts when using GCP KMS

isabelatkinson:
This is going to be a slightly messy rebase, so I'm going to hold off on doing that until this PR is approved and I can squash my git history.

### DIFF
--- a/src/client/csfle/state_machine.rs
+++ b/src/client/csfle/state_machine.rs
@@ -214,7 +214,7 @@ impl CryptExecutor {
                 State::NeedKmsCredentials => {
                     let ctx = result_mut(&mut ctx)?;
                     #[allow(unused_mut)]
-                    let mut out = rawdoc! {};
+                    let mut kms_providers = rawdoc! {};
                     let credentials = self.kms_providers.credentials();
                     if credentials
                         .get(&KmsProvider::Aws)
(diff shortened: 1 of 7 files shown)


isabelatkinson:
The evergreen failures should be fixed after rebasing.

src/client/csfle/state_machine.rs
```diff
@@ -240,7 +239,72 @@ impl CryptExecutor {
...
+                                    code: None,
+                                };
+                                error.into()
+                            }
+
+                            let http_client = HttpClient::default();
+                            let host = std::env::var("GCE_METADATA_HOST")
+                                .unwrap_or_else(|_| "metadata.google.internal".into());
+                            let uri = format!(
+                            "http://{}/computeMetadata/v1/instance/service-accounts/default/token",
```
abr-egn, line 276: nit: looks like these lines are missing an indent

src/runtime/http.rs
```diff
@@ -1,17 +1,17 @@
-#[cfg(feature = "aws-auth")]
+#[cfg(any(feature = "aws-auth", feature = "in-use-encryption-unstable"))]
```
abr-egn, line 1: I assume you're planning to switch this over to the setup I used for Azure in the merge?

abr-egn (approved):
LGTM!

src/runtime/http.rs
isabelatkinson: oh yes I left a review comment for this but forgot to post it :) yep, I was holding off on updating this since it'll need refactoring during the rebase anyway

PR #878, opened 2023-05-18
DRIVERS-2617 Test splitting of large changestream events

abr-egn:
DRIVERS-2617

This adds a PoC for a simple prose test to validate that the driver handles the new `$changeStreamSplitLargeEvent` stage.

### DIFF
--- a/src/test/change_stream.rs
+++ b/src/test/change_stream.rs
@@ -1,11 +1,11 @@
 use bson::{doc, Bson, Document};
-use futures_util::StreamExt;
+use futures_util::{StreamExt, TryStreamExt};
 use semver::VersionReq;
 
 use crate::{
     change_stream::{
         event::{ChangeStreamEvent, OperationType},
-        options::ChangeStreamOptions,
+        options::{ChangeStreamOptions, FullDocumentBeforeChangeType},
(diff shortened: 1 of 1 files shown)


src/test/change_stream.rs
```diff
@@ -709,3 +710,68 @@ async fn create_coll_pre_post() -> Result<()> {
...
+        .inserted_id;
+    let stream = coll
+        .watch(
+            [doc! { "$changeStreamSplitLargeEvent": {} }],
+            ChangeStreamOptions::builder()
+                .full_document_before_change(Some(FullDocumentBeforeChangeType::Required))
+                .build(),
+        )
+        .await?
+        .with_type::<Document>();
```
abr-egn, line 757: Unfortunately, the stream needs to produce `Document`s rather than `ChangeStreamEvent`s because the fields are arbitrarily split up across the parts, so there are no guarantees that required fields are present in any particular fragment.  This also means that there's no need to add the `splitEvent: { fragment: 1, of: 2 }` structure to `ChangeStreamEvent` (and it seems like doing so would be misleading, since you couldn't actually use that type for these events).

isabelatkinson (approved):

PR #880, opened 2023-05-19
RUST-1649 Deprecate sharded-replicaset test topology / sync outdated spec tests

isabelatkinson:
In addition to making the changes for the linked ticket, this PR pulls in a lot of other tests for specs we haven't synced in a while. I fixed up the ones that seemed like they would be easy and otherwise filed tickets/added TODOs.


src/test/serde_helpers.rs
```diff
@@ -0,0 +1,18 @@
+use serde::de::{Deserialize, DeserializeOwned, Deserializer};
```
isabelatkinson, line 1: I didn't port all of our helpers over here but I thought it would be nice to have a central place to put methods like this to minimize clutter in test files.

src/test/spec/unified_runner/mod.rs
```diff
@@ -101,30 +101,74 @@ impl IntoFuture for RunUnifiedTestsAction {
...
+        "collectionData-createOptions.json",
+        "expectedError-errorResponse.json",
+        "entity-cursor-iterateOnce.json",
+        "matches-lte-operator.json",
+        "poc-transactions-convenient-api.json",
+        "kmsProviders-placeholder_kms_credentials.json",
+        "kmsProviders-unconfigured_kms.json",
+        "kmsProviders-explicit_kms_credentials.json",
+        "kmsProviders-mixed_kms_credential_fields.json",
+    ];
```
isabelatkinson, line 132: couldn't come up with a more graceful way to do this without violating lifetime rules

src/test/spec/unified_runner/mod.rs
```diff
@@ -101,30 +101,71 @@ impl IntoFuture for RunUnifiedTestsAction {
...
+        "poc-transactions-convenient-api.json",
+    ];
+    #[cfg(not(feature = "in-use-encryption-unstable"))]
+    let skipped_files = &[
+        "collectionData-createOptions.json",
+        "expectedError-errorResponse.json",
+        "entity-cursor-iterateOnce.json",
+        "matches-lte-operator.json",
+        "poc-transactions-convenient-api.json",
+        "kmsProviders-placeholder_kms_credentials.json",
```
abr-egn, line 128: Is there a ticket for these?

abr-egn (approved):
LGTM, thank you for tracking down all the updates needed for this!

src/test/spec/unified_runner/mod.rs
isabelatkinson: the test runner code needed to deserialize these is gated behind in-use-encryption-unstable, so they're skipped when that flag isn't on. i'll add a comment explaining as much

PR #882, opened 2023-05-25
minor: improve unified test skipping functionality

isabelatkinson:
I originally used `&[&str]`s to represent skipped files/tests, but I've since realized that this type makes adding skips based on features and client information difficult. This PR updates the lists to be vecs instead.

### DIFF
--- a/src/test/spec/client_side_encryption.rs
+++ b/src/test/spec/client_side_encryption.rs
@@ -32,6 +32,6 @@ async fn run_legacy() {
     }
 
     run_v2_tests(&["client-side-encryption", "legacy"])
-        .skip_files(skipped_files)
+        .skip_files(&skipped_files)
         .await;
 }
--- a/src/test/spec/client_side_encryption.rs
+++ b/src/test/spec/client_side_encryption.rs
@@ -10,10 +10,11 @@ use crate::test::{
 async fn run_unified() {
     let _guard: RwLockWriteGuard<()> = LOCK.run_exclusively().await;
 
-    #[cfg(not(feature = "openssl-tls"))]
-    let skipped_tests = &["create datakey with KMIP KMS provider"];
-    #[cfg(feature = "openssl-tls")]
-    let skipped_tests = &[];
+    let skipped_tests = if cfg!(not(feature = "openssl-tls")) {
+        Some(vec!["create datakey with KMIP KMS provider"])
+    } else {
+        None
+    };
 
     run_unified_tests(&["client-side-encryption", "unified"])
         .skip_tests(skipped_tests)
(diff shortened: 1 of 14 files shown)


src/test/spec/client_side_encryption.rs
```diff
@@ -25,11 +26,11 @@ async fn run_unified() {
...
     let _guard: RwLockWriteGuard<()> = LOCK.run_exclusively().await;
 
-    // TODO RUST-528: Unskip timeoutMS.json when CSOT is implemented.
-    #[cfg(not(feature = "openssl-tls"))]
-    let skipped_files = &["timeoutMS.json", "kmipKMS.json"];
-    #[cfg(feature = "openssl-tls")]
-    let skipped_files = &["timeoutMS.json"];
+    // TODO RUST-528: unskip this file
+    let mut skipped_files = vec!["timeoutMS.json"];
+    if cfg!(not(feature = "openssl-tls")) {
```
isabelatkinson, line 31: using the runtime `cfg!` macro rather than conditional compilation allows us to avoid unused mut warnings

src/test/spec/unified_runner/mod.rs
```diff
@@ -43,25 +43,25 @@ pub(crate) fn run_unified_tests(spec: &'static [&'static str]) -> RunUnifiedTest
...
+    skipped_files: Option<Vec<&'static str>>,
+    skipped_tests: Option<Vec<&'static str>>,
     file_transformation: Option<FileTransformation>,
 }
 
 impl RunUnifiedTestsAction {
     /// The files to skip deserializing. The provided filenames should only contain the filename and
     /// extension, e.g. "unacknowledged-writes.json". Filenames are matched case-sensitively.
-    pub(crate) fn skip_files(self, skipped_files: &'static [&'static str]) -> Self {
+    pub(crate) fn skip_files(self, skipped_files: impl Into<Option<Vec<&'static str>>>) -> Self {
```
abr-egn, line 54: I think this could be slightly improved (and avoid the noise of `vec!` where it's not needed) by having the signature be
`fn skip_files(self, skipped_files: &[&'static str]) -> Self`
and then storing it via
`skipped_files: Some(skipped_files.to_vec())`.

abr-egn (commented):
Looks good!  One small suggestion.

src/test/spec/unified_runner/mod.rs
isabelatkinson: The benefit of using vecs comes at the time of construction: when we need to skip conditionally based on a feature or server version/topology, a vec lets us create a list and then append to it rather than create two distinct lists that might have duplication (e.g. see the diff for `valid_pass` in this file).

abr-egn: Right - the signature I propose doesn't have the `'static` bound that the original `skip_files` had on the slice, so those cases can just pass `&local_vec` (or `local_vec.as_slice()` if type inference is confused).

isabelatkinson: Ah I see what you mean - good idea. updated

abr-egn (approved):

PR #886, opened 2023-06-06
RUST-1659 Implement basic OIDC SASL support

abr-egn:
RUST-1659

This implements the basic OIDC authentication protocol and exercises it via the first prose test.


src/client/mod.rs
```diff
@@ -108,6 +108,16 @@ pub struct Client {
     inner: Arc<ClientInner>,
 }
 
+#[allow(dead_code, unreachable_code, clippy::diverging_sub_expression)]
```
abr-egn, line 111: Accidentally making `Client` non-`Send` or -`Sync` causes confusing errors in random other parts of the code where those constraints are implicitly used; this little hack (which appears to be the standard way to assert auto traits) means you get a more clear error.

src/client/auth/mod.rs
```diff
@@ -191,22 +228,23 @@ impl AuthMechanism {
...
         match self {
             AuthMechanism::ScramSha1 | AuthMechanism::ScramSha256 | AuthMechanism::MongoDbCr => {
                 uri_db.unwrap_or("admin")
             }
             AuthMechanism::MongoDbX509 => "$external",
             AuthMechanism::Plain => uri_db.unwrap_or("$external"),
+            AuthMechanism::MongoDbOidc => "$external",
             #[cfg(feature = "aws-auth")]
             AuthMechanism::MongoDbAws => "$external",
-            _ => "",
```
abr-egn, line 209: The catchall here caused me a bit of lost time debugging an unhelpful server `invalid database name ""` error; since the only thing it caught was `Gssapi`, updated to that to avoid repeating the problem if we add more auth methods.

src/client/options/mod.rs
```diff
@@ -547,6 +548,12 @@ pub struct ClientOptions {
     #[builder(default)]
     pub write_concern: Option<WriteConcern>,
 
+    /// The token callbacks for OIDC authentication.
+    #[builder(default)]
+    #[serde(skip)]
+    #[derivative(Debug = "ignore", PartialEq = "ignore")]
+    pub oidc_callbacks: Option<oidc::Callbacks>,
```
isabelatkinson, line 555: Unless there are complications I'm not thinking of, I think it would be a little more consistent with other authentication methods to store these on the credential rather than the client options directly. WDYT?

As an aside, it would be nice to refactor our credential to be an enum with each mechanism as a variant that stores the information that's needed so that we can make more static guarantees about the absence/presence of fields, but that is a 3.0 concern 🙂

src/client/options/mod.rs
```diff
@@ -1890,6 +1898,9 @@ impl ConnectionString {
                         Some(index) => {
                             let (k, v) = exclusive_split_at(kvp, index);
                             let key = k.ok_or_else(err_func)?;
+                            if key == "ALLOWED_HOSTS" {
+                                return Err(err_func());
```
isabelatkinson, line 1902: "improperly formatted" might be confusing here -- can we add a message indicating that this val has to be specified through client options (if I'm understanding the spec correctly)?

src/client/mod.rs
isabelatkinson: nice 😎

src/client/auth/oidc.rs
```diff
@@ -0,0 +1,153 @@
...
+pub struct IdpServerInfo {
+    pub issuer: String,
+    #[serde(rename = "clientId")]
+    pub client_id: String,
+    #[serde(rename = "requestScopes")]
+    pub request_scopes: Option<Vec<String>>,
+}
+
+#[non_exhaustive]
+pub struct IdpServerResponse {
```
isabelatkinson, line 73: I think we'll need to add a builder to this so that users can construct it due to non_exhaustive, ditto anything else that users will need to construct

src/client/auth/oidc.rs
```diff
@@ -0,0 +1,153 @@
...
+#[derive(Clone)]
+pub struct Callbacks {
+    inner: Arc<CallbacksInner>,
+}
+
+impl Callbacks {
+    /// Create a new instance with a token request callback.
+    pub fn new<F>(on_request: F) -> Self
+    where
+        F: Fn(IdpServerInfo, RequestParameters) -> BoxFuture<'static, Result<IdpServerResponse>>
```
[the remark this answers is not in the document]
abr-egn: Makes sense, vendored.

Applying a lifetime would require propagating it all the way back up through `ClientOptions` and would also complicate `Clone` quite a bit, unfortunately.

src/client/auth/mod.rs
```diff
@@ -176,6 +181,38 @@ impl AuthMechanism {
...
+                    return Err(Error::invalid_argument(
+                        "password must not be set for MONGODB-OIDC authentication",
+                    ));
+                }
+                if credential.mechanism != Some(AuthMechanism::MongoDbOidc) {
+                    return Err(Error::invalid_argument(
+                        "mechanism must be MONGODB-OIDC for MONGODB-OIDC authentication",
+                    ));
+                }
+                Ok(())
```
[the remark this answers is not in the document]
abr-egn: You're right, that's unreachable - removed.  Was just me following the spec without paying attention.

src/client/auth/oidc.rs
```diff
@@ -0,0 +1,153 @@
...
+    on_request: Box<
+        dyn Fn(IdpServerInfo, RequestParameters) -> BoxFuture<'static, Result<IdpServerResponse>>
+            + Send
+            + Sync,
+    >,
+    // on_refresh: Option<Box<dyn Fn(&IdpServerInfo) -> IdpServerResponse + Send + Sync>>,
+}
+
+#[derive(Debug, Deserialize)]
+#[non_exhaustive]
```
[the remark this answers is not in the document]
abr-egn: Done.

src/client/auth/oidc.rs
abr-egn: It's just this one, done :)

drshika (approved):

src/client/auth/oidc.rs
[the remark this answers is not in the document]
isabelatkinson: ah that makes sense, I think that's fine, then.

isabelatkinson (approved):
lgtm!

PR #888, opened 2023-06-07
minor: skip session tests if mongocryptd fails to spawn

abr-egn:
`mongocryptd` doesn't need to be installed for 99% of work on the driver, and this way those tests don't fail in that case if run incidentally (i.e. as part of a full test run).

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -93,6 +93,8 @@ functions:
 
               export AZURE_IMDS_MOCK_PORT=44175
 
+              export SESSION_TEST_REQUIRE_MONGOCRYPTD=1
+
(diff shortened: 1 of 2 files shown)


isabelatkinson:
Is there a specific variant on which this test has been failing? Or is it just flaky? I'm concerned that if this starts failing unconditionally we won't have any indication.

abr-egn:
> Is there a specific variant on which this test has been failing? Or is it just flaky? I'm concerned that if this starts failing unconditionally we won't have any indication.

I actually intended this for convenience when running the test suite manually, but good point - added an env var so evergreen tests treat this case as a failure still.

PR #889, opened 2023-06-07
RUST-1421 Implement FromStr for Namespace

drshika:


abr-egn (commented):
Looks good overall!  I think it'd be good to add a test to make sure multiple `.`s work (i.e. `"foo.bar.baz"` parses into `"foo"` and `"bar.baz"`) and then it'll be good to go.

src/coll/mod.rs
```diff
@@ -1501,3 +1501,23 @@ impl Serialize for Namespace {
...
+impl FromStr for Namespace {
+    type Err = Error;
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        let mut parts = s.split('.');
+
+        let db = parts.next();
+        let coll = parts.collect::<Vec<_>>().join(".");
+
+        match (db, coll) {
+            (Some(db), coll) if !coll.is_empty() => std::result::Result::Ok(Self {
```
abr-egn, line 1514: You can just use `Ok` here instead of `std::result::Result::Ok` (and likewise for `std::result::Result::Err` below) - `Ok` and `Err` are common enough that they're part of the [Rust prelude](https://doc.rust-lang.org/std/prelude/index.html).

src/test/coll.rs
```diff
@@ -1201,3 +1202,16 @@ fn assert_duplicate_key_error_with_utf8_replacement(error: &ErrorKind) {
         ),
     }
 }
+
+#[test]
+fn test_namespace_fromstr() {
+    let t: Namespace = "something.somethingelse".parse().unwrap();
+    assert_eq!(t.db, "something");
+    assert_eq!(t.coll, "somethingelse");
+    let t2: std::result::Result<Namespace, <Namespace as std::str::FromStr>::Err> =
```
abr-egn, line 1211: Likewise, this can just be `let t2: Result<Namespace> = ...`

abr-egn (approved):
LGTM!  Tagging in Isabel.

PR #892, opened 2023-06-09
RUST-1370 Only check specific fields for hello equality

abr-egn:


isabelatkinson (commented):
It looks like the only place we were using our `PartialEq` implementation for `HelloCommandResponse` was the deleted code on line 151. I'm not sure the reasoning behind the way it was implemented the first time around, but I would guess it was custom logic for server description equality checks so I think we'd be fine to revamp that impl rather than add a new method for equality checking. We'd just want to add a comment similar to the one you added in this PR. As a side note, something I happened to discover today was this pattern:

```
#[derive(Derivative)]
#[derivative(PartialEq)]
struct A {
    pub x: i32,
    pub y: String,
    #[derivative(PartialEq = "ignore")]
    pub z: Option<f64>,
}
```

which might be a little bit more maintainable than a manual implementation of the trait if we're not skipping too many fields.

abr-egn:
I think the reason it's not using `Derivative` is because `server_type()` has non-trivial logic so it's not just a field-by-field compare.  I'd like to keep the logic for server description comparison specific to that rather than making it the general `PartialEq` impl since that seems easy to accidentally use elsewhere.

Good point that the impl was only used in the removed code, deleted it :)

isabelatkinson (approved):

PR #893, opened 2023-06-12
minor: remove unused ocsp test configuration

abr-egn:
This has been unused longer than I've been on the team, removing it will mean we don't bump into it while otherwise updating the config.  I'll add a comment pointing RUST-361 at this PR in case what's there is useful reference for when we do get around to that.


isabelatkinson (approved):
lgtm, thanks for cleaning this up!

PR #894, opened 2023-06-12
RUST-1571 Update ServerId to be int64

drshika:
ServerId is assigned to ConnectionId here: [https://github.com/mongodb/mongo-rust-driver/blob/main/src/cmap/establish/handshake/mod.rs#L457](https://github.com/mongodb/mongo-rust-driver/blob/main/src/cmap/establish/handshake/mod.rs#L457 ). To fix the int32 overflow issue, I added a new ServerId int64 value.

### DIFF
--- a/src/cmap/conn/mod.rs
+++ b/src/cmap/conn/mod.rs
@@ -46,8 +46,8 @@ pub struct ConnectionInfo {
     pub id: u32,
 
     /// A server-generated identifier that uniquely identifies the connection. Available on server
-    /// versions 4.2+. This may be used to correlate driver connections with server logs. This
-    /// could be a truncated value based on input server_id.
+    /// versions 4.2+. This may be used to correlate driver connections with server logs. 
+    /// If the connection ID sent by the server is too large for an i32, this will be a truncated value.
     pub server_id: Option<i32>,
 
     /// A server-generated identifier that uniquely identifies the connection. Available on server
--- a/src/cmap/conn/mod.rs
+++ b/src/cmap/conn/mod.rs
@@ -51,8 +51,8 @@ pub struct ConnectionInfo {
     pub server_id: Option<i32>,
 
     /// A server-generated identifier that uniquely identifies the connection. Available on server
-    /// versions 4.2+. This may be used to correlate driver connections with server logs. Reference
-    /// this for the full server_id
+    /// versions 4.2+. This may be used to correlate driver connections with server logs. This value
+    /// will not be truncated and should be used rather than `server_id`.
     pub server_id_i64: Option<i64>,
 
     /// The address that the connection is connected to.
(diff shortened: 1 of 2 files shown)


src/cmap/conn/mod.rs
```diff
@@ -165,7 +171,8 @@ impl Connection {
     pub(crate) fn info(&self) -> ConnectionInfo {
         ConnectionInfo {
             id: self.id,
-            server_id: self.server_id,
```
abr-egn, line 168: I'm of two minds here whether truncation or omission is better when the server_id doesn't fit in an i32.  @isabelatkinson do you have an opinion here?

src/cmap/conn/mod.rs
```diff
@@ -46,9 +46,15 @@ pub struct ConnectionInfo {
     pub id: u32,
 
     /// A server-generated identifier that uniquely identifies the connection. Available on server
-    /// versions 4.2+. This may be used to correlate driver connections with server logs.
+    /// versions 4.2+. This may be used to correlate driver connections with server logs. This
+    /// could be a truncated value based on input server_id.
```
abr-egn, line 50: Nit: how about "If the connection ID sent by the server is too large for an i32, this will be a truncated value." (or "will be None." if we go the other way) for wording?  I think it's useful for our users when we can be more precise about the behavior.

abr-egn (commented):
Looks good!  Waiting to see what Isabel's thoughts are before giving the official stamp.

src/cmap/conn/mod.rs
isabelatkinson: I think it would be a breaking/unexpected change for users who are currently relying on this field to be `Some` (even if it's a truncated value) to stop populating it.

drshika: I agree with Isabel. The i64 overflow issue affects a small amount of users and would bother more users if we just removed support for the original server_id without warning.

abr-egn (approved):
LGTM!

src/cmap/conn/mod.rs
```diff
@@ -46,9 +46,15 @@ pub struct ConnectionInfo {
...
 
     /// A server-generated identifier that uniquely identifies the connection. Available on server
-    /// versions 4.2+. This may be used to correlate driver connections with server logs.
+    /// versions 4.2+. This may be used to correlate driver connections with server logs. 
+    /// If the connection ID sent by the server is too large for an i32, this will be a truncated value.
     pub server_id: Option<i32>,
 
+    /// A server-generated identifier that uniquely identifies the connection. Available on server
+    /// versions 4.2+. This may be used to correlate driver connections with server logs. Reference
+    /// this for the full server_id
```
isabelatkinson, line 55: (feel free to tweak the phrasing if you'd like)
```suggestion
    /// versions 4.2+. This may be used to correlate driver connections with server logs. This value
    /// will not be truncated and should be used rather than `server_id`.
```

isabelatkinson (approved):

PR #895, opened 2023-06-15
RUST-1330 Consolidate test script core functions

abr-egn:
RUST-1330

This brings all the various places we were building options for `cargo test` into a single shared set of functions; as a side effect, a bunch of test targets that didn't have the nice granular per-test output in evergreen now do.  I also folded together `configure-rust.sh` and `env.sh` since the former called the latter and they're both just setting a bunch of variables.


.evergreen/cargo-test.sh
```diff
@@ -0,0 +1,44 @@
...
+
+CARGO_OPTIONS=()
+TEST_OPTIONS=("-Z unstable-options" "--format json" "--report-time")
+FEATURE_FLAGS=()
+CARGO_RESULT=0
+
+use_single_thread() {
+    if [ "${SINGLE_THREAD}" = true ]; then
+        TEST_OPTIONS+=("--test-threads=1")
+    fi
```
isabelatkinson, line 11: it looks like we were only doing this for some tests, but I think the original intention of this flag was to run all of our tests single-threaded when it's set. maybe we should just inline this in `cargo_test`?

isabelatkinson (commented):
lgtm, this is a nice refactor! just a few suggestions

.evergreen/run-csfle-tests.sh
```diff
@@ -4,15 +4,13 @@ set -o errexit
 set -o pipefail
 
 source ./.evergreen/env.sh
+source .evergreen/cargo-test.sh
 
 set -o xtrace
 
-FEATURE_FLAGS="in-use-encryption-unstable,aws-auth,azure-kms,${TLS_FEATURE}"
-OPTIONS="-- -Z unstable-options --format json --report-time"
+FEATURE_FLAGS+=("in-use-encryption-unstable" "aws-auth" "azure-kms" "${TLS_FEATURE}")
```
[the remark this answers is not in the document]
abr-egn: Good catch - filed RUST-1690 to address that.

isabelatkinson:
hm, looks like the macOS tests were broken by inlining the single-thread check 🤔

.evergreen/cargo-test.sh
abr-egn: Done.

abr-egn:
> hm, looks like the macOS tests were broken by inlining the single-thread check 🤔

Yeah, turns out appending an option in a loop isn't a great plan, fixed :)

isabelatkinson (approved):

PR #901, opened 2023-06-23
RUST-1636 Add RunCommand Test Specification

drshika:
1. Removed `write_concern` and `read_concern` from RunCommand API to update with the new specification: https://github.com/mongodb/specifications/commit/ab66eaf6f7e2bbeb5218672bebaf5118faf8209c
2. Added in new spec tests for RunCommand

### DIFF
--- a/src/operation/run_command/mod.rs
+++ b/src/operation/run_command/mod.rs
@@ -1,10 +1,8 @@
 #[cfg(test)]
 mod test;
 
-use std::convert::TryInto;
-
-use crate::operation::WriteConcern;
 use bson::{RawBsonRef, RawDocumentBuf};
+use std::convert::TryInto;
 
 use super::{CursorBody, OperationWithDefaults};
 use crate::{
@@ -45,17 +43,10 @@ impl<'conn> RunCommand<'conn> {
         selection_criteria: Option<SelectionCriteria>,
         pinned_connection: Option<&'conn PinnedConnectionHandle>,
     ) -> Result<Self> {
-        let write_concern = command
-            .get("writeConcern")?
-            .and_then(|b| b.as_document())
-            .map(|doc| bson::from_slice::<WriteConcern>(doc.as_bytes()))
-            .transpose()?;
-
         Ok(Self {
             db,
             command,
             selection_criteria,
-            write_concern,
             pinned_connection,
         })
     }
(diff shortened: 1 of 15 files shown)


isabelatkinson (commented):
looks good, just a small suggestion. we can also keep an eye on whether we can remove the sharded-replicaset topology from the spec tests; if so, I'll make a PR on the specifications repo doing so (or you can do so if interested!) and we can resync the tests once those changes are merged. also, can you add the ticket number to the PR description?

src/test/spec/unified_runner/test_file.rs
```diff
@@ -84,6 +84,7 @@ pub(crate) struct RunOnRequirement {
 pub(crate) enum Topology {
     Single,
     ReplicaSet,
+    #[serde(alias = "sharded-replicaset")]
```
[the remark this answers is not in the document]
isabelatkinson: update: I filed a ticket for this so we can remove this in a future PR. no action needed here

src/operation/run_command/mod.rs
```diff
@@ -1,17 +1,15 @@
 #[cfg(test)]
 mod test;
 
-use std::convert::TryInto;
-
 use bson::{RawBsonRef, RawDocumentBuf};
+use std::convert::TryInto;
```
isabelatkinson, line 5: nit: can we revert this change? usually imports from std go at the top, then imports from external crates, then internal imports (with a blank line between each section)

isabelatkinson (commented):
one very small comment, otherwise code changes look good to me! a few other things:

- can we add the ticket number to the PR title?
- is there anything we need to do to address unacknowledged writes? it seems like we should be in the clear based on Kevin's comment in the channel but want to make sure we've thought that through

src/test/spec/run_command.rs
```diff
@@ -10,6 +10,8 @@ async fn run_unified() {
         .skip_tests(&[
             // TODO re: RUST-1649: fix withTransaction for new test runner
             "attaches transaction fields to given command",
+            "always attaches $db and implicit lsid to given command and omits default readPreference",
```
abr-egn, line 13: "attaches transaction fields to given command" was skipped because we don't have an implementation of `withTransaction` for the unified runner, but it doesn't look like these use that - I think these were valid failures that need investigation.

abr-egn (approved):
LGTM!

isabelatkinson (approved):

PR #905, opened 2023-06-26
RUST-1420 Cache AWS credentials received from endpoints

isabelatkinson:


isabelatkinson:
I'm fixing up the MSRV failure but this is ready for review otherwise.

src/test/auth_aws.rs
```diff
@@ -13,3 +16,118 @@ async fn auth_aws() {
...
+// caching tests use a regular client instead to avoid that noise.
+async fn get_client() -> Client {
+    Client::with_uri_str(DEFAULT_URI.clone()).await.unwrap()
+}
+
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn credential_caching() {
+    // This test should only be run when authenticating using AWS endpoints.
+    if var("SKIP_CREDENTIAL_CACHING_TESTS").is_ok() {
```
abr-egn, line 30: It might make the evergreen config a little less noisy if this were an enable rather than a skip.  That's just a vague inclination, though, so feel free to disregard if you feel otherwise :)

abr-egn (approved):
LGTM!

src/test/auth_aws.rs
isabelatkinson: I thought about doing that originally (agreed this is more noisy), but I'd prefer for these tests to be opt-out rather than opt-in so that we don't forget to enable them if we add a new mechanism using an endpoint in the future.

PR #907, opened 2023-06-29
RUST-1604 Add custom bucketing fields to timeseries options

isabelatkinson:
The drivers ticket description only details a test sync, but the new tests were using options we hadn't implemented yet.

### DIFF
--- a/src/db/options.rs
+++ b/src/db/options.rs
@@ -98,11 +98,7 @@ pub struct CreateCollectionOptions {
     /// Used to automatically delete documents in time series collections. See the [`create`
     /// command documentation](https://www.mongodb.com/docs/manual/reference/command/create/) for more
     /// information.
-    #[serde(
-        default,
-        deserialize_with = "serde_util::deserialize_duration_option_from_u64_seconds",
-        serialize_with = "serde_util::serialize_duration_option_as_int_secs"
-    )]
+    #[serde(default, with = "serde_util::duration_option_as_int_seconds")]
     pub expire_after_seconds: Option<Duration>,
 
     /// Options for supporting change stream pre- and post-images.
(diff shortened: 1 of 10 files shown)


src/db/options.rs
```diff
@@ -192,8 +188,10 @@ pub struct IndexOptionDefaults {
 }
 
 /// Specifies options for creating a timeseries collection.
+#[skip_serializing_none]
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, TypedBuilder)]
 #[serde(rename_all = "camelCase")]
+#[builder(field_defaults(default, setter(into)))]
```
abr-egn, line 194: Is this a backwards-compatible change?

isabelatkinson: Good question. I did some testing and it looks like `setter(into)` will break this:

```
let options = TimeseriesOptions::builder()
        .time_field("x".into())
        .meta_field(Some("y".into()))
        .granularity(Some(TimeseriesGranularity::Hours))
        .build();
```

so we'll need to leave that out for now. We could also override the default setting for the existing fields using `#[builder(setter(!into))]` (see [docs](https://docs.rs/typed-builder/latest/typed_builder/derive.TypedBuilder.html)), but the inconsistency between fields might be confusing.

The lack of `default` here, however, is currently pretty prohibitive, as the following won't compile:

```
let options = TimeseriesOptions::builder().time_field("x".into()).build();
```

because the builder doesn't default the optional fields to `None`. This means that users currently need to specify all fields, and therefore adding new fields will break code. I can't think of a scenario in which `default` will break anything so I _think_ it's a fine addition, but if we want to play it safe we could just defer these fields until 3.0.

As an aside, I know we decided a while back that these we should use `default` and `setter(into)` for all of our `TypedBuilder` implementations, but I can't find any documentation of this choice outside of (#368). I remember some convos about this in the team channel and the Prisma channel, but those would've been cleared a long time ago because of the slack retention policy. It looks like there are some structs in our API that don't have these options set, so I filed RUST-1695 to fix this for 3.0 and document/enforce the policy to avoid running into this in the future.

abr-egn: Thanks for testing this!  I also can't see any way `default` would be breaking, so that seems like the right way here.

abr-egn (approved):
LGTM!

PR #912, opened 2023-07-10
RUST-1588: Add RunCursorCommand

drshika:
added operation `RunCursorCommand` (spec: https://github.com/mongodb/specifications/blob/c09f979ad296400552a98c9b784197ec648c2096/source/run-command/run-command.rst#runcursorcommand)
- implemented Operation which uses default RunCommand behavior for every method but handle_response

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -501,18 +501,23 @@ impl Database {
         options: RunCursorCommandOptions,
         session: &mut ClientSession,
     ) -> Result<SessionCursor<Document>> {
+        let mut selection_criteria = SelectionCriteria::ReadPreference(options.read_preference.clone().unwrap()).into();
         match session.transaction.state {
             TransactionState::Starting | TransactionState::InProgress => {
-                if command.contains_key("readConcern") {
-                    return Err(ErrorKind::InvalidArgument {
-                        message: "Cannot set read concern after starting a transaction".into(),
+                selection_criteria = match selection_criteria {
+                    Some(selection_criteria) => Some(selection_criteria),
+                    None => {
+                        if let Some(ref options) = session.transaction.options {
+                            options.selection_criteria.clone()
+                        } else {
+                            None
+                        }
                     }
-                    .into());
-                }
+                };     
             }
             _ => {}
         }
-        let rcc = RunCommand::new(self.name().to_string(), command, options.read_preference.clone(), None)?;
+        let rcc = RunCommand::new(self.name().to_string(), command, selection_criteria, None)?;
         let rc_command = RunCursorCommand {
             run_command: rcc,
             options,
--- a/src/db.rs
+++ b/src/db.rs
@@ -501,7 +501,9 @@ impl Database {
         options: RunCursorCommandOptions,
         session: &mut ClientSession,
     ) -> Result<SessionCursor<Document>> {
-        let mut selection_criteria = SelectionCriteria::ReadPreference(options.read_preference.clone().unwrap()).into();
+        let mut selection_criteria = options.read_preference
+            .clone()
+            .map(SelectionCriteria::ReadPreference);        
         match session.transaction.state {
             TransactionState::Starting | TransactionState::InProgress => {
                 selection_criteria = match selection_criteria {
--- a/src/db.rs
+++ b/src/db.rs
@@ -485,7 +485,10 @@ impl Database {
         command: Document,
         options: RunCursorCommandOptions,
     ) -> Result<Cursor<Document>> {
-        let rcc = RunCommand::new(self.name().to_string(), command, None, None)?;
+        let selection_criteria = options.read_preference
+            .clone()
+            .map(SelectionCriteria::ReadPreference);
+        let rcc = RunCommand::new(self.name().to_string(), command, selection_criteria, None)?;
         let rc_command = RunCursorCommand {
             run_command: rcc,
             options,
@@ -744,7 +747,7 @@ impl<'conn> Operation for RunCursorCommand<'conn> {
             cursor_info,
             description.server_address.clone(),
             self.options.batch_size,
-            self.options.max_time_ms,
+            self.options.max_time,
             self.options.comment.clone(),
         ))
     }
--- a/src/db.rs
+++ b/src/db.rs
@@ -482,7 +482,7 @@ impl Database {
     pub async fn run_cursor_command(
         &self,
         command: Document,
-        options: RunCursorCommandOptions,
+        options: impl Into<Option<RunCursorCommandOptions>>,
     ) -> Result<Cursor<Document>> {
         let rcc = RunCommand::new(self.name().to_string(), command, options.read_preference.clone(), None)?;
         let rc_command = RunCursorCommand::new(rcc, options)?; 
--- a/src/db.rs
+++ b/src/db.rs
@@ -48,8 +48,6 @@ use crate::{
     SessionCursor,
 };
 
-use self::options::RunCursorCommandWithSessionOptions;
-
 /// `Database` is the client-side abstraction of a MongoDB database. It can be used to perform
 /// database-level operations or to obtain handles to specific collections within the database. A
 /// `Database` can only be obtained through a [`Client`](struct.Client.html) by calling either
@@ -500,9 +498,9 @@ impl Database {
     pub async fn run_cursor_command_with_session(
         &self,
         command: Document,
-        options: RunCursorCommandWithSessionOptions,
+        options: RunCursorCommandOptions,
         session: &mut ClientSession,
-    ) -> Result<Cursor<Document>> {
+    ) -> Result<SessionCursor<Document>> {
         match session.transaction.state {
             TransactionState::Starting | TransactionState::InProgress => {
                 if command.contains_key("readConcern") {
@@ -515,13 +513,12 @@ impl Database {
             _ => {}
         }
         let rcc = RunCommand::new(self.name().to_string(), command, options.read_preference.clone(), None)?;
-        let rc_command = RunCursorCommandWithSession {
+        let rc_command = RunCursorCommand {
             run_command: rcc,
             options,
-            session,
         };
         let client = self.client();
-        client.execute_cursor_operation(rc_command).await
+        client.execute_session_cursor_operation(rc_command, session).await
     }
 
     /// Runs a database-level command using the provided `ClientSession`.
@@ -669,90 +666,6 @@ pub(super) struct RunCursorCommand<'conn> {
 }
 
 impl<'conn> Operation for RunCursorCommand<'conn> {
-    type O = CursorSpecification;
-    type Command = RawDocumentBuf;
-
-    const NAME: &'static str = "run_cursor_command";
-
-    fn build(&mut self, _description: &StreamDescription) -> Result<Command<Self::Command>> {
-        self.run_command.build(_description)
-    }
-
-    fn serialize_command(&mut self, cmd: Command<Self::Command>) -> Result<Vec<u8>> {
-        self.run_command.serialize_command(cmd)
-    }
-
-    fn extract_at_cluster_time(
-        &self,
-        _response: &bson::RawDocument,
-    ) -> Result<Option<bson::Timestamp>> {
-        self.run_command.extract_at_cluster_time(_response)
-    }
-
-    fn handle_error(&self, error: Error) -> Result<Self::O> {
-        Err(error)
-    }
-
-    fn selection_criteria(&self) -> Option<&SelectionCriteria> {
-        self.run_command.selection_criteria()
-    }
-
-    fn is_acknowledged(&self) -> bool {
-        self.run_command.is_acknowledged()
-    }
-
-    fn write_concern(&self) -> Option<&WriteConcern> {
-        self.run_command.write_concern()
-    }
-
-    fn supports_read_concern(&self, _description: &StreamDescription) -> bool {
-        self.run_command.supports_read_concern(_description)
-    }
-
-    fn supports_sessions(&self) -> bool {
-        self.run_command.supports_sessions()
-    }
-
-    fn retryability(&self) -> crate::operation::Retryability {
-        self.run_command.retryability()
-    }
-
-    fn update_for_retry(&mut self) {
-        self.run_command.update_for_retry()
-    }
-
-    fn pinned_connection(&self) -> Option<&PinnedConnectionHandle> {
-        self.run_command.pinned_connection()
-    }
-
-    fn name(&self) -> &str {
-        self.run_command.name()
-    }
-
-    fn handle_response(
-        &self,
-        response: RawCommandResponse,
-        description: &StreamDescription,
-    ) -> Result<Self::O> {
-        let doc = Operation::handle_response(&self.run_command, response, description)?;
-        let cursor_info = bson::from_document(doc)?;
-        Ok(CursorSpecification::new(
-            cursor_info,
-            description.server_address.clone(),
-            self.options.batch_size,
-            self.options.max_time_ms,
-            self.options.comment.clone(),
-        ))
-    }
-}
-
-pub(super) struct RunCursorCommandWithSession<'conn> {
-    run_command: RunCommand<'conn>,
-    options: RunCursorCommandWithSessionOptions,
-    session: &'conn mut ClientSession,
-}
-
-impl<'conn> Operation for RunCursorCommandWithSession<'conn> {
     type O = CursorSpecification;
     type Command = RawDocumentBuf;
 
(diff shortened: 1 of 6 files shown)


src/db/options.rs
```diff
@@ -312,3 +312,34 @@ pub struct ChangeStreamPreAndPostImages {
...
+    CursorLifetime,
+}
+
+/// Specifies the options to a
+/// [`Database::RunCursorCommand`](../struct.Database.html#method.run_cursor_command) operation.
+#[derive(Clone, Debug, Default, Deserialize, TypedBuilder)]
+#[serde(rename_all = "camelCase")]
+#[builder(field_defaults(default, setter(into)))]
+#[non_exhaustive]
+pub struct RunCursorCommandOptions {
```
abr-egn, line 332: I just noticed that [the spec](https://github.com/mongodb/specifications/blob/master/source/run-command/run-command.rst#runcursorcommand) has `RunCursorCommandOptions` extending `RunCommandOptions`.  We don't actually have `RunCommandOptions` in the Rust driver (this was a mistake, one of the things we want to fix in 3.0), so we'll need to fold [those options](https://github.com/mongodb/specifications/blob/master/source/run-command/run-command.rst#runcommand) from the spec into `RunCursorCommandOptions`:
* `readPreference`: this can just be directly added to `RunCursorCommandOptions`, it'll be used to supply the `SelectionCriteria` for `RunCommand::new`
* `session`: because of Rust lifetime restrictions, this can't be directly included in the options; instead, we have to kind of awkwardly duplicate most methods to provide one that doesn't accept a session and one that does.  See `run_command` / `run_command_with_session` for an example.
* `timeoutMS`: this is for client-side operation timeouts (CSOT), which the Rust driver doesn't support, so you can just skip it.

src/db.rs
```diff
@@ -607,3 +632,88 @@ impl Database {
...
+}
+
+impl<'conn> Operation for RunCursorCommand<'conn> {
+    type O = CursorSpecification;
+    type Command = RawDocumentBuf;
+
+    const NAME: &'static str = "run_cursor_command";
+
+    fn build(&mut self, _description: &StreamDescription) -> Result<Command<Self::Command>> {
+        Operation::build(&mut self.run_command, _description)
```
abr-egn, line 648: You should be able to call this (and the other places you use `Operation::<foo>`) directly, i.e. `self.run_command.build(_description)`.  What you've got here is the [fully qualified syntax](https://doc.rust-lang.org/reference/expressions/call-expr.html#disambiguating-function-calls), which will always work, but usually isn't needed.

src/db.rs
```diff
@@ -607,3 +632,88 @@ impl Database {
...
+    }
+
+    fn handle_response(
+        &self,
+        response: RawCommandResponse,
+        description: &StreamDescription,
+    ) -> Result<Self::O> {
+        let doc = Operation::handle_response(&self.run_command, response, description)?;
+        let cursor_info = bson::from_document(doc)?;
+        // let response: CursorBody = response.body()?;
```
abr-egn, line 709: These commented lines can be removed now :)

src/db.rs
```diff
@@ -469,6 +479,48 @@ impl Database {
...
+    /// Runs a database-level command and returns a cursor to the response.
+    pub async fn run_cursor_command_with_session(
+        &self,
+        command: Document,
+        options: RunCursorCommandOptions,
+        session: &mut ClientSession,
+    ) -> Result<SessionCursor<Document>> {
+        match session.transaction.state {
+            TransactionState::Starting | TransactionState::InProgress => {
+                if command.contains_key("readConcern") {
```
abr-egn, line 506: The spec actually says
> Drivers MUST NOT attempt to check the command document for the presence of a readConcern and writeConcern field.

so I think this check should be removed.  I do think that using the transaction's `selection_criteria` if it's not set by the user like `run_command_with_session` does is also applicable here.

src/db/options.rs
```diff
@@ -312,3 +312,37 @@ pub struct ChangeStreamPreAndPostImages {
...
+    pub cursor_type: Option<CursorType>,
+    /// Number of documents to return per batch.
+    pub batch_size: Option<u32>,
+    /// Optional non-negative integer value. Use this value to configure the maxTimeMS option sent
+    /// on subsequent getMore commands.
+    pub max_time_ms: Option<Duration>,
+    /// Optional BSON value. Use this value to configure the comment option sent on subsequent
+    /// getMore commands.
+    pub comment: Option<Bson>,
+    /// The session to run this command with.
```
abr-egn, line 346: I don't think session should be here (is this field used anywhere?)

src/db/options.rs
```diff
@@ -312,3 +312,37 @@ pub struct ChangeStreamPreAndPostImages {
...
+/// Specifies the options to a
+/// [`Database::RunCursorCommand`](../struct.Database.html#method.run_cursor_command) operation.
+#[derive(Clone, Debug, Default, TypedBuilder)]
+#[builder(field_defaults(default, setter(into)))]
+#[non_exhaustive]
+pub struct RunCursorCommandOptions {
+    /// Optional string enum value, one of 'iteration' | 'cursorLifetime'.
+    pub timeout_mode: Option<TimeoutMode>,
+    /// The default read preference for operations.
+    pub read_preference: Option<SelectionCriteria>,
```
abr-egn, line 335: I think this should be an `Option<ReadPreference>` - that can be wrapped in a `SelectionCriteria` when it needs to be passed to `RunCommand::new`.

src/operation.rs
```diff
@@ -74,7 +74,7 @@ pub(crate) use run_command::RunCommand;
 pub(crate) use update::Update;
 
 const SERVER_4_2_0_WIRE_VERSION: i32 = 8;
-const SERVER_4_4_0_WIRE_VERSION: i32 = 9;
+pub const SERVER_4_4_0_WIRE_VERSION: i32 = 9;
```
abr-egn, line 77: I don't know why this change is here?

src/db.rs
```diff
@@ -501,18 +501,23 @@ impl Database {
         options: RunCursorCommandOptions,
         session: &mut ClientSession,
     ) -> Result<SessionCursor<Document>> {
+        let mut selection_criteria = SelectionCriteria::ReadPreference(options.read_preference.clone().unwrap()).into();
```
abr-egn, line 504: Breaking down what's going on here:
* `options.read_preference.clone()` - creates a temporary value, of type `Option<ReadPreference>`.  Not using `clone()` here would move the value of `read_preference` out of `options`, leaving the latter partially-initialized, so yup, makes sense.
* `<above>.unwrap()` - this consumes the temporary `Option<ReadPreference>` and produces a `ReadPreference`.  It does this by checking whether the value is `Some` (it can just return the value) or `None` (it terminates the program with a panic).  Since it would be quite impolite of us to crash the user's program, using `unwrap` is something we avoid in almost all circumstances.
* `SelectionCriteria::ReadPreference(<above>)` - straightforward.
* `<above>.into()` - the Rust standard library provides the `Into` trait (and its mirror `From`) for general type conversions.  What it actually does depends on both the type being converted from and the type being converted to.  In this case, the compiler can tell that the source type is `SelectionCriteria` (because you just constructed it and are calling `into()` directly on it) and can infer that the destination type is `Option<SelectionCriteria>` (because you `match` against it below with `Some()` and `None`).  Converting some value `v` of type `T` into a new value of type `Option<T>` is very simple - it just constructs `Some(v)`.

So, putting that all together, what this line will do:
* if `options.read_preference` contains a `Some(v)` value, this will construct a `Some(SelectionCriteria::ReadPreference(v))` - yup, that's what we want 👍 
* if `options.read_preference` contains a `None` value, this will crash! 👎 

What we *actually* want is to turn a `Some(v)` into a `Some(SelectionCriteria::ReadPreference(v))` and a `None` into, well, `None`.  This is an excellent place for a `match`, since that's exactly what match is for 🙂:
```
let mut selection_criteria = match options.read_preference.clone() {
  Some(v) => Some(SelectionCriteria::ReadPreference(v)),
  None => None,
};
```

It turns out that with `Option` this kind of "wrap it if it's `Some`, do nothing if it's `None`" pattern is _very_ common, so there's a standard library function to do just that, [`Option::map`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map).  Using that, this becomes:
```
let mut selection_criteria = options.read_preference
    .clone()
    .map(|v| SelectionCriteria::ReadPreference(v));
```
As a bit of a side note, because the closure `|v| SelectionCriteria::ReadPreference(v)` is just passing its argument directly through, this can be simplified slightly:
```
let mut selection_criteria = options.read_preference
    .clone()
    .map(SelectionCriteria::ReadPreference);
```

Working with values like this can definitely feel a bit foreign if you're not used to languages that use _algebraic data types_ as a core feature.  I'd encourage you to approach situations like this by starting with a "what data do I have, and what data do I need" analysis, seeing if you can translate that into a `match` statement (or multiple, if needed), and then looking to see if there are utility functions for what you've ended up doing.

abr-egn (approved):
LGTM!  Tagged Isabel in.

src/db/options.rs
```diff
@@ -312,3 +312,35 @@ pub struct ChangeStreamPreAndPostImages {
...
+
+/// Determines the lifetime of a cursor.
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub enum TimeoutMode {
+    /// Times out after one Iteration.
+    Iteration,
+    /// Times out after the end of the Cursor Lifetime.
+    CursorLifetime,
+}
```
isabelatkinson, line 324: I think we should omit this (and the field below) for now. Timeout mode is only relevant for CSOT, which we don't yet support in the driver. We'll add it when we do RUST-582.

src/db/options.rs
```diff
@@ -312,3 +312,35 @@ pub struct ChangeStreamPreAndPostImages {
...
+/// Specifies the options to a
+/// [`Database::RunCursorCommand`](../struct.Database.html#method.run_cursor_command) operation.
+#[derive(Clone, Debug, Default, TypedBuilder)]
+#[builder(field_defaults(default, setter(into)))]
+#[non_exhaustive]
+pub struct RunCursorCommandOptions {
+    /// Optional string enum value, one of 'iteration' | 'cursorLifetime'.
+    pub timeout_mode: Option<TimeoutMode>,
+    /// The default read preference for operations.
+    pub read_preference: Option<ReadPreference>,
```
isabelatkinson, line 335: Why use a `ReadPreference` here rather than a `SelectionCriteria`?

src/db/options.rs
```diff
@@ -312,3 +312,35 @@ pub struct ChangeStreamPreAndPostImages {
...
+    /// Optional string enum value, one of 'tailable' | 'tailableAwait' | 'nonTailable'.
```
isabelatkinson, line 336:
```suggestion
    /// The type of cursor to return.
```
I think we can make this a little simpler -- the fact that this option is serialized as a string in the command is more of an implementation detail, and the possible options are already documented by the enum. (This is the doc comment for the `cursor_type` field on `FindOptions`.)

src/db.rs
```diff
@@ -607,3 +666,86 @@ impl Database {
...
+    options: RunCursorCommandOptions,
+}
+
+impl<'conn> Operation for RunCursorCommand<'conn> {
+    type O = CursorSpecification;
+    type Command = RawDocumentBuf;
+
+    const NAME: &'static str = "run_cursor_command";
+
+    fn build(&mut self, _description: &StreamDescription) -> Result<Command<Self::Command>> {
```
isabelatkinson, line 681:
```suggestion
    fn build(&mut self, description: &StreamDescription) -> Result<Command<Self::Command>> {
```
(no need to have the underscore here since this isn't an unused variable)

isabelatkinson, line 681: We'll also need to apply the user's options in this method.

src/db.rs
```diff
@@ -469,6 +479,55 @@ impl Database {
...
+                        if let Some(ref options) = session.transaction.options {
+                            options.selection_criteria.clone()
+                        } else {
+                            None
+                        }
+                    }
+                };     
+            }
+            _ => {}
+        }
```
isabelatkinson, line 521: I think you should be able to use the `resolve_selection_criteria_with_session` macro here ([e.g.](https://docs.rs/mongodb/latest/src/mongodb/coll/mod.rs.html#342-358)).

src/db.rs
```diff
@@ -607,3 +666,86 @@ impl Database {
         GridFsBucket::new(self.clone(), options.into().unwrap_or_default())
     }
 }
+
+pub(super) struct RunCursorCommand<'conn> {
```
isabelatkinson, line 670: We have each of our operation structs in their own file under `src/operation` -- can we move this one there as well as `src/operation/run_cursor_command.rs`?

src/db/options.rs
drshika: I had that originally but as per Abraham's comment, we also want to also add the session transaction state to the Selection Criteria which would be easier to do if we didn't already have one in here.

src/db.rs
drshika: It seems like the option types don't match up. Do we want to switch to that type or use the one we are currently using.
image

drshika: image
seems like the macro won't work out since the options types don't match up.

src/db.rs
```diff
@@ -469,6 +478,41 @@ impl Database {
...
+        command: Document,
+        options: impl Into<Option<RunCursorCommandOptions>>,
+        session: &mut ClientSession,
+    ) -> Result<SessionCursor<Document>> {
+        let mut options: Option<RunCursorCommandOptions> = options.into().clone();
+        resolve_selection_criteria_with_session!(self, options, Some(&mut *session))?;
+        let selection_criteria = match options.clone() {
+            Some(options) => options.selection_criteria,
+            None => None,
+        };
```
isabelatkinson, line 505: Rather than cloning the entire options struct, it would be more efficient only to clone the `selection_criteria`. You should be able to do this by using a reference to the options, which can be done either like this:

```
let selection_criteria = match options {
    Some(ref options) => options.selection_criteria.clone(),
    None => None,
};
```

or this: 

```
let selection_criteria = match &options {
    Some(options) => options.selection_criteria.clone(),
    None => None,
}
```

(I slightly prefer the first one for no good reason; either is totally fine.)

You can also use `and_then` to achieve the same thing:

```
let selection_criteria = options
    .as_ref()
    .and_then(|options| options.selection_criteria.clone());
```

We use `as_ref` to get a reference to the struct inside the `Option` rather than an owned version of it so that we can continue to use it later. `and_then` does something called a flat map (you can read more about it [here](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then)), which is useful when you want to avoid getting data that's wrapped in multiple layers of `Option`s/`Result`s.

All three of these options would work here, but it's generally considered best practice to use `and_then` rather than an explicit `match` in situations like these.

src/db.rs
```diff
@@ -469,6 +478,41 @@ impl Database {
...
+    }
+
+    /// Runs a database-level command and returns a cursor to the response.
+    pub async fn run_cursor_command_with_session(
+        &self,
+        command: Document,
+        options: impl Into<Option<RunCursorCommandOptions>>,
+        session: &mut ClientSession,
+    ) -> Result<SessionCursor<Document>> {
+        let mut options: Option<RunCursorCommandOptions> = options.into().clone();
```
isabelatkinson, line 500: cloning shouldn't be necessary here

src/db.rs
```diff
@@ -469,6 +478,41 @@ impl Database {
...
+        resolve_selection_criteria_with_session!(self, options, Some(&mut *session))?;
+        let selection_criteria = match options.clone() {
+            Some(options) => options.selection_criteria,
+            None => None,
+        };
+        let option = match options.clone() {
+            Some(options) => options,
+            None => RunCursorCommandOptions::default(),
+        };
```
isabelatkinson, line 509: Rather than creating an empty options struct when the user hasn't provided any options, we can just make the `options` field on `RunCursorCommand` an `Option<RunCursorCommandOptions>`. This will save us an allocation when `options` is `None`. (You'll probably need to do similar calls to `and_then` in `handle_response` as I described above to extract fields from the options).

isabelatkinson:
I think the changes still need to be pushed :) it looks like there are also some compilation errors with CSFLE on the CI run

drshika:
@isabelatkinson changes should be in here: https://github.com/mongodb/mongo-rust-driver/pull/912/commits/54185f8879625569838c0fda3cd91d54fd3dcd3f

isabelatkinson (approved):
one tiny suggestion but otherwise lgtm! no need for a re-review

isabelatkinson:
Ah it actually looks like there are some clippy and cargo-deny failures, those will need to be fixed up as well.

PR #919, opened 2023-07-24
Force non-human readable serialization in collections

### LINKED ISSUE #917: Replacements in collections use different serializer than insertions
## Versions/Environment
1. What version of Rust are you using?
   ```
   cargo 1.68.2 (6feb7c9cf 2023-03-26)
   ```
3. What operating system are you using?
   ```
   Linux 6.4.3-arch1-2 x86_64 GNU/Linux
   ```
4. What versions of the driver and its dependencies are you using? (Run `cargo pkgid mongodb` & `cargo pkgid bson`)
   ```
   https://github.com/rust-lang/crates.io-index#mongodb@2.6.0
   https://github.com/rust-lang/crates.io-index#bson@2.6.1
   ```
5. What version of MongoDB are you using? (Check with the MongoDB shell using `db.version()`)
   ```
   6.0
   ```
6. What is your MongoDB topology (standalone, replica set, sharded cluster, serverless)?
   ```
   replica set
   ```

## Describe the bug

For types that use different a serializer for human-readable serialization, we face the following issue:

When inserted, values are serialized by the non-human readable serializer on account of using `bson::to_vec` under the hood.

However, if we try to replace such values all `replace` methods in [Collection][] use internally `bson::to_document` that uses the human-readable serializer.

This implies that if we attempt to find such a value that has been replaced and deserialize, we will either get a deserialization error, or incorrect data.

**To Reproduce**
1. Define a type that implements `serde_with::SerializeAs` using different serializers for human-readable (e.g. `String`) and non-readable (e.g.  `Bytes`).
2. Then insert a value of this type into a collection. Store it as `original`
4. After doing that, replace that value with a clone of itself using any of the [Collection] methods that replace values (e.g. `replace_one`).
5. Get the replaced value `replaced` from the database and verify that `!assert_eq!(original, replaced)`.

[Collection]: https://docs.rs/mongodb/latest/mongodb/struct.Collection.html

kodemartin:
## Description

This patch resolves #917

### DIFF
--- a/src/coll/mod.rs
+++ b/src/coll/mod.rs
@@ -15,7 +15,7 @@ use serde::{
 
 use self::options::*;
 use crate::{
-    bson::{doc, to_document, Bson, Document},
+    bson::{doc, to_document_with_options, Bson, Document, SerializerOptions},
     bson_util,
     change_stream::{
         event::ChangeStreamEvent,
@@ -1119,7 +1119,10 @@ where
         options: impl Into<Option<FindOneAndReplaceOptions>>,
         session: impl Into<Option<&mut ClientSession>>,
     ) -> Result<Option<T>> {
-        let replacement = to_document(replacement.borrow())?;
+        let replacement = to_document_with_options(
+            replacement.borrow(),
+            SerializerOptions::builder().human_readable(false).build(),
+        )?;
 
         let session = session.into();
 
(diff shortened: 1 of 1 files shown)


abr-egn:
Hi!  Thanks for sending this fix in.  Once I've verified that there are no unexpected test failures I'll merge this in.  We'll also need to include a way to override this behavior in the same release in case someone is accidentally depending on it, so it won't be immediate.

PR #920, opened 2023-07-24
RUST-1608 Clean shutdown for Client

abr-egn:
RUST-1608

This adds the `shutdown` and `shutdown_immediate` methods on `Client`.  Terminating the workers was straightforward but closing outstanding resources (cursors and sessions) was very much not.  Initially, I was hoping that `shutdown` could actively track and clean up those when called, but the shared ownership required for that would have required API changes that were both backwards-incompatible and generally much less easy to work with.

What I've ended up with is using the pre-created tasks we discussed in RUST-1439; while this doesn't mean they're guaranteed to run when the runtime shuts down, it _does_ mean that the `Client` can track and await them.  Concretely, this allows `shutdown` to wait for all pending resources to have been dropped and complete async cleanup.  This also fixes RUST-1099.

This solution does have a couple of downsides:
* It pushes part of the "orderly shutdown" responsibility on to the user, requiring that all resources be dropped before `shutdown` will return.  For normal usage, this shouldn't be an issue - Rust's typical scoped memory management should mean that most of the time that happens automatically, and when it doesn't it should be easy to fix.  In case that's not possible, `shutdown_immediate` provides an escape hatch that does just the worker thread termination without the resource wait.
* There's a potential race condition where the user could create a cursor or session after `shutdown` has waited for resources but before it closes out the workers.  I think this mostly falls into the "please don't do that" sort of misbehavior.


src/client.rs
```diff
@@ -653,16 +711,35 @@ impl Client {
...
+
+#[derive(Derivative)]
+#[derivative(Debug)]
+pub(crate) struct AsyncDropToken {
+    #[derivative(Debug = "ignore")]
+    tx: Option<tokio::sync::oneshot::Sender<BoxFuture<'static, ()>>>,
+}
+
+impl AsyncDropToken {
+    pub(crate) fn spawn(&mut self, fut: impl Future<Output = ()> + Send + 'static) {
```
abr-egn, line 733: Ideally this would consume the token, but since `drop` only gives a `&mut self` this follows suit and uses an internal `Option`.

src/id_set.rs
```diff
@@ -0,0 +1,42 @@
+use std::collections::HashMap;
+
+/// A set that provides removal tokens when an item is added.
+#[derive(Debug, Clone)]
+pub(crate) struct IdSet<T> {
```
abr-egn, line 5: It feels like there _must_ be an existing and more robust solution for this, but a quick crate search didn't turn anything up.

src/runtime/async_resource.rs
```diff
@@ -0,0 +1,102 @@
+use std::sync::{Arc, Weak};
```
abr-egn, line 1: This module turned out not to be the right way to go, but it could be useful in other circumstances and getting it right was tricky, so I left it in.  Happy to remove it if you feel it's not worth the code space :)

src/tracking_arc.rs
```diff
@@ -0,0 +1,140 @@
+#[cfg(feature = "internal-track-arc")]
+use crate::id_set::{self, IdSet};
+use std::sync::Arc;
+#[cfg(feature = "internal-track-arc")]
+use std::sync::Mutex as SyncMutex;
+
+/// An `Arc` that records the backtraces of construction of live clones.  When not compiled
+/// with the `internal-track-arc` feature, a zero-cost `Arc` wrapper.
+#[derive(Debug)]
+pub(crate) struct TrackingArc<T> {
```
abr-egn, line 10: This is largely a relic of the `AsyncResource` approach, but being able to track what `Client` clones are active and where they were created seemed useful enough to preserve.

src/gridfs/upload.rs
```diff
@@ -376,7 +377,7 @@ impl Drop for GridFsUploadStream {
         if !matches!(self.state, State::Closed) {
             let chunks = self.bucket.chunks().clone();
             let id = self.id.clone();
-            runtime::execute(async move {
+            self.drop_token.spawn(async move {
```
isabelatkinson, line 380: I wrote the following test because I was curious about what "bad" usage of `shutdown` would cause in a situation like this:

```
async fn stream_shutdown() {
    use futures_util::AsyncWriteExt;

    let client = TestClient::new().await.into_client();

    let db = client.database("db");
    let bucket = db.gridfs_bucket(None);
    let mut stream = bucket.open_upload_stream("test", None);

    let buf = vec![0u8; 100];
    stream.write(&buf[..]).await.unwrap();

    client.shutdown().await;

    stream.close().await.unwrap();
}
```
The result was a hang, due to the fact that `shutdown` will wait either for a cleanup future to be registered for the stream or for its sender to be dropped, neither of which can happen if `stream` is used after `shutdown` is called. IIUC this relates to your first bullet point in the overall comment for this PR (that correct ordering is on the user), but I do think that, if a user fumbles this in a situation in which guaranteeing order is somehow difficult, diagnosing a hang like this will be challenging.

I wonder if we can somehow flip the direction of these messages: the stream passes its cleanup future to `register_async_drop` so that the client has a handle to it immediately, `register_async_drop` returns a receiver that will get an ack if `shutdown` is called before `GridFsUploadStream::drop`, and `drop` uses the state of the receiver to determine whether it should do the cleanup itself. Then `shutdown` will always be guaranteed to complete and I *think* the call to `stream.close()` would return an error that the client has already been shut down. (This was just a quick idea so it's very possible I'm missing some complications that would make it infeasible.)

Also, will this drop code be run if a user doesn't call either shutdown method?

src/error.rs
```diff
@@ -624,6 +625,10 @@ pub enum ErrorKind {
     /// A custom value produced by user code.
     #[error("Custom user error")]
     Custom(Arc<dyn Any + Send + Sync>),
+
+    /// A method was called on a client that was shut down.
+    #[error("Client has been shut down")]
+    Shutdown,
```
isabelatkinson, line 631: I don't see this error kind constructed anywhere; should we be returning this somewhere in `execute_operation`?

src/gridfs/upload.rs
abr-egn: Registering the cleanup future at construction with `register_async_drop` runs into the same shared-state issue that my initial approach did - that future will need to co-own some subset of the internal state of the stream along with the `Stream` struct itself, which in turn means access only through a mutex, which in turn means nothing can return a plain `&ref`.

Ideally the behavior would be what you laid out: `shutdown` or `drop` would both check to see if cleanup had already happened, and if not do it immediately.  That's what I spent most of the time on this ticket trying to build, but I haven't been able to find a way to make that work with the Rust ownership model and our current API.

The current state of this PR is my stab at a second-best sort of solution.  It starts with the assumption that ownership of resource state will lie solely with the associated struct, which rules out any kind of pre-emptive cleanup when `shutdown` executes.  Waiting for outstanding resources to be cleaned up at least means we can give a clear guarantee to the user: "When `shutdown()` has completed, everything is tidily closed out."

I agree that a hang is not a great user experience.  My hypothesis is that the "natural" way to use this library is unlikely to run into that, and where the user is *expecting* `shutdown` to guarantee everything is tidied up a hang is at least a clear signal that cleanup isn't happening.  As an escape hatch, `shutdown_immediate` exists for situations where that's not possible.  I've gone back and updated the comment on `shutdown` to be more clear about what's needed there. 

> Also, will this drop code be run if a user doesn't call either shutdown method?

Yup, the cleanup code runs as soon as `AsyncDropToken::spawn` runs (and the task scheduler decides to pick up that task).  `shutdown` doesn't trigger that execution, it just waits for it.

src/error.rs
abr-egn: Whoops, yup, forgot to follow through here.  Fixed.

Cargo.toml
```diff
@@ -86,8 +86,11 @@ in-use-encryption-unstable = ["mongocrypt", "rayon", "num_cpus"]
 # TODO: pending https://github.com/tokio-rs/tracing/issues/2036 stop depending directly on log.
 tracing-unstable = ["tracing", "log"]
 
+internal-track-arc = ["backtrace"]
```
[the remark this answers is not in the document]
abr-egn: Good call, done.

src/test/client.rs
```diff
@@ -845,3 +845,21 @@ async fn retry_commit_txn_check_out() {
         .await
         .expect("should see checked out event");
 }
+
+/// Verifies that `Client::shutdown` succeeds.
+#[cfg_attr(feature = "tokio-runtime", tokio::test)]
+#[cfg_attr(feature = "async-std-runtime", async_std::test)]
+async fn manual_shutdown() {
+    let _guard = LOCK.run_concurrently().await;
+    let client = Client::test_builder().build().await.into_client();
```
isabelatkinson, line 854: Can we add some extra logic here to create/drop the objects which register async drop tokens on the client and make sure shutdown works as intended? And ditto below to ensure that not dropping those things won't cause a hang for `shutdown_immediate`.

Also, this might be overkill, but it could be nice to use command monitoring to make sure that the cleanup stuff actually happens when shutdown is/isn't called; I think we could just listen for a killCursors (for cursor drop) and a delete (for gridfsuploadstream drop).

I trust that everything works now, but I find this async stuff to be tricky to get right so it would be nice to have some sanity checks for any mucking around we do with these implementations in the future 🙂

src/client.rs
```diff
@@ -461,6 +476,81 @@ impl Client {
...
+        AsyncDropToken {
+            tx: Some(cleanup_tx),
+        }
+    }
+
+    /// Shut down this `Client`, terminating background thread workers and closing connections.
+    /// This will wait for any live handles to server-side resources (cursors, sessions) to be
+    /// dropped and any associated server-side operations to finish.  IMPORTANT: this means that
+    /// any live cursor or session values that are not dropped will cause this method to wait
+    /// indefinitely.  It's strongly recommended to structure your usage to avoid this, e.g. by
```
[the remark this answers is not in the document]
isabelatkinson: nice, this comment and the overall comment look great

src/gridfs/upload.rs
isabelatkinson: Ah that makes a lot more sense. I forgot that the task is still executing in the background regardless of whether `shutdown` is called.

Yeah, I think it would be nice to revisit some of this when we can make breaking changes -- can you file a ticket with a 3.0 fix version so we remember to think about it more then?

isabelatkinson (commented):
code changes look good, although it looks like there are some lint/test failures in CI

Cargo.toml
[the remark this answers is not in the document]
abr-egn: It turns out that enabling this for tests by default causes a lot of tests on Windows to time out - apparently collecting backtraces is particularly expensive there.  I've updated this to be behind the raw cfg value `mongodb_internal_tracking_arc` - this isn't a feature that will show up in Cargo, and to invoke it you need the awkward `RUSTFLAGS="--cfg mongodb_internal_tracking_arc" cargo ...` invocation.

isabelatkinson (approved):
lgtm! nice work with this

PR #921, opened 2023-07-24
minor: bump clippy version to `1.71.0`

isabelatkinson:


src/test/spec/connection_stepdown.rs
```diff
@@ -116,10 +115,9 @@ async fn get_more() {
         assert_eq!(client.count_pool_cleared_events(), 0);
     }
 
-    run_test(function_name!(), get_more_test).await;
+    run_test("get_more", get_more_test).await;
```
isabelatkinson, line 118: This line was failing the [`needless_return`](https://rust-lang.github.io/rust-clippy/rust-1.71.0/index.html#needless_return) lint (and ditto below). The error must've been coming from the call to `function_name!` because removing it made the error go away (and the lint isn't relevant here otherwise). External macro calls aren't supposed to cause clippy failures so this must've been an edge case. I'm also not sure why the error didn't pop up in other calls to `function_name!` 🤔

abr-egn (approved):

PR #922, opened 2023-07-24
RUST-1588: Implementing Testing for RunCursorCommand

drshika:


src/client/executor.rs
```diff
@@ -117,7 +117,6 @@ impl Client {
                     }
                     .into());
                 }
-
```
abr-egn, line 120: Any particular reason for the despacing?  I find that putting blank lines between groups of related statements helps when visually skimming.

src/test/spec/unified_runner/operation.rs
```diff
@@ -1556,12 +1559,14 @@ impl TestOperation for CreateCollection {
                         .await
                 })
                 .await?;
+                collection = Entity::Collection(database.collection(&self.collection));
```
abr-egn, line 1562: It looks like this is the same in both `if` branches - could the assignment be moved to just after the if?

abr-egn (commented):
Looks excellent!  Just a few stylistic questions.

abr-egn (approved):
LGTM!

src/operation/run_cursor_command.rs
```diff
@@ -96,24 +99,31 @@ impl<'conn> Operation for RunCursorCommand<'conn> {
...
+            match options.cursor_type {
+                Some(CursorType::Tailable) | Some(CursorType::TailableAwait) => {
+                    match &self.options {
+                        Some(options) => options.max_time,
+                        None => None,
+                    };
+                }
+                _ => max_time = None,
+            };
+        }
```
isabelatkinson, line 121: I think (based on [this test](https://github.com/mongodb/mongo-rust-driver/blob/76cf34921fc23facce470e4aef253bb5d88ae003/src/test/spec/json/run-command/unified/runCursorCommand.json#L575-L576) that I suggested unskipping below) we actually don't need to worry about the user's cursor type here. A cursor type is not configured in that call to `run_cursor_command` and the `max_time` field is still expected to be attached to the calls to getMore.

src/test/spec/unified_runner/operation.rs
```diff
@@ -1644,6 +1649,147 @@ impl TestOperation for RunCommand {
...
+    command: Document,
+    // We don't need to use this field, but it needs to be included during deserialization so that
+    // we can use the deny_unknown_fields tag.
+    #[serde(rename = "commandName")]
+    _command_name: String,
+    selection_criteria: Option<SelectionCriteria>,
+    cursor_type: Option<String>,
+    batch_size: Option<u32>,
+    max_time_m_s: Option<u64>,
+    comment: Option<Bson>,
```
isabelatkinson, line 1664: If you take a look at most of our other operation implementations you'll see we use `#[serde(flatten)]` to deserialize options directly into their respective options structs rather than specifying them at the top-level. This makes it easier to introduce new options in the future. Can we update this implementation to do that instead?

src/serde_util.rs
```diff
@@ -58,6 +58,27 @@ where
     Ok(millis.map(Duration::from_millis))
 }
 
+pub(crate) fn deserialize_cursor_type<'de, D>(
```
isabelatkinson, line 61: This implementation should be equivalent to adding the following tags to the cursor type:
```
#[derive(Deserialize)]
pub enum CursorType { ... }
```
and adding the following tag to the `cursor_type` field on the options:
```
#[serde(default)]
pub cursor_type: Option<CursorType>,
```
`default` here is necessary to deserialize into an optional type properly: if there is a value specified, then the `Deserialize` implementation for `CursorType` will be used, and if not, the deserializer will fall back to the default value of `None`.

src/test/spec/unified_runner/operation.rs
```diff
@@ -1674,25 +1672,21 @@ impl TestOperation for RunCursorCommand {
...
-                _ => None,
-            };
+
+            let max_time = self.options.as_ref().and_then(|opts| opts.max_time);
+            let cursor_type = self.options.as_ref().and_then(|opts| opts.cursor_type);
+            let selection_criteria = self.options.as_ref().and_then(|opts| opts.selection_criteria.clone());
+            let comment = self.options.as_ref().and_then(|opts| opts.comment.clone());
+            let batch_size = dbg!(self.options.as_ref().and_then(|opts| opts.batch_size));
+
             let options = RunCursorCommandOptions::builder()
```
isabelatkinson, line 1682: Rather than reconstructing the options here you should be able to pass in `self.options.clone()` directly where this variable is used instead.

src/test/spec/unified_runner/operation.rs
```diff
@@ -1644,6 +1649,142 @@ impl TestOperation for RunCommand {
...
+    command: Document,
+    // We don't need to use this field, but it needs to be included during deserialization so that
+    // we can use the deny_unknown_fields tag.
+    #[serde(rename = "commandName")]
+    _command_name: String,
+    selection_criteria: Option<SelectionCriteria>,
+    cursor_type: Option<String>,
+    batch_size: Option<u32>,
+    max_time_m_s: Option<u64>,
+    comment: Option<Bson>,
```
isabelatkinson, line 1724: Once the deserialization issues with `RunCursorCommandOptions` are ironed out can we make the equivalent update here?

isabelatkinson:
some of the lint failures might be fixed by rebasing, I'd recommend doing that and then re-running CI

PR #925, opened 2023-07-31
RUST-1638 Fix duplicate CMAP logging test descriptions

isabelatkinson:


isabelatkinson:
Closing this out as pulling in the changes for `connection-logging.json` post-https://github.com/mongodb/specifications/pull/1374 results in errors due to the bug described in RUST-1437. The work for RUST-1638 itself is very trivial, so we can hold off on syncing `connection-logging.json` until we do RUST-1437.

PR #927, opened 2023-08-03
Update README.md with update

adrien3d:


kevinAlbs:
@adrien3d is this intended for review? If so, I suggest moving this from Draft to Open.

PR #928, opened 2023-08-05
RUST-1719 Fix panic after `SessionCursor::with_type` is called

mlokr:
Hello there,

when using the transaction API, I encountered unconditional panic and also panic while dropping in `SessionCursor::with_type`. The problem occurs because the `.state` is moved into new cursor and subsequently, `.unwrap()` is called on it when marking it exhausted and also when checking for exhaustion. This is the minimal code change to fix this problem

### DIFF
--- a/src/cursor/session.rs
+++ b/src/cursor/session.rs
@@ -350,7 +350,9 @@ impl<T> SessionCursor<T> {
 
 impl<T> SessionCursor<T> {
     fn mark_exhausted(&mut self) {
-        self.state.as_mut().unwrap().exhausted = true;
+        if let Some(state) = self.state.as_mut() {
+            state.exhausted = true;
+        }
     }
 
     pub(crate) fn is_exhausted(&self) -> bool {
--- a/src/cursor/session.rs
+++ b/src/cursor/session.rs
@@ -354,7 +354,7 @@ impl<T> SessionCursor<T> {
     }
 
     pub(crate) fn is_exhausted(&self) -> bool {
-        self.state.as_ref().unwrap().exhausted
+        self.state.as_ref().map_or(true, |state| state.exhausted)
     }
 
     #[cfg(test)]
--- a/src/cursor/session.rs
+++ b/src/cursor/session.rs
@@ -321,7 +321,6 @@ impl<T> SessionCursor<T> {
             #[cfg(test)]
             kill_watcher: self.kill_watcher.take(),
         };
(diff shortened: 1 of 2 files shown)


isabelatkinson:
Hi @mlokr thank you for reporting this issue! Would you mind providing a reproduction of the code that caused this panic and let us know what version of the driver you're using? I will look into this in the meantime.

src/cursor/session.rs
```diff
@@ -347,11 +347,13 @@ impl<T> SessionCursor<T> {
 
 impl<T> SessionCursor<T> {
     fn mark_exhausted(&mut self) {
-        self.state.as_mut().unwrap().exhausted = true;
+        if let Some(state) = self.state.as_mut() {
+            state.exhausted = true;
+        }
```
isabelatkinson, line 352: I think we can just remove the call to `mark_exhausted` in `with_type` (since it'll be a no-op with this change) and keep this method as-is.

isabelatkinson (changes requested):
Update: I was able to reproduce this issue with the following test:
```
async fn session_cursor_with_type() {
    let _guard: RwLockReadGuard<()> = LOCK.run_concurrently().await;
    let client = TestClient::new().await;

    let mut session = client.start_session(None).await.unwrap();
    let coll = client.database("db").collection("coll");
    coll.drop_with_session(None, &mut session).await.unwrap();

    coll.insert_many_with_session(
        vec![doc! { "x": 1 }, doc! { "x": 2 }, doc! { "x": 3 }],
        None,
        &mut session,
    )
    .await
    .unwrap();

    let mut cursor: crate::SessionCursor<bson::Document> = coll
        .find_with_session(doc! {}, None, &mut session)
        .await
        .unwrap();

    let _ = cursor.next(&mut session).await.unwrap().unwrap();

    let mut cursor_with_type: crate::SessionCursor<bson::RawDocumentBuf> = cursor.with_type();

    let _ = cursor_with_type.next(&mut session).await.unwrap().unwrap();
}

```
Would you mind adding a test to verify the changes here fix this issue (feel free to paste in the one above)?

Otherwise I just have one small comment!

mlokr:
@isabelatkinson I added the test and made other requested changes, the newly added test passes for me. Thanks for providing it!

isabelatkinson (approved):
Hey @mlokr, I rebased this onto main and fixed a small clippy failure. The changes look great! I am going to tag in the other member of the team for review; once he approves we'll merge this in and put out a patch release. Thanks again for bringing this bug to our attention!

PR #929, opened 2023-08-09
RUST-1674 Use cargo-nextest as test runner

abr-egn:
RUST-1674

This switches to using `cargo-nextest`, with the immediate benefit of removing the need for the unstable output options and corresponding Rust version pinning.  Because `cargo-nextest` runs each test in its own process, our existing test lock no longer works; as a workaround, I've set `cargo-nextest` to only run one test at a time.  Future work for which I've filed tickets:
* RUST-1721 Use test profiles to regain parallelism
* RUST-1722 Remove now-defunct test locks
* RUST-1723 Take advantage of cargo-nextest flaky test detection

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,9 @@
+[profile.default]
+retries = 1
+
+[profile.ci]
+failure-output = "final"
+test-threads = 1
+
+[profile.ci.junit]
+path = "junit.xml"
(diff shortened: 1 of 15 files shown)


.config/nextest.toml
```diff
@@ -0,0 +1,9 @@
+[profile.default]
+retries = 1
+
+[profile.ci]
```
isabelatkinson, line 4: Can we also set the `no-fail-fast` option here? Otherwise I think the tests will terminate once a failure is encountered.

isabelatkinson (commented):
just one suggestion, otherwise looks great! I closed out RUST-1280 and RUST-1546 as well.

PR #932, opened 2023-08-14
RUST-1712 Provide a connection pool warmup method

abr-egn:
RUST-1712

This works by spawning a task that sends check out requests in a loop until the receiving worker reports that the pool has hit `min_pool_size`.  This is a little indirect but seems to have the fewest drawbacks compared to other methods I considered:
* because they're mostly-normal checkout requests, any constraints or other handling of those continues to apply
* allows the worker to continue to process other requests while the pool is warming; initially I was going to implement this loop directly in the worker but that would block requests during the warmup.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -619,7 +619,7 @@ impl Client {
     /// 
     /// Note that topology changes require rebuilding the connection pool, so this method cannot 
     /// guarantee that the pool will always be filled for the lifetime of the `Client`.
-    pub async fn warm_connection_pool(self) {
+    pub async fn warm_connection_pool(&self) {
         if !self.inner.options.min_pool_size.map_or(false, |s| s > 0) {
             // No-op when min_pool_size is zero.
             return;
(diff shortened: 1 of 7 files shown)


src/client.rs
```diff
@@ -612,6 +612,22 @@ impl Client {
...
 
+    /// Add connections to the connection pool up to `min_pool_size`.  This is normally not needed -
+    /// the connection pool will be filled in the background, and new connections created as needed
+    /// up to `max_pool_size`.  However, it can sometimes be preferable to pay the (larger) cost of
+    /// creating new connections up-front so that individual operations execute as quickly as
+    /// possible.
+    ///
+    /// Note that topology changes require rebuilding the connection pool, so this method cannot
+    /// guarantee that the pool will always be filled for the lifetime of the `Client`.
+    pub async fn warm_connection_pool(&self) {
```
abr-egn, line 623: I'm open to naming suggestions here.  "warm" seems a little vague but I couldn't think of a better one that wasn't super verbose.

isabelatkinson: yeah this seems fine to me. can you note in the comment for this method that nothing will happen if `min_pool_size` is 0?

isabelatkinson (approved):
one small comment, no need for re-review!

PR #936, opened 2023-08-22
RUST-1588 Implement `IterateOnce` unified test runner operation

isabelatkinson:
This PR implements the `IterateOnce` unified test operation and unskips the "does not close the cursor when receiving an empty batch" `run_cursor_command` test. This operation required adding a test-only `try_advance` method to `Cursor` and `SessionCursor`, which only performs one getMore rather than looping until the buffer has been refilled. I considered adding this method to the public API, but chose not to for a few reasons:

- No users have asked for this functionality.
- Depending upon what we decided the method should return, we might need to add a new type (`AdvanceResult`) to our API as well.
- The existence of `advance` and `try_advance` might be confusing to users.

If we do receive requests for this functionality we can further discuss API and consider making it public then.

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -208,6 +208,13 @@ impl<T> Cursor<T> {
         self.wrapped_cursor.as_mut().unwrap().advance().await
     }
 
+    #[cfg(test)]
+    #[allow(dead_code)]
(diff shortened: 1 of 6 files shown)


abr-egn (approved):
LGTM, thank you!  One very minor comment only, no need for re-review.

PR #939, opened 2023-08-24
RUST-1539 Restructure and refine contents of Evergreen configuration

isabelatkinson:
### Patch vs. Commit Builds
Several build variants have `patchable: false` configured (currently commented out for the sake of development), which means that they will not run on patch builds. These are tasks that a) represent less common use cases and/or b) are testing code paths that we don't touch during normal development. We will need to monitor the waterfall more closely to make sure the `patchable: false` tasks don't break and perhaps set up some kind of automation to alert us if something fails.

### Formatting
I ran `shfmt -i 2` on all of our shell scripts. When reviewing, it will be useful to hide whitespace changes to avoid the noise from this.

### Omitted Work
The original ticket for this work mentioned using generate.tasks. I decided against this as the existing tools to generate the JSON file used by generate.tasks use Go and Python, languages which I am not (and wouldn't necessarilly expect current/future Rust team members to be) proficient in. Writing a similar Rust tool would be an interesting project (maybe for next Skunkworks), but wasn't in scope for this project.

Some other notable omitted work (not done for the sake of time):
- Convert all uses of shell.exec to subprocess.exec (RUST-1728)
- Use the include feature to break up the configuration file (RUST-1740)
- Remove all uses of the legacy shell (RUST-1525; blocked, but has some work ready to be done)


.evergreen/config.yml
```diff
@@ -1,2116 +1,1945 @@
...
+    run_on:
+      - rhel87-small
+    expansions:
+      AUTH: auth
+      SSL: ssl
+    tasks:
+      - test-plain-auth
+
+  - name: serverless
+    # TODO: Enable this build variant when the serverless issues are fixed.
```
abr-egn, line 330: I don't suppose we have an issue to follow for this?

.evergreen/config.yml
```diff
@@ -1,2116 +1,1945 @@
...
+      - func: "windows fix"
+      - func: "fix absolute paths"
+      - func: "init test-results"
+      - func: "make files executable"
+      - func: "install rust"
+      - func: "install libmongocrypt"
+      - command: shell.exec
+        params:
+          shell: bash
+          script: |-
```
abr-egn, line 469: For my own education, what's the difference between `script: |` and `script: |-` ?

abr-egn (commented):
This is *so much* better, thank you!

kevinAlbs (approved):

.evergreen/config.yml
kevinAlbs: > For my own education, what's the difference between `script: |` and `script: |-` ?

From https://yaml.org/spec/1.2.2/#8112-block-chomping-indicator: 

> Stripping is specified by the “-” chomping indicator. In this case, the final [line break](https://yaml.org/spec/1.2.2/#line-break-characters) and any trailing [empty lines](https://yaml.org/spec/1.2.2/#empty-lines) are excluded from the [scalar’s content](https://yaml.org/spec/1.2.2/#scalar).

https://yaml-multiline.info has a neat interactive explainer.

abr-egn (approved):
LGTM!

PR #941, opened 2023-08-28
RUST-1743 Fix gridfs numeric type serialization

abr-egn:
RUST-1743

This fixes the ``Chunk.n`` and ``FilesCollectionDocument.chunk_size_bytes`` to serialize as ``i32`` rather than ``i64``.  Happily, deserialization will accept either, so this is backwards-compatible.

### DIFF
--- a/src/gridfs.rs
+++ b/src/gridfs.rs
@@ -31,9 +31,17 @@ pub(crate) struct Chunk<'a> {
     #[serde(rename = "_id")]
     id: ObjectId,
     files_id: Bson,
+    #[serde(serialize_with = "u32_as_i32")]
     n: u32,
     #[serde(borrow)]
     data: RawBinaryRef<'a>,
+}
+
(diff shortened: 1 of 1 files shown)


src/gridfs.rs
```diff
@@ -51,7 +59,7 @@ pub struct FilesCollectionDocument {
     pub length: u64,
 
     /// The size of the file's chunks in bytes.
-    #[serde(rename = "chunkSize")]
+    #[serde(rename = "chunkSize", serialize_with = "u32_as_i32")]
```
isabelatkinson, line 62: I believe you should be able to use [this method](https://docs.rs/bson/latest/bson/serde_helpers/fn.serialize_u32_as_i32.html) from `bson::serde_helpers` here

isabelatkinson (approved):
lgtm, just one small suggestion!

src/gridfs.rs
abr-egn: Entirely forgot we had those, fixed :)

isabelatkinson (approved):
lgtm, thanks for the quick fix!

PR #942, opened 2023-08-28
RUST-877 Delay replacement document serialization until `Operation::build`

isabelatkinson:


src/operation/update/test.rs
isabelatkinson, line 1: ditto above

src/bson_util.rs
```diff
@@ -119,6 +125,17 @@ pub(crate) fn read_document_bytes<R: Read>(mut reader: R) -> Result<Vec<u8>> {
     Ok(bytes)
 }
 
+pub(crate) fn extend_raw_document_buf(
```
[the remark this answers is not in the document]
abr-egn: Agreed, filed RUST-1754 for this.

src/coll/options.rs
```diff
@@ -74,12 +74,20 @@ pub enum Hint {
...
 impl Hint {
+    #[cfg(test)]
     pub(crate) fn to_bson(&self) -> Bson {
         match self {
             Hint::Keys(ref d) => Bson::Document(d.clone()),
             Hint::Name(ref s) => Bson::String(s.clone()),
         }
     }
+
+    pub(crate) fn to_raw_bson(&self) -> Result<RawBson> {
```
abr-egn, line 85: It seems like this could just be `to_bson().try_into()?` ?

src/operation/find_and_modify.rs
```diff
@@ -21,89 +19,86 @@ use crate::{
...
 
-pub(crate) struct FindAndModify<T = Document>
-where
-    T: DeserializeOwned,
-{
+pub(crate) struct FindAndModify<'a, R, T: DeserializeOwned> {
     ns: Namespace,
     query: Document,
+    modification: Modification<'a, R>,
+    human_readable_serialization: Option<bool>,
```
abr-egn, line 36: Could this be an un-serialized field in `FindAndModifyOptions` rather than adding it as an explicit parameter?

src/coll/options.rs
isabelatkinson: I added this as a minor optimization to avoid an intermediary allocation of a regular `Bson` via `to_bson`. It did seem unnecessary to have both `to_bson` and `to_raw_bson` though so I just refactored the few uses of `to_bson` and removed it.

Ideally we'd have a `to_raw_bson(value: impl Serialize)` method in bson, but I'm not sure how that would actually work; the raw serializer spits out raw document bytes as opposed to the non-raw serializer, which generates BSON values.

abr-egn (approved):
LGTM!

PR #948, opened 2023-08-31
fixing the find or delete serde inconsistency

mlokr:
Hello,

I found an inconsistency in the way mongo documents are deserialized when using `serde(with)`  and `find_one_and_delete`. My fix might nor be ideal because it breaks
```
< operation::find_and_modify::test::handle_no_value_delete                                                                                                                                                                                    
< operation::find_and_modify::test::handle_no_value_replace                      
< operation::find_and_modify::test::handle_no_value_update
```
The tests no longer return `Err(_)` but just `Ok(None)`. I am not sure how I can assert that Field is present but can be null, when deserializing `Option<T>` and could not find anything either.

I added a tests to verify my changes fix the issue (see changes)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -195,6 +195,7 @@ time = "0.3.9"
 tokio = { version = ">= 0.0.0", features = ["fs"] }
 tracing-subscriber = "0.3.16"
 regex = "1.6.0"
+serde-hex = "0.1.0"
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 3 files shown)


abr-egn:
I couldn't find a way to distinguish between `null` and "missing" in autogenerated Serde impls either; fortunately, we can just swap the `Bson` for a `RawBson` and check directly :)

abr-egn (approved):

isabelatkinson (approved):

PR #949, opened 2023-09-05
RUST-1698 Use standardized os and db version for benchmarks

abr-egn:
RUST-1698

Some of the benchmark runs fail, and have been failing for quite some time; fix for that to come in a follow-up PR.

### DIFF
--- a/.evergreen/benchmarks.yml
+++ b/.evergreen/benchmarks.yml
@@ -132,6 +132,7 @@ functions:
             AUTH=${AUTH} \
             SSL=${SSL} \
             REQUIRE_API_VERSION=${REQUIRE_API_VERSION} \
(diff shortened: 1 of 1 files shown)


isabelatkinson (commented):
changes look good. I believe we discussed moving the benchmarks to our main evergreen project. did you decide against this/to do it in a separate PR?

abr-egn:
> changes look good. I believe we discussed moving the benchmarks to our main evergreen project. did you decide against this/to do it in a separate PR?

Ah, sorry, should have mentioned - with the other stuff around benchmarks I think it's better to defer that.  Filed RUST-1759 to track.

isabelatkinson (approved):

PR #951, opened 2023-09-06
RUST-1757 Fix final cursor batch handling

abr-egn:
RUST-1757

Looks like this broke as part of a refactor a few weeks ago, so the only release it's live in is 2.7.0-beta :)

### DIFF
--- a/src/cursor/common.rs
+++ b/src/cursor/common.rs
@@ -135,12 +135,14 @@ where
         let result = self.provider.execute(spec, client, pin).await;
         self.handle_get_more_result(result)?;
 
-        if self.is_exhausted() {
-            Ok(AdvanceResult::Exhausted)
-        } else {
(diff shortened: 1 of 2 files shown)


isabelatkinson (commented):
good catch! this seems like something we should have test coverage of outside of the benchmarks, would you mind adding a repro of what broke so it doesn't happen again?

abr-egn:
> good catch! this seems like something we should have test coverage of outside of the benchmarks, would you mind adding a repro of what broke so it doesn't happen again?

Added a test :)  Was a little harder to poke this behavior than I thought until I realized that `advance` and the `Stream` impl don't share logic here - I think this is also why this didn't break any spec tests, as those all use `Stream` methods.  I filed RUST-1761 to merge these.

isabelatkinson (approved):

PR #952, opened 2023-09-07
add upsert option to employee update

tsondru:


abr-egn:
Hi!

Thanks for the contribution!  However, this example needs to stay equivalent to the examples provided for [other languages](https://www.mongodb.com/docs/manual/core/transactions-in-applications/#core-api), which doesn't include this particular option.

PR #957, opened 2023-09-18
RUST-1763 deprecate CollectionOptions::human_readable_serialization

abr-egn:


isabelatkinson (approved):
code changes look good, can you put the ticket I filed for this in the commit message (RUST-1763)?

CryZe:
I'm having a tough time figuring out why this was deprecated. What's the alternative way to do this then?

PR #964, opened 2023-09-27
minor: pin async-executor

abr-egn:


abr-egn:
See https://jira.mongodb.org/browse/RUST-1768 for information.

PR #968, opened 2023-09-28
RUST-935 direct retries to different mongos

abr-egn:
RUST-935

I opted to introduce the `SelectionRetry` wrapper so that the actual details of what it is and how it's used only need to be known within `attempt_to_select_server` - callers just need to preserve it and pass it back in.  I did consider making this part of `SelectionCriteria`, but that would have been a pretty major refactoring since that would have to move from the current `enum` to a `struct` with a "kind" enum.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -52,7 +52,11 @@ use crate::{
         SessionOptions,
     },
     results::DatabaseSpecification,
-    sdam::{server_selection, SelectedServer, Topology},
+    sdam::{
+        server_selection::{self, SelectionRetry},
+        SelectedServer,
+        Topology,
+    },
     tracking_arc::TrackingArc,
     ClientSession,
 };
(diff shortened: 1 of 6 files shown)


isabelatkinson:
> I opted to introduce the `SelectionRetry` wrapper so that the actual details of what it is and how it's used only need to be known within `attempt_to_select_server`

just personal preference, but I think the indirection here makes it a little less clear what's going on in the executor; I would probably prefer including the field directly or naming the wrapper something a bit more specific, but up to you 🙂 

looks like there are also some prose tests that need to be implemented as part of this ticket: https://github.com/mongodb/specifications/commit/86d961fee4c5e92fbcff76a62abe1aea3fafd451

abr-egn:
> just personal preference, but I think the indirection here makes it a little less clear what's going on in the executor; I would probably prefer including the field directly or naming the wrapper something a bit more specific, but up to you 🙂

Yeah, good call.  I've got a bee in my bonnet about trying to keep reasoning as local as possible, but that didn't work out here.

> looks like there are also some prose tests that need to be implemented as part of this ticket: [mongodb/specifications@86d961f](https://github.com/mongodb/specifications/commit/86d961fee4c5e92fbcff76a62abe1aea3fafd451)

Derp.  Added.

isabelatkinson (approved):

PR #974, opened 2023-10-11
RUST-1654 Enable QEv2 tests on serverless

abr-egn:
RUST-1654

This mostly just required shuffling around some evergreen config, which also ended up implementing RUST-1754.


.evergreen/config.yml
```diff
@@ -324,12 +324,14 @@ buildvariants:
       - test-plain-auth
 
   - name: serverless
-    # TODO RUST-1745: Enable this build variant when the serverless issues are fixed.
-    disable: true
-    patchable: false
+    #patchable: false
```
abr-egn, line 327: I'll revert this line before submitting.

.evergreen/config.yml
```diff
@@ -381,16 +383,6 @@ buildvariants:
     tasks:
       - test-in-use-encryption-openssl
 
-  - name: in-use-encryption-serverless
```
isabelatkinson, line 384: Would keeping this variant separate have any impact on the timeout problems?

isabelatkinson: makes sense!

PR #977, opened 2023-10-19
RUST-933 Add support for the srvMaxHosts option

abr-egn:
RUST-933

Pretty straightforward.  The shuffling needs to be in two places (initial host discovery and polling), unfortunately, because the logic at those two places is a bit different.

### DIFF
--- a/src/client/options.rs
+++ b/src/client/options.rs
@@ -917,12 +917,19 @@ impl Default for HostInfo {
 }
 
 impl HostInfo {
-    async fn resolve(self, resolver_config: Option<ResolverConfig>) -> Result<ResolvedHostInfo> {
+    async fn resolve(
+        self,
+        resolver_config: Option<ResolverConfig>,
+        srv_max_hosts: Option<u32>,
+    ) -> Result<ResolvedHostInfo> {
         Ok(match self {
             Self::HostIdentifiers(hosts) => ResolvedHostInfo::HostIdentifiers(hosts),
             Self::DnsRecord(hostname) => {
                 let mut resolver =
-                    SrvResolver::new(resolver_config.clone().map(|config| config.inner)).await?;
+                    SrvResolver::new(
+                        resolver_config.clone().map(|config| config.inner),
+                        srv_max_hosts,
+                    ).await?;
                 let config = resolver.resolve_client_options(&hostname).await?;
                 ResolvedHostInfo::DnsRecord { hostname, config }
             }
@@ -1217,7 +1224,7 @@ impl ClientOptions {
         let mut options = Self::from_connection_string(conn_str);
         options.resolver_config = resolver_config.clone();
 
-        let resolved = host_info.resolve(resolver_config).await?;
+        let resolved = host_info.resolve(resolver_config, options.srv_max_hosts).await?;
         options.hosts = match resolved {
             ResolvedHostInfo::HostIdentifiers(hosts) => hosts,
             ResolvedHostInfo::DnsRecord {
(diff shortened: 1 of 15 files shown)


src/client/options.rs
```diff
@@ -1721,6 +1750,26 @@ impl ConnectionString {
             ConnectionStringParts::default()
         };
 
+        if let Some(srv_max_hosts) = conn_str.srv_max_hosts {
```
abr-egn, line 1753: Similarly, the spec says that the validation here _has_ to happen both for the initial URI parse and after SRV resolution, so that gets duplicated too :/

src/srv.rs
```diff
@@ -24,7 +24,7 @@ pub(crate) struct ResolvedConfig {
 
 #[derive(Debug, Clone)]
 pub(crate) struct LookupHosts {
-    pub(crate) hosts: Vec<Result<ServerAddress>>,
```
abr-egn, line 27: This being a `Vec<Result<ServerAddress>>` rather than `Vec<ServerAddress>` with error values either being propagated or ignored depending on context made me stumble a bit when reading this code, so I switched it to the (IMO) more understandable simple value vec with an explicit behavior toggle.

src/srv.rs
```diff
@@ -103,15 +109,19 @@ impl SrvResolver {
...
+                             name {}",
+                            original_hostname,
+                            address,
+                            domain_name.join(".")
+                        ),
+                    }
+                    .into());
                 }
-                .into()));
+                continue;
```
abr-egn, line 124: Previously there was no `continue` here, which was kind of a nasty bug-in-waiting - invalid hosts would still have been added to the list.

src/srv.rs
```diff
@@ -103,15 +109,19 @@ impl SrvResolver {
...
 
             if !&hostname_parts[1..].ends_with(domain_name) {
-                srv_addresses.push(Err(ErrorKind::DnsResolve {
-                    message: format!(
-                        "SRV lookup for {} returned result {}, which does not match domain name {}",
-                        original_hostname,
-                        address,
-                        domain_name.join(".")
-                    ),
+                if matches!(dm, DomainMismatch::Error) {
```
isabelatkinson, line 112: Based on the language in the spec (see the language in the second bullet [here](https://github.com/mongodb/specifications/blob/master/source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst#implementation)), it seems like we should never actually return an error here and instead log a warning. I think these errors were just being constructed for the sake of logging them at some point in the future? But I think it would be simpler to do the logging here directly rather than propagate the error. Based on that I think we can just remove this logic entirely and file a ticket to log a warning when this happens.

src/srv.rs
isabelatkinson: good catch!

src/client/options.rs
```diff
@@ -1252,6 +1262,24 @@ impl ClientOptions {
...
+                            ));
+                        }
+                        if options.load_balanced == Some(true) {
+                            return Err(Error::invalid_argument(
+                                "srvMaxHosts and loadBalanced=true cannot both be present",
+                            ));
+                        }
+                        use rand::prelude::SliceRandom;
+                        config.hosts.shuffle(&mut rand::thread_rng());
+                        config.hosts.truncate(max as usize);
```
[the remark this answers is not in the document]
abr-egn: Done.  And yeah, AFAICT rand's implementation of `shuffle` does actually use fisher-yates, but `choose_multiple` doesn't, but I think the idea is to make sure it's a valid and efficient selection, for which I'm happy to trust rand.

isabelatkinson (approved):

PR #980, opened 2023-10-23
RUST-1231 Run tests using `DEFAULT_MAX_POOL_SIZE`

stIncMale:
The tasks [`compile-on-msrv`](https://spruce.mongodb.com/task/mongo_rust_driver_compile_compile_on_msrv_patch_135c9f0a45275f638a42dfa71a560cdeafec7fdc_6536d372850e612d3d22e1a2_23_10_23_20_11_31?execution=0) and [`check-cargo-deny`](https://spruce.mongodb.com/task/mongo_rust_driver_lint_check_cargo_deny_patch_135c9f0a45275f638a42dfa71a560cdeafec7fdc_6536d372850e612d3d22e1a2_23_10_23_20_11_31?execution=0) fail because of RUST-1780.

RUST-1231


abr-egn:
Swapped Patrick out for Isabel since he's no longer on the Rust driver team :)

abr-egn (approved):
LGTM!

stIncMale:
Merged `main` to make all tests pass.

src/test.rs
```diff
@@ -114,7 +113,7 @@ lazy_static! {
 
 pub(crate) fn update_options_for_testing(options: &mut ClientOptions) {
     if options.max_pool_size.is_none() {
-        options.max_pool_size = Some(MAX_POOL_SIZE);
+        options.max_pool_size = Some(DEFAULT_MAX_POOL_SIZE);
```
isabelatkinson, line 116: I think we can remove this logic entirely since the driver will already apply this default internally.

src/sdam/description/topology/server_selection/test/in_window.rs
```diff
@@ -216,7 +217,7 @@ async fn load_balancing_test() {
     let mut handler = EventHandler::new();
     let mut subscriber = handler.subscribe();
     let mut options = CLIENT_OPTIONS.get().await.clone();
-    let max_pool_size = 10;
+    let max_pool_size = DEFAULT_MAX_POOL_SIZE;
```
isabelatkinson, line 220: ditto here, I think we just want min/max pool size to be the same, so line 224 can just assign to `Some(DEFAULT_MAX_POOL_SIZE)`

src/test.rs
stIncMale: Indeed! Done.

src/sdam/description/topology/server_selection/test/in_window.rs
stIncMale: I removed `options.max_pool_size = Some(max_pool_size)`, but the `max_pool_size` variable is still needed nonetheless (it seems better to use it in multiple places in the method, rather than using `DEFAULT_MAX_POOL_SIZE` in those same places, and it is how it was done originally).

isabelatkinson (approved):

stIncMale:
@isabelatkinson Could you please also merge the PR? Asking because I don't have the permission to do that.

PR #981, opened 2023-10-24
RUST-1780 Bump MSRV to 1.61.0, upgrade `ahash` to 0.8.5 in `MSRV-Cargo.lock`

stIncMale:
`ahash` 0.8.3 and older were yanked because of a security vulnerability, and 0.8.4 (with a bug) and 0.8.5 were released to address that. `ahash` 0.8.5 requires Rust 1.61+. As a result, the Evergreen task for `mongo-rust-driver` that checks MSRV compilation started to fail.

There were also a change in the `zerocopy` license, so `deny.toml` had to be updated.


docs/manual/searchindex.js
stIncMale, line 1: Here and in `searchindex.json` only the MSRV was updated to 1.61.0. But because both files are huge one-liners, the change is as rendered by GitHub is not reviewable. How do you usually update the [manual](https://mongodb.github.io/mongo-rust-driver/manual/)? Maybe you build it from `./manual`, and then replace the whole `./docs/manual` in a separate commit?

abr-egn: Generally we regenerate the contents of `docs/manual` in the same PR that updates `manual/src`; the generated files don't need to be reviewed so it's not an issue that github doesn't render them.

As a side note, we'll likely be dropping this manual entirely in the nearish future since we have official docs now :)

abr-egn (approved):
LGTM!

docs/manual/searchindex.js
stIncMale: Thanks! Since this PR updates `manual/src/README.md`, I built the book and committed the result.

isabelatkinson (approved):

PR #986, opened 2023-11-01
minor: utility to update version numbers for release

abr-egn:
This adds a simple utility to keep track of the various places version numbers need to be updated so it's a single command rather than a manual search and replace.  This can be used both for the branch update for release (updating the crate version and the two git dependencies to point to crates.io)
```
mongo-rust-driver/etc/update_version$ cargo run -- --version 2.8.0 --bson 2.7.0 --mongocrypt 0.2
```
and the mainline update post-release (just updating the crate version)
```
mongo-rust-driver/etc/update_version$ cargo run -- --version 2.8.0
```

I wrote this to be robust in various ways:
* the regexes match against specific chunks of the file rather than just "things that look like a version number"
* any of the regexes not matching is a fatal error
* no file changes are written until all updates have succeeded

If this seems reasonable I'll update the release instructions to use it.


isabelatkinson (approved):
lgtm! let's add this to the release process

PR #989, opened 2023-11-08
RUST-1667 Add search index management helpers

abr-egn:
RUST-1667

Nothing particularly unusual.  New commands, new wrappers for them, new unified tests.  The only thing of note is that the commands are proxied server-side to a different process, so stuff like write concerns and sessions don't apply.


src/search_index.rs
```diff
@@ -0,0 +1,130 @@
...
+        self.client().execute_operation(op, None).await
+    }
+
+    /// Gets index information for one or more search indexes in the collection.
+    ///
+    /// If name is not specified, information for all indexes on the specified collection will be
+    /// returned.
+    pub async fn list_search_indexes(
+        &self,
+        name: impl Into<Option<&str>>,
```
isabelatkinson, line 70: It's a bummer that nested `impl`s aren't allowed, because I think this type would ideally be something like `impl Into<Option<impl AsRef<str>>>`. With this type definition, passing in the `name` field of a `SearchIndexModel` will require `model.name.as_deref()` as `Option<String>` don't pass the `impl` test. An outright `String` requires `s.as_str()`.

I wonder if it would be more ergonomic to make this type `Option<impl AsRef<str>>`. That would allow passing `model.name` directly, but users would need to wrap a non-optional `String` or `&str` in `Some`. I'm not sure which use case is better to optimize for, WDYT?

src/coll.rs
```diff
@@ -183,6 +183,16 @@ impl<T> Collection<T> {
         }
     }
 
+    pub(crate) fn clone_unconcerned(&self) -> Self {
```
isabelatkinson, line 186: Is this necessary? IIRC these options are propagated from the client when a db/coll is instantiated and then resolved with the `resolve_options!` macro, so I think they will only get applied if they're explicitly folded into the options provided to a method (which we're not doing in these helpers)

src/test/spec/unified_runner/operation/search_index.rs
```diff
@@ -0,0 +1,151 @@
+use bson::{to_bson, Bson, Document};
```
isabelatkinson, line 1: good call on moving these to their own file, `operation.rs` is huge and could probably use some modularization

src/search_index.rs
```diff
@@ -0,0 +1,130 @@
...
+                }],
+                aggregation_options,
+            )
+            .await
+    }
+}
+
+/// Specifies the options for a search index.
+#[derive(Debug, Clone, Default, TypedBuilder, Serialize, Deserialize)]
+#[non_exhaustive]
```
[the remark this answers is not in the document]
abr-egn: Good catch, done.

src/test/spec/unified_runner/operation/search_index.rs
abr-egn: Filed RUST-1800 for this.

src/coll.rs
abr-egn: It's needed for `list_search_indexes`, which is implemented as an `aggregate` call.

src/search_index.rs
abr-egn: I went down a rabbit hole here - I originally tried having both types be full parameters with `where` clauses, but that doesn't work either because it requires annotations for the `None` case.  It seemed like providing a default for the inner type would be the way to go, but that's not supported on functions because the semantics are actually [less obvious](https://github.com/rust-lang/rust/issues/36887#issuecomment-342667728) than it seems.

I'd lean towards keeping the `Into<Option<..>>` for consistency with the rest of the API (including the other parameters of this `fn`).

src/search_index.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Gets index information for one or more search indexes in the collection.
+    ///
+    /// If name is not specified, information for all indexes on the specified collection will be
+    /// returned.
+    pub async fn list_search_indexes(
+        &self,
+        name: impl Into<Option<&str>>,
+        aggregation_options: impl Into<Option<AggregateOptions>>,
+        _list_index_options: impl Into<Option<ListSearchIndexOptions>>,
+    ) -> Result<Cursor<Document>> {
```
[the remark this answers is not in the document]
abr-egn: The spec doesn't actually say, it just gives the return as `Document`.  Looking at other drivers, C++ and Go don't even have support for a type-parameterized cursor; Java provides both a `Cursor<Document>` and a `Cursor<T>` but no specific concrete type.  Given that, I'd like to leave it at the more general.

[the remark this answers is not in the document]
isabelatkinson: Oh interesting, I was going to say that I would prefer consistency with our existing API but that's not much of a concern if the return types are so different. Sticking with `Document` SGTM.

isabelatkinson (approved):

PR #992, opened 2023-11-24
Replace `async_once` with `tokio::sync::OnceCell`

Expyron:
`async_once` is unmaintained and [has been reported unsound](https://github.com/hjiayz/async_once/issues/5).

`tokio`'s `parking_lot` feature has been added in order to have access to `OnceCell::const_new()`. This is only necessary in older `tokio` versions.

### DIFF
--- a/.evergreen/MSRV-Cargo.lock
+++ b/.evergreen/MSRV-Cargo.lock
@@ -244,12 +244,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async_once"
-version = "0.2.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2ce4f10ea3abcd6617873bae9f91d1c5332b4a778bd9ce34d0cd517474c1de82"
-
-[[package]]
 name = "atomic-waker"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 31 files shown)


abr-egn:
Thank you for contributing this!  I've authorized an evergreen run and assuming that passes I'll merge this in.

abr-egn:
It looks like this doesn't compile with the `sync` feature enabled, and also needs a pass through `rustfmt +nightly --unstable-features`.  Once those are fixed I can merge this :)

Expyron:
I pushed an update:
- Ran `rustfmt`
- I'm not sure what happened with the `sync` feature, but I don't think it was the changes in that PR. I tried to fix it by updating the `log` dependency, which was what caused an issue on my machine. If it fails again, can you share the error message?

abr-egn:
Sorry, forgot that evergreen runs aren't publicly visible!  Running `cargo check --tests --features tokio-sync` will reproduce the errors:
```
error[E0618]: expected function, found `get_client_options`
  --> src/sync/test.rs:64:19
   |
39 | / lazy_static! {
40 | |     static ref get_client_options: ClientOptions =
41 | |         runtime::block_on(async { crate::test::get_client_options().await.clone() });
42 | | }
   | |_- `sync::test::get_client_options` defined here
...
64 |       let options = get_client_options().clone();
   |                     ^^^^^^^^^^^^^^^^^^--
   |                     |
   |                     call expression requires function

... etc ...
```

Expyron:
Pushed an update fixing the sync tests

abr-egn (approved):
LGTM!  Tagging in Isabel.

PR #1001, opened 2023-12-21
RUST-1512 Preview of helper method conversion

abr-egn:
RUST-1512

I don't intend to merge this yet, but I'd like a review since the process of converting `list_databases`/`list_database_names` has turned up a bunch of points of potential conversation for the general shape of the conversion.  I'll call out the specific areas with inline comments.


src/client/action/list_databases.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+}
+
+#[cfg(any(feature = "sync", feature = "tokio-sync"))]
+impl SyncClient {
+    /// Gets information about each database present in the cluster the Client is connected to.
+    pub fn list_databases(
+        &self,
+    ) -> ListDatabases {
+        self.async_client.list_databases()
```
abr-egn, line 43: Since this is now literally just a pass-through (the runtime wrapping is handled by `ListDatabases::run`), it hints towards a tantalizing future where there's no need for a distinct `sync::Client` type.  That might not ever be possible, though, due to things like the async nature of construction and whatnot.

src/client/action/list_databases.rs
```diff
@@ -0,0 +1,162 @@
...
+        }.boxed()
+    }
+}
+
+#[cfg(any(feature = "sync", feature = "tokio-sync"))]
+impl<'a, M: Mode> ListDatabases<'a, M>
+    where Self: IntoFuture
+{
+    /// Synchronously execute this action.
+    pub fn run(self) -> <Self as IntoFuture>::Output {
```
abr-egn, line 165: It's unfortunate that sync users will now always need a `.run()` in there, but I don't see any way around that.

src/db/options.rs
```diff
@@ -284,25 +284,6 @@ pub struct ListCollectionsOptions {
     pub comment: Option<Bson>,
 }
 
-/// Specifies the options to a
```
abr-egn, line 287: I moved this to `operation/list_databases.rs` since it's now describing what options the server expects in the operation, not part of the public API.

src/client/action/list_databases.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    /// Filters the query.
+    pub fn filter(mut self, value: impl Into<Option<Document>>) -> Self {
+        self.options().filter = value.into();
+        self
+    }
+
+    /// Determines which databases to return based on the user's access privileges. This option is
+    /// only supported on server versions 4.0.5+.
+    pub fn authorized_databases(mut self, value: impl Into<Option<bool>>) -> Self {
```
abr-egn, line 88: I'd rather have these values be the simple types to keep things clean and keep compile time down, but that makes it a hassle for any calling code to have optional values for these.

src/client/action/list_databases.rs
```diff
@@ -0,0 +1,140 @@
...
+            options: Default::default(),
+            session: None,
+            mode: PhantomData,
+        }
+    }
+}
+
+/// Gets information about each database present in the cluster the Client is connected to.
+#[must_use]
+pub struct ListDatabases<'a, M: Mode = Full> {
```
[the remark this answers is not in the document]
isabelatkinson: agreed, I'd rather minimize duplication

src/client/action/list_databases.rs
isabelatkinson: Is there a real use case for passing `None` into any of these setter methods? For most (if not all) of our options `None` is the default and means "don't send anything," so I'm not sure why users would need to do so. If not I think we should consider taking only non-optional types instead to simplify the API. Maybe that's something we could flesh out during the pre-release period.

src/client/action/list_databases.rs
```diff
@@ -0,0 +1,168 @@
...
+impl SyncClient {
+    /// Gets information about each database present in the cluster the Client is connected to.
+    pub fn list_databases(&self) -> ListDatabases {
+        self.async_client.list_databases()
+    }
+}
+
+/// Gets information about each database present in the cluster the Client is connected to.
+#[must_use]
+pub struct ListDatabases<'a, M: Mode = Full> {
```
isabelatkinson, line 49: Have you thought at all about combining this type with the `operation::ListDatabases` type (and ditto for future action types we introduce)? It would require some refactoring to store the client/session as part of the operation's state, but it might end up making the operation layer a little bit more streamlined if the compiler allows it...

src/client/action/list_databases.rs
```diff
@@ -0,0 +1,140 @@
...
+    client: &'a Client,
+    options: Option<op::Options>,
+    session: Option<&'a mut ClientSession>,
+    mode: PhantomData<M>,
+}
+
+
+mod private {
```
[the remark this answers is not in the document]
isabelatkinson: can you explain what the trait is needed for?

src/test/spec/faas.rs
```diff
@@ -39,7 +39,7 @@ async fn check_faas_handshake(vars: &[(&'static str, &str)]) -> Result<()> {
     let _tv = TempVars::set(vars);
 
     let client = Client::test_builder().build().await;
-    client.list_database_names(doc! {}, None).await?;
+    client.list_database_names().filter(doc! {}).await?;
```
isabelatkinson, line 42: nit: can the filter here be omitted entirely? I'd suspect empty filter is equivalent to no filter set

src/client/action/list_databases.rs
```diff
@@ -0,0 +1,140 @@
...
+pub struct Full;
+pub struct Names;
+
+impl<'a, M: Mode> ListDatabases<'a, M> {
+    fn options(&mut self) -> &mut op::Options {
+        self.options.get_or_insert_with(op::Options::default)
+    }
+
+    #[cfg(test)]
+    pub(crate) fn with_options(mut self, value: impl Into<Option<op::Options>>) -> Self {
```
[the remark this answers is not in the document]
isabelatkinson: minor concern: we won't actually be testing the builder codepath if we're setting options via this method in tests. probably not a huge deal since the setters are very straightforward, but it is worth considering whether we want any test coverage of those methods

src/client/action/list_databases.rs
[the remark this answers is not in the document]
abr-egn: The `Mode` trait has to be public because it's part of the API.  However, we don't want users adding impls of `Mode` to new types because it's just there to allow the two different paths for `ListDatabases`.  This hack is the typical way to do that in Rust: https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed

src/client/action/list_databases.rs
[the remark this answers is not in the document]
abr-egn: Yeah, good to think about, I don't think it's worth testing specifically though.  I actually suspect that once I've written a couple of these I'll be able to put together a declarative macro to automate the boilerplate, which would make tests even less useful.

src/client/action/list_databases.rs
abr-egn: The extremely hand-wavy hypothetical is when the client is writing something where they want to expose the optional nature of a setting in their own API.  In that case, if we're not accepting an `Option` they'd have to go through an awkward little dance:
```
let user_auth_param: Option<bool> = ...;// something external
let mut act = client.list_databases();
if let Some(auth) = user_auth_param {
  act = act.authorized_databases(auth);
}
let result = act.await;
```
rather than just
```
let user_auth_param: Option<bool> = ...;// something external
let result = client.list_databases().authorized_databases(user_auth_param).await;
```

Given that (a) this seems pretty unlikely and (b) it's still doable if needed, I think you're right and I'll go with the simple option and we can see if this comes up in pre-release.

src/client/action/list_databases.rs
abr-egn: That's a really interesting possibility.  I'll definitely give this a try and see if I can make the refactor work.

src/test/spec/faas.rs
abr-egn: Oh, yup, good catch.  Rote translation...

src/client/action/list_databases.rs
abr-egn: Unfortunately, there are some significant hurdles to this approach :(  I've filed RUST-1816 for this - it _might_ be possible but it'd be a huge refactor of the executor, change streams, and cursors.

abr-egn:
Closing this out as I continue work.

PR #1002, opened 2024-01-04
RUST-1512 Fluent API: `list_databases`, `watch`

abr-egn:
RUST-1512

In the interest of not having a monster PR, I'm going to split up this conversion.  Notable updates since the preview PR:
* The `option_setters!` macro, which automates generating the setters on the action.  Happily, it turns out to be possible to propagate doc comments from the macro to the generated code :)
* The implementation for `watch`; specific details have their own comments.


src/change_stream/options.rs
```diff
@@ -13,95 +12,28 @@ use crate::{
...
-    /// This feature is only available on MongoDB 4.2+.
-    ///
-    /// See the documentation [here](https://www.mongodb.com/docs/master/changeStreams/#change-stream-start-after) for more
-    /// information.
-    #[builder(default)]
-    pub start_after: Option<ResumeToken>,
-
-    /// If `true`, the change stream will monitor all changes for the given cluster.
-    #[builder(default, setter(skip))]
+pub(crate) struct ChangeStreamOptions {
```
abr-egn, line 19: Unlike `list_databases`, the situation with options and operations is a little more messy here, so I opted to simply make this non-`pub` and defer further internal refactoring.

src/client/action/watch.rs
```diff
@@ -0,0 +1,350 @@
...
+    /// [`ChangeStream`](change_stream/struct.ChangeStream.html)
+    ///
+    /// If the pipeline alters the structure of the returned events, the parsed type will need to be
+    /// changed via [`ChangeStream::with_type`].
+    pub fn watch(&self) -> Watch {
+        Watch::new_cluster(&self)
+    }
+}
+
+impl Database {
```
abr-egn, line 50: Since the impls for `Database` and `Collection` are just different `AggregateTargets`, I opted to just bundle all of them into this file.

src/client/action/watch.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+}
+
+mod private {
+    pub trait Sealed {}
+    impl Sealed for super::Implicit {}
+    impl<'a> Sealed for super::Explicit<'a> {}
+}
+
+pub trait Session: private::Sealed {}
```
abr-egn, line 148: I expect this will be needed for other methods that have a different return type when a session is present; when I get to that point I'll bump this up to the parent `action` module.

src/client/action/watch.rs
```diff
@@ -0,0 +1,350 @@
...
+        self.pipeline = value.into_iter().collect();
+        self
+    }
+
+    /// Specifies the logical starting point for the new change stream. Note that if a watched
+    /// collection is dropped and recreated or newly renamed, `start_after` should be set instead.
+    /// `resume_after` and `start_after` cannot be set simultaneously.
+    ///
+    /// For more information on resuming a change stream see the documentation [here](https://www.mongodb.com/docs/manual/changeStreams/#change-stream-resume-after)
+    pub fn resume_after(mut self, value: impl Into<Option<ResumeToken>>) -> Self {
```
abr-egn, line 209: This was one scenario where it's clearly better to accept an `impl Into<Option<..>>` - if more of these show up I'm going to lean more towards having that be the standard setter parameter.

src/client/action/list_databases.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// Gets the names of the databases present in the cluster the Client is connected to.
+    pub fn list_database_names(&self) -> ListDatabases<'_, Names> {
+        self.async_client.list_database_names()
+    }
+}
+
+/// Gets information about each database present in the cluster the Client is connected to.
+#[must_use]
+pub struct ListDatabases<'a, M: Mode = Full> {
```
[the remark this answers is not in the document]
abr-egn: Good call - you're right, it's not actually needed.  Removed it and updated the names of the unit structs to make their purpose more clear.  (That also led me on a tangent to update the rustdoc so that the usage pattern should be clear.)

src/client/action/watch.rs
abr-egn: Oh, sorry - because the user will get this value from `ChangeStream::resume_token`, which returns an `Option` ([example](https://github.com/abr-egn/mongo-rust-driver/blob/RUST-1512/better-helpers/src/test/documentation_examples.rs#L1813)).

isabelatkinson (approved):
lgtm! very excited for this new API

PR #1003, opened 2024-01-08
fix: list_collections fails with MongoDB Atlas database with time-series collection

hallettj:
When creating a time-series collection on MongoDB Atlas the `clusteredIndex` field of the collection options is given as `true` instead of as an object that deserializes to `ClusteredIndex`. The rust driver does not expect a boolean in that position, so parsing `CreateCollectionOptions` fails. This causes methods such as `db.list_collections()` to fail.

This change adds a custom deserializer that handles the case where `clusteredIndex: true` by filling in the default value for `ClusteredIndex`.

I don't know if `clusteredIndex: false` is a valid state so in the interest of being conservative I have made that an error case in the custom deserializer.


src/db/options.rs
```diff
@@ -176,6 +177,35 @@ impl Default for ClusteredIndex {
...
+        enum ValueUnion {
+            Bool(bool),
+            ClusteredIndex(ClusteredIndex),
+        }
+
+        let value_option: Option<ValueUnion> = Deserialize::deserialize(deserializer)?;
+        value_option
+            .map(|value| match value {
+                ValueUnion::Bool(value) if value == true => Ok(ClusteredIndex::default()),
+                ValueUnion::Bool(_) => Err(serde::de::Error::custom(
```
isabelatkinson, line 200: Suggestion to fix clippy error:
```suggestion
                ValueUnion::Bool(true) => Ok(ClusteredIndex::default()),
                ValueUnion::Bool(false) => Err(serde::de::Error::custom(
```

src/db/options.rs
```diff
@@ -336,3 +366,24 @@ pub struct RunCursorCommandOptions {
...
+        let options = from_value::<CreateCollectionOptions>(input)?;
+        let clustered_index = options
+            .clustered_index
+            .expect("deserialized options include clustered_index");
+        let default_index = ClusteredIndex::default();
+        assert_eq!(clustered_index.key, default_index.key);
+        assert_eq!(clustered_index.unique, default_index.unique);
+        assert_eq!(clustered_index.name, default_index.name);
+        assert_eq!(clustered_index.v, default_index.v);
+        Ok(())
```
isabelatkinson, line 387: Would you mind moving this test to `src/test/db.rs`? This module should no longer be necessary with that change.

hallettj:
I made the requested changes. Sorry about the long delay!

isabelatkinson (commented):
Changes look good, looks like there's just a formatting issue in `src/test/db.rs`. Would you mind running the following to fix it?

```
rustfmt +nightly --unstable-features src/test/db.rs
```

After that this should be ready to go!

isabelatkinson:
Hey @hallettj, this will be fixed in #1077. Thank you for the bug report!

PR #1011, opened 2024-01-18
RUST-1810 Clean up concern helpers

abr-egn:


src/concern.rs
```diff
@@ -104,8 +81,8 @@ impl ReadConcern {
     /// Creates a read concern with a custom read concern level. This is present to provide forwards
     /// compatibility with any future read concerns which may be added to new versions of
     /// MongoDB.
-    pub fn custom(level: String) -> Self {
-        ReadConcernLevel::from_str(level.as_str()).into()
+    pub fn custom(level: &str) -> Self {
```
isabelatkinson, line 84:
```suggestion
    pub fn custom(level: impl AsRef<str>) -> Self {
```

src/concern.rs
```diff
@@ -322,6 +295,21 @@ impl From<String> for Acknowledgment {
...
+        Acknowledgment::Nodes(v).into()
+    }
+
+    /// A `WriteConcern` requesting [`Acknowledgment::Majority`].
+    pub fn majority() -> Self {
+        Acknowledgment::Majority.into()
+    }
+
+    /// A `WriteConcern` with a custom acknowledgment.
+    pub fn custom(s: &str) -> Self {
```
isabelatkinson, line 309:
```suggestion
    pub fn custom(s: impl AsRef<str>) -> Self {
```

src/concern.rs
abr-egn: Updated.

src/concern.rs
abr-egn: Updated.

isabelatkinson (approved):

PR #1012, opened 2024-01-19
RUST-1505 Convert remaining event handler types to new API

abr-egn:
RUST-1505

This updates `cmap_event_handler` and `sdam_event_handler` to use the new API.  There are a lot of little changes but this is a fairly mechanical PR, it's all following the same pattern as `command_event_handler` did.


src/test/util/event.rs
```diff
@@ -93,20 +91,6 @@ impl Event {
     }
 }
 
-#[derive(Clone, Debug, Serialize)]
-#[serde(untagged)]
-pub(crate) enum SdamEvent {
```
abr-egn, line 98: I don't know why we had two of these.

isabelatkinson (approved):

PR #1013, opened 2024-01-23
RUST-1512 Macro for fluent API definition, Database::drop conversion

abr-egn:
RUST-1512

I think we want to leave the door open to using something like [stackfuture](https://crates.io/crates/stackfuture) if it turns out there are methods where the small overhead of the `BoxFuture` heap alloc are an issue.  To that end, this updates the futures returned by the action `IntoFuture` impls to be opaque wrappers.

This, in turn, pushed the things over the threshold where I think it's worthwhile to macro-generate the boilerplate surrounding the actual execution method for actions, so I put one together and updated existing definitions to use it.

This also includes the conversion for `Database::drop`, and moves the `action` module up a level because it's not `Client`-specific (hasn't been since `watch`, really).


src/action.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        #[cfg(any(feature = "sync", feature = "tokio-sync"))]
+        impl$(<$lt>)? $action {
+            /// Synchronously execute this action.
+            pub fn run(self) -> $out {
+                crate::runtime::block_on(std::future::IntoFuture::into_future(self))
+            }
+        }
+    };
+    // Generate with a sync type conversion
```
abr-egn, line 74: Having to duplicate the definition here is annoying, but (AFAICT) necessary - while you can define optional matchers for a macro, you can't explicitly branch on empty or provide default values.  This means that to handle both the no-conversion and conversion case two distinct match arms are needed.  I reduced the duplication by moving the non-varying part to `action_impl_inner`.

src/action.rs
```diff
@@ -0,0 +1,140 @@
...
+        impl$(<$lt>)? std::future::Future for $f_ty$(<$lt>)? {
+            type Output = $out;
+
+            fn poll(mut self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> std::task::Poll<Self::Output> {
+                self.0.as_mut().poll(cx)
+            }
+        }
+    }
+}
+pub(crate) use action_impl_inner;
```
abr-egn, line 128: Unfortunately, these helpers can't just be hidden in the module because they're technically being invoked in the expanded code elsewhere.

src/action.rs
```diff
@@ -0,0 +1,140 @@
...
+
+macro_rules! action_impl_future_wrapper {
```
abr-egn, line 130: This exists for the same reason as the multiple arms on `action_impl` - if there's no lifetime the wrapped type would expand to `BoxFuture<, $out>` when it actually needs to be `BoxFuture<'static, $out>`.

src/action/perf.rs
```diff
@@ -0,0 +1,51 @@
...
+
+/// Add connections to the connection pool up to `min_pool_size`.  Create by calling
+/// [`Client::warm_connection_pool`] and execute with `await` (or [`run`](WarmConnectionPool::run)
+/// if using the sync client).
+#[must_use]
+pub struct WarmConnectionPool<'a> {
+    pub(crate) client: &'a Client,
+}
+
+// IntoFuture impl in src/client/action/perf.rs
```
abr-egn, line 51: Some of the implementation of the methods used things (e.g. `Client::inner`) that aren't exposed to the rest of the crate.  With the `action` module no longer a submodule of `client`, that didn't work, so the options were:
* make the relevant fields `pub(crate)`
* implement the relevant bits in a submodule and re-export it

While we generally haven't been too careful with crate-interior visibility I went with the second because I'm generally a fan of visibility boundaries as a dev aid.  Not entirely convinced this was the right choice though.

isabelatkinson (approved):
lgtm! appreciate how easy the macro implementations are to read

PR #1016, opened 2024-01-25
RUST-1525 Update AWS tests to use new scripts

isabelatkinson:
Removes all uses of the legacy shell by migrating to the new AWS scripts in drivers-evergreen-tools. Also completes RUST-1777 by upgrading to Ubuntu 20.04 (the new arn is no longer relevant now that we're using the secrets manager).

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -257,7 +257,7 @@ buildvariants:
 
   - name: aws-auth
     display_name: "AWS Authentication"
-    # patchable: false
+    patchable: false
     run_on:
       - ubuntu2004-small
     expansions:
@@ -1356,9 +1356,6 @@ functions:
           ECS_SRC_DIR=$AUTH_AWS_DIR/src
 
           mkdir -p $ECS_SRC_DIR/.evergreen
-
-          # fix issue with `TestData` in SERVER-46340
-          sed -i '1s+^+TestData = {};\n+' $AUTH_AWS_DIR/lib/ecs_hosted_test.js
 
           # compile mini test project
           cd $PROJECT_DIRECTORY/.evergreen/aws-ecs-test
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -79,9 +79,6 @@ buildvariants:
     run_on:
       - ubuntu2204-arm64-small
     expansions:
-      # Needed to skip installation of the legacy shell in boostrap mongo-orchestration, which is
-      # not available on Ubuntu 22.04.
-      SKIP_LEGACY_SHELL: true
       AUTH: auth
       SSL: ssl
     tasks:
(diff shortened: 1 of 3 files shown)


.evergreen/config.yml
```diff
@@ -79,9 +79,6 @@ buildvariants:
     run_on:
       - ubuntu2204-arm64-small
     expansions:
-      # Needed to skip installation of the legacy shell in boostrap mongo-orchestration, which is
-      # not available on Ubuntu 22.04.
-      SKIP_LEGACY_SHELL: true
```
isabelatkinson, line 84: This variable is no longer respected by drivers-evergreen-tools so I removed our uses of it

.evergreen/config.yml
```diff
@@ -266,13 +257,10 @@ buildvariants:
 
   - name: aws-auth
     display_name: "AWS Authentication"
-    patchable: false
+    # patchable: false
```
isabelatkinson, line 260: I'll uncomment this before merging

.evergreen/config.yml
```diff
@@ -266,13 +257,10 @@ buildvariants:
...
     run_on:
-      # Use ubuntu1804. AWS tasks require legacy `mongo` shell. Legacy `mongo` shell is not available on newer distros.
-      - ubuntu1804-small
+      - ubuntu2004-small
```
isabelatkinson, line 262: Using an older Ubuntu version here because the ECS test doesn't work on 22.04

abr-egn (approved):
LGTM!

.evergreen/env.sh
```diff
@@ -9,6 +9,7 @@ export NVM_DIR="${NODE_ARTIFACTS_PATH}/nvm"
 
 . ${CARGO_HOME}/env
 
+set +u # allow undefined variable for $OS
 if [[ "$OS" == "Windows_NT" ]]; then
```
kevinAlbs, line 13:
```suggestion
if [[ "$OSTYPE" == "cygwin" ]]; then
```

Suggest using the Bash variable `OSTYPE`. It is a variable set by Bash: https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html and this script defines bash in the shebang.

`OSTYPE` is also used in the C driver scripts. Example: https://github.com/mongodb/mongo-c-driver/blob/5e5784d814b2e3912c2855ac2b9a172267b8235f/.evergreen/scripts/add-build-dirs-to-paths.sh#L3-L4

.evergreen/config.yml
```diff
@@ -1476,46 +1370,22 @@ functions:
           cp $PROJECT_DIRECTORY/.evergreen/run-mongodb-aws-ecs-test.sh $ECS_SRC_DIR/.evergreen
```
kevinAlbs, line 1370: Suggest removing the `sed` line above. I expect it is no longer needed. Adding `TestData = {};\n` does not appear in the [C driver config](https://github.com/mongodb/mongo-c-driver/blob/5e5784d814b2e3912c2855ac2b9a172267b8235f/.evergreen/scripts/run-aws-tests.sh#L116-L133).

kevinAlbs (approved):
LGTM. I left two minor suggestions.

.evergreen/config.yml
isabelatkinson: updated

.evergreen/env.sh
isabelatkinson: updated

PR #1017, opened 2024-01-26
RUST-1831 Use secret manager for serverless tests

abr-egn:
This updates the serverless tests to use the aws secret manager (which includes `SERVERLESS_REGION`).

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -324,7 +324,7 @@ buildvariants:
       - test-plain-auth
 
   - name: serverless-passthrough
-    patchable: false
+    #patchable: false
     display_name: "Serverless (Passthrough)"
     run_on:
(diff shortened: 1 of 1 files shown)


.evergreen/config.yml
```diff
@@ -324,7 +324,7 @@ buildvariants:
       - test-plain-auth
 
   - name: serverless-passthrough
-    patchable: false
+    #patchable: false
```
abr-egn, line 327: Will revert this before submitting.

abr-egn:
The failure in serverless-terminating is apparently expected.

kevinAlbs (approved):
> The failure in serverless-terminating is apparently expected.

DRIVERS-2812 appears to track investigating the failing "errors during the initial connection hello are ignored" test.

PR #1019, opened 2024-01-30
RUST-1512 Bring back options as part of the visible API

abr-egn:
RUST-1512

As I do more conversions (not included in this PR) the use cases for these continue to accumulate:
* Utility functions ("do this thing with passed-in options and then do something afterward")
* Testing code (compare test execution results to option values)
* Specific one-offs (`CreateCollectionOptions` is returned as part of the record for `list_collections`)
* Complicated conditional option-setting logic
* Legacy migration path - it'll be easier for users to just update code to `.with_options` than to move around the individual setter functions.

I've placed them in the corresponding `action` modules because I prefer grouping distinct chunks of functionality together rather than the sort-by-type of having a global 'options' module, and added compatibility re-exports to `options` so existing code won't break.  Happy to reconsider if you feel strongly the other way.


src/action.rs
```diff
@@ -7,34 +7,37 @@ mod session;
...
-        pub(crate) fn with_options(mut self, value: impl Into<Option<$opt_field_ty>>) -> Self {
+        /// Set all options.  Note that this will replace all previous values set.
+        pub fn with_options(mut self, value: impl Into<Option<$opt_field_ty>>) -> Self {
             self.options = value.into();
             self
         }
 
         $(
-            $(#[$($attrss)*])*
+            #[doc = concat!("Set the [`", stringify!($opt_field_ty), "::", stringify!($opt_name), "`] option.")]
```
abr-egn, line 40: With the options struct back to the place where there's canonical docs for the individual options, this means that the option_setters invocation don't need to duplicate the docs.

isabelatkinson:
> I've placed them in the corresponding action modules because I prefer grouping distinct chunks of functionality together rather than the sort-by-type of having a global 'options' module, and added compatibility re-exports to options so existing code won't break. Happy to reconsider if you feel strongly the other way.

I agree with the sentiment here, but I think the `actions` module (and the new concept of actions as a whole) will be less known by users since the various action types are abstracted away by the `IntoFuture` implementations -- for that reason I'd be more in favor of maintaining the existing `options` mod structure. I also think it can get a little confusing when there are multiple places to import a type from. Don't feel _super_ strongly here though.

abr-egn:
> I agree with the sentiment here, but I think the `actions` module (and the new concept of actions as a whole) will be less known by users since the various action types are abstracted away by the `IntoFuture` implementations -- for that reason I'd be more in favor of maintaining the existing `options` mod structure. I also think it can get a little confusing when there are multiple places to import a type from. Don't feel _super_ strongly here though.

Yeah, that makes sense.  Put them back in the original locations.

isabelatkinson (approved):

PR #1022, opened 2024-02-04
chore: move `lazy_static` to `once_cell`

attila-lin:


abr-egn:
Thanks for the contribution!  I've merged this in and it'll go out with our next release.

PR #1023, opened 2024-02-05
RUST-1498 Automatically test FaaS behavior

abr-egn:
RUST-1498

This spins up an aws lambda function using the standard template that tracks and reports some stats.


.evergreen/config.yml
```diff
@@ -395,12 +394,20 @@ buildvariants:
...
-    patchable: true  # TODO RUST-1684 change to false before merge
+    patchable: false
     run_on:
       - rhel87-small
     tasks:
       - name: search-index-task-group
 
+  - name: aws-lambda
+    display_name: "AWS Lambda"
+    patchable: true
```
abr-egn, line 405: ... but this time I will, swearsies.

.evergreen/run-aws-lambda-test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Needed for cargo-lambda
+pip3 install ziglang
```
kkloberdanz, line 4: Are we using Zig for its cross compilation toolchain, i.e., `zig cc`?

kkloberdanz (approved):
Looks good!

.evergreen/run-aws-lambda-test.sh
abr-egn: I believe that's what `cargo-lambda` uses it for, yeah.  The important part from my PoV is it refuses to run if it isn't installed :)

isabelatkinson (approved):

PR #1025, opened 2024-02-06
RUST-1841 Allow double-valued connectionIds

abr-egn:
RUST-1841

It turns out that when we implemented RUST-1571 we missed that the value could be a double as well as an int64.

I'll cherry-pick this to `main` as well.


kkloberdanz (approved):
Looks good!

isabelatkinson (approved):

PR #1027, opened 2024-02-08
RUST-1789 Clarify exactly what code/label fields drivers should inspect to determine retryability

kkloberdanz:
[Evergreen build](https://spruce.mongodb.com/version/65c52f8cd6d80abc78780f57/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


src/error.rs
```diff
@@ -181,16 +185,29 @@ impl Error {
     /// indicates a 4.4+ server, a label should only be added if the error is a network error.
     /// Otherwise, a label should be added if the error is a network error or the error code
     /// matches one of the retryable write codes.
-    pub(crate) fn should_add_retryable_write_label(&self, max_wire_version: i32) -> bool {
+    pub(crate) fn should_add_retryable_write_label(
+        &self,
+        max_wire_version: i32,
+        conn: Option<&Connection>,
```
abr-egn, line 191: I'd probably pass in `Option<ServerType>` instead here, since that's all this needs and it'll avoid the need for the return to be `Result`.

src/error.rs
```diff
@@ -181,16 +185,29 @@ impl Error {
...
+        is_reply_ok: Option<bool>,
```
abr-egn, line 192: The spec change doesn't look like it mentions additional logic based on `ok` values?

kkloberdanz: This is following along with how the change was made in the [C driver](https://github.com/mongodb/mongo-c-driver/commit/da1f98cdd0732813ac33f8616c1dda0a4fd6fbf6#diff-d96d8295b32780116338ab4eacb607f962ac00cbb900b17667493e8ec2072eacR123-R128), which was the reference implementation for this change.

src/error.rs
kkloberdanz: Done

abr-egn (approved):
LGTM!

src/error.rs
isabelatkinson: Is the `is_reply_ok` param here meant to indicate whether the error was a write concern error? If so, we should be able to simplify this by checking the `kind` field of the error this method is being called on, i.e.
```
// could also add this as a method on Error
let is_write_concern_error = match *self.kind {
    ErrorKind::Write(WriteFailure::WriteConcernError(_)) => true,
    _ => false,
};

if server_type == Some(ServerType::Mongos) && is_write_concern_error {
    return false;
}
```
This feels a little cleaner to me than needing to propagate the value of the `ok` field.

kkloberdanz: Unfortunately, this proposed change would cause the tests to fail, see this [patch build failure](https://spruce.mongodb.com/version/65d618567742aed7860e3326/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) with this change applied for details. If we look at the [C driver](https://github.com/mongodb/mongo-c-driver/commit/da1f98cdd0732813ac33f8616c1dda0a4fd6fbf6#diff-d96d8295b32780116338ab4eacb607f962ac00cbb900b17667493e8ec2072eacR123-R128) implementation, which was the reference implementation for this change, we can see that it was the intention of the author of this spec update to not retry if the server is mongos and the reply from the server is `ok`.

isabelatkinson: Ah I missed in my suggestion that we wrap errors returned from `insert_many` in `ErrorKind::BulkWrite`. The following method worked for me when testing this locally as a replacement for the check against `is_reply_ok`:

```
// method on Error
fn is_write_concern_error(&self) -> bool {
    match *self.kind {
        ErrorKind::Write(WriteFailure::WriteConcernError(_)) => true,
        ErrorKind::BulkWrite(ref bulk_write_error)
            if bulk_write_error.write_concern_error.is_some() =>
        {
            true
        }
        _ => false,
    }
}
```

It seems like the C implementation is using the `ok` value as a proxy for whether the error is a write concern error, but since we have type info about write concern errors directly on our `Error` type I think it's a little more readable to use that here instead.

isabelatkinson (approved):
lgtm!

PR #1029, opened 2024-02-12
RUST-1512 Fluent `Collection` methods, part 1

abr-egn:
RUST-1512

`Collection` has a lot of methods (about as much as the other major types combined) so I'm breaking this up across multiple PRs.


src/action.rs
```diff
@@ -169,3 +174,19 @@ macro_rules! action_impl_future_wrapper {
     };
 }
 pub(crate) use action_impl_future_wrapper;
+
+use crate::Collection;
+
+pub(crate) struct CollRef<'a> {
```
abr-egn, line 180: The context for this is that there are a bunch of `Collection` methods that don't actually use the document type parameter.  If the action struct for those methods just contained a `&Collection<T>` the action struct would need to have the type parameter as well, which means it inherits a bunch of constraints from `T` that it doesn't actually need and cause various hassle.

The action can just contain a `Collection<Document>` directly, but that means that unlike all the other actions it doesn't have a lifetime parameter binding it to the original call site; the `CollRef` struct here synthesizes that parameter.  I think this is valuable for consistency, and it also means that if the added cost of the `Arc::clone` somehow turns out to be a bottleneck we can potentially change the action to use an (unsafe) ref transmute to avoid the cost.

src/action/aggregate.rs
```diff
@@ -113,14 +151,20 @@ action_impl! {
...
         async fn execute(mut self) -> Result<SessionCursor<Document>> {
-            resolve_options!(
-                self.db,
-                self.options,
-                [read_concern, write_concern, selection_criteria]
-            );
-
-            let aggregate = crate::operation::aggregate::Aggregate::new(self.db.name().to_string(), self.pipeline, self.options);
-            let client = self.db.client();
+            if matches!(self.target, AggregateTargetRef::Collection(_)) {
```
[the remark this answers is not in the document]
isabelatkinson: this seems like an oversight to me

src/coll/options.rs
```diff
@@ -622,6 +622,10 @@ pub struct AggregateOptions {
 #[builder(field_defaults(default, setter(into)))]
 #[non_exhaustive]
 pub struct CountOptions {
+    /// The filter to use for the operation.
+    #[serde(skip_serializing)]
+    pub filter: Option<Document>,
```
isabelatkinson, line 627: Is the plan to move `filter` into options for all methods that accept one? This is at odds with the definitions in the CRUD spec, which have `filter` as a required parameter adjacent to the options. I'm not sure I have strong feelings about which design is better, but we might want to make sure there isn't a strong reason the CRUD definitions were written this way before deviating. (edit: looks like we already deviate by accepting filters as optional docs in our existing API...)

isabelatkinson: good point about silent defaults, sgtm

isabelatkinson (approved):
lgtm, no need for re-review on the filter stuff unless you want another look!

src/coll/options.rs
abr-egn: So it turns out that the spec has `filter` as an explicit non-optional parameter for [`countDocuments`](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#L171), but as just another option for [`listDatabases`](https://github.com/mongodb/specifications/blob/master/source/enumerate-databases.rst#filters) and [`listCollections`](https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst#post-mongodb-280-rc3-versions).  I'm going to stick to mirroring that for the Rust driver; happy to do a follow-up PR if you think it'd be better to be non-optional for all three.

isabelatkinson: yeah going by the spec seems like the right approach here + for the other methods that accept a filter in future PRs

PR #1030, opened 2024-02-15
RUST-1631 Always use polling monitoring when running in a FaaS environment

abr-egn:
RUST-1631

The core change is pretty simple - add some behavior to switch back to polling when FaaS is detected - but there's a lot of stuff around that (tests, URI config, etc.)

### DIFF
--- a/.evergreen/aws-lambda-test/mongodb/src/main.rs
+++ b/.evergreen/aws-lambda-test/mongodb/src/main.rs
@@ -34,10 +34,12 @@ impl Stats {
 
     fn handle_sdam(&mut self, event: &SdamEvent) {
         match event {
-            SdamEvent::ServerHeartbeatStarted(_) => {
+            SdamEvent::ServerHeartbeatStarted(ev) => {
+                assert!(!ev.awaited);
                 self.heartbeats_started += 1;
             }
             SdamEvent::ServerHeartbeatFailed(ev) => {
+                assert!(!ev.awaited);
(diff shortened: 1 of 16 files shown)


.evergreen/aws-lambda-test/mongodb/src/main.rs
```diff
@@ -34,10 +34,12 @@ impl Stats {
 
     fn handle_sdam(&mut self, event: &SdamEvent) {
         match event {
-            SdamEvent::ServerHeartbeatStarted(_) => {
+            SdamEvent::ServerHeartbeatStarted(ev) => {
+                assert!(!ev.awaited);
```
abr-egn, line 38: Rather than setting `patchable: true` so this is included in the evergreen runs and then (forget to) change it back before submitting, I just did an [independent run](https://spruce.mongodb.com/version/65ce39a6d6d80a58a79688fb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) for that.

src/sdam/monitor.rs
```diff
@@ -102,13 +117,19 @@ impl Monitor {
         while self.is_alive() {
             let check_succeeded = self.check_server().await;
 
+            if self.topology_version.is_some() && self.allow_streaming {
```
abr-egn, line 120: The spec changed from allowing drivers to use a dedicated RTT monitor even when polling to requiring that one not be used when polling, so starting up that task had to be deferred to the point where we know whether polling is happening or not.

isabelatkinson (approved):

PR #1031, opened 2024-02-15
RUST-1839 Avoid logging secrets when running OIDC or serverless tests

abr-egn:
RUST-1839 / RUST-1838

Since they're not patchable, [here's](https://spruce.mongodb.com/version/65ce5ae3a4cf473f1634d93e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) a run with OIDC and [here's](https://spruce.mongodb.com/version/65ce62a8850e61402789c00b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) one with serverless.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1566,13 +1566,11 @@ functions:
       params:
         working_dir: src
         shell: bash
+        include_expansions_in_env:
+          - SERVERLESS_ATLAS_USER
+          - SERVERLESS_ATLAS_PASSWORD
         script: |
           ${PREPARE_SHELL}
-
-          # Export without xtrace to avoid leaking credentials.
-          set +o xtrace
-          export SERVERLESS_ATLAS_USER=${SERVERLESS_ATLAS_USER}
-          export SERVERLESS_ATLAS_PASSWORD=${SERVERLESS_ATLAS_PASSWORD}
 
           export SINGLE_MONGOS_LB_URI=${SERVERLESS_URI}
           . .evergreen/generate-uri.sh
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1726,9 +1726,6 @@ functions:
           cd ${DRIVERS_TOOLS}/.evergreen/auth_oidc
           set +o xtrace
 
-          export AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
-          export AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
-          export AWS_SESSION_TOKEN=${AWS_SESSION_TOKEN}
           export OIDC_TOKEN_DIR=/tmp/tokens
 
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -1566,14 +1566,12 @@ functions:
...
         script: |
           ${PREPARE_SHELL}
 
-          # Export without xtrace to avoid leaking credentials.
-          set +o xtrace
-          export SERVERLESS_ATLAS_USER=${SERVERLESS_ATLAS_USER}
-          export SERVERLESS_ATLAS_PASSWORD=${SERVERLESS_ATLAS_PASSWORD}
-
           export SINGLE_MONGOS_LB_URI=${SERVERLESS_URI}
           . .evergreen/generate-uri.sh
```
kevinAlbs, line 1576: Remove `. .evergreen/generate-uri.sh`? Sourcing the script enables `xtrace`. Later logs [include REDACTED output](https://parsley.mongodb.com/evergreen/mongo_rust_driver_serverless_passthrough_test_serverless_patch_abbbbd46730a3afd08a7c7f92d8c772446dad2d1_65ce62a8850e61402789c00b_24_02_15_19_14_57/0/task?bookmarks=0,3426&shareLine=1068).

I expect the output of `generate-uri.sh` is not being used.

abr-egn: I'm pretty sure those are false positives (matching on the path to a script in the tools repo).

I haven't actually run it without to test, but SSL is enabled for the serverless tests and the test is using `SINGLE_MONGOS_LB_URI` which `generate-uri.sh` will update.

kevinAlbs: > SSL is enabled for the serverless tests

I thought so too. Then I saw [this log line](https://parsley.mongodb.com/evergreen/mongo_rust_driver_serverless_passthrough_test_serverless_patch_abbbbd46730a3afd08a7c7f92d8c772446dad2d1_65ce62a8850e61402789c00b_24_02_15_19_14_57/0/task?bookmarks=0,3426&shareLine=1054) which suggests the SSL variable is empty:
```
[[ '' = \s\s\l ]]
```

I now expect `SINGLE_MONGOS_LB_URI` does not need to be updated for serverless. The serverless URI is a `mongodb+srv` URI. I expect the driver [implicitly enables TLS](https://github.com/mongodb/specifications/blob/e51a7bcef0d342cf764817f02b618520e0e1d2ed/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#validation-before-querying-dns):

> If mongodb+srv is used, a driver MUST implicitly also enable TLS.

All that said, `generate-uri.sh` appears to a general "prepare the URI for the test runner" script. Continuing to use it seems OK for consistency with other tests.

I suggest removing the `set -o xtrace` from `generate-uri.sh` as a precaution. Sourcing `generate-uri.sh` enables xtrace for the parent shell and may log more than desired.

PR #1036, opened 2024-02-23
RUST-1512 Collection fluent API, part 2

abr-egn:
RUST-1512

No surprises here, this is entirely following the patterns from previous PRs.


src/concern/test.rs
```diff
@@ -363,34 +351,24 @@ async fn command_contains_write_concern_update_many() {
...
-                    .w(Acknowledgment::Nodes(1))
-                    .journal(false)
-                    .build(),
-            )
-            .build(),
-    )
-    .await
-    .unwrap();
+    coll.update_many(doc! { "foo": "bar" }, doc! { "$set": { "foo": "baz" } })
+        .write_concern(
```
abr-egn, line 355: Sidebar: I wonder if it would be worth also having convenience methods for those options where the value itself has a builder.  It could pre-construct a builder and accept a closure that can update it, e.g.:
```
coll.update_many(doc! { "foo": "bar" }, doc! { "$set": { "foo": "baz" } })
    .build_write_concern(|wc| wc.w(Acknowledgement::Nodes(1)).journal(true))
    .await
```
Better to keep it out of this update just to streamline the release, but if you think that'd be a useful addition I'll file a ticket for it.

isabelatkinson (approved):

src/concern/test.rs
abr-egn: Yup, agreed.  Filed RUST-1865 for this.

PR #1046, opened 2024-03-11
RUST-1512 Convert the remainder of `Collection` methods to use actions

abr-egn:
RUST-1512

This also necessitated converting `action_impl!` into a proc_macro.  I think this is an overall positive - while the code for it is now more verbose, it's also now fairly simple normal Rust code.


src/operation/update.rs
```diff
@@ -11,58 +11,48 @@ use crate::{
...
     results::UpdateResult,
-    serde_util::to_raw_document_buf_with_options,
     Namespace,
 };
 
 #[derive(Clone, Debug)]
-pub(crate) enum UpdateOrReplace<'a, T = ()> {
+pub(crate) enum UpdateOrReplace {
     UpdateModifications(UpdateModifications),
-    Replacement(&'a T),
```
isabelatkinson, line 21: I think we should leave this as-is. Keeping a reference to the user's input here means that we only need to allocate the byte buffer for the replacement document once, i.e. on line 34 below. Storing a `RawDocumentBuf` requires allocating the byte buffer once when the `ReplaceOne` action is constructed and then again when it's cloned on line 34 in the new diff. (This was a perf ticket a little while ago that I worked on, see RUST-877.)

abr-egn: Derp, I missed that this was reverting some of your performance work, sorry!

That said, I like the other effects of using a `RawDocumentBuf` here instead of a `&T` - it makes types simpler without the lifetime and parameter, and means the `human_readable_serialization` option doesn't need to be propagated.  I believe with https://github.com/mongodb/bson-rust/pull/463 we get the best of both worlds; it avoids the intermediate buffer without having to carry a ref around.

PR #1049, opened 2024-03-14
RUST-1882 Consolidate and de-flake test event handling

abr-egn:
RUST-1882

This consolidates event handling across all the various tests into `EventBuffer` (renamed from `EventHandler` to avoid clashing with the public API).  To avoid race conditions between async tasks, the core buffer is stored as a (non-async) mutex protected `Vec`, with the event handler callback immediately appending to that vec.

In addition, there are two modes that our tests use for events:
* _bufffer and query_, where a bunch of things are run and then the events generated are checked afterwards
* _streaming_, where test operations are interleaved with event validation

`EventSubscriber` is the interface for the latter; where before it was implemented via broadcast channel (which could drop events if no readers existed), it's now a cursor on an underlying `EventBuffer`, with a generational marker to cause concurrent modifications to be fail-fast.

I moved convenience methods from `EventClient` to `EventBuffer` so that there aren't multiple surfaces for the same tasks.

There's a fair amount of follow-up work possible:
* `EventBuffer` inherited some methods with surprising mutable behavior; those now require a `&mut self` receiver to make it more obvious at the call site, but I think they could be entirely eliminated in favor of an immutable view onto the buffered events with a little work.
* As an extension of that, I suspect that we could entirely consolidate on one of the buffer or streaming modes of event-handling and get rid of the other.
* `EventClient` is now just "`TestClient` with bundled `EventBuffer`".  I'm not sure there's any real utility there over places that want to track events just having their own `EventBuffer` and passing it in to client construction.

I'll file tickets for those unless you object :)

Sadly, this hasn't removed all flakes, but in my testing has shown much less incidence of the "expected events [X, Y, Z] got [Y, Z]" kind of failure.


src/test/csfle.rs
```diff
@@ -548,9 +548,9 @@ async fn bson_size_limits() -> Result<()> {
 
     // Setup: encrypted client.
     let mut opts = get_client_options().await.clone();
-    let handler = Arc::new(EventHandler::new());
-    let mut events = handler.subscribe();
-    opts.command_event_handler = Some(handler.clone().into());
+    let buffer = EventBuffer::<Event>::new();
```
abr-egn, line 551: A type annotation was required here; it's not clear to me what made this harder to infer than any of the other call sites.

src/test/util/buffer.rs
```diff
@@ -0,0 +1,464 @@
...
+        if events.data.len() > self.index {
+            let event = events.data[self.index].0.clone();
+            self.index += 1;
+            return Some(event);
+        }
+        None
+    }
+
+    /// Consume and pass events to the provided closure until it returns Some or the timeout is hit.
+    pub(crate) async fn filter_map_event<F, R>(
```
abr-egn, line 327: This and the following methods are preserved from the previous `EventSubscriber` implementation; to avoid introducing bugs I left the implementation as is, but it's likely there are much more efficient ways to implement most of these given the underlying cursor representation.

src/test/util/buffer.rs
isabelatkinson, line 1: nit: I think `event_buffer.rs` would be a bit more descriptive for this file

isabelatkinson (commented):
Proposed follow-up work SGTM. In the meantime, can you add some documentation on the various event handling test types we have indicating a) if they should be used in new code (e.g. we probably want to stop using `EventClient`) and b) what they're best suited for? I think the latter would be particularly useful for discerning between `EventBuffer` and `EventSubscriber`. I'm worried that adding more internal test API without ripping out all of the old stuff will cause more disparate implementations, but hopefullly having some documented guidance will discourage that.

src/test/util/buffer.rs
abr-egn: Done.

abr-egn:
> Proposed follow-up work SGTM. In the meantime, can you add some documentation on the various event handling test types we have indicating a) if they should be used in new code (e.g. we probably want to stop using `EventClient`) and b) what they're best suited for? I think the latter would be particularly useful for discerning between `EventBuffer` and `EventSubscriber`. I'm worried that adding more internal test API without ripping out all of the old stuff will cause more disparate implementations, but hopefullly having some documented guidance will discourage that.

Created RUST-1888, RUST-1889, and RUST-1890 for follow-up.

For what it's worth, the only new API here is `watch_all`, the rest is just consolidating what we already had :)  Your point is quite valid, though, so I've added documentation and deprecation tags.

isabelatkinson (approved):

PR #1053, opened 2024-03-21
RUST-1855 Propagate cluster time for unified tests

abr-egn:
RUST-1855

This fixes some flaky unified tests; note that we'll still see failures in legacy tests until RUST-817 is done.

### DIFF
--- a/src/test/util.rs
+++ b/src/test/util.rs
@@ -25,6 +25,7 @@ use self::event_buffer::EventBuffer;
 #[cfg(feature = "in-use-encryption-unstable")]
 use crate::client::EncryptedClientBuilder;
 use crate::{
+    action::DropCollection,
     bson::{doc, Bson},
     client::options::ServerAddress,
     hello::{hello_command, HelloCommandResponse},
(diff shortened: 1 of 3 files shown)


isabelatkinson (approved):

src/test/util.rs
```diff
@@ -494,7 +484,16 @@ pub(crate) async fn drop_collection<T>(coll: &Collection<T>)
 where
     T: Serialize + DeserializeOwned + Unpin + Debug + Send + Sync,
 {
-    match coll.drop().await.map_err(|e| *e.kind) {
+    drop_collection_opt(coll, |a| a).await
+}
+
+pub(crate) async fn drop_collection_opt<'a, T>(
```
isabelatkinson, line 490: Just happened to notice that we already suppress namespace not found errors (i.e. code 26, RUST-392) in our drop collection implementation, so this wrapper shouldn't be necessary at all anymore

abr-egn: Oh, nice.  Removed!

PR #1055, opened 2024-03-25
RUST-1306 Move `Compression` enum behind compression feature flags

isabelatkinson:


src/cmap/establish/handshake.rs
```diff
@@ -315,8 +320,11 @@ pub(crate) struct Handshaker {
     /// given the same pool options, so it can be created at the time the Handshaker is created.
     command: Command,
 
-    // This field is not read without a compression feature flag turned on.
-    #[allow(dead_code)]
+    #[cfg(any(
```
abr-egn, line 323: We could avoid having to repeat this list everywhere (and update it everywhere if we add a new compressor) by making the `*-compression` features enable a `compression` feature.  Up to you if you think that's worthwhile.

abr-egn (approved):
LGTM!  No re-review needed if you decide to update with a `compression` feature.

src/cmap/establish/handshake.rs
isabelatkinson: Yeah, the repetition is annoying. I think it's manageable enough for now that it doesn't feel worth adding a technically-public feature flag, but I would definitely consider that if we do more compression-related work in the future.

It'd be really nice if we could define some kind of `#[compression]` attribute that expanded to this conditional compilation expression, but I don't think that's possible after cursory investigation of proc macro capabilities.

PR #1056, opened 2024-03-26
RUST-1792 Test order of heartbeat events

abr-egn:
RUST-1792

The Rust driver was already emitting the events in the correct order, so this is just implementing the prose test.


abr-egn:
As it turns out, this would be a flaky test :(  I've filed RUST-1895 for the root cause, I figure it'll be useful to have the disabled test checked in so it can just be uncommented once that's fixed.

isabelatkinson (approved):

PR #1058, opened 2024-03-27
RUST-1872 Convert index management methods to fluent style

abr-egn:
RUST-1872

Nothing unusual in this one.

### DIFF
--- a/src/action/create_collection.rs
+++ b/src/action/create_collection.rs
@@ -12,10 +12,10 @@ impl Database {
     ///
     /// `await` will return d[`Result<()>`].
     #[deeplink]
-    pub fn create_collection(&self, name: impl AsRef<str>) -> CreateCollection {
+    pub fn create_collection(&self, name: impl Into<String>) -> CreateCollection {
         CreateCollection {
             db: self,
-            name: name.as_ref().to_owned(),
+            name: name.into(),
             options: None,
             session: None,
         }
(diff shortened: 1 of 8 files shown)


src/action/create_collection.rs
```diff
@@ -12,10 +12,10 @@ impl Database {
     ///
     /// `await` will return d[`Result<()>`].
     #[deeplink]
-    pub fn create_collection(&self, name: impl AsRef<str>) -> CreateCollection {
+    pub fn create_collection(&self, name: impl Into<String>) -> CreateCollection {
```
isabelatkinson, line 15: This seems more idiomatic when we're converting into an owned `String` -- should we do this across the board? I think there are a few more inputs we should change outside the changes in this PR (e.g. the `WriteConcern` construction methods), but I can just file a separate ticket to do a full audit

src/action/gridfs/rename.rs
```diff
@@ -7,11 +7,11 @@ impl GridFsBucket {
     /// error if the `id` does not match any files in the bucket.
     ///
     /// `await` will return [`Result<()>`].
-    pub fn rename(&self, id: Bson, new_filename: impl AsRef<str>) -> Rename {
+    pub fn rename(&self, id: Bson, new_filename: impl Into<String>) -> Rename {
```
isabelatkinson, line 10: unrelated, but it'd be nice to accept `impl Into<Bson>` for these kinds of inputs to allow passing in something like an objectID directly. I can file a ticket for that as well as that sounds good

isabelatkinson (approved):
changes look good, just a few proposals for unrelated work

src/action/create_collection.rs
abr-egn: Yeah, I think a ticket for a general pass would make sense.  I did this here because I realized `name.as_ref().to_owned()` was silly.

src/action/gridfs/rename.rs
abr-egn: Yup, thank you!

PR #1060, opened 2024-03-28
RUST-1873 Transaction API fluency

abr-egn:
RUST-1873

This conversion PR has a couple of wrinkles.  First up, I converted `action_impl` from a regular macro to an attribute macro.  Regular macros can take as input any arbitrary string of tokens, which means that rustfmt will ignore their contents; attribute macros have to take input in the shape of normal Rust toplevel items, so rustfmt will happily process them, and that's a strong positive for the action impls.

The other is that I changed the convenient transaction API to chain off of `start_transaction` rather than being its own toplevel thing.  My motivation here was that this flows nicely especially when options are present:
```
session
    .start_transaction()
    .some_option(value)
    .and_run(...)
    .await
```
That reads well IMO and puts the options before the body of the transaction, which makes more sense to me than putting the options after as they would be if `with_transaction` were its own action.

### DIFF
--- a/src/action/transaction.rs
+++ b/src/action/transaction.rs
@@ -29,7 +29,7 @@ impl ClientSession {
     /// # let client = Client::with_uri_str("mongodb://example.com").await?;
     /// # let coll = client.database("foo").collection::<Document>("bar");
     /// # let mut session = client.start_session().await?;
-    /// session.start_transaction(None).await?;
+    /// session.start_transaction().await?;
     /// let result = coll.insert_one(doc! { "x": 1 }).session(&mut session).await?;
     /// session.commit_transaction().await?;
     /// # Ok(())
@@ -60,7 +60,7 @@ impl ClientSession {
     /// # let client = Client::with_uri_str("mongodb://example.com").await?;
     /// # let coll = client.database("foo").collection::<Document>("bar");
     /// # let mut session = client.start_session().await?;
-    /// session.start_transaction(None).await?;
+    /// session.start_transaction().await?;
     /// let result = coll.insert_one(doc! { "x": 1 }).session(&mut session).await?;
     /// session.commit_transaction().await?;
     /// # Ok(())
@@ -87,7 +87,7 @@ impl ClientSession {
     /// # let client = Client::with_uri_str("mongodb://example.com").await?;
     /// # let coll = client.database("foo").collection::<Document>("bar");
     /// # let mut session = client.start_session().await?;
-    /// session.start_transaction(None).await?;
+    /// session.start_transaction().await?;
     /// match execute_transaction(&coll, &mut session).await {
     ///     Ok(_) => session.commit_transaction().await?,
     ///     Err(_) => session.abort_transaction().await?,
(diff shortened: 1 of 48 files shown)


src/client/session/action.rs
```diff
@@ -0,0 +1,400 @@
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+#[cfg(feature = "sync")]
+impl<'a> StartTransaction<SyncSession<'a>> {
+    /// Synchronously execute this action.
```
abr-egn, line 224: I implemented this manually rather than having an `impl<'a> Action for StartTransaction<SyncSession<'a>>` because it doesn't make much sense to have a distinct future type or `IntoFuture` impl for the sync version.

src/action/transaction.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// `defaultTransactionOptions` will be used. This session must be passed into each operation
+    /// within the transaction; otherwise, the operation will be executed outside of the
+    /// transaction.
+    ///
+    /// Errors returned from operations executed within a transaction may include a
+    /// [`crate::error::TRANSIENT_TRANSACTION_ERROR`] label. This label indicates that the entire
+    /// transaction can be retried with a reasonable expectation that it will succeed.
+    ///
+    /// Transactions are supported on MongoDB 4.0+. The Rust driver currently only supports
+    /// transactions on replica sets.
```
isabelatkinson, line 23:
```suggestion
    /// Transactions on replica sets are supported on MongoDB 4.0+. Transactions on sharded
    /// clusters are supported on MongoDB 4.2+.
```

(this was probably just a copy-paste but that info is outdated)

src/action/transaction.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub fn start_transaction(&mut self) -> StartTransaction<AsyncSession> {
+        StartTransaction {
+            session: AsyncSession(self),
+            options: None,
+        }
+    }
+
+    /// Commits the transaction that is currently active on this session.
+    ///
+    ///
```
isabelatkinson, line 49:
```suggestion

```

src/action/transaction.rs
```diff
@@ -0,0 +1,240 @@
...
+/// Start a new transaction.  Construct with [`ClientSession::start_transaction`].
+#[must_use]
+pub struct StartTransaction<S> {
+    pub(crate) session: S,
+    pub(crate) options: Option<TransactionOptions>,
+}
+
+pub struct AsyncSession<'a>(pub(crate) &'a mut ClientSession);
+#[cfg(feature = "sync")]
+pub struct SyncSession<'a>(pub(crate) &'a mut crate::sync::ClientSession);
```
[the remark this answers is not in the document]
isabelatkinson: any reason we can't just use the wrapped reference types directly? i.e. return `StartTransaction<&'a mut ClientSession>` from `start_transaction` and so on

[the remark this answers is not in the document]
abr-egn: ... nope, thank you.  Done!

PR #1062, opened 2024-04-03
RUST-1208 Future-proof features

abr-egn:
RUST-1208 / RUST-1680 / RUST-1083

This PR:
* removes the various `bson-*` features
* adds the forward-compatibility feature
* makes `rustls` and `trust-dns` optional

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,10 +64,6 @@ features = ["sync"]
 |:---------------------|:--------------------------------------------------------------------------------------------------------------------------------------|:--------------------------------|:--------|
 | `sync`               | Expose the synchronous API (`mongodb::sync`).                                                                                         | n/a                             | no      |
 | `aws-auth`           | Enable support for the MONGODB-AWS authentication mechanism.                                                                          | `reqwest`                       | no      |
-| `bson-uuid-0_8`      | Enable support for v0.8 of the [`uuid`](docs.rs/uuid/0.8) crate in the public API of the re-exported `bson` crate.                    | n/a                             | no      |
-| `bson-uuid-1`        | Enable support for v1.x of the [`uuid`](docs.rs/uuid/1.0) crate in the public API of the re-exported `bson` crate.                    | n/a                             | no      |
-| `bson-chrono-0_4`    | Enable support for v0.4 of the [`chrono`](docs.rs/chrono/0.4) crate in the public API of the re-exported `bson` crate.                | n/a                             | no      |
-| `bson-serde_with`    | Enable support for the [`serde_with`](docs.rs/serde_with/latest) crate in the public API of the re-exported `bson` crate.             | `serde_with`                    | no      |
 | `zlib-compression`   | Enable support for compressing messages with [`zlib`](https://zlib.net/)                                                              | `flate2`                        | no      |
 | `zstd-compression`   | Enable support for compressing messages with [`zstd`](http://facebook.github.io/zstd/).                                               | `zstd`                          | no      |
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -27,46 +27,37 @@ exclude = [
...
-# Enable support for v1.x of the uuid crate in the public API of the BSON library.
-bson-uuid-1 = ["bson/uuid-1"]
+rustls-tls = ["dep:rustls", "dep:rustls-pemfile", "dep:tokio-rustls"]
+openssl-tls = ["dep:openssl", "dep:openssl-probe", "dep:tokio-openssl"]
+dns-resolver = ["dep:trust-dns-resolver", "dep:trust-dns-proto"]
 
 # Enable support for MONGODB-AWS authentication.
 # This can only be used with the tokio-runtime feature flag.
-aws-auth = ["reqwest"]
+aws-auth = ["dep:reqwest"]
```
abr-egn, line 39: Using the `dep:foo` format prevents the optional dependencies from automatically becoming their own feature flags.

src/client/options.rs
```diff
@@ -45,7 +44,10 @@ use crate::{
     srv::{OriginalSrvInfo, SrvResolver},
 };
 
+#[cfg(feature = "dns-resolver")]
 pub use resolver_config::ResolverConfig;
+#[cfg(not(feature = "dns-resolver"))]
+pub(crate) use resolver_config::ResolverConfig;
```
abr-egn, line 50: Use of DNS resolution is ... not neatly encapsulated in the codebase, so to keep this from being a giant refactor PR I made the various machinery crate-internal when the feature is disabled rather than being feature-gated entirely.

Cargo.toml
```diff
@@ -27,46 +27,37 @@ exclude = [
...
-bson-serde_with = ["bson/serde_with"]
-
-# Enable support for v0.8 of the uuid crate in the public API of the BSON library.
-bson-uuid-0_8 = ["bson/uuid-0_8"]
-
-# Enable support for v1.x of the uuid crate in the public API of the BSON library.
-bson-uuid-1 = ["bson/uuid-1"]
+rustls-tls = ["dep:rustls", "dep:rustls-pemfile", "dep:tokio-rustls"]
+openssl-tls = ["dep:openssl", "dep:openssl-probe", "dep:tokio-openssl"]
+dns-resolver = ["dep:trust-dns-resolver", "dep:trust-dns-proto"]
```
abr-egn, line 35: I'm not sure whether giving this feature a generic name is the right way (treating the actual dependency as an internal detail) or if it should just explicitly be called "trust-dns-resolver" in case we want to add other resolver crates in the future.

Cargo.toml
isabelatkinson: nice, I think the need to do this will go away in the 2024 edition https://github.com/rust-lang/cargo/issues/12826

Cargo.toml
isabelatkinson: Hmm, I think I'm in favor of keeping it generic. I was curious what the status of `trust-dns-resolver` was and it looks like they've actually "rebranded" to a different crate (see [here](https://docs.rs/trust-dns-resolver/latest/trust_dns_resolver/)) that we might want to switch over to at some point. Keeping the name neutral will make it easier for us to swap out the internal dep.

Worst case, if we did want to introduce more resolver crates we could treat this flag as our "default" choice and then allow users to opt into a different one if they want to.

isabelatkinson (approved):
lgtm!

PR #1067, opened 2024-04-05
RUST-1860 Accept ergonomic conversions for specific option types

abr-egn:
RUST-1860

The hammer of `proc_macro` looked like a good fit for this nail as well, since it means that the `option_setters!` macro can just have special handling for the specific types rather than having to handcode it for every action.

### DIFF
--- a/action_macro/src/lib.rs
+++ b/action_macro/src/lib.rs
@@ -1,6 +1,5 @@
 extern crate proc_macro;
 
-use proc_macro2::Span;
 use quote::{quote, ToTokens};
 use syn::{
     braced,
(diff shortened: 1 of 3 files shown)


action_macro/src/lib.rs
```diff
@@ -55,23 +57,10 @@ pub fn action_impl(
...
-                    }
-                }
-            }
-        }
-        syn::visit_mut::visit_type_mut(&mut Visitor, &mut formal);
         quote! {
             /// Synchronously execute this action.
             pub fn run(self) -> Result<#sync_type> {
-                crate::sync::TOKIO_RUNTIME.block_on(std::future::IntoFuture::into_future(self)).map(#formal::new)
+                crate::sync::TOKIO_RUNTIME.block_on(std::future::IntoFuture::into_future(self)).map(<#sync_type>::new)
```
abr-egn, line 63: Coincidentally, when reading the documentation for `quote` I discovered that this is a much easier way of handling the type-position vs expr-position issue for qualified types.

isabelatkinson (approved):
nice!

PR #1070, opened 2024-04-09
RUST-1364 Collapse comment and comment_bson

abr-egn:


src/coll/options.rs
```diff
@@ -776,22 +776,11 @@ pub struct FindOptions {
...
-
     /// Tags the query with an arbitrary [`Bson`] value to help trace the operation through the
     /// database profiler, currentOp and logs.
     ///
-    /// This option is only supported on server versions 4.4+. Use the `comment` option on
-    /// older server versions.
-    // TODO RUST-1364: Remove this field
-    #[serde(rename(serialize = "comment"))]
-    pub comment_bson: Option<Bson>,
+    /// This option is only supported on server versions 4.4+.
```
isabelatkinson, line 782:
```suggestion
    /// For server versions less than 4.4, only a string value may be provided.
```

Ditto below. `comment` was added as string-only in 3.6 and updated to be any BSON value in 4.4, but we can rely on the server to return an error if a non-string value is used for a server < 4.4.

Same suggestion for the `comment` option in `AggregateOptions` -- looks like we've removed `comment_bson` already but the docs weren't updated.

PR #1073, opened 2024-04-10
RUST-1890 Remove some uses of EventClient

abr-egn:
RUST-1890

This removes uses of `EventClient::with_options` / `with_additional_options` in favor of using `Client::test_builder` and explicitly passing an `EventBuffer` around with it.

I'm not sure this change is worth the churn, honestly.  Given that it just means things have to pass around `(TestClient, EventBuffer)` anyway it seems better to keep them bundled in a struct.  I'm inclined to undo that part of this PR and update the rest to be removing the direct constructors in favor of the builder (so keep `EventClient` and `TestClientBuilder::event_client()` but get rid of `EventClient::new` / `with_options` / `with_additional_options`).  Thoughts?

### DIFF
--- a/src/client/session/test.rs
+++ b/src/client/session/test.rs
@@ -386,12 +386,15 @@ async fn session_usage() {
     #[allow(deprecated)]
     async fn session_usage_test<F, G>(command_name: &str, operation: F)
     where
-        F: Fn(EventClient) -> G,
+        F: Fn(Client) -> G,
         G: Future<Output = ()>,
     {
-        let client = EventClient::new().await;
-        operation(client.clone()).await;
-        let mut events = client.events.clone();
+        let client = Client::test_builder()
+            .monitor_command_events()
+            .build()
+            .await;
+        operation(client.clone().into_client()).await;
+        let mut events = client.event_buffer().clone();
         #[allow(deprecated)]
         let (command_started, _) = events.get_successful_command_execution(command_name);
         assert!(
--- a/src/cmap/test.rs
+++ b/src/cmap/test.rs
@@ -8,8 +8,6 @@ use tokio::sync::{Mutex, RwLock};
 
 use self::file::{Operation, TestFile, ThreadedOperation};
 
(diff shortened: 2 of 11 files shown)


isabelatkinson:
>  Given that it just means things have to pass around `(TestClient, EventBuffer)` anyway it seems better to keep them bundled in a struct.

That's fair. My qualms with the `EventClient` type are mostly related to the divergent construction methods and implementations, but if everything is unified under one builder pattern then I think it'll be less of a pain. An alternative could be adding a generic type on `TestClient` to consolidate into one type, e.g.:

```
struct NoEventBuffer;

struct TestClient<T = NoEventBuffer> {
    event_buffer: T,
    ...
}

impl Client {
    pub(crate) fn test_builder() -> TestClientBuilder<NoEventBuffer> { ... }
}

impl<T> TestClientBuilder<T> {
    // duplicate for sdam and cmap events -- only monitoring the events that are relevant will probably
    // be a small performance improvement
    pub(crate) fn monitor_command_events(mut self) -> TestClientBuilder<EventBuffer> {
        // set flag to monitor command events
    }

    ...rest of builder methods
}

impl TestClientBuilder<NoEventBuffer> {
    pub(crate) async fn build(self) -> TestClient { ... }
}

impl TestClientBuilder<EventBuffer> {
    pub(crate) async fn build(self) -> TestClient<EventBuffer> {
        // create EventBuffer buffer
        // set buffer as handler for event types requested on client options
        // set buffer as event_buffer field on TestClient
        // create inner client and return
    }
}

impl TestClient<EventBuffer> {
    pub(crate) fn event_buffer(&self) -> &EventBuffer {
        self.event_buffer
    }
}

#[tokio::test]
async fn example_use() {
    let client = Client::test_builder().monitor_command_events().build().await;
    let event_buffer = client.event_buffer();
}
```

Not sure this would have a huge impact since `EventClient` derefs into `TestClient` anyway but it might feel simpler?

PR #1074, opened 2024-04-10
RUST-1890 Consolidate test event monitoring and client construction

abr-egn:
RUST-1890

This makes `EventClient` the sole way for tests to track events, and the builder the sole way to construct an `EventClient`.

I did attempt the parameterized `TestClient` path but the type machinery ended up more complicated there than it seemed like it was worth; the approach in this PR also has the advantage that the event handling code is nicely separated from the base `TestClient` code.

### DIFF
--- a/src/client/session/test.rs
+++ b/src/client/session/test.rs
@@ -389,7 +389,7 @@ async fn session_usage() {
         F: Fn(EventClient) -> G,
         G: Future<Output = ()>,
     {
-        let client = EventClient::new().await;
+        let client = Client::test_builder().monitor_events().build().await;
         operation(client.clone()).await;
         let mut events = client.events.clone();
         #[allow(deprecated)]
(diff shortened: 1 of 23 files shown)


src/client/session/test.rs
```diff
@@ -389,7 +389,7 @@ async fn session_usage() {
         F: Fn(EventClient) -> G,
         G: Future<Output = ()>,
     {
-        let client = EventClient::new().await;
+        let client = Client::test_builder().monitor_events().build().await;
```
abr-egn, line 392: This is a little more verbose, but only slightly, and I think that's okay because it makes it easy to include other builder setup and removes the temptation for further constructors in `EventClient`.

src/test/util/event.rs
```diff
@@ -122,94 +118,64 @@ impl std::ops::Deref for EventClient {
...
         &mut self.client
     }
 }
 
 impl TestClientBuilder {
-    #[deprecated = "use EventBuffer directly"]
-    #[allow(deprecated)]
-    pub(crate) fn event_client(self) -> EventClientBuilder {
-        EventClientBuilder { inner: self }
+    pub(crate) fn monitor_events(self) -> EventClientBuilder {
```
abr-egn, line 128: I renamed this because it makes it more clear what's going on at the call site.

PR #1079, opened 2024-04-16
RUST-1775 Remove manual in favor of a link to the docs page

abr-egn:


docs/manual/index.html
abr-egn, line 1: This is the only file retained; everything else is a deletion.

isabelatkinson (approved):

PR #1084, opened 2024-04-18
RUST-817 Convert transactions tests to unified format

abr-egn:
RUST-817

Along with the test file conversion, there were a few bugs found and fixed and some substantial test runner implementation wrinkles - the latter I've noted inline.


src/client/executor.rs
```diff
@@ -543,6 +543,9 @@ impl Client {
                     TransactionState::Starting => {
                         cmd.set_start_transaction();
                         cmd.set_autocommit();
+                        if session.causal_consistency() {
```
abr-egn, line 546: The spec requires that the first operation in a causally consistent transaction have `afterClusterTime` attached, even if it's a write; previously the driver would only do that if it was a read.  This wasn't caught in the legacy tests because we had a skip in there pre-dating the implementation of causal consistency that never got removed :)

src/test/spec/unified_runner/entity.rs
```diff
@@ -52,6 +53,12 @@ pub(crate) enum Entity {
     None,
 }
 
+#[derive(Debug)]
+pub(crate) struct SessionPtr(pub(crate) *mut ClientSession);
```
abr-egn, line 57: This is really ugly and I tried a lot of things to avoid it but none of them ended up being feasible.  For the legacy runner, this was avoidable because there's a fixed number of sessions created for the tests and those could just be put into their own structure+lifetime and pulled out of the rest of the test runner.  For the unified runner, sessions are created dynamically, with no statically known number or lifetime.

Storing a raw pointer lets us elide the lifetime; when it's retrieved it's cast to a `&mut ClientSession` with a strictly shorter lifetime than the "real" one, but not in a way that I could prove to the compiler, so ... `unsafe` :(

src/test/spec/unified_runner/operation.rs
```diff
@@ -136,21 +137,47 @@ macro_rules! with_mut_session {
...
-                .write()
-                .await
-                .insert(id.to_string(), Entity::Session(session_owned));
-            out
+            }
+        }
+    };
+}
+
+macro_rules! with_opt_session {
```
abr-egn, line 171: This is an unrelated convenience; many of the operation impls work in exactly the same way, so this captures that pattern and makes 'em easier to read.

src/test/spec/unified_runner/operation.rs
```diff
@@ -164,13 +191,19 @@ pub(crate) struct Operation {
 }
 
 impl Operation {
-    pub(crate) async fn execute<'a>(&self, test_runner: &TestRunner, description: &str) {
+    pub(crate) async fn execute(&self, test_runner: &TestRunner, description: &str) {
+        let _ = self.execute_fallible(test_runner, description).await;
+    }
+
+    async fn execute_fallible(&self, test_runner: &TestRunner, description: &str) -> Result<()> {
```
abr-egn, line 198: The test `withTransaction` operation needs to have any error from the child operations propagated, independently of their `expectError` fields, so this makes that available.

src/test/spec/unified_runner/operation.rs
```diff
@@ -1891,8 +1901,59 @@ impl TestOperation for AbortTransaction {
...
+        async move {
+            with_mut_session!(test_runner, id, |session| async move {
+                session
+                    .start_transaction()
+                    .with_options(self.options.clone())
+                    .and_run(
+                        (&self.callback, test_runner),
+                        |session, (callback, test_runner)| {
+                            async move {
+                                test_runner.entities.write().await.insert(
```
abr-egn, line 1937: This is the crux of the entity issue:
* SessionEntity is an owned value, but the callback to `start_transaction().and_run` (necessarily) receives a `&mut ClientSession`
* For the child operations to be able to access the session, it needs to be in the session map
* There's no expressable lifetime for the entity map to be parameterized on

To work around all of that, `WithTransaction` provides a raw pointer session for the entity map that `with_mut_session` and `as_mut_session` translate back into a `&mut`.

src/test/util/matchable.rs
```diff
@@ -157,21 +157,9 @@ impl Matchable for Document {
             if k == "upsertedCount" {
                 continue;
             }
-            // TODO RUST-48: Remove this logic to bypass read concerns with an afterClusterTime
```
abr-egn, line 160: Even though the relevant tests are no longer using the legacy runner, I removed these skips in case other tests happen to be affected by them.

isabelatkinson (approved):

PR #1095, opened 2024-05-03
RUST-1921 Sign crate on release

abr-egn:
RUST-1921

This works similarly to the papertrail integration, but there's no custom evergreen command this time :)


.evergreen/releases.yml
```diff
@@ -137,19 +138,38 @@ functions:
...
+          - ARTIFACTORY_USERNAME
+          - ARTIFACTORY_PASSWORD
+          - GARASIGN_USERNAME
+          - GARASIGN_PASSWORD
+          - CRATE_VERSION
+        binary: bash
+        args:
+          - .evergreen/release-sign.sh
+
+  "save signature":
```
abr-egn, line 159: This is a hack; once DEVPROD-7047 is done this can just be an `s3.put` of the signature file specifically and then we can include it in github releases / link to the URL / whatever.

isabelatkinson (approved):

PR #1101, opened 2024-05-10
Fix #1099 and allow `&mut ClientSession` to be used without moving

### LINKED ISSUE #1099: `.session(...)` builder func moves `&mut ClientSession` preventing it from being reused
## Versions/Environment
1. What version of Rust are you using? rustc 1.78.0 (9b00956e5 2024-04-29)
2. What operating system are you using? macOS M1
3. What versions of the driver and its dependencies are you using? (Run 
`cargo pkgid mongodb` & `cargo pkgid bson`) main branch
4. What version of MongoDB are you using? (Check with the MongoDB shell using `db.version()`) N/A
5. What is your MongoDB topology (standalone, replica set, sharded cluster, serverless)? N/A

## Describe the bug

After switching to the main branch, I can no longer use session on two separate 

```rust
col.find_one(doc! {}).session(session).await;
col.find_one(doc! {}).session(session).await;
```

```
error[E0382]: use of moved value: `session`
  --> file.rs:21:48
   |
11 |     session: &mut ClientSession,
   |     ------- move occurs because `session` has type `&mut mongodb::ClientSession`, which does not implement the `Copy` trait
...
20 |     let supply = col.find_one(doc! {}).session(session).await;
   |                                                ------- value moved here
21 |     let supply = col.find_one(doc! {}).session(session).await;
   |                                                ^^^^^^^ value used here after move
```

For reference, I can get this to work

```rust
use_session(session);
use_session(session);

fn use_session(session: &mut ClientSession) {
    let session = session;
}
```

but I can reproduce the error with this

```rust
use_session(session);
use_session(session);

fn use_session<'a>(session: impl Into<&'a mut ClientSession>) {
    let session = session.into();
}
```

**BE SPECIFIC**:
* What is the _expected_ behavior and what is _actually_ happening? I would expect to be able to use the session on multiple queries, but due to the `Into` semantics, Rust thinks the `.session(...)` method is moving the session.
* Do you have any particular output that demonstrates this problem? N/A
* Do you have any ideas on _why_ this may be happening that could give us a
clue in the right direction? Example illustrates why this is happening, and one potential fix.
* Did this issue arise out of nowhere, or after an update (of the driver,
server, and/or Rust)? After switching to the main branch.

clarkmcc:
This is a breaking change, though in most cases I suspect it won't actually break code unless people are truly passing non-`ClientSession` types that are `Into<&mut ClientSession>`. None of the tests in the repo were doing this that I could find.

In any case, not sure if this is a good idea, just want to get some eyes on the issue.

### DIFF
--- a/src/action/aggregate.rs
+++ b/src/action/aggregate.rs
@@ -123,13 +123,13 @@ impl<'a> Aggregate<'a, ImplicitSession> {
     /// Use the provided session when running the operation.
     pub fn session(
         self,
-        value: impl Into<&'a mut ClientSession>,
+        value: &'a mut ClientSession,
     ) -> Aggregate<'a, ExplicitSession<'a>> {
         Aggregate {
             target: self.target,
             pipeline: self.pipeline,
             options: self.options,
-            session: ExplicitSession(value.into()),
+            session: ExplicitSession(value),
         }
     }
 }
--- a/src/action/count.rs
+++ b/src/action/count.rs
@@ -133,8 +133,8 @@ impl<'a> CountDocuments<'a> {
     );
 
     /// Use the provided session when running the operation.
-    pub fn session(mut self, value: impl Into<&'a mut ClientSession>) -> Self {
-        self.session = Some(value.into());
+    pub fn session(mut self, value: &'a mut ClientSession) -> Self {
+        self.session = Some(value);
         self
     }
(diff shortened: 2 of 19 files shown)


isabelatkinson:
Closing this based on the conversation in #1099

PR #1103, opened 2024-05-15
RUST-1875 Define separate types for summary and verbose bulk write results

isabelatkinson:


src/error/bulk_write.rs
```diff
@@ -4,27 +4,46 @@ use std::collections::HashMap;
...
+    Verbose(VerboseBulkWriteResult),
+}
+
+impl PartialBulkWriteResult {
+    pub(crate) fn merge(&mut self, other: Self) {
+        match (self, other) {
+            (Self::Summary(this), Self::Summary(other)) => this.merge(other),
+            (Self::Verbose(this), Self::Verbose(other)) => this.merge(other),
+            // The operation execution path makes this an unreachable state
+            _ => {}
```
abr-egn, line 32: Maybe `unreachable!()` here?

src/results/bulk_write.rs
```diff
@@ -2,84 +2,180 @@
...
+    pub deleted_count: i64,
+}
+
+#[cfg_attr(test, serde_with::skip_serializing_none)]
+#[derive(Clone, Debug, Default)]
+#[cfg_attr(test, derive(serde::Serialize))]
+#[cfg_attr(test, serde(rename_all = "camelCase"))]
+#[non_exhaustive]
+pub struct VerboseBulkWriteResult {
+    pub inserted_count: i64,
```
abr-egn, line 29: I wonder if it'd be worth having `pub summary: SummaryBulkWriteResult` instead of duplicating the fields?  API-wise it makes the superset relationship clear, and it would mean the impl would have a little less repeated code.

abr-egn (approved):
LGTM!  A couple of potential changes but I don't feel strongly about either, and no need for a re-review either way.

src/results/bulk_write.rs
isabelatkinson: yeah I think that makes sense. this might be the first time I've wanted subclasses in Rust...

PR #1104, opened 2024-05-15
RUST-1858 Add builders for bulk write models

isabelatkinson:
This also has the changes from #1103, filtering by the last commit will show the relevant changes for this PR.

### DIFF
--- a/src/test/spec/unified_runner/operation/bulk_write.rs
+++ b/src/test/spec/unified_runner/operation/bulk_write.rs
@@ -3,12 +3,19 @@ use futures_util::FutureExt;
 use serde::Deserialize;
 
 use crate::{
-    bson::{to_bson, Array, Bson, Document},
-    coll::options::UpdateModifications,
+    bson::to_bson,
     error::Result,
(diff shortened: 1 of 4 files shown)


src/test/spec/unified_runner/operation/bulk_write.rs
```diff
@@ -31,136 +38,22 @@ impl<'de> Deserialize<'de> for WriteModel {
         #[derive(Debug, Deserialize)]
         #[serde(rename_all = "camelCase")]
         enum WriteModelHelper {
```
abr-egn, line 40: Is this impl still needed?  This looks equivalent to what I think `#[derive(Deserialize)]` on `WriteModel` would produce now.

isabelatkinson: [this limitation](https://github.com/mongodb/mongo-rust-driver/pull/1034#discussion_r1499779896) still holds unfortunately, the test files use a "tagged" representation whereas the `ops` array doesn't

abr-egn (approved):

PR #1105, opened 2024-05-16
RUST-1827 remove generic parameter to `Command`

abr-egn:
I thought we had a jira ticket for this, but I didn't find one.

This has been bugging me for a while - the type parameter here was always `Document` or `RawDocumentBuf`, and there's no actual reason for it to differ across commands.  In most cases this was a nearly drop-in update, in some cases I had to do a bit more refactoring, and in a very few I kept the construction as `Document` and converted at the end because it seemed like it would be a giant hassle to do otherwise.

### DIFF
--- a/src/bson_util.rs
+++ b/src/bson_util.rs
@@ -1,7 +1,10 @@
 use std::{
+    collections::HashSet,
     convert::TryFrom,
     io::{Read, Write},
 };
+
+use serde::Serialize;
 
 use crate::{
     bson::{
@@ -15,7 +18,7 @@ use crate::{
         RawDocumentBuf,
     },
     checked::Checked,
-    error::{ErrorKind, Result},
+    error::{Error, ErrorKind, Result},
     runtime::SyncLittleEndianRead,
 };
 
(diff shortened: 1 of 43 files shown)


src/bson_util.rs
```diff
@@ -162,13 +172,42 @@ pub(crate) fn extend_raw_document_buf(
     this: &mut RawDocumentBuf,
     other: RawDocumentBuf,
 ) -> Result<()> {
+    let mut keys: HashSet<String> = HashSet::new();
```
abr-egn, line 175: I got nervous about this mostly in the context of `append_options_to_raw_document`, where it would silently generate invalid documents if there were duplicate keys, whereas `append_options` would silently overwrite the old values.  This does introduce the overhead of building the key map but I think that won't be significant, since it'll still be less overhead than constructing the body as a `Document` would have been.

isabelatkinson (approved):
lgtm!

> I thought we had a jira ticket for this, but I didn't find one.

RUST-1827 🙂

PR #1108, opened 2024-05-17
Macro for field projection proposal

yourlogarithm:
Usage example

```bash
cargo add mongodb -F projection_macro
```

```rust
use mongodb::{
    bson::Document,
    options::{FindOptions, Project}
};
use serde::Deserialize;

#[derive(Project, Deserialize)]
struct Person {
    #[serde(rename = "n")]
    name: String,
    age: u8,
    male: bool,
    #[serde(rename = "h.a")]
    address: String,
}

#[tokio::main]
async fn main() {
    let client = mongodb::Client::with_uri_str("mongodb://localhost:27017").await.unwrap();
    let projection = Person::project().name().age().address().build();
    println!("{:?}", projection);  // mongodb::bson::Document({"n": Int32(1), "age": Int32(1), "h.a": Int32(1)})
    let options = FindOptions::builder().projection(projection).build();
    let document: Document = client.database("test").collection("people").find_one(None, options).unwrap().unwrap();
}
```


abr-egn:
If I understand correctly, the use case here is to build the document for projection with field names matching those that will be generated by serde.  I have two concerns:
* It looks like this is doing its own parsing of serde attributes, so for example it'll be incorrect if `rename_all` is used.  Is there any way to get the information from serde directly?
* Is there any reason this needs to be part of the mongodb driver crate?  If it could feasibly be its own crate that would likely be better all around.

PR #1109, opened 2024-05-21
RUST-1934 Publish internal macros as their own crate

abr-egn:
RUST-1934

Rust requires that macros be in a crate distinct from where they're used, and cargo doesn't allow crates with internal sub-crates to be published, so published as its own crate it shall be :(

My aim here is for this to be as invisible as possible; the `mongodb-internal-macros` crate will always have the same version number as the main `mongodb` crate, and will be published/signed/attested by the same scripts, so unless something goes wrong we won't ever have to think about it.


etc/update_version/src/main.rs
```diff
@@ -74,45 +74,53 @@ struct Args {
 }
 
 fn main() {
+    let zero = std::env::current_exe().unwrap();
```
abr-egn, line 77: This is unrelated but I got annoyed at having to prepend "../.." to every path.

etc/update_version/src/main.rs
```diff
@@ -74,45 +74,54 @@ struct Args {
...
         ),
     ];
-    let bson_version_loc =
-        Location::new("../../Cargo.toml", r#"bson = (?<target>\{ git = .*? \})\n"#);
-    let mongocrypt_version_loc = Location::new(
-        "../../Cargo.toml",
-        r#"mongocrypt = (?<target>\{ git = .*? \})\n"#,
+    let quote_version_loc = Location::new(
+        "Cargo.toml",
+        r#"mongodb-internal-macros = (?<target>\{ path = .* \})\n"#,
```
kevinAlbs, line 104: I expect the regex will no longer match once `mongodb-internal-macros` is added to Cargo. Suggest adding a comment to remind:
```suggestion
        // TODO: after `mongodb-internal-macros` is added to Cargo, update regex to:
        // r#"mongodb-internal-macros = (?<target>.*?)\n"#
        r#"mongodb-internal-macros = (?<target>\{ path = .* \})\n"#,
```

kevinAlbs (approved):
LGTM. Noted a possible comment.

etc/update_version/src/main.rs
abr-egn: I don't think this will be needed - `mongodb-internal-macros` will only be updated to point to crates.io for releases, on the dev branch it'll always be a `path` dependency.  We treat `bson` similarly, except with a `git` dependency rather than `path` since it's not in the same repo.

PR #1110, opened 2024-05-22
RUST-1941 Test against server 8.0

abr-egn:
RUST-1941

This PR is actually mostly about using a generation script for that chunk of the evergreen config, but it does add 8.0 to the version list :)

I tried using evergreen's `generate.tasks` for this initially, but because those tasks with their associated tags don't exist on configuration load it fails validation.  I think `include` is a better match here anyway since it's static generation rather than at test time.


kevinAlbs (commented):
I like the generation script. LGTM!

kevinAlbs (approved):

PR #1116, opened 2024-06-04
minor: fix rustdoc failures

abr-egn:


Cargo.toml
```diff
@@ -174,3 +174,6 @@ serde_path_to_error = "0.1"
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
 all-features = true
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(mongodb_internal_tracking_arc)'] }
```
isabelatkinson, line 179: Thanks for fixing this, I started looking into it and got stuck on how to actually compile using this flag. I tried various ways of passing `--cfg mongodb_internal_tracking_arc` to `cargo build`/`rustc` and couldn't get it to work. Have you been able to use this?

abr-egn: Yup - the way to do this is via the `RUSTFLAGS` env var:
```
RUSTFLAGS="--cfg mongodb_internal_tracking_arc" cargo check
```

abr-egn:
The many runs with `SYSTEM FAILED` are due to an unrelated mongo-orchestration bug.

Cargo.toml
isabelatkinson: nice - thanks!

isabelatkinson (approved):

PR #1120, opened 2024-06-05
RUST-1678 Switch junit merge tools

abr-egn:
RUST-1678

Turns out someone beat me to it.  Also removed some shell scripts that were no longer being used that I forgot to remove in earlier PRs.


isabelatkinson (approved):
nice!

PR #1121, opened 2024-06-05
RUST-1965 Use the implicit runtime when spawning tasks via the sync API

abr-egn:
RUST-1965

Without this, anything using handles that are tracked by the graceful shutdown support causes a panic.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -197,7 +197,7 @@ buildvariants:
 
   - name: sync-api
     display_name: "Sync API"
-    patchable: false
+    #patchable: false
     run_on:
       - rhel87-small
     expansions:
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -197,7 +197,7 @@ buildvariants:
 
   - name: sync-api
     display_name: "Sync API"
-    patchable: false
+    #patchable: false
```
abr-egn, line 200: Will revert before merge.

isabelatkinson: Maybe we should be running this task on all PRs? The tests don't take that long and it'd be nice to catch stuff like this earlier

isabelatkinson (approved):
lgtm with one suggestion, can go either way on that

PR #1123, opened 2024-06-05
RUST-1967 Fix cloud provider test binaries

abr-egn:
RUST-1967

These also got hit by my rustfmt-on-save editor configuration; I can revert that if it's annoying.  The actual updates are all just tweaking some call sites for the fluent API changes.

### DIFF
--- a/.evergreen/aws-ecs-test/src/main.rs
+++ b/.evergreen/aws-ecs-test/src/main.rs
@@ -1,14 +1,17 @@
-use mongodb::{bson::Document, Client};
+use mongodb::{
+    bson::{doc, Document},
+    Client,
+};
 
 #[tokio::main]
 async fn main() {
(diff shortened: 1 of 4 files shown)


abr-egn:
> I _think_ the lambda test is still failing based on this log line: https://evergreen.mongodb.com/task_log_raw/mongo_rust_driver_aws_lambda_test_aws_lambda_deployed_patch_19d0460b895ef1f7a9f08401f551546017b2eccb_6660aee253db1b000738fd95_24_06_05_18_31_01/1?type=T#L1316

You're right, but at least the test binary is actually being run now :)  I'd like to merge this PR and then follow up on that test in a different one if that works for you.

isabelatkinson (approved):
> I'd like to merge this PR and then follow up on that test in a different one if that works for you.

sounds good!

PR #1147, opened 2024-06-25
RUST-1845 Support search index type field

abr-egn:
RUST-1845

Pretty straightforward.  The test isn't patchable so [here's](https://spruce.mongodb.com/version/667af50538cea5000730c50e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) the passing run.

### DIFF
--- a/src/test/spec/index_management.rs
+++ b/src/test/spec/index_management.rs
@@ -7,8 +7,10 @@ use bson::{doc, oid::ObjectId, Document};
 use futures_util::TryStreamExt;
 
 use crate::{
+    search_index::SearchIndexType,
     test::{log_uncaptured, spec::unified_runner::run_unified_tests},
     Client,
+    Collection,
     SearchIndexModel,
 };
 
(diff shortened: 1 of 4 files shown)


src/test/spec/index_management.rs
```diff
@@ -272,3 +274,117 @@ async fn search_index_drop_not_found() {
...
+    let coll_name = ObjectId::new().to_hex();
+    let db = client.database("search_index_test");
+    db.create_collection(&coll_name).await.unwrap();
+    let coll0 = db.collection::<Document>(&coll_name);
+
+    let result = coll0
+        .create_search_index(
+            SearchIndexModel::builder()
+                .name(String::from("test-search-index-case7-vector"))
+                .index_type(SearchIndexType::VectorSearch)
```
isabelatkinson, line 371: The [definition](https://github.com/mongodb/specifications/blob/master/source/index-management/tests/README.md#case-8-driver-requires-explicit-type-to-create-a-vector-search-index) for this test doesn't include the `type` field in the index model -- this probably shouldn't be passing written as-is? I think the `create_search_index` call here is equivalent to the last one in the test above

abr-egn: As it turns out, the test _was_ failing (i.e. the index creation was succeeding), but when I kicked off the evergreen run I accidentally left in a restriction that filtered out the test entirely.  I've fixed the test and also updated the evergreen config to include the test-level breakdown which would have made that more obvious.  New passing run is [here](https://spruce.mongodb.com/task/mongo_rust_driver_search_index_test_search_index_patch_45736f241311cf2e2d0a44708e421c8cd9282203_667d96433aed0b0007e32def_24_06_27_16_41_44?execution=0&sortBy=STATUS&sortDir=ASC) :)

isabelatkinson (approved):

PR #1155, opened 2024-06-27
RUST-1982 Fix aggregate with_type with explicit session

abr-egn:
RUST-1982

We accidentally overspecialized the impls here, meaning both `aggregate(..).session(..).with_type::<T>()` and `aggregate(..).with_type::<T>().session(..)` would fail to compile for slightly different reasons.

### DIFF
--- a/src/action/aggregate.rs
+++ b/src/action/aggregate.rs
@@ -218,8 +218,9 @@ impl<'a, T> Action for Aggregate<'a, ExplicitSession<'a>, T> {
             self.options,
         );
         let client = self.target.client();
+        let session = self.session;
         client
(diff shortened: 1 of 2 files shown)


isabelatkinson (approved):
changes LGTM, just looks like there's an MSRV failure that requires an explicit `Send` bound on the action type

abr-egn:
> changes LGTM, just looks like there's an MSRV failure that requires an explicit `Send` bound on the action type

Happily, it turned out just being explicit that 'self' doesn't actually cross the await point fixed this without the need for a bound.  I guess the newer compiler version is smart enough to notice that itself.

PR #1156, opened 2024-06-27
minor: migrate from `trust-dns` to `hickory`, bump MSRV to 1.67

isabelatkinson:
The `trust-dns` libraries have migrated to new namespaces and are no longer being updated; this switches our dependencies to the new ones so that we can continue to pull in updates/fixes.


abr-egn:
It looks like hickory requires rust 1.67; I'd be fine with bumping our msrv to that since that's over a year and a half old.

abr-egn (approved):

PR #1162, opened 2024-07-03
RUST-1986 Use Collection type parameter for change streams

abr-egn:
RUST-1986

This is technically a breaking change but I believe that the circumstances that would cause breakage here are sufficiently narrow that fixing a functionality regression from 2.x is more important.


isabelatkinson (approved):
lgtm, just a docs suggestion!

PR #1168, opened 2024-07-16
RUST-1879 Convert read/write concern operation tests to unified format

isabelatkinson:
With this change, the CSFLE tests are the only remaining tests that use the v2 runner 🙂 Unfortunately we're using some of the v2 machinery for other tests, so we can't put the whole module behind the `in-use-encryption-unstable` feature flag.


abr-egn (approved):
Nice!

PR #1179, opened 2024-07-31
RUST-1997 Fix evergreen benchmarks config

abr-egn:
RUST-1997

Passing run [here](https://spruce.mongodb.com/version/66aa826b58c08f00070bf7ef/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) (except for gridfs on a sharded deployment, for which I've filed RUST-2010).

### DIFF
--- a/.evergreen/benchmarks.yml
+++ b/.evergreen/benchmarks.yml
@@ -138,10 +138,7 @@ functions:
     - command: expansions.update
       params:
         file: mo-expansion.yml
-
-  "run driver benchmarks":
     - command: shell.exec
-      type: test
       params:
         shell: bash
         working_dir: "src"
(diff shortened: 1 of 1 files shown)


.evergreen/benchmarks.yml
```diff
@@ -150,7 +147,20 @@ functions:
           export MONGODB_URI="${MONGODB_URI}"
           export SSL="${SSL}"
           . .evergreen/generate-uri.sh
+    - command: expansions.update
+      params:
+        file: src/uri-expansions.yml
 
+  "run driver benchmarks":
+    - command: shell.exec
```
isabelatkinson, line 155: can this be subprocess.exec instead?

abr-egn: Diff to the contrary, this is just preserved from the existing stanza; switching to subprocess.exec will require sorting out `${PREPARE_SHELL}` (it does more than just re-exporting variables) and I think that'd be better done for all the commands in this file as its own chunk of work.

isabelatkinson (approved):

PR #1184, opened 2024-08-12
RUST-1926 Support Range Indexes as GA

abr-egn:
RUST-1926 / RUST-1999

Minimal code changes plus a new prose test and a lotta spec test syncs.


src/test/csfle.rs
```diff
@@ -3436,6 +3436,69 @@ async fn bind(addr: &str) -> Result<TcpListener> {
     Ok(TcpListener::bind(addr.parse::<std::net::SocketAddr>()?).await?)
 }
 
+// Prose test 23. Range explicit encryption applies defaults
+#[tokio::test]
+async fn range_explicit_encryption_defaults() -> Result<()> {
+    let name = "range_explicit_encryption_defaults";
+    if !check_env(name, false) || !fle2v2_ok(name).await {
```
kevinAlbs, line 3443:
```suggestion
    if !check_env(name, false) {
```

Suggest removing unneeded server checks. Test only creates local payloads.

kevinAlbs (approved):

PR #1185, opened 2024-08-16
minor: bump derive_more to 1.0

jayvdb:


isabelatkinson:
Hi @jayvdb, thanks for creating this pull request! `derive_more`'s MSRV for 1.0 is 1.75, which is quite a bit higher than our current MSRV of 1.67. Do you have an immediate need for these changes? Otherwise, I'd prefer to wait until our MSRV is compatible.

jayvdb:
No immediate need. Happy for this to be put on ice.

PR #1186, opened 2024-08-16
RUST-2008 Remove almost all uses of `add_expansions_to_env`

abr-egn:
RUST-2008

The "almost" here is the serverless tests, which are already failing so there's no way to test if edits here would work :(  Filed RUST-2016 for that.

Passing runs for the non-patchable variants:
* [search index](https://spruce.mongodb.com/version/66bd01bb138cb1000799558a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
* [aws lambda](https://spruce.mongodb.com/version/66be25c90c608400077d709c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -50,7 +50,8 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// [`ErrorKind`](enum.ErrorKind.html) is wrapped in an `Arc` to allow the errors to be
 /// cloned.
 #[derive(Clone, Debug, Error)]
-#[error("Kind: {kind}, labels: {labels:?}, backtrace: {bt}")]
+#[cfg_attr(test, error("Kind: {kind}, labels: {labels:?}, backtrace: {bt}"))]
+#[cfg_attr(not(test), error("Kind: {kind}, labels: {labels:?}"))]
 #[non_exhaustive]
 pub struct Error {
     /// The type of error that occurred.
@@ -59,6 +60,7 @@ pub struct Error {
     pub(crate) wire_version: Option<i32>,
     #[source]
     pub(crate) source: Option<Box<Error>>,
+    #[cfg(test)]
     bt: Arc<std::backtrace::Backtrace>,
 }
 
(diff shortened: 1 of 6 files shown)


src/error.rs
```diff
@@ -50,7 +50,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// [`ErrorKind`](enum.ErrorKind.html) is wrapped in an `Arc` to allow the errors to be
 /// cloned.
 #[derive(Clone, Debug, Error)]
-#[error("Kind: {kind}, labels: {labels:?}")]
+#[error("Kind: {kind}, labels: {labels:?}, backtrace: {bt}")]
```
abr-egn, line 53: Having a backtrace from the point of error _construction_ rather than the point of handling was very useful for debugging a lot of failures here, and it doesn't cost anything to construct unless `RUST_BACKTRACE` is set.  This does make error messages quite a bit more verbose; open to thoughts on how this should work.

src/test/csfle.rs
```diff
@@ -3220,6 +3216,7 @@ async fn range_explicit_encryption_test(
         KV_NAMESPACE.clone(),
         vec![LOCAL_KMS.clone()],
     )?
+    .extra_options(EXTRA_OPTIONS.clone())
```
abr-egn, line 3219: Without this (and the change to the v2 runner), the tests were accidentally using mongocryptd instead of crypt_shared.

src/error.rs
isabelatkinson: This seems valuable for us, but maybe put it behind `[cfg(test)]`? This could introduce a lot of noise for users who are running with `RUST_BACKTRACE=1` for other reasons. e.g. I got this error running an example main function that tries to create a client with a bad URI with `RUST_BACKTRACE=1 cargo run`:

Screenshot 2024-08-16 at 12 10 23

abr-egn: Good call, updated.

isabelatkinson (approved):
lgtm!

Issue #799, opened 2022-12-31
Querying for a document and getting "invalid utf-8 sequence of 1 bytes from index 119" despite not even trying to deserialize anything


LukasDeco:
## Versions/Environment
1. What version of Rust are you using?  - 1.65.0
2. What operating system are you using? - WSL (Linux on windows)
3. What versions of the driver and its dependencies are you using? (Run 
`cargo pkgid mongodb` & `cargo pkgid bson`) - mongodb@2.3.1 & bson@2.4.0
4. What version of MongoDB are you using? (Check with the MongoDB shell using `db.version()`)  - 5.0.14
5. What is your MongoDB topology (standalone, replica set, sharded cluster, serverless)? - Replica Set - 3 nodes

## Describe the bug

I am attempting to query for a document that is giving me an error about "invalid utf-8 sequence". I've done a lot of googling on this issue but no luck so far. Nothing related to mongodb :( 

The document is quite large, so that might be a potential issue, but I'm not sure. 

I'm able to query for another document from the same collection without issue, and that document is also quite large, so I'm not sure if the size is the issue. 

I have removed all the properties from the struct so I'm not trying to deserliaze anything at this point, just get the document successfully - and I still get this error. 😢 

My next move is to manually delete much of the data out of the document or query for a different document... but obviously none of this is ideal. I'd just like to get someone to point me in the right direction on what the cause of this error might be. 

Also important to note is I use Mongodb App Services(formerly Realm, formerly Stitch) and I ran schema validations across these documents and everything passes.

Here's the code, but I don't think it helps much: 
```
// manually setting ID for the query
let id = ObjectId::from_str("6116dc1633616dc8924e1050").unwrap();
        let filter_document = doc! {"userId": id};
        let find_one_options = FindOneOptions::builder().build();
        let profile = self
            .profiles_repository
            .find_one::<Profile >(filter_document, find_one_options)
            .await;
```

```
//inside find_one
async fn find_one<'b, T: DeserializeOwned + Sync + Send + Unpin>(
        &self,
        filter: Document,
        find_one_options: FindOneOptions,
    ) -> Result<Option<T>, Error> {
        let collection = self
            .client
            .database("Occasionally")
            .collection::<T>("Profiles");

        collection.find_one(filter, find_one_options).await
    }
```

```
// Profile Struct, commented out all the props :(
#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
[...]

PR #1187, opened 2024-08-20
Add deserialize-utf8-lossy feature to always deserialize using lossy UTF-8 conversion

tyilo:
Useful if you need to read from a collection created by a driver for another programming language.

See #799

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,8 @@ snappy-compression = ["dep:snap"]
 # The In Use Encryption API is unstable and may have backwards-incompatible changes in minor version updates.
 in-use-encryption-unstable = ["dep:mongocrypt", "dep:rayon", "dep:num_cpus"]
 
+deserialize-utf8-lossy = []
+
 # Enables support for emitting tracing events.
 # The tracing API is unstable and may have backwards-incompatible changes in minor version updates.
 # TODO: pending https://github.com/tokio-rs/tracing/issues/2036 stop depending directly on log.
--- a/src/change_stream.rs
+++ b/src/change_stream.rs
@@ -159,7 +159,7 @@ where
     /// ```
     pub async fn next_if_any(&mut self) -> Result<Option<T>> {
         Ok(match NextInBatchFuture::new(self).await? {
-            BatchValue::Some { doc, .. } => Some(bson::from_slice(doc.as_bytes())?),
+            BatchValue::Some { doc, .. } => Some(crate::de::from_slice(doc.as_bytes())?),
             BatchValue::Empty | BatchValue::Exhausted => None,
         })
     }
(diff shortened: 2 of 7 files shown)


tyilo:
Also consider making this the default to improve interoperability with other drivers.

abr-egn:
Thanks for the contribution!  I don't think we'll want to make this the default to avoid breaking anyone relying on the existing validation, but it seems quite helpful as an opt-in feature.

If you're willing, could you share more information about the situation that motivates this?  The bson spec says that strings are UTF-8, so at least in theory drivers shouldn't be writing values that require this.

tyilo:
> If you're willing, could you share more information about the situation that motivates this? The bson spec says that strings are UTF-8, so at least in theory drivers shouldn't be writing values that require this.

Yes, I agree that in a perfect world this would be rejected by all drivers and also the MongoDB server itself.

However, at least the Java driver doesn't always produce valid UTF-8 when writing to a collection.
See https://jira.mongodb.org/projects/JAVA/issues/JAVA-5575 which has been closed as "Won't Fix".

abr-egn (approved):
LGTM!

tyilo:
@isabelatkinson Doesn't seem to work for me:

```rust
use std::fmt::Debug;
use std::{error::Error, sync::LazyLock};

use bson::serde_helpers::Utf8LossyDeserialization;
use futures_util::TryStreamExt;
use mongodb::{
    bson::{doc, Document, RawDocumentBuf},
    Database,
};
use serde::de::DeserializeOwned;

const COLLECTION_NAME: &str = "document-registry:documents";
const FILTER: LazyLock<Document> = LazyLock::new(|| {
    doc! {
        ... // problem document
    }
});

async fn log_all<T: Send + Sync + DeserializeOwned + Debug>(database: &Database) -> Result<()> {
    let collection = database.collection::<T>(COLLECTION_NAME);
    let mut stream = collection.find(FILTER.clone()).await?;
    while let Some(raw) = stream.try_next().await? {
        println!("Raw document: {:#?}", raw);
    }
    Ok(())
}

#[tokio::main]
async fn main() -> Result<()> {
    let database = ...;

    log_all::<Utf8LossyDeserialization<Document>>(&database).await?;

    Ok(())
}
```

Output:
```
Error: Error { kind: BsonDeserialization(DeserializationError { message: "invalid utf-8 sequence of 1 bytes from index 0" }), labels: {}, wire_version: None, source: None }
```

tyilo:
@isabelatkinson Seems to work great.

isabelatkinson:
@tyilo great to hear. I just merged in the addition to the BSON library, and it will be included in that crate's next release. Going to close this out - thanks for bringing this issue to our attention!

PR #1188, opened 2024-08-26
RUST-1425 Clarify test event handling

abr-egn:
RUST-1425

I've spent some time trying various approaches here, and I've come to the conclusion that the status quo is good, actually.

The status quo:
* `EventBuffer`: a vector of timestamped events, with utilities for registering with a `Client` and fetching subsets of events (e.g. `CommandStarted`).
* `EventSubscriber`: a streaming view of events in an `EventBuffer`, with utilities for watching for events matching predicates.

Some approaches I've tried to boil this down to a single unified handler:
* Remove `EventSubscriber` entirely, convert all tests using it to check assumptions post-facto.  This would be an _enormous_ amount of work and there are quite a few tests that (subjectively) are easier to follow with the streaming approach.  Also, some tests use values from earlier events in subsequent test logic.
* Remove `EventSubscriber`, provide a `pop` method to pull events from the beginning of the buffer to replace streaming.  Along with the readability issues of the above, this makes it harder to reason about and validate the total state of the buffer without maintaining a parallel copy.
* Remove `EventSubscriber`, move its methods to `EventBuffer` and have them accept a `Cursor` that tracks the index and return a new `Cursor` where relevant.  This just makes things worse; the two views on the data are still there but the distinction is less clear, and dealing with passing the `Cursor` around is a hassle.
* Remove `EventSubscriber`, move its methods to `EventBuffer` using an internal cursor.  This has the same obfuscation issues as above; additionally, various tests spawn multiple `EventSubscribers` from a single `EventBuffer` which can't be easily expressed with them merged.

My position now is that buffering and streaming are both reasonable views of events generated during a span of action and which is more natural for a given test will depend quite a bit on the particulars.  What I've done in this PR is (hopefully) clarify that distinction and how it should be used:
* Rename `EventSubscriber` (and related methods/bindings) to `EventStream`.  "Subscriber" was reflective of the old implementation using a broadcast channel and linked subscribers; now that it's iterating over a buffer, I think "stream" is a clearer reflection of the core "one at a time" nature.
* Remove `EventBuffer::watch_all`, which I originally wrote hoping that it would be a replacement for `EventSubscriber` but turned out to be too awkward to use in practice.
* Update documentation to make recommended usage clear (and remove obsolete mentions of broadcast/subscription).

### DIFF
--- a/src/test/util/event_buffer.rs
+++ b/src/test/util/event_buffer.rs
@@ -20,7 +20,9 @@ use super::Event;
 /// A buffer of events that provides utility methods for querying the buffer and awaiting new event
 /// arrival.
 ///
-/// New test code should prefer to use this over `EventSubscriber`.
+/// If working with events one at a time as they arrive is more convenient for a particular test,
+/// register an `EventBuffer` with the `Client` and then use `EventBuffer::stream` or
+/// `EventBuffer::stream_all`.
 #[derive(Clone, Debug)]
 pub(crate) struct EventBuffer<T = Event> {
     inner: Arc<EventBufferInner<T>>,
@@ -72,22 +74,22 @@ impl<T> EventBuffer<T> {
             .collect()
     }
 
-    /// Create a new asynchronous subscription to events generated after the point of this call.
-    pub(crate) fn subscribe(&self) -> EventSubscriber<'_, T> {
+    /// Create a new stream view of events buffered after the point of this call.
+    pub(crate) fn stream(&self) -> EventStream<'_, T> {
         let (index, generation) = {
             let events = self.inner.events.lock().unwrap();
             (events.data.len(), events.generation)
         };
-        EventSubscriber {
+        EventStream {
             buffer: self,
             index,
             generation,
         }
     }
 
-    /// Create a new asynchronous subscription to all events contained in the buffer.
-    pub(crate) fn subscribe_all(&self) -> EventSubscriber<'_, T> {
-        EventSubscriber {
+    /// Create a new stream view, starting with events already contained in the buffer.
+    pub(crate) fn stream_all(&self) -> EventStream<'_, T> {
+        EventStream {
             buffer: self,
             index: 0,
             generation: self.inner.events.lock().unwrap().generation,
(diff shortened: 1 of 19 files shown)


src/test/util/event_buffer.rs
```diff
@@ -305,12 +275,25 @@ impl<'a, T: Clone> EventSubscriber<'a, T> {
         None
     }
 
-    /// Consume and pass events to the provided closure until it returns Some or the timeout is hit.
-    pub(crate) async fn filter_map_event<F, R>(
-        &mut self,
-        timeout: Duration,
-        mut filter_map: F,
-    ) -> Option<R>
+    /// Get the next unread event from the underlying buffer, waiting for a new one to arrive if all
```
abr-egn, line 278: I moved this after `try_next` so reading down the file goes in the order of increasing layers.

src/test/util/event_buffer.rs
```diff
@@ -305,12 +275,25 @@ impl<'a, T: Clone> EventSubscriber<'a, T> {
...
+                notified.await;
+            }
+        })
+        .await
+        .unwrap_or(None)
+    }
+
+    /// Get the next unread event for which the provided closure returnes `Some`, waiting for new
+    /// events to arrive if all current ones have been read.
+    pub(crate) async fn next_map<F, R>(&mut self, timeout: Duration, mut filter_map: F) -> Option<R>
```
abr-egn, line 296: Along with renaming the struct, I renamed a bunch of its method to better match the iteration/streaming model.

src/test/util/event_buffer.rs
```diff
@@ -350,31 +336,22 @@ impl<'a, T: Clone> EventSubscriber<'a, T> {
...
+            while let Some(event) = self.next_map(timeout, &mut filter).await {
                 events.push(event);
             }
         })
         .await;
         events
     }
 
-    #[cfg(feature = "in-use-encryption-unstable")]
-    pub(crate) async fn clear_events(&mut self, timeout: Duration) {
```
abr-egn, line 372: This was a now nearly nonsensical holdover from previous iterations of event cleanup, only used in a few places in CSFLE tests that were easily replaced with just making a fresh `EventStream`.

isabelatkinson (approved):
lgtm! agreed that it makes sense to keep these two concepts

PR #1189, opened 2024-08-28
RUST-1449 Simplify test client creation

abr-egn:
RUST-1449

Two main themes of changes in this PR:
* Removed the remaining uses of older `TestClient` construction methods in favor of `Client::test_builder`
* Cleaned up the `additional_options` tangle: moved the LB-specific initialization to be internal to the v2 test runner (the only thing that actually needed it / was using it) and provided the much simpler `sharded_use_first_host` for all the other places it was being called.

With this done, there's now only one way to make a `TestClient` and only one way to set options for it :)

### DIFF
--- a/src/test/db.rs
+++ b/src/test/db.rs
@@ -40,7 +40,7 @@ async fn get_coll_info(db: &Database, filter: Option<Document>) -> Vec<Collectio
 #[tokio::test]
 #[function_name::named]
 async fn list_collections() {
-    let client = Client::test_builder().await;
+    let client = Client::for_test().await;
     let db = client.database(function_name!());
     db.drop().await.unwrap();
 
@@ -73,7 +73,7 @@ async fn list_collections() {
 #[tokio::test]
 #[function_name::named]
 async fn list_collections_filter() {
-    let client = Client::test_builder().await;
+    let client = Client::for_test().await;
     let db = client.database(function_name!());
     db.drop().await.unwrap();
 
@@ -108,7 +108,7 @@ async fn list_collections_filter() {
 #[tokio::test]
 #[function_name::named]
 async fn list_collection_names() {
-    let client = Client::test_builder().await;
+    let client = Client::for_test().await;
     let db = client.database(function_name!());
     db.drop().await.unwrap();
 
@@ -136,7 +136,7 @@ async fn list_collection_names() {
 #[tokio::test]
 #[function_name::named]
 async fn collection_management() {
-    let client = Client::test_builder().await;
+    let client = Client::for_test().await;
     let db = client.database(function_name!());
     db.drop().await.unwrap();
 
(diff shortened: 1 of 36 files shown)


src/test/spec/v2_runner.rs
```diff
@@ -299,6 +299,44 @@ impl TestContext {
     }
 }
 
+impl crate::test::util::TestClientBuilder {
+    async fn options_for_multiple_mongoses(
```
abr-egn, line 303: This is a merged version of `additional_options` / the now-removed `TestClient::options_for_multiple_mongoses`, the main difference being it's now internal to this module.  This logic could be inlined directly into the single call site above but this seemed a little better for readability.

src/test/util.rs
```diff
@@ -87,17 +83,9 @@ impl TestClientBuilder {
...
-        mut self,
-        options: impl Into<Option<ClientOptions>>,
-        use_multiple_mongoses: bool,
-    ) -> Self {
-        let options = options.into();
-        assert!(self.options.is_none() || options.is_none());
-        self.options =
-            Some(TestClient::options_for_multiple_mongoses(options, use_multiple_mongoses).await);
+    /// When running against a sharded topology, only use the first configured host.
+    pub(crate) fn sharded_use_first_host(mut self) -> Self {
```
abr-egn, line 87: This is definitely an awkward name, I welcome suggestions.

src/test/util.rs
```diff
@@ -131,6 +119,16 @@ impl TestClientBuilder {
             options.test_options_mut().min_heartbeat_freq = Some(freq);
         }
 
+        if self.sharded_use_first_host {
```
abr-egn, line 122: Deferring the action to here means there's no issue with order of calls with setting options and it avoids the need for an additional `.await` in the builder call chain.

src/test/db.rs
```diff
@@ -41,7 +40,7 @@ async fn get_coll_info(db: &Database, filter: Option<Document>) -> Vec<Collectio
 #[tokio::test]
 #[function_name::named]
 async fn list_collections() {
-    let client = TestClient::new().await;
+    let client = Client::test_builder().build().await;
```
isabelatkinson, line 43: WDYT about making `build` an `IntoFuture` impl instead? It'd be nice to make the no-configuration case a little less verbose, although I'm not sure what the right name for `test_builder` would be if there's no corresponding `build`.

src/test/util.rs
isabelatkinson: Maybe `use_single_mongos`? That's in line with the `use_multiple_mongoses` language from the unified tests

src/test/db.rs
abr-egn: Ah, that cleans things up nicely :)  Renamed the method to `Client::for_test` which (IMO) reads well both with and without chained methods.

isabelatkinson (approved):

PR #1197, opened 2024-09-05
RUST-2030 Add more event fields: lsid, txnNumber and disambiguatedPaths

arthurprs:
Add more event fields: lsid, txnNumber and disambiguatedPaths

### DIFF
--- a/src/test/change_stream.rs
+++ b/src/test/change_stream.rs
@@ -668,7 +668,17 @@ async fn transaction_fields() -> Result<()> {
             None => return Ok(()),
         };
     if client.is_sharded() {
-        log_uncaptured("skipping change stream test on unsupported topology");
+        log_uncaptured("skipping change stream test transaction_fields on unsupported topology");
+        return Ok(());
+    }
+    if !VersionReq::parse(">=5.0")
+        .unwrap()
+        .matches(&client.server_version)
+    {
+        log_uncaptured(format!(
+            "skipping change stream test transaction_fields on unsupported version {:?}",
+            client.server_version
+        ));
         return Ok(());
     }
     if !client.supports_transactions() {
--- a/src/test/change_stream.rs
+++ b/src/test/change_stream.rs
@@ -659,6 +659,48 @@ async fn split_large_event() -> Result<()> {
     Ok(())
 }
 
+/// Test that transaction fields are parsed correctly
+#[tokio::test]
+async fn transaction_fields() -> Result<()> {
+    let (client, coll, mut stream) =
+        match init_stream("chang_stream_transaction_fields", true).await? {
+            Some(t) => t,
+            None => return Ok(()),
+        };
+    if client.is_sharded() {
+        log_uncaptured("skipping change stream test on unsupported topology");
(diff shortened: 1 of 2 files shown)


src/test/change_stream.rs
```diff
@@ -659,6 +659,48 @@ async fn split_large_event() -> Result<()> {
...
+
+    let next_event = stream.next().await.transpose()?;
+    assert!(matches!(next_event,
+        Some(ChangeStreamEvent {
+            operation_type: OperationType::Insert,
+            document_key: Some(key),
+            lsid: Some(lsid),
+            txn_number: Some(1),
+            ..
+        }) if key == doc! { "_id": 1 } && lsid.get("id") == session_id.as_ref()
```
arthurprs, line 698: The lsid contains a second field "uid", which I'm familiar with. But the test should still be valid.

isabelatkinson:
Hey @arthurprs, thanks for opening this PR! We have an open ticket for adding the `disambiguated_paths` field (RUST-1423) which also includes an option and some additional testing, so we can certainly address that.

What is your use case for adding the `txn_number` and `lsid` fields? These are not part of the specification for the `ChangeStreamEvent` type, so I'm hesitant to make this addition. Would it be possible for you to retrieve these values from [`CommandStartedEvent`](https://docs.rs/mongodb/latest/mongodb/event/command/struct.CommandStartedEvent.html)s emitted via command monitoring/logging instead? They would both be present as top-level fields in the [`document`](https://docs.rs/mongodb/latest/mongodb/event/command/struct.CommandStartedEvent.html#structfield.command) field for events emitted during a transaction.

arthurprs:
Maybe I'm misinterpreting [the manual page](https://www.mongodb.com/docs/manual/reference/change-events/update/#mongodb-data-update) but `lsid` and `txn_number` seemed documented and even referred to by `clusterTime` description.

Our use cases include detecting if a certain actor was the source of the write by its session ID (or previous ones).

abr-egn:
You're quite right, those fields are in the manual!  The confusing part here on our end is that they're not part of the [specification](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.md#server-specification), so I'm working to track down that discrepancy.  My concern is that I don't want to add fields to the public API if they're not actually intended to be part of the stable interface 🙂 

While we're tracking this down, you should be able to unblock anything that needs those fields by defining your own event type and using it with [`ChangeStream::with_type`](https://docs.rs/mongodb/latest/mongodb/change_stream/struct.ChangeStream.html#method.with_type).

abr-egn (approved):
LGTM!

abr-egn:
Sorry for the delay here, it looks like it was just a simple oversight that those fields weren't included in the spec.  I've started the test run and assuming there are no unexpected failures I'll merge this in.  Thanks again for your patience!

abr-egn:
It looks like the test is using some logic that's not supported in server version 4.x, can you add a check to restrict it to >=5.0?

arthurprs:
Done in 
bba9d00

PR #1200, opened 2024-09-10
RUST-2024 Fix serverless CSFLE tests

abr-egn:
RUST-2024

There were three independent problems happening here:
* The serverless tests and serverless csfle tests were both trying to merge results into `results.xml`, and the second one would fail because the merge tool (appropriately) won't overwrite a file without an additional flag set.  Fixed this by changing the test script to do a rolling merge, which also has the nice effect that our tests don't need to include output file names or an explicit merge step any more.
* Many of the serverless tests were failing because `CRYPT_SHARED_LIB_PATH` wasn't being set; I tracked this down to a bug in the serverless setup scripts in drivers-evergreen-tools.  I've sent https://github.com/mongodb-labs/drivers-evergreen-tools/pull/491 for that and in the meanwhile pointed our testing at my branch with the fix.
* A few of the legacy tests were failing because of CLOUDP-267864; this just needed a test file sync to pick up the serverless skip for those tests.

### DIFF
--- a/.evergreen/cargo-test.sh
+++ b/.evergreen/cargo-test.sh
@@ -29,7 +29,12 @@ cargo_test() {
   TAIL_PID=$!
   LOG_UNCAPTURED=${LOG_PATH} RUST_BACKTRACE=1 cargo nextest run --profile ci $(cargo_test_options $1)
   ((CARGO_RESULT = ${CARGO_RESULT} || $?))
-  mv target/nextest/ci/junit.xml $2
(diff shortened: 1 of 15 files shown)


abr-egn:
https://spruce.mongodb.com/version/66e06df35524ab00078b4b00/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC has a run with these fixes.  The Ubuntu 22.04 (ARM64) failures are unrelated, those fail on `main` as well.  I suspect a script breakage.

PR #1201, opened 2024-09-11
track task spawn location for tokio instrumentation

hds:
The tracing instrumentation present in tokio (which feeds into
tokio-console) tracks the spawning location of each task. This can be
useful in understanding what each task is doing, especially if the tasks
aren't named (which is a `tokio_unstable` feature so not usually present
in published crates).

The location tracking isn't as useful when tasks are spawned from some
central location like the `mongodb` crate does, because all tasks have
the same spawn location. However, since the instrumentation uses the
`panic:Location` to determine the spawn location, the `#[track_caller]`
attribute macro can be used to ensure that the actual location is picked
up instead of the common one.

A side effect of using the `#[track_caller]` attribute is that the
single line output in the case of a panic (when `RUST_BACKTRACE` isn't
set) will point to the callsite in the code, and not the common site -
although this is often desirable anyway.

This change adds `#[track_caller]` to the `pub(crate)` function
`crate::runtime::spawn` as well as the method `AsyncJoinHandle::spawn`.

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -54,6 +54,7 @@ pub(crate) use tls::TlsConfig;
 ///
 /// This must be called from an async block
 /// or function running on a runtime.
+#[track_caller]
 pub(crate) fn spawn<F, O>(fut: F) -> AsyncJoinHandle<O>
 where
     F: Future<Output = O> + Send + 'static,
--- a/src/runtime/join_handle.rs
+++ b/src/runtime/join_handle.rs
@@ -9,6 +9,7 @@ use std::{
 pub(crate) struct AsyncJoinHandle<T>(tokio::task::JoinHandle<T>);
 
 impl<T> AsyncJoinHandle<T> {
+    #[track_caller]
     pub(crate) fn spawn<F>(fut: F) -> Self
     where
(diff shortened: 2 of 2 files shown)


hds:
A couple of notes on the MR. This came about after receiving a question about the behaviour of an application which was using the `mongodb` crate and had this screenshot from tokio-console:

console-mongodb_tasks

All those tasks would have a more useful value in the Location column with this patch.

No tests are included because testing `#[track_caller]` is a little bit involved - and testing it with the tokio instrumentation enabled even more so. However, I'm happy to add tests by forcing a panic and recording the panic location, they would look something like the ones here:

https://github.com/tokio-rs/tokio/blob/91169992b2ed0cf8844dbaa0b4024f9db588a629/tokio/tests/rt_panic.rs

with the following helper function:

https://github.com/tokio-rs/tokio/blob/91169992b2ed0cf8844dbaa0b4024f9db588a629/tokio/tests/support/panic.rs

abr-egn:
This looks like a very nice enhancement, thank you!  I've kicked off the test run and if there aren't any known failures I'll merge this in :)

abr-egn (approved):

PR #1211, opened 2024-10-01
RUST-2001 Allow SRV hostnames with less than three parts

abr-egn:
RUST-2001

The actual changed logic is in `LookupHosts::validate` (which used to be in `SrvResolver::get_srv_hosts`), and is the conditional bit near the top checking the length of `original_hostname_parts`.  The rest of this is refactoring to make the new logic testable.

### DIFF
--- a/src/client/options.rs
+++ b/src/client/options.rs
@@ -265,15 +265,6 @@ impl ServerAddress {
             Self::Unix { path } => path.to_string_lossy(),
         }
     }
-
-    #[cfg(feature = "dns-resolver")]
-    pub(crate) fn port(&self) -> Option<u16> {
-        match self {
-            Self::Tcp { port, .. } => *port,
-            #[cfg(unix)]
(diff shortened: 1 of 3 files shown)


src/client/options.rs
```diff
@@ -1455,39 +1446,34 @@ impl ConnectionString {
             host_list.push(address);
         }
 
-        let hosts = if srv {
-            if host_list.len() != 1 {
-                return Err(ErrorKind::InvalidArgument {
-                    message: "exactly one host must be specified with 'mongodb+srv'".into(),
+        let host_info = if !srv {
```
abr-egn, line 1449: This isn't directly related to the change, but when I was reading through to find understand the original code I had a hard time skimming this section.  This changes it to put the success cases up front and have the error cases nicely enumerated.

isabelatkinson (approved):

PR #1221, opened 2024-10-10
RUST-2010 Work around evergreen agent bug(?)

abr-egn:
RUST-2010

This disables the gridfs multi upload benchmark for now ([passing run](https://spruce.mongodb.com/version/6707e544bb7df200076311b8/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)); I'll be filing an evergreen bug to follow up.

I've verified that it's specifically that benchmark - running that one alone is enough to trigger the "system unresponsive" status, and disabling it causes the suite to complete successfully.  Running the suite manually on a host with the same distro as the evergreen run works fine and the host remains responsive throughout, so there's definitely some weird interaction happening with the evergreen agent heartbeat.  Unfortunately, I could find very little documentation on how that works :(

### DIFF
--- a/.evergreen/run-driver-benchmark-unresponsive.sh
+++ b/.evergreen/run-driver-benchmark-unresponsive.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -o errexit
+
+source ./.evergreen/env.sh
+
+cd benchmarks
+cargo run \
+  --release \
+  -- --output="../benchmark-results.json" -i 21
+
+cat ../benchmark-results.json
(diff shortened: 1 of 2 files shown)


.evergreen/run-driver-benchmark-unresponsive.sh
abr-egn, line 1: This is just to make reproducing the failure easier.

isabelatkinson (approved):

abr-egn:
Update: filed DEVPROD-11980.

PR #1228, opened 2024-10-16
minor: refactor faas tests to validate metadata

abr-egn:
This is preliminary to RUST-2000, which will need this kind of observability.  (And we probably should have been doing it from the start anyway.)

### DIFF
--- a/src/cmap/establish/handshake.rs
+++ b/src/cmap/establish/handshake.rs
@@ -336,7 +336,8 @@ pub(crate) struct Handshaker {
 }
 
 #[cfg(test)]
-pub(crate) static METADATA: std::sync::OnceLock<ClientMetadata> = std::sync::OnceLock::new();
+#[allow(clippy::incompatible_msrv)]
+pub(crate) static TEST_METADATA: std::sync::OnceLock<ClientMetadata> = std::sync::OnceLock::new();
 
 impl Handshaker {
     /// Creates a new Handshaker.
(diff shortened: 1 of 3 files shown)


src/cmap/establish/handshake.rs
```diff
@@ -335,6 +335,9 @@ pub(crate) struct Handshaker {
     http_client: crate::runtime::HttpClient,
 }
 
+#[cfg(test)]
+pub(crate) static METADATA: std::sync::OnceLock<ClientMetadata> = std::sync::OnceLock::new();
```
abr-egn, line 339: This gives me the heebie-jeebies but since it's explicitly test-restricted and a lot less disruptive than trying to propagate the computed metadata back up the ownership/worker hierarchy it seemed the less-bad option.

abr-egn:
evergreen retry

src/cmap/establish/handshake.rs
```diff
@@ -335,6 +335,10 @@ pub(crate) struct Handshaker {
     http_client: crate::runtime::HttpClient,
 }
 
+#[cfg(test)]
+#[allow(clippy::incompatible_msrv)]
```
abr-egn, line 339: This should be fine since we only test on the latest release and `cfg(test)` isn't transitive, so someone using <1.70 (when `OnceLock` was introduced) can still use the `mongodb` crate fine, including in their tests.

isabelatkinson (approved):

PR #1233, opened 2024-10-28
DRIVERS-2949 Test happy eyeballs behavior

abr-egn:
DRIVERS-2949

With happy eyeballs on its way to being a driver spec, it needs a way for the behavior to be tested.  This PR implements a (relatively) simple python server that provides pairs of IPv4 and IPv6 ports on demand, with one of the pair being slow to start accepting incoming connections, and the corresponding Rust driver test using that server.

Originally I was hoping to have a static set of ports exposed by the test server (slow v4, fast v4, slow v6, fast v6), but it turns out that once you start accepting connections there's no way to delay the TCP handshake.  Instead, each test gets its own ephemeral pair of ports and the delay happens between binding to the port and accepting connections.

This also demonstrates the use of cargo nextest's `default-filter` to prevent these tests from running during normal interactive usage.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -30,7 +30,6 @@ pre:
 
 # Functions to run after all tasks (except those in task groups).
 post:
-  - func: "stop happy eyeballs server"
   - func: "stop csfle servers"
   - func: "stop load balancer"
   - func: "stop mongo orchestration"
@@ -381,6 +380,20 @@ buildvariants:
     tasks:
       - name: test-aws-lambda-task-group
 
+  - name: happy-eyeballs-macos
+    display_name: "Happy Eyeballs (MacOS)"
+    run_on:
+      - macos-1100
+    tasks:
+      - happy-eyeballs-task-group
+
+  - name: happy-eyeballs-windows
+    display_name: "Happy Eyeballs (Windows)"
+    run_on:
+      - windows-64-vs2017-small
+    tasks:
+      - happy-eyeballs-task-group
+
 ###############
 # Task Groups #
 ###############
@@ -711,6 +724,26 @@ task_groups:
     setup_group_timeout_secs: 1800
     tasks:
       - oidc-auth-test-gcp-latest
+
+  - name: happy-eyeballs-task-group
+    setup_group_can_fail_task: true
+    setup_group_timeout_secs: 1800
+    setup_group:
+      - func: "fetch source"
+      - func: "create expansions"
+      - func: "prepare resources"
+      - func: "windows fix"
+      - func: "fix absolute paths"
+      - func: "init test-results"
+      - func: "make files executable"
+      - func: "install rust"
+      - func: "install junit dependencies"
+      - func: "start happy eyeballs server"
+    tasks:
+      - test-happy-eyeballs
+    teardown_task:
+      - func: "stop happy eyeballs server"
+      - func: "upload test results"
 
 #########
 # Tasks #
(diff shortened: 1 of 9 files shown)


.evergreen/config.yml
```diff
@@ -1378,7 +1379,17 @@ functions:
...
           - .evergreen/fetch-drivers-tools.sh
-
+    - command: subprocess.exec
+      params:
+        working_dir: src
+        include_expansions_in_env:
+          - DRIVERS_TOOLS
+        binary: bash
+        args:
+          - .evergreen/find-python3.sh
```
abr-egn, line 1389: This addition means I don't have to wrap the server startup/shutdown in shell scripts (and might be useful for similar later things).

.evergreen/config.yml
```diff
@@ -1452,6 +1463,20 @@ functions:
...
+        binary: ${PYTHON3}
+        args:
+          - .evergreen/happy-eyeballs-server.py
+    - command: subprocess.exec
+      params:
+        working_dir: src
+        binary: ${PYTHON3}
+        args:
+          - .evergreen/happy-eyeballs-server.py
+          - --wait
```
abr-egn, line 1479: I tried having a "fork background server, wait in foreground" mode to avoid the need for a second command here, and got one that worked for interactive invocations, but Evergreen seems to follow the fork and wait for the child process so that didn't work where it needed to :(

.evergreen/happy-eyeballs-client.py
abr-egn, line 1: I used this as a development aid but it seems useful to include as a demo program.

.evergreen/happy-eyeballs-server.py
```diff
@@ -0,0 +1,135 @@
...
+    srv6.close()
+    await asyncio.wait([
+        asyncio.create_task(srv4.wait_closed()),
+        asyncio.create_task(srv6.wait_closed()),
+    ])
+    
+    print(f'{PREFIX}: connection complete, test ports closed', file=sys.stderr)
+    print(f'{PREFIX}: ========================', file=sys.stderr)
+
+async def test_listen(name: str, srv, delay: bool, connected: asyncio.Event):
```
abr-egn, line 87: Fun fact: `srv` can't have a type annotation because the version of python3 on some Evergreen hosts is old enough that the correct type (`asyncio.Server`) isn't public.

src/runtime/stream.rs
```diff
@@ -77,21 +84,14 @@ async fn tcp_try_connect(address: &SocketAddr) -> Result<TcpStream> {
     Ok(TcpStream::from_std(std_stream)?)
 }
 
-async fn tcp_connect(address: &ServerAddress) -> Result<TcpStream> {
+pub(crate) async fn tcp_connect(resolved: Vec<SocketAddr>) -> Result<TcpStream> {
```
abr-egn, line 87: This is just a minor refactor so that the behavior of TCP connection can be tested independently of DNS resolution.

.evergreen/config.yml
```diff
@@ -380,6 +380,20 @@ buildvariants:
     tasks:
       - name: test-aws-lambda-task-group
 
+  - name: happy-eyeballs-macos
```
abr-egn, line 383: The happy eyeballs test server relies on underspecified TCP handshake behavior that the MacOS (well, BSD kernel) and Windows networking stacks interpret differently than the Linux one, i.e. this test doesn't work on Linux and AFAICT there's no way to make it do so.

.evergreen/happy-eyeballs-server.py
```diff
@@ -0,0 +1,145 @@
...
+    # Reply to control request with success byte and test server ports
+    writer.write(b'\x01')
+    writer.write(ipv4_port.to_bytes(2, 'big'))
+    writer.write(ipv6_port.to_bytes(2, 'big'))
+    await writer.drain()
+    writer.close()
+    await writer.wait_closed()
+
+    # Start test servers listening in parallel
+    # Hold a reference to the tasks so they aren't GC'd
```
abr-egn, line 73: Other fun fact: there's a better API for this idiom (`asyncio.TaskGroup`) but it's also not in all evergreen python installations.

isabelatkinson (approved):
driver changes LGTM. I didn't review the python stuff as I expect that will be covered by the spec PR but LMK if you want another set of eyes on that

jadalilleboe (approved):

PR #1235, opened 2024-10-30
RUST-911: Add srvServiceName URI option

jadalilleboe:
[RUST-911](https://jira.mongodb.org/browse/RUST-911)


isabelatkinson (commented):
looking good so far - we'll also need to implement the prose test described [here](https://github.com/mongodb/specifications/tree/master/source/polling-srv-records-for-mongos-discovery/tests#srv-polling-with-srvservicename-mongoclient-option)

isabelatkinson:
It looks like we're also currently skipping some URI options tests for `srvServiceName` that we should unskip as part of this. See [here](https://github.com/mongodb/mongo-rust-driver/blob/842c536381107f6f181196cf38a36fc8da136ba5/src/client/options/test.rs#L25) and [here](https://github.com/mongodb/mongo-rust-driver/blob/842c536381107f6f181196cf38a36fc8da136ba5/src/test/spec/initial_dns_seedlist_discovery.rs#L67). (I think the comment on the second one points to the wrong ticket.)

src/sdam/srv_polling/test.rs
```diff
@@ -182,7 +203,22 @@ async fn srv_max_hosts_random() {
...
     assert!(actual.contains(&localhost_test_build_10gen(27017)));
 }
+
+#[tokio::test]
+async fn srv_service_name() {
+    let hosts = vec![
+        localhost_test_build_10gen(27019),
+        localhost_test_build_10gen(27020),
+    ];
+    run_test_srv(
```
isabelatkinson, line 217: `run_test_srv` just returns the addresses on the topology, so we'll need to add assertions on what gets returned here (see some of the other tests for examples)

isabelatkinson:
disregard my comment about assertions, I misread which method was being called!

isabelatkinson (approved):
lgtm!

src/client/options.rs
```diff
@@ -900,11 +910,16 @@ impl Default for HostInfo {
 }
 
 impl HostInfo {
-    async fn resolve(self, resolver_config: Option<ResolverConfig>) -> Result<ResolvedHostInfo> {
+    async fn resolve(
+        self,
+        resolver_config: Option<ResolverConfig>,
+        srv_service_name: Option<String>,
```
abr-egn, line 916: Would it make sense to bundle `srv_service_name` into `resolver_config`?  Mechanically, they're always passed around as a pair but I could see going either way on whether it's the right thing conceptually.

abr-egn (approved):
LGTM, with one judgement call I leave up to you :)  No need for re-review if you decide to keep it as is.

src/client/options.rs
jadalilleboe: Looked into this a bit, decided to keep it as is to reduce complexity in how the srv_service_name is passed through the codepath (from being parsed as part of the connection string down to the srv resolver- adding the parameter to ResolverConfig disrupts this a little bit). Thank you for pointing this out!

PR #1241, opened 2024-11-11
RUST-1454 Add a `PooledConnection` type

isabelatkinson:
Adds a separate type for connections within the connection pool, as well as an enum marking the state of a pooled connection. There are no functional CMAP changes in this PR.


src/cmap/conn/pooled.rs
```diff
@@ -0,0 +1,355 @@
...
+    #[derivative(Debug = "ignore")]
+    event_emitter: CmapEventEmitter,
+
+    /// The state of this connection.
+    state: PooledConnectionState,
+}
+
+/// The state of a pooled connection.
+#[derive(Clone, Debug)]
+enum PooledConnectionState {
```
isabelatkinson, line 55: My original goal for this PR was to make `PooledConnection` generic over its state so that we could assert at compile time that a connection is in the correct state; however, that proved to be a pain because drop specialization is not allowed, and a lot of our connection transformation is done by mutation. This was a compromise solution that required much less overall refactoring, but I still think the additional type information is useful.

abr-egn: That's a real shame about drop specialization.  This is still _way_ more readable than before, though.

abr-egn (approved):
LGTM!

abr-egn:
> I _think_ the happy eyeballs failure is unrelated as well

Yeah, I can repro this - it looks like the test is flaky on windows.

PR #1242, opened 2024-11-11
RUST-229 Parse IPv6 addresses in the connection string

isabelatkinson:


src/client/options.rs
```diff
@@ -185,74 +206,99 @@ impl FromStr for ServerAddress {
...
+                        address
+                    ),
                 }
-                part
-            }
-            None => {
+                .into());
+            };
+
+            if let Err(parse_error) = Ipv6Addr::from_str(hostname) {
```
isabelatkinson, line 243: was kind of back-and-forth on whether this was valuable, but I think it's worth it because otherwise we may strip the brackets from something that was invalid and attempt to connect to it, e.g. "[localhost]:27017"

src/client/options.rs
```diff
@@ -185,74 +206,99 @@ impl FromStr for ServerAddress {
...
         let address = address.as_ref();
-        // checks if the address is a unix domain socket
+
         #[cfg(unix)]
         {
             if address.ends_with(".sock") {
-                return Ok(ServerAddress::Unix {
-                    path: PathBuf::from(address),
-                });
+                #[cfg(unix)]
```
abr-egn, line 216: This is already inside the `#[cfg(unix)]` on line 213, so I don't think this is needed.

src/client/options.rs
```diff
@@ -185,74 +206,99 @@ impl FromStr for ServerAddress {
...
+                {
+                    let address =
+                        percent_decode(address, "unix domain sockets must be URL-encoded")?;
+                    return Ok(Self::Unix {
+                        path: PathBuf::from(address),
+                    });
+                }
+                #[cfg(not(unix))]
```
abr-egn, line 224: Similarly, I don't think this will ever trigger.

abr-egn (approved):
LGTM!  One nit, no need for re-review.

pdhanoya:
@isabelatkinson The changes are incorrectly parsing localhost ipv6 ` "mongodb://[::1]:27017";`

isabelatkinson:
The following test:

```
#[tokio::test]
async fn uri() {
    let uri = "mongodb://[::1]:27017";
    let options = ClientOptions::parse(uri).await.unwrap();
    dbg!("{}", options.hosts);

    let uri = "mongodb://[0:0:0:0:0:0:0:1]:27017";
    let options = ClientOptions::parse(uri).await.unwrap();
    dbg!("{}", options.hosts);
} 
```

gives me the following output:

```
[src/client/options/test.rs:369:5] "{}" = "{}"
[src/client/options/test.rs:369:5] options.hosts = [
    Tcp {
        host: "::1",
        port: Some(
            27017,
        ),
    },
]
[src/client/options/test.rs:373:5] "{}" = "{}"
[src/client/options/test.rs:373:5] options.hosts = [
    Tcp {
        host: "0:0:0:0:0:0:0:1",
        port: Some(
            27017,
        ),
    },
]
```

Is there a different output you're expecting?

Note that the `ServerAddress::parse` method only parses the "[::1]:27017" portion of the URI, so passing the entire URI to that method will fail.

pdhanoya:
@isabelatkinson My local client fails like this when run with uri `"mongodb://[::1]:27017";`
```C02F43RPMD6M:mongo-rust-driver pdhanoya$ cargo run  --bin test_client
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.09s
     Running `target/debug/test_client`
Error: Error { kind: InvalidArgument { message: "port must be valid 16-bit unsigned integer, instead got: " }, labels: {}, wire_version: None, source: None }```

pdhanoya:
@isabelatkinson I was just behind a few commits after doing a pull.. I am able to test without issue and can confirm ipv6 localhost is working as expected.

PR #1245, opened 2024-11-12
RUST-2097 Migrate from `derivative` to `derive-where`

isabelatkinson:
Migrates from `derivative` to `derive-where`. In the long term, I'd like to switch over to `derive_more`, which we already depend on for some `Display`-related utilities. `derive_more` has the largest feature set of crates of its ilk; however, its MSRV for the features we'd need is 1.75 (compared to our current 1.67), and it does not yet support `PartialEq`. We can make the swap once those concerns are resolved. This is tracked in RUST-2102.


abr-egn (approved):
LGTM!

abr-egn:
> In the long term, I'd like to switch over to `derive_more`, which we already depend on for some `Display`-related utilities. `derive_more` has the largest feature set of crates of its ilk; however, its MSRV for the features we'd need is 1.75 (compared to our current 1.67), and it does not yet support `PartialEq`. We can make the swap once those concerns are resolved.

Can you file a ticket so we remember to do this?

TroyKomodo:
Can we make a minor release of this `3.1.1` on crates.io @isabelatkinson

isabelatkinson:
@TroyKomodo 3.1.1 was just released!

PR #1248, opened 2024-11-20
RUST-2103 Preliminary improvement for action rustdocs: `find`

abr-egn:
RUST-2103

There's enough machinery here that I thought it would be better to do it across multiple PRs - this one has the core added functionality and a single action converted to use it.  I'll follow up with PRs implementing conversions for the rest of the actions and the (hopefully minor) functionality updates those will likely need.

This is what the rustdocs for `find` look like with these changes ([old](https://docs.rs/mongodb/latest/mongodb/struct.Collection.html#method.find) for comparison):

Screenshot 2024-11-22 at 11 28 20 AM

And the new rustdocs for the setters ([old](https://docs.rs/mongodb/latest/mongodb/action/struct.Find.html#method.max_await_time)):

Screenshot 2024-11-20 at 2 27 27 PM

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -1,5 +1,7 @@
 extern crate proc_macro;
 
+use std::collections::HashMap;
+
 use macro_magic::import_tokens_attr;
 use quote::{quote, ToTokens};
 use syn::{
@@ -14,12 +16,14 @@ use syn::{
     Block,
     Error,
     Expr,
+    Fields,
     GenericArgument,
     Generics,
     Ident,
     ImplItem,
     ImplItemFn,
     ItemImpl,
+    ItemStruct,
     Lifetime,
     Lit,
     Meta,
@@ -224,24 +228,11 @@ pub fn options_doc(
     let mut impl_fn = parse_macro_input!(item as ImplItemFn);
 
     // Collect a list of names from the setters impl
-    let option_setters_path = parse_quote! { option_setters };
     let mut setter_names = vec![];
     for item in &setters.items {
         match item {
-            // bare fn
             ImplItem::Fn(item) => {
                 setter_names.push(item.sig.ident.to_token_stream().to_string());
-            }
-            // option_setters! call
-            ImplItem::Macro(item) if item.mac.path == option_setters_path => {
-                let tokens = item.mac.tokens.clone().into();
-                let list = parse_macro_input!(tokens as OptionSettersList);
-                if list.opt_field_name.is_some() {
-                    setter_names.push("with_options".to_owned());
-                }
-                for setter in &list.setters {
-                    setter_names.push(setter.name.to_token_stream().to_string());
-                }
             }
             _ => continue,
         }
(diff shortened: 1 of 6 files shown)


macros/src/lib.rs
```diff
@@ -322,20 +332,18 @@ pub fn option_setters(input: proc_macro::TokenStream) -> proc_macro::TokenStream
         setters,
     } = parse_macro_input!(input as OptionSettersList);
 
-    let extras = opt_field_name.map(|name| {
-        quote! {
-            #[allow(unused)]
-            fn options(&mut self) -> &mut #opt_field_type {
-                self.#name.get_or_insert_with(<#opt_field_type>::default)
-            }
+    let extras = quote! {
```
abr-egn, line 335: There was leftover unused functionality to omit the `options` and `with_options` from generated code; stripping that out let me simplify assumptions for the newer iteration.

macros/src/lib.rs
```diff
@@ -458,3 +461,140 @@ impl Parse for OptionSetter {
         Ok(Self { attrs, name, type_ })
     }
 }
+
+#[import_tokens_attr]
```
abr-egn, line 465: The core magic that the `macro_magic` crate gives us is that while this proc macro is attached to an `impl` block that it edits to add the action option setters, it _also_ receives the full "exported" token stream for the option struct as its `attr` argument (instead of just the path as it would seem from the call site).

This means that we can remove _all_ of the duplicated information we had to have in `option_setters!` invocations (names, types, cfg attrs) _and_ bring forward more information (doc comments).

macros/src/lib.rs
```diff
@@ -458,3 +461,140 @@ impl Parse for OptionSetter {
...
+        })
+    }
+
+    // All done.
+    impl_in.to_token_stream().into()
+}
+
+#[import_tokens_attr]
+#[proc_macro_attribute]
+pub fn options_doc(
```
abr-egn, line 558: This uses the same trick, but in a much simpler way to make the list of setters available in the doc comment for the main action method call itself.  It uses the token stream of the impl block rather than the option struct so we pick up things like `session` and any custom setters we happen to write.

src/action/find.rs
```diff
@@ -81,32 +84,9 @@ pub struct Find<'a, T: Send + Sync, Session = ImplicitSession> {
     session: Session,
 }
 
+#[option_setters_2(crate::coll::options::FindOptions)]
```
abr-egn, line 87: The token stream is exported by creating a macro with a mangled name and the same visibility, so normal imports don't actually pick it up and you have to use the full path if you're not in the same module.

macros/src/lib.rs
```diff
@@ -458,3 +459,141 @@ impl Parse for OptionSetter {
         Ok(Self { attrs, name, type_ })
     }
 }
+
+#[import_tokens_attr]
+#[proc_macro_attribute]
+pub fn option_setters_2(
```
jadalilleboe, line 465: Is `option_setters` still being used somewhere or can it be deleted after making this `option_setters_2`?

abr-egn: Once all of the actions have been moved over I'll delete `option_setters` and rename `option_setters_2` to replace it.

jadalilleboe (approved):

macros/src/lib.rs
```diff
@@ -458,3 +459,141 @@ impl Parse for OptionSetter {
...
+
+    // Add the list of setters to the rustdoc for the fn
+    impl_fn.attrs.push(parse_quote! {
+        #[doc = ""]
+    });
+    impl_fn.attrs.push(parse_quote! {
+        #[doc = "These methods can be chained before calling `.await` to set options:"]
+    });
+    for name in setter_names {
+        let docstr = format!("  * [{0}]({1}::{0})", name, doc_path);
```
isabelatkinson, line 593: nit: would it be possible to code format the method names (i.e. "`session`" instead of "session")? not sure if that's doable in a format string like this

isabelatkinson (approved):
looks great, just one suggestion! I usually find macro code hard to understand but this is quite readable

PR #1253, opened 2024-11-27
RUST-564 Implement socketTimeoutMs option

jadalilleboe:
[RUST-564](https://jira.mongodb.org/browse/RUST-564)

There were three spec tests being skipped due to `socketTimeoutMs not being implemented`, but on further investigation it seems like other drivers are skipping these tests due to them being flaky. The C# driver is [skipping "Ignore network timeout error on find" and "Reset server and pool after network timeout error during authentication"](https://jira.mongodb.org/browse/CSHARP-4459) due to flakiness and the Python driver has [a BF associated with the spec test "add RetryableWriteError and UnknownTransactionCommitResult labels to connection errors"](https://jira.mongodb.org/browse/PYTHON-4987). Therefore, I think these tests being unskipped should be part of a separate ticket and I changed the description of why they are being skipped to "flaky tests".

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -114,6 +114,7 @@ tracing = { version = "0.1.36", optional = true }
 typed-builder = "0.10.0"
 webpki-roots = "0.25.2"
 zstd = { version = "0.11.2", optional = true }
+pin-project = "1.1.7"
 
 [dependencies.pbkdf2]
 version = "0.11.0"
--- a/src/action/client_options.rs
+++ b/src/action/client_options.rs
@@ -44,7 +44,7 @@ impl ClientOptions {
     ///   * `retryWrites`: not yet implemented
     ///   * `retryReads`: maps to the `retry_reads` field
     ///   * `serverSelectionTimeoutMS`: maps to the `server_selection_timeout` field
-    ///   * `socketTimeoutMS`: unsupported, does not map to any field
+    ///   * `socketTimeoutMS`: maps to `socket_timeout` field
     ///   * `ssl`: an alias of the `tls` option
     ///   * `tls`: maps to the TLS variant of the `tls` field`.
     ///   * `tlsInsecure`: relaxes the TLS constraints on connections being made; currently is just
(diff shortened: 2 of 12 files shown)


abr-egn:
So it turns out that I'd completely misunderstood what `read_timeout`/`write_timeout` do on sockets - when they're in nonblocking mode (which is what async I/O is built on), it boils down to "essentially nothing".   They definitely *don't* cause timeout errors when that duration is passed, at most they cause the future to be polled an extra time.

TL;DR - this PR is exactly correct according to the specifications I gave, but those were wrong :)

I'm looking to see if I can implement a timeout at the level of our existing `AsyncStream` wrapper without too much hassle.  This is made a pain by the very low-level async nature of the stream APIs here, if it gets to be too much effort this probably won't be worth it since it was intended as a quick workaround to the lack of CSOT.

abr-egn:
TL;DR - it looks like this won't be worth the engineer time it would take.

Longer version: it would be possible to implement this.  However, because of technical details*, doing so in a performant way would be a challenging and nontrivial project.  In the process of investigating this, I came across information that led me to revisit our conclusions that CSOT (RUST-582) wouldn't be feasible, and I now believe that is reasonably doable and the better path.

\* The Tokio [Sleep](https://docs.rs/tokio/latest/tokio/time/struct.Sleep.html) future doesn't implement the `Unpin` trait, which would transitively make any stream implementation using it to implement a timeout also not `Unpin`.  However, `Self: Unpin` is required for [most uses](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read) of async I/O.  The potential solutions for this are:
* Wrap the `Sleep` future in a `Box`, which is always `Unpin` no matter the contents.  However, this means that our hypothetical socket-with-timeout stream would potentially incur a heap allocation on every socket read or write, which doesn't seem great.
* Implement our own timing machinery that is `Unpin`.  Presumably, if it were easy to do this in a general, performant way Tokio would have done so already

PR #1256, opened 2024-12-02
RUST-226 Support tlsCertificateKeyFilePassword

abr-egn:
RUST-226

This will (probably; I'm not sure how to test it) allow using `tlsCertificateKeyFilePassword` or the corresponding in-code option field.  Should work for both `rustls-tls` and `openssl-tls`.


src/runtime/tls_openssl.rs
```diff
@@ -26,16 +28,19 @@ impl TlsConfig {
     /// Create a new `TlsConfig` from the provided options from the user.
     /// This operation is expensive, so the resultant `TlsConfig` should be cached.
     pub(crate) fn new(options: TlsOptions) -> Result<TlsConfig> {
+        if options.tls_certificate_key_file_password.is_some() {
+            return Err(ErrorKind::InvalidArgument {
+                message: "'tlsCertificateKeyFilePassword' can't be used with 'openssl-tls'".into(),
+            }
+            .into());
+        }
+
```
kevinAlbs, line 37:
```suggestion
```

src/runtime/tls_openssl.rs
```diff
@@ -66,25 +71,43 @@ pub(super) async fn tls_connect(
...
     if let Some(path) = cert_key_file_path {
-        builder.set_certificate_file(path.clone(), SslFiletype::PEM)?;
-        builder.set_private_key_file(path, SslFiletype::PEM)?;
+        builder
+            .set_certificate_file(path.clone(), SslFiletype::PEM)
+            .map_err(openssl_err)?;
+        if let Some(key_pw) = tls_certificate_key_file_password {
+            let contents = std::fs::read(&path)?;
+            let key_bytes = decrypt_private_key(&contents, &key_pw)?;
+            let key = openssl::pkey::PKey::private_key_from_pem(&key_bytes).map_err(openssl_err)?;
```
kevinAlbs, line 104:
```suggestion
            let key = openssl::pkey::PKey::private_key_from_der(&key_bytes).map_err(openssl_err)?;
```

Fixes observed error testing locally:

```
Error: Error { kind: InvalidTlsConfig { message: "error:1E08010C:DECODER routines:OSSL_DECODER_from_bio:unsupported:crypto/encode_decode/decoder_lib.c:102:No supported data to decode. Input type: PEM" }, labels: {}, wire_version: None, source: None }
```

IIUC, PEM is the base64 encoding, DER is the binary encoding.

src/client/options.rs
```diff
@@ -1047,6 +1047,10 @@ pub struct TlsOptions {
     /// The default value is to error on invalid hostnames.
     #[cfg(feature = "openssl-tls")]
     pub allow_invalid_hostnames: Option<bool>,
+
+    /// If set, the key in `cert_key_file_path` must be encrypted with this password. Only
+    /// supported with `rustls`.
```
kevinAlbs, line 1052:
```suggestion
    /// If set, the key in `cert_key_file_path` must be encrypted with this password.
```

kevinAlbs (approved):
LGTM with additional suggestions applied.

Testing with the [drivers-evergreen-tools certificates](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/x509gen/client-pkcs8-encrypted.pem) initially got this error:
```
Error { kind: InvalidTlsConfig { message: "Invalid encrypted private key: PKCS#8 ASN.1 error: unknown/unsupported OID: 1.2.840.113549.3.7 at DER byte 66" }, labels: {}, wire_version: None, source: None }
```

OID [1.2.840.113549.3.7](https://oidref.com/1.2.840.113549.3.7) refers to a DES algorithm. The error was fixed by adding features `3des`, `des-insecure`, and `sha1-insecure`:
```
pkcs8 = { version = "0.10.2", features = [
    "encryption",
    "pkcs5",
    "3des",
    "des-insecure",
    "sha1-insecure",
] }
```

However, https://docs.rs/pkcs8/latest/pkcs8/ notes:

> DES support (gated behind the des-insecure feature) is implemented to allow for decryption of legacy PKCS#8 files only.

I expect that the drivers-evergreen-tools certificate using outdated algorithms. I suggest keeping those feature flags out for now (as is done in current changes).

With those feature flags added, I was able to connect (both with `openssl-tls` and without) with [this test](https://github.com/kevinAlbs/rust-bootstrap/tree/ce1e7d02ca7cf0b897ee9d3f1cb42ac9364bdde9/test-tlsCertificateKeyFilePassword).

abr-egn:
> I expect that the drivers-evergreen-tools certificate using outdated algorithms. I suggest keeping those feature flags out for now (as is done in current changes).

Yup, that makes sense.  Downstream users can enable those features if they happen to be using keys with those algorithms.

> With those feature flags added, I was able to connect (both with `openssl-tls` and without) with [this test](https://github.com/kevinAlbs/rust-bootstrap/tree/ce1e7d02ca7cf0b897ee9d3f1cb42ac9364bdde9/test-tlsCertificateKeyFilePassword).

Thank you!

abr-egn:
Update: I've put this behind an optional feature flag, since my expectation is that the very large majority of users will never want this and that avoids pulling in extra dependencies for them.

isabelatkinson (commented):
Code changes LGTM. There's a parsing test in `src/client/options/test.rs` that can be unskipped as part of this as well. Would it be feasible to add something like Kevin's [test](https://github.com/kevinAlbs/rust-bootstrap/tree/ce1e7d02ca7cf0b897ee9d3f1cb42ac9364bdde9/test-tlsCertificateKeyFilePassword) to CI? I'd like to avoid inadvertently breaking this in the future

abr-egn:
> Code changes LGTM. There's a parsing test in `src/client/options/test.rs` that can be unskipped as part of this as well. Would it be feasible to add something like Kevin's [test](https://github.com/kevinAlbs/rust-bootstrap/tree/ce1e7d02ca7cf0b897ee9d3f1cb42ac9364bdde9/test-tlsCertificateKeyFilePassword) to CI? I'd like to avoid inadvertently breaking this in the future

Unskipped the test; I'd like to leave the CI integration test for a follow-up PR for expediency and to make merging into the backport branch simpler.

PR #1261, opened 2024-12-06
RUST-2103 Better documentation for action options, batch 2

abr-egn:
RUST-2103

This updates more actions to use the doc generation macros for option setters.  Some of these (notably `aggregate`) have setters split across multiple `impl` blocks with different type constraints that we want to show as a merged list in the documentation, so this required some changes to the machinery.

In particular, `option_setters_2` now accepts the name to export tokens for doc use under, and an optional list of extra setter names to include in those exported tokens.  This could be done at least two other ways:
* the extra names could be passed in to the `options_doc` macro; I didn't go with that because that would mean repeating them for each use of the macro, which is always at least twice (async/sync) and sometimes more if an action type is shared.
* the impls could be pairwise-merged via another macro ('option_setters_merge_export'?) for doc consumption; that seems like it'd get messy fast.

Since this required introducing custom argument parsing for `option_setters_2`, I made it use named `key = value` arguments for readability.

Unrelated to any of the above, I also introduced an optional `sync` arg to `options_doc` which just changes its wording slightly to be appropriate.

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -1,15 +1,19 @@
 extern crate proc_macro;
 
 use macro_magic::{import_tokens_attr, mm_core::ForeignPath};
+use proc_macro2::extra;
 use quote::{quote, ToTokens};
 use syn::{
     braced,
+    bracketed,
     parenthesized,
     parse::{Parse, ParseStream},
     parse_macro_input,
     parse_quote,
     parse_quote_spanned,
+    punctuated::Punctuated,
     spanned::Spanned,
+    token::Bracket,
     Attribute,
     Block,
     Error,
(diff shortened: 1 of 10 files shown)


macros/src/lib.rs
```diff
@@ -545,21 +550,134 @@ pub fn option_setters_2(
...
+}
+
+#[derive(Debug)]
+struct OptionSettersArgsExtra {
+    extra_text: (Ident, Token![=]), // extra =
+    bracket: Bracket,
+    names: Punctuated<Ident, Token![,]>,
+}
+
+impl Parse for OptionSettersArgs {
```
abr-egn, line 593: `Parse` + `ToTokens` + `ForeignPath` is required for `macro_magic` custom arguments, it's a bit of a song and dance.  At least it's invisible at the call site, though.

src/action/count.rs
```diff
@@ -1,13 +1,18 @@
-use bson::Document;
+use bson::{Bson, Document};
+use mongodb_internal_macros::{option_setters_2, options_doc};
+use std::time::Duration;
 
 use crate::{
-    coll::options::{CountOptions, EstimatedDocumentCountOptions},
+    coll::options::{CountOptions, EstimatedDocumentCountOptions, Hint},
```
abr-egn, line 6: This is kind of an annoying consequence - because the setters are taken from the options struct as a token stream, the types have to be re-imported here.  Actual compile-time reflection with type information rather than just textual macros would be the answer but that got dramabombed so who knows if rust will ever get that :(

macros/src/lib.rs
```diff
@@ -545,21 +550,134 @@ pub fn option_setters_2(
...
+    fn foreign_path(&self) -> &syn::Path {
+        &self.foreign_path
+    }
+}
+
+impl Parse for OptionSettersArgsExtra {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let extra_text = (parse_name(input, "extra")?, input.parse::<Token![=]>()?);
+        let content;
+        let bracket = bracketed!(content in input);
```
isabelatkinson, line 651: this was very confusing to read before I remembered `in` is a keyword...

isabelatkinson (approved):

PR #1268, opened 2024-12-11
RUST-2117 Pin dependencies for MSRV

abr-egn:
RUST-2117

Turns out that there were a bunch beyond `hickory-resolver` that had crept up.


isabelatkinson (approved):
lgtm! i'll kick off the release tomorrow morning

PR #1272, opened 2024-12-20
RUST-2103 Update the rest of actions to document options

abr-egn:
RUST-2013

This did involve one more rework of how the macros handle things (sorry!); with that, there's now a family of three related macros:
* `option_setters`: takes the path to an options struct and generates setters for the fields of that struct in the `impl` block to which it's attached.  Can be optionally told to skip setters for some of the fields.
* `export_doc`: generates a notional named "doc item" for use by `options_doc` from an `impl` block.  Can be optionally told to add entries to the doc item that aren't present in the code of the `impl`.
* `options_doc`: consumes a "doc item" to add rustdoc listing option setters to a `fn`.  Can be optionally told it's a sync fn to adjust the rustdoc output.  (unchanged from last PR)

With this split, I think it's both easier to read and understand (no more multiline kwarg macro invocations) and it much better handles edge cases, of which we have surprisingly many.

Oh, and I also split up the contents of `macros/src/lib.rs` since that had gotten quite big.


src/action/csfle/encrypt.rs
```diff
@@ -142,12 +148,10 @@ pub struct RangeOptions {
     pub precision: Option<i32>,
 }
 
-impl<Mode> Encrypt<'_, Mode> {
-    option_setters!(options: EncryptOptions;
-        contention_factor: i64,
-        range_options: RangeOptions,
-    );
-}
+#[option_setters(EncryptOptions, skip = [query_type])]
```
abr-egn, line 151: Here's one of the fun corner cases!  The `Encrypt` action is shared across both `encrypt` and `encrypt_expression`; we want to expose the `query_type` option for the former, both in code and documentation, but not the latter.  So:
* For the generic `impl` block, the generated setter for `query_type` is skipped, and it's manually implemented for `Encrypt<'_, Value>` only.
* For the generated documentation for `encrypt`, the manually written setter for `query_type` is added to the list
* For the generated documentation for `encrypt_expr`, it isn't

src/action/shutdown.rs
```diff
@@ -130,6 +134,7 @@ pub struct Shutdown {
     pub(crate) immediate: bool,
 }
 
+#[export_doc(shutdown)]
```
abr-egn, line 137: Another corner case: we want to generate documentation for the options for `shutdown`, but there's no corresponding options object.

src/action/csfle/encrypt.rs
isabelatkinson: Somewhat tangential, but this led me to find a niche corner case that can lead to bad behavior when using `with_options` in a call to `encrypt_expression`:

```
let options = EncryptOptions::builder()
    .contention_factor(0)
    .range_options(range_options)
    .build();
let result = client_encryption
    .encrypt_expression(...)
    .with_options(options)
    .await?;
```

The call to `encrypt_expression` sets the `query_type` to the correct value on the options [internally](https://docs.rs/mongodb/latest/src/mongodb/action/csfle/encrypt.rs.html#60), but chaining `with_options` wipes this value, and the following error is returned: `EncryptExpression may only be used for range queries`.

We should add the following validation to the `IntoFuture` impl for `Encrypt<Expression>`:
- if `query_type` is set to `Some(<string not equal to "range">)`, return an error indicating that `query_type` cannot be set via `with_options` for calls to `encrypt_expression`
- if `query_type` is set to `None`, manually re-set the value to `Some("range")` as the user inadvertently overrode it by the call to `with_options`

I can make a follow-up PR to fix this; we're also missing a `TypedBuilder` impl for `EncryptOptions` so I'll add that too.

isabelatkinson (approved):
the updated docs look great!

src/action/csfle/encrypt.rs
abr-egn: Nice catch!  Follow up PR makes sense, thank you :)

PR #1276, opened 2025-01-07
RUST-1905 Add Kubernetes Support for OIDC

abr-egn:
RUST-1905

This adds support for the "k8s" environment for OIDC, and tests to exercise it.  The driver change is pretty minimal, but the tests required doing some refactoring of the evergreen config for how we were doing OIDC testing to pull out common functionality, and there's some ritual around running things in the k8s providers.


src/client/auth/oidc.rs
```diff
@@ -144,7 +157,7 @@ impl Callback {
 
     /// Create azure callback.
     #[cfg(feature = "azure-oidc")]
-    fn azure_callback(client_id: Option<&str>, resource: &str) -> CallbackInner {
+    fn azure_callback(client_id: Option<&str>, resource: &str) -> Function {
```
abr-egn, line 160: This signature change wasn't needed for the functionality here, it just gets rid of a level of indentation and means each of the `*_callback` functions doesn't need to construct its own fresh cache that can be done just as easily at the call site.

src/client/auth/oidc.rs
```diff
@@ -886,3 +916,90 @@ async fn send_sasl_command(
         response.auth_response_body(MONGODB_OIDC_STR)?,
     )
 }
+
+pub(super) fn validate_credential(credential: &Credential) -> Result<()> {
```
abr-egn, line 920: This is mostly just copied over from its old location, with a few new validation cases to catch things we hadn't been that the newly synced spec tests showed up.

src/test/spec/oidc.rs
```diff
@@ -1286,3 +1285,25 @@ mod gcp {
...
+        bson::{doc, Document},
+        Client,
+    };
+
+    // There's no spec test for K8s, so we run this simple sanity check.
+    #[tokio::test]
+    async fn successfully_authenticates() -> anyhow::Result<()> {
+        get_env_or_skip!("OIDC");
+
+        let client = Client::with_uri_str(mongodb_uri_single!()).await?;
```
isabelatkinson, line 1300: `mongodb_uri_single` looks for the "MONGODB_URI_SINGLE" environment var and skips if it's unset - where is this variable being defined? I can't seem to find it anywhere

isabelatkinson (commented):
lgtm!

src/test/spec/oidc.rs
abr-egn: It's set in the drivers-evergreen-tools [k8s/setup.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/f8420d2cf735a1d574e9be48428ac481c5572688/.evergreen/auth_oidc/k8s/setup.sh#L16) script.

isabelatkinson (approved):

PR #1279, opened 2025-01-08
RUST-2125 Run tests using LB URI when present

abr-egn:
RUST-2125

Turns out we were, in fact, not running the driver test suite in loadbalanced mode.  [This run](https://spruce.mongodb.com/version/677eab6d0a5b3c00075cbda7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) shows a test properly failing with a pinning error with this fix in place and the earlier bugfix to the pinning logic reverted.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1510,6 +1510,8 @@ functions:
         include_expansions_in_env:
           - PROJECT_DIRECTORY
           - OPENSSL
+          - SINGLE_MONGOS_LB_URI
+          - MULTI_MONGOS_LB_URI
           - MONGODB_URI
           - MONGODB_API_VERSION
           - PATH


src/operation/bulk_write.rs
```diff
@@ -292,9 +292,12 @@ where
                 None,
                 self.options.and_then(|options| options.comment.clone()),
             );
-            let pinned_connection = self
-                .client
-                .pin_connection_for_cursor(&specification, context.connection)?;
+
+            let pinned_connection = self.client.pin_connection_for_cursor(
```
abr-egn, line 296: This was a distinct bug from RUST-2131, where the original code should have been conditionally using `pin_connection_for_session` when a session was present; not doing so caused tests of bulk write on load balancer running inside a transaction to fail with an internal "already pinned" error.  I merged the two `pin_connection_for_*` methods to prevent this kind of easy mistake in the future and make it easier to reason about exactly when a connection for a cursor gets pinned.

src/test/spec/unified_runner/operation/session.rs
```diff
@@ -78,7 +78,7 @@ impl TestOperation for AssertSessionPinned {
         async move {
             let is_pinned =
                 with_mut_session!(test_runner, self.session.as_str(), |session| async {
-                    session.transaction.pinned_mongos().is_some()
+                    session.transaction.is_pinned()
```
abr-egn, line 81: Also a distinct bug, just missing that there are now more cases than `pinned_mongos` :)

src/cmap/conn.rs
```diff
@@ -327,16 +327,26 @@ impl PinnedConnectionHandle {
...
     pub(crate) async fn take_connection(&self) -> Result<PooledConnection> {
+        use tokio::sync::mpsc::error::TryRecvError;
         let mut receiver = self.receiver.lock().await;
-        let mut connection = receiver.recv().await.ok_or_else(|| {
-            Error::internal(format!(
-                "cannot take connection after unpin (id={})",
-                self.id
-            ))
-        })?;
+        let mut connection = match receiver.try_recv() {
```
abr-egn, line 335: Using `try_recv` rather than `recv` means the re-entrant case becomes an error rather than a deadlock, which seems much better behavior to me.  The hypothetical downside is that it could fail in a race condition, but since any individual pinned connection is used in a strictly linear sequence that doesn't apply here.

isabelatkinson (approved):
tiny doc fix, otherwise looks good!

PR #1283, opened 2025-01-14
RUST-2112 Test connecting with tlsCertificateKeyFilePassword

abr-egn:
RUST-2112

I was originally hoping this could just be another test function in the suite with logic about when to run, but since it needs to set features on `pkcs8`, it gets its own binary and buildvariant.


etc/tls_cert_password/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+};
+
+#[tokio::main]
+async fn main() {
+    let mut certpath = PathBuf::from(std::env::var("DRIVERS_TOOLS").unwrap());
+    certpath.push(".evergreen/x509gen");
+    let mut options = mongodb::options::ClientOptions::parse(std::env::var("MONGODB_URI").unwrap())
+        .await
+        .unwrap();
+    options.tls = Some(Tls::Enabled(
```
abr-egn, line 16: Since we're testing parsing elsewhere, I figure setting the options directly gives us slightly more coverage.

isabelatkinson (commented):
> it needs to set features on pkcs8

How about adding an entry to `dev-dependencies` for `pkcs8` with the additional feature flags enabled? I did a test of this locally and it seems to work as long as the non-dev and dev dependencies don't have conflicting features and/or versions

abr-egn:
> How about adding an entry to `dev-dependencies` for `pkcs8` with the additional feature flags enabled? I did a test of this locally and it seems to work as long as the non-dev and dev dependencies don't have conflicting features and/or versions

... yup, that makes it much simpler, thank you!

isabelatkinson (approved):

PR #1285, opened 2025-01-15
Skip KMS retry test on serverless

isabelatkinson:
The certs used in `kms_retry` are not valid for connecting with serverless


isabelatkinson:
having trouble getting a successful patch for serverless due to other flakiness, but this fixes the KMS retry test at least :)

PR #1288, opened 2025-01-22
make from_raw publicly accessible

wangyk414:
This allows the client to persist and re-construct a ResumeToken, which is particularly useful when the change stream consumption is interrupted and the client is able to restart from a persisted ResumeToken without losing any changes.

### DIFF
--- a/src/change_stream/event.rs
+++ b/src/change_stream/event.rs
@@ -38,7 +38,8 @@ impl ResumeToken {
         }
     }
 
-    pub(crate) fn from_raw(doc: Option<RawDocumentBuf>) -> Option<ResumeToken> {
+    /// A utility method to create a `ResumeToken` from a raw BSON document.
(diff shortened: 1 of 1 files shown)


isabelatkinson:
Hey @wangyk414, thanks for opening this PR!

> to persist and re-construct a ResumeToken

What is your use case for the intermediary document rather than storing the `ResumeToken` directly? The driver treats `ResumeToken` as an opaque type throughout, so you shouldn't need to use documents for resume tokens anywhere in the API.

wangyk414:
Hi @isabelatkinson, good day.
I am currently utilizing the change stream to monitor changes in the mongo documents in a long running service, which produces gcp pubsub messages if some specific fields get changed. To ensure continuity, this service need to persist the latest used resume token, so that upon a restart or temporary offline period, it could resume consuming the change steam from that timestamp. And no document changes are lost.

PR #1289, opened 2025-01-23
RUST-1891 Implement `Default` on results structs

isabelatkinson:
RUST-1891

Implements `Default` on the types contained in the `results` module. I don't think there's anything of value to test here.

### DIFF
--- a/src/results.rs
+++ b/src/results.rs
@@ -19,7 +19,7 @@ pub use bulk_write::*;
 
 /// The result of a [`Collection::insert_one`](../struct.Collection.html#method.insert_one)
 /// operation.
-#[derive(Clone, Debug, Serialize)]
+#[derive(Clone, Debug, Serialize, Default)]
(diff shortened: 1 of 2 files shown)


src/results.rs
```diff
@@ -120,14 +112,15 @@ pub(crate) struct GetMoreResult {
...
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
 #[serde(rename_all = "camelCase")]
 #[non_exhaustive]
 pub enum CollectionType {
     /// Indicates that the data store is a view.
     View,
 
     /// Indicates that the data store is a collection.
+    #[default]
```
isabelatkinson, line 123: this felt like a reasonable default; don't think the choice here is that critical, though, since these impls are for mocking

abr-egn (approved):

PR #1293, opened 2025-01-29
RUST-2004 Benchmark client bulk write

isabelatkinson:
RUST-2004

This also includes some clippy fixes/general cleanup in the second and third commits.


benchmarks/src/bench.rs
```diff
@@ -57,6 +62,38 @@ pub static TARGET_ITERATION_COUNT: Lazy<usize> = Lazy::new(|| {
         .expect("invalid TARGET_ITERATION_COUNT")
 });
 
+static SMALL_DOC: OnceCell<Document> = OnceCell::const_new();
```
abr-egn, line 65: Nit: I think these statics can be moved into the scope of the `get_` functions that use them.

benchmarks/src/bench.rs
```diff
@@ -57,6 +62,38 @@ pub static TARGET_ITERATION_COUNT: Lazy<usize> = Lazy::new(|| {
...
+pub async fn get_small_doc() -> Document {
+    SMALL_DOC
+        .get_or_init(|| async {
+            let data_path = DATA_PATH
+                .join("single_and_multi_document")
+                .join("small_doc.json");
+            let mut file = spawn_blocking_and_await!(std::fs::File::open(data_path))
```
abr-egn, line 72: Nit: It might be worth factoring out a utility function like
```
async fn from_json_file<T: DeserializeOwned>(path: impl AsRef<Path>) -> Result<T>
```

benchmarks/src/bench.rs
```diff
@@ -69,7 +106,7 @@ pub trait Benchmark: Sized {
         Ok(())
     }
 
-    async fn do_task(&self) -> Result<()>;
+    async fn do_task(&mut self) -> Result<()>;
```
abr-egn, line 109: My understanding is that the introduction of `mut` here is to support copying the write models in `before_task` rather than `do_task` (which absolutely makes sense).

Personally, I'd rather avoid the `mut` on `do_task` because I think it makes it harder to reason about; how would you feel about, instead, updating the `Benchmark` trait with
```
type TaskState = ();
async fn before_task(&mut self) -> Result<TaskState> {
    Ok(())
}
async fn do_task(&self, state: TaskState) -> Result<()>;
```
That avoids the need for `mut` and the `Option`/`take` ceremony, but might be more type-level complication than it's worth.  WDYT?

isabelatkinson (commented):
new patch: https://spruce.mongodb.com/version/67a1255a9006ea0007996014/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

abr-egn (approved):

PR #1303, opened 2025-02-07
RUST-2055 Use cargo-nextest to filter tests

isabelatkinson:
This is an initial pass over tests we skip with environment variables set in test scripts. I'm putting up this PR now to minimize size and get feedback on the initial strategy. The two pieces remaining are:

- CSFLE tests (currently in progress, will get a separate PR)
- A pass over the inline shell scripts in `.evergreen/config.yml`

The changes in the planned maintenance testing file will also require a PR on drivers-atlas-testing.


src/test/index_management.rs
```diff
@@ -1,14 +1,35 @@
...
     IndexModel,
 };
 
+#[tokio::test]
+async fn run_unified() {
+    let client = Client::for_test().await;
+
+    let mut skipped_files = Vec::new();
+    let mut skipped_tests = Vec::new();
+    // TODO DRIVERS-2794: unskip these tests
```
isabelatkinson, line 21: I think changing around the test names changed the order in which the index management tests are being run such that we now hit the issue described in DRIVERS-2794

src/test.rs
```diff
@@ -2,10 +2,11 @@
 #![allow(clippy::cast_possible_wrap)]
 
 #[cfg(feature = "dns-resolver")]
-mod atlas_connectivity;
-mod atlas_planned_maintenance_testing;
-#[cfg(feature = "aws-auth")]
-mod auth_aws;
+#[path = "test/atlas_connectivity.rs"]
```
abr-egn, line 5: This is handy, I had no idea there was an annotation to allow it.

abr-egn (approved):
LGTM, much cleaner :)

PR #1306, opened 2025-02-18
RUST-2055 Use `cargo-nextest` to filter CSFLE tests

isabelatkinson:
RUST-2055

The changes here are mostly shuffling things around (sorry for the very noisy diff). I split the prose tests into two files based on whether they require the KMIP server to be running, and added a `skip_local` filter for the KMIP tests. I also added in-file modules for the prose tests that have multiple test functions and/or their own utility functions for readability. `src/test/csfle.rs` contains common utilities.

I removed the `check_env` utility so that the prose tests will fail if the environment variables they need are not set. I didn't add a `skip_local` filter to the non-KMIP tests because I never test the driver with the `in-use-encryption` feature enabled if I don't intend to run the CSFLE tests with the proper credentials set. Same goes for tests behind additional feature flags. I can change this, though, if you have a different workflow.


abr-egn (commented):
Code structure is much better, thank you!

I do occasionally run the full test suite with all features enabled when I've changed something core to the driver, so I think having the prose tests behind a `skip_local` filter as well would be good.  LGTM otherwise though!

isabelatkinson (commented):
> I do occasionally run the full test suite with all features enabled when I've changed something core to the driver, so I think having the prose tests behind a `skip_local` filter as well would be good.

Makes sense! I put all of `test::csfle` behind `skip_local` and organized the prose tests into separate modules based on the environment they require.

PR #1312, opened 2025-02-21
minor: skip sessions not supported tests on 8.1+

isabelatkinson:
Running `buildInfo` on `mongocryptd` 8.1+ causes errors (perhaps related to [SERVER-87874](https://jira.mongodb.org/browse/SERVER-87874)?) for the sessions-not-supported tests. I skipped the tests on 8.1+ for now, but we should do RUST-1447 so that we don't need to call `buildInfo` and other metadata-retrieving commands for each test client we create.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -238,7 +238,7 @@ buildvariants:
 
   - name: gcp-kms
     display_name: "GCP KMS"
-    # patchable: false
+    patchable: false
     run_on:
       # The GCP CLI is not available on RHEL/Ubuntu machines.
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -238,7 +238,7 @@ buildvariants:
 
   - name: gcp-kms
     display_name: "GCP KMS"
-    # patchable: false
+    patchable: false
```
isabelatkinson, line 241: this snuck in from my last PR

abr-egn (approved):

PR #1316, opened 2025-02-26
RUST-2140 Support the nsType field for change streams

abr-egn:


abr-egn:
MSRV and clippy failures are unrelated (RUST-2165 and RUST-2164 respectively).

PR #1317, opened 2025-02-27
RUST-2055 Filter OIDC tests with `cargo nextest`

isabelatkinson:


src/test/spec/oidc.rs
```diff
@@ -507,22 +494,22 @@ mod basic {
     }
 
     #[tokio::test]
+    #[cfg(target_os = "linux")] // MONGODB_URI_MULTI is only set when running on linux
```
isabelatkinson, line 497: from the setup script: https://github.com/mongodb-labs/drivers-evergreen-tools/blob/b65459ec9bb6785c731f16b0218d4cc1ced05670/.evergreen/auth_oidc/setup.sh#L26

abr-egn (approved):

PR #1318, opened 2025-03-04
RUST-1447 Cache test client metadata

isabelatkinson:
RUST-1447

Removes all server round-trips from test client creation and instead caches client metadata in a number of lazy static variables in `src/test.rs`. The changes outside of `src/test.rs` and `src/test/util.rs` are largely search-and-replace; I also removed some checks related to 4.0 now that we don't support/test any lower server versions.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -55,16 +55,19 @@ use tokio::sync::OnceCell;
 #[cfg(feature = "tracing-unstable")]
 use self::util::TracingHandler;
 use crate::{
+    bson::{doc, Document},
     client::{
         auth::Credential,
         options::{ServerApi, ServerApiVersion},
     },
-    options::ClientOptions,
+    hello::HelloCommandResponse,
+    options::{ClientOptions, ServerAddress},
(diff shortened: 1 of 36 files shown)


src/test.rs
```diff
@@ -73,6 +76,168 @@ pub(crate) async fn get_client_options() -> &'static ClientOptions {
         })
         .await
 }
+pub(crate) async fn auth_enabled() -> bool {
+    get_client_options().await.credential.is_some()
+}
+
+async fn get_server_version() -> &'static semver::Version {
+    static SERVER_VERSION: OnceCell<semver::Version> = OnceCell::const_new();
```
abr-egn, line 84: I wonder if it might be better to bundle the `static` values into a single struct with a single lazy initializer; as it is a new `Client` has to be constructed for each group of related functions and that would collapse it down to a single one.

This is well into micro-optimization territory, though, so up to you if you want to implement that or merge as-is :)

abr-egn (approved):

PR #1319, opened 2025-03-04
HELP-68823 Fix invalid_me comparison and normalize host strings

abr-egn:
HELP-68823

This wasn't the fix that I was intending to put into mainline, but that turns out to have a lot more in the way of wrinkles than I expected.  In particular, we have some public APIs (e.g. [`ToplogyDescription::servers`](https://docs.rs/mongodb/latest/mongodb/event/sdam/struct.TopologyDescription.html#method.servers)) that hand out `&ServerDescription` values, which locks us into having the underlying data structure storing that type either as the primary (how it's currently done) or as some kind of internally-mutable cache (ew).

There may be some clever path forward for split types that maintains API compatibility that I didn't see, but for the moment I think it's best to get the simple fix in and unblock downstream users.  This PR, while minimal, fixes the core issue and explicitly normalizes anything that parses as an IP address, so it should be reasonably robust, and I've gone through and audited all the other uses of `ServerAddress`'s `Display` impl and they're all test code or error messages so I think the chance of another recurrence of this problem is low.

### DIFF
--- a/src/client/options.rs
+++ b/src/client/options.rs
@@ -11,7 +11,7 @@ use std::{
     convert::TryFrom,
     fmt::{self, Display, Formatter, Write},
     hash::{Hash, Hasher},
-    net::Ipv6Addr,
+    net::{Ipv4Addr, Ipv6Addr},
     path::PathBuf,
     str::FromStr,
     time::Duration,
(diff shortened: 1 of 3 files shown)


isabelatkinson (approved):
LGTM! This seems like a good fix for now.

> In particular, we have some public APIs (e.g. [ToplogyDescription::servers](https://docs.rs/mongodb/latest/mongodb/event/sdam/struct.TopologyDescription.html#method.servers)) that hand out &ServerDescription values

Would this make sense to revisit in 4.0? i.e. return owned `ServerDescription`s in the public API and convert from the internal type in those methods

PR #1336, opened 2025-03-19
RUST-2181 Fix rustup installation on windows

abr-egn:
RUST-2181

This works around the cygwin symlink bug (https://github.com/rust-lang/rustup/issues/4239) with a small script that replaces the symlinks with actual copies.  This wouldn't be a good workaround for a long-lived installation because it'd mess with the rustup upgrade process, but because we're doing a clean install every time and then discarding it once tests are done it's fine here.

I went with python for the script rather than bash because I think it ends up being substantially more readable.


kevinAlbs (approved):
LGTM

PR #1339, opened 2025-03-20
RUST-2172 Test on Graviton processor

abr-egn:
RUST-2172

Happily, no changes were needed.  Full test against older versions is [here](https://spruce.mongodb.com/version/67dc3757e942f7000766eea2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) - I left those evergreen configs present but commented out to make it easy to re-run in the future if needed.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -376,6 +376,26 @@ buildvariants:
     tasks:
       - happy-eyeballs-task-group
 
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
LGTM. Having the other versions commented out seems handy.

PR #1343, opened 2025-03-25
RUST-2107 Remove deprecated fields from gridfs tests

abr-egn:


isabelatkinson (approved):
there's some logic related to these fields we should be able to remove now, but otherwise LGTM
- test skips: https://github.com/mongodb/mongo-rust-driver/blob/458228f940ae0962bdb5846e538ca7e989a7b519/src/test/spec/gridfs.rs#L24
- deserialization fields: https://github.com/mongodb/mongo-rust-driver/blob/458228f940ae0962bdb5846e538ca7e989a7b519/src/test/spec/unified_runner/operation/gridfs.rs#L97

PR #1358, opened 2025-04-11
RUST-2131 Fix bulk write cursor iteration on load balanced topologies

isabelatkinson:
RUST-2131

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -31,6 +31,7 @@ use crate::{
         },
         ConnectionPool,
         RawCommandResponse,
+        StreamDescription,
     },
     cursor::{session::SessionCursor, Cursor, CursorSpecification},
     error::{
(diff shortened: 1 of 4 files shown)


src/client/executor.rs
```diff
@@ -105,42 +106,55 @@ impl Client {
...
+            {
+                return Err(ErrorKind::Transaction {
+                    message: "read preference in a transaction must be primary".into(),
                 }
+                .into());
             }
-            self.execute_operation_with_retry(op, session).await
-        })
-        .await
+            // If the current transaction has been committed/aborted and it is not being
```
abr-egn, line 144: I like having this here instead of in `execute_operation_with_retry` - it makes the purpose of that method cleaner - but for my own understanding, was there a functional reason for this change beyond that?

isabelatkinson: nope, this and the reshuffling above were to simplify `execute_operation_with_retry` and `execute_operation_on_connection`

src/operation/bulk_write.rs
```diff
@@ -293,27 +371,28 @@ where
...
-            let iteration_result = match context.session {
-                Some(session) => {
-                    let mut session_cursor =
-                        SessionCursor::new(self.client.clone(), specification, pinned_connection);
-                    self.iterate_results_cursor(
-                        session_cursor.stream(session),
-                        &mut result,
-                        &mut error,
-                    )
+            let iteration_result = if self.client.is_load_balanced() {
```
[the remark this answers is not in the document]
isabelatkinson: [discussed this in standup, leaving as-is because connection pinning is only required in load balanced mode, and checking out new connections for getMore operations may be more resilient to connection errors]

PR #1359, opened 2025-04-15
RUST-2195 add CODEOWNERS

kevinAlbs:
To meet requirement of DRIVERS-3098. Follows pattern from other drivers: [PHP](https://github.com/mongodb/mongo-php-library/blob/a4ccccd9011bad978e6f0d52efd1a512ef64d35a/.github/CODEOWNERS), [C#](https://github.com/mongodb/mongo-csharp-driver/blob/main/CODEOWNERS).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# Listing code owners is required by DRIVERS-3098
+* @mongodb/dbx-rust


abr-egn (approved):

kevinAlbs:
> LGTM. should we also add this for https://github.com/mongodb/bson-rust and https://github.com/mongodb/libmongocrypt-rust?

Good catch. Yes I expect so. Additional PRs incoming.

PR #1360, opened 2025-04-17
RUST-663 Support $merge and $out executing on secondaries

abr-egn:
RUST-663

My original plan to have a new method on `Operation` that could update the stored selection criteria based on operation-specific logic given a topology snapshot didn't work out because server selection might need to wait for handshake to finish and if so the snapshot won't have the version information.

I did consider making `Client::select_server` take an `Operation` rather than a string operation name, but it's called from various contexts that don't have an operation to give which would need to synthesize one, and that seemed worse.

In the end, what this means is:
* server selection needs to know if it's selecting a server for an Aggregate-with-$merge/$out operation, and
* server command construction needs to know what criteria server selection _actually_ ended up using, not just the one the operation reports

So this PR does the plumbing for that.

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -309,6 +309,8 @@ impl Client {
         let mut retry: Option<ExecutionRetry> = None;
         let mut implicit_session: Option<ClientSession> = None;
         loop {
+            op.update_for_topology(&self.inner.topology.description());
+
             if retry.is_some() {
                 op.update_for_retry();
             }
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -498,7 +498,7 @@ impl Client {
         self.inner.topology.update_command_with_read_pref(
             connection.address(),
             &mut cmd,
-            Some(&effective_criteria),
+            &effective_criteria,
         );
 
         match session {
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -29,6 +29,7 @@ use crate::{
             wire::{next_request_id, Message},
             PinnedConnectionHandle,
         },
+        Command,
         ConnectionPool,
         RawCommandResponse,
     },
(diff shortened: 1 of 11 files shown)


src/client/executor.rs
```diff
@@ -651,78 +661,9 @@ impl Client {
         let start_time = Instant::now();
         let command_result = match connection.send_message(message).await {
             Ok(response) => {
-                async fn handle_response<T: Operation>(
```
abr-egn, line 654: I pulled this out and renamed it for readability, but the body is unchanged.

src/sdam/description/topology.rs
```diff
@@ -205,7 +205,7 @@ impl TopologyDescription {
         &self,
         address: &ServerAddress,
         command: &mut Command,
-        criteria: Option<&SelectionCriteria>,
+        criteria: &SelectionCriteria,
```
abr-egn, line 208: Another side effect: this method no longer receives an `Option`, so it could be slightly simplified.

src/client/executor.rs
isabelatkinson: ha, I also did this in #1358 (minus the rename)

src/client/executor.rs
```diff
@@ -309,6 +310,8 @@ impl Client {
         let mut retry: Option<ExecutionRetry> = None;
         let mut implicit_session: Option<ClientSession> = None;
         loop {
+            //op.update_for_topology(&self.inner.topology.description());
```
isabelatkinson, line 313:
```suggestion

```
(guessing this is left over from another design)

isabelatkinson (approved):
lgtm!

abr-egn:
Rebase required a little bit of rewriting, so please do another look over.

isabelatkinson (approved):
LGTM. we might not be able to merge this til the cargo-deny error is fixed because of the rules I added for RUST-1907

PR #1365, opened 2025-04-29
minor: include CDLA-Permissive-2.0 in allowed licenses

abr-egn:


isabelatkinson:
is there any guidance for which licenses are permissible? It seems like we got a list from the legal team when this initial work was done (https://github.com/mongodb/mongo-rust-driver/pull/643)

abr-egn:
Update: we got a "this is a little odd but it's fine" from legal :)

isabelatkinson (approved):

PR #1366, opened 2025-04-29
RUST-2080 spec tests (and fix) for gridfs rename

abr-egn:
RUST-2080

This syncs the tests and adds the missing file-not-found logic that the tests caught.

This did require making the `rename` test operation a little odd because it can either target a collection, in which case it accepts arguments of the form `{ to: String }`, or it can target a gridfs bucket, where the arguments are `{ id: Bson, newFilename: String }`.  We parse the test file json into `TestOperation` values at file load time, where entity type isn't known, so we can't have distinct operation types for "rename collection" and "rename bucket", so instead there's now a wrapper layer that defers parsing the args document until the test is actually executed at which point it checks entity type and parses appropriately.

### DIFF
--- a/src/test/spec/unified_runner/operation/rename.rs
+++ b/src/test/spec/unified_runner/operation/rename.rs
@@ -1,0 +1,79 @@
+use bson::{doc, Bson, Document};
+use futures::FutureExt;
+use serde::Deserialize;
+
+use crate::{
+    error::Result,
+    gridfs::GridFsBucket,
+    test::spec::unified_runner::{Entity, TestRunner},
+    BoxFuture,
+    Collection,
+};
+
+use super::TestOperation;
+
+#[derive(Debug, Deserialize)]
+#[serde(transparent)]
+pub(super) struct Rename(Document);
+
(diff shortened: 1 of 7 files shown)


src/test/spec/unified_runner/operation/rename.rs
```diff
@@ -0,0 +1,79 @@
...
+    ) -> Result<Option<Entity>> {
+        let ns = target.namespace();
+        let mut to_ns = ns.clone();
+        to_ns.coll.clone_from(&self.to);
+        let cmd = doc! {
+            "renameCollection": crate::bson::to_bson(&ns)?,
+            "to": crate::bson::to_bson(&to_ns)?,
+        };
+        let admin = test_runner.internal_client.database("admin");
+        admin.run_command(cmd).await?;
```
isabelatkinson, line 62: I think we could technically drop support for this operation and skip tests that use it since it's not part of our API. That said, I can't think of a good reason to reduce test coverage here so probably fine to leave as-is

isabelatkinson (approved):

PR #1372, opened 2025-05-08
RUST-2166 Update convenient transactions API to use async closures

abr-egn:
RUST-2166

This adds the slightly-awkwardly-named-but-much-easier-to-use `and_run2` method, and updates almost all* of the callers to use the new method.  It also pulls out the method body into a macro so rather than being duplicated across the async and sync (and now new-async) implementations it's shared across all three.

\* A couple of test operations ran into a rustc bug that mean they don't compile with the new method due to really fun `implementation of 'std::marker::Send' is not general enough` errors, so I left those with the old method.  I think it pretty unlikely actual library users will run into this (the trigger condition here is the transaction closure capturing a value that has a manual `Send` impl when `and_run2` is being called inside multiple layers of `.box()`ed futures AFAICT).

### DIFF
--- a/.evergreen/check-clippy.sh
+++ b/.evergreen/check-clippy.sh
@@ -5,7 +5,7 @@ set -o errexit
 source ./.evergreen/env.sh
 
 # Pin clippy to the latest version. This should be updated when new versions of Rust are released.
-CLIPPY_VERSION=1.84.0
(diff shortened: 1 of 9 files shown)


isabelatkinson (approved):
the new async closures are so much nicer, lgtm!

PR #1375, opened 2025-05-19
RUST-2216 bump MSRV to 1.81.0

kevinAlbs:
# Summary

Bump MSRV to 1.81.0.

# Motivation

To fix failing [compile-on-msrv](https://spruce.mongodb.com/task/mongo_rust_driver_compile_compile_on_msrv_5902d420b159a3ecc2d8a05e08d00da10143ded3_25_05_09_16_14_40/logs?execution=0) task:
```
error: package `bson v2.14.0 (https://github.com/mongodb/bson-rust?branch=main#806d68f5)` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.74.0
```

Tested with this patch build: https://spruce.mongodb.com/version/682b3c30e30473000763fd53

https://github.com/mongodb/bson-rust/commit/806d68f5800f5baf9fab3c8e367ab6780e06873f added `rust-version = "1.81"` to Cargo.toml in bson-rust.

https://releases.rs/docs/1.81.0/ was released 5 September, 2024. I expect bumping to 1.81.0 is OK with the [MSRV policy](https://github.com/mongodb/mongo-rust-driver?tab=readme-ov-file#minimum-supported-rust-version-msrv-policy):

> will be to a Rust version at least six months old.


isabelatkinson:
I actually put up a fix for this last week (#1373) - the MSRV resolver doesn't work for the way we specify these dependencies, but we technically don't need to do a bump here

kevinAlbs:
Ah perfect. Thank you for the fix. Closing.

PR #1380, opened 2025-05-22
RUST-2217 Optionally support bson crate 3.0

abr-egn:
RUST-2217

This adds the `bson-2` and `bson-3` features, allowing users to select which version of the bson crate they'd like the driver crate to work with.  `bson-2` remains the default for compatibility reasons, we can't change that until driver 4.0.

There are actually quite a few user scenarios here and I think we managed to cover them nicely:
* _User does nothing_: Nothing changes and there are no surprises; `bson` 2.x continues to be used.
* _User had disabled default features_: They'll have been required to enable the `compat-3-0-0` feature (compile error otherwise), which at the time did nothing; now that feature enables `bson-2`, so this user will also continue to get `bson` 2.x.
* _User adds the `bson-3` feature flag_: The driver API will switch over to using `bson` 3.x, nothing else will change.  Under the hood, the `bson` 2.x crate will continue to be pulled in as a dependency; this is a consequence of Rust features needing to be additive.  If the user cares about the compile time/binary size, they can go with:
* _User disables default features, selects the `bson-3` feature flag_: They'll get _just_ `bson` 3.x, no 2.x pulled in.


Cargo.toml
```diff
@@ -29,7 +29,10 @@ exclude = [
 
 [features]
 default = ["compat-3-0-0", "rustls-tls", "dns-resolver"]
-compat-3-0-0 = []
+compat-3-0-0 = ["compat-3-3-0", "bson-2"]
+compat-3-3-0 = []
```
abr-egn, line 33: Keeping a rolling `compat-VERSION` feature allows us to do this again for other things if we need to, e.g. if we need to go back to supporting multiple async runtimes.

abr-egn:
cargo-deny and msrv failures are both unrelated; I've filed RUST-2222 and RUST-2223 for them.

Cargo.toml
kevinAlbs: Nice. That seems like a good precaution in case there is a future need to make a required feature optional.

kevinAlbs (approved):
LGTM! Testing scenarios locally match the PR description.

kevinAlbs (approved):

.evergreen/compile-only.sh
```diff
@@ -9,9 +9,6 @@ source ./.evergreen/env.sh
 if [ "$RUST_VERSION" != "" ]; then
   rustup toolchain install $RUST_VERSION
   TOOLCHAIN="+${RUST_VERSION}"
-  # Remove the local git dependencies for bson and mongocrypt, which don't work properly with the MSRV resolver.
-  sed -i "s/bson =.*/bson = \"2\"/" Cargo.toml
```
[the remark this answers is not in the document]
abr-egn: @isabelatkinson I'd like your input here - I can't find any brokenness but I'm deeply wary of things that work when I don't expect them to :)

Cargo.toml
```diff
@@ -120,6 +121,27 @@ zstd = { version = "0.11.2", optional = true }
...
+branch = "2.15.x"
+package = "bson"
+version = "2.15.0"
+optional = true
+
+[dependencies.bson3]
+git = "https://github.com/mongodb/bson-rust"
+branch = "main"
+package = "bson"
+version = "3.0.0"
```
isabelatkinson, line 135: Mostly out of curiosity: will this break if we try to release the driver before bson 3.0 is released?

.evergreen/compile-only.sh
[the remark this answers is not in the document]
isabelatkinson: This originally had to be fixed when the driver's MSRV (1.74) was less than bson's (1.81) - the MSRV resolver couldn't look back for an older MSRV-compatible version with a git dependency instead of a published version dependency. That problem goes away with the driver's MSRV being >= bson's MSRV (which is done in this PR) since an older version doesn't need to be selected, but it will come up again if we bump bson without bumping the driver. I think we can just figure that out if/when it comes up again, though - we should be able to do something similar once 3.0 is actually published.

Cargo.toml
abr-egn: `cargo publish` will fail since there's no non-git version, yeah.  If we do need to do a driver release before 3.0 we can either push a patch to the release branch that strips out the `bson-3` feature entirely or we could push an alpha release of the bson crate to cargo.

isabelatkinson (approved):

PR #1389, opened 2025-06-05
Bump dependencies to fix `cargo deny`

isabelatkinson:
The version of `tower-http` currently in the lockfile was yanked


isabelatkinson:
will be fixed by #1391

PR #1396, opened 2025-06-10
RUST-2104 Add From trait for SocketAddr to ServerAddress

JamieTsai1024:
Implemented and updated tests for to convert `std::net::SocketAddr` to `ServerAddress` (`Tcp` variant) using the `From` trait

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -237,7 +237,7 @@ dependencies = [
 [[package]]
 name = "bson"
 version = "3.0.0"
-source = "git+https://github.com/mongodb/bson-rust?branch=main#328d540df28d01afa71da55d485e4b83e61913c7"
+source = "git+https://github.com/mongodb/bson-rust?branch=main#8389d37175a6e9018ac82dacd62b20415e9c4469"
 dependencies = [
  "ahash",
  "base64 0.22.1",
--- a/src/cmap/conn/command.rs
+++ b/src/cmap/conn/command.rs
@@ -209,16 +209,6 @@ impl RawCommandResponse {
         })
     }
 
-    /// Used to handle decoding responses where the server may return invalid UTF-8 in error
-    /// messages.
-    pub(crate) fn body_utf8_lossy<'a, T: Deserialize<'a>>(&'a self) -> Result<T> {
-        crate::bson::from_slice_utf8_lossy(self.raw.as_bytes()).map_err(|e| {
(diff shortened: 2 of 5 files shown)


src/test/client.rs
```diff
@@ -982,3 +987,17 @@ async fn ipv6_connect() {
         .unwrap();
     assert_eq!(result.get_f64("ok").unwrap(), 1.0);
 }
+
+#[tokio::test]
+async fn server_address_from_socket_addr() {
```
isabelatkinson:
```suggestion
#[test]
fn server_address_from_socket_addr() {
```

This isn't doing any async work, so it can be a regular test

src/test/client.rs
```diff
@@ -1,7 +1,12 @@
...
-
 use crate::bson::Document;
 use serde::{Deserialize, Serialize};
+use std::{
+    borrow::Cow,
+    collections::HashMap,
+    future::IntoFuture,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    time::Duration,
+};
```
isabelatkinson, line 7: Can you move this block back to the top of the file? Our general convention for imports is:
```
// standard library imports
use std::{...};

// third-party imports
use futures::{...};
use serde::{...};

/// crate imports
use crate::{...};
```

the formatter doesn't enforce this so not every file does this perfectly, but it's a good convention to try to stick to

isabelatkinson (commented):
looks good so far! a few minor suggestions

src/test/client.rs
```diff
@@ -982,3 +987,17 @@ async fn ipv6_connect() {
...
+async fn server_address_from_socket_addr() {
+    let socket_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 27017);
+    let server_address = ServerAddress::from(socket_addr);
+
+    match server_address {
+        ServerAddress::Tcp { host, port } => {
+            assert_eq!(host, "127.0.0.1", "Host was not correctly converted");
+            assert_eq!(port, Some(27017), "Port was not correctly converted");
+        }
+        _ => panic!("ServerAddress should have been Tcp variant"),
```
isabelatkinson, line 1002: Can you add a second test case here for an IPv6 address?

isabelatkinson (approved):
lgtm! going to tag in Abraham for review as well

PR #1401, opened 2025-06-18
RUST-1992 Driver changes to track bson serde-optional API changes

abr-egn:
RUST-1992

An awful lot of search and replace here; the only real substance is in the `bson_compat` module.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -137,8 +137,8 @@ optional = true
 features = ["serde"]
 
 [dependencies.mongocrypt]
-git = "https://github.com/abr-egn/libmongocrypt-rust.git"
-branch = "RUST-1992/bson-optional-serde"
+git = "https://github.com/mongodb/libmongocrypt-rust.git"
+branch = "main"
 version = "0.3.1"
 default-features = false
 optional = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1525,7 +1525,7 @@ dependencies = [
 [[package]]
 name = "mongocrypt"
 version = "0.3.1"
-source = "git+https://github.com/abr-egn/libmongocrypt-rust.git?branch=RUST-1992%2Fbson-optional-serde#0cf9c34b092e4dba6daa1a1480fd45afd5e411cf"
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -134,11 +134,12 @@ branch = "main"
 package = "bson"
 version = "3.0.0"
 optional = true
+features = ["serde"]
 
 [dependencies.mongocrypt]
-git = "https://github.com/mongodb/libmongocrypt-rust.git"
-branch = "main"
-version = "0.3.0"
+git = "https://github.com/abr-egn/libmongocrypt-rust.git"
```
abr-egn, line 140: I'll revert this once https://github.com/mongodb/libmongocrypt-rust/pull/45 is merged.

src/bson_compat.rs
```diff
@@ -1,31 +1,131 @@
-#[cfg(feature = "bson-3")]
-pub(crate) trait RawDocumentBufExt {
-    fn append_ref<'a>(
+use crate::bson::RawBson;
+
+pub(crate) trait RawDocumentBufExt: Sized {
+    fn append_err(&mut self, key: impl AsRef<str>, value: impl Into<RawBson>) -> RawResult<()>;
```
abr-egn, line 4: Since `append` and similar are now fallible in 3.0, the compatibility layer needed to make them fallible (at least at the type level) for 2.x as well, even if they never actually fail.  The other option would be to wrap the 3.x variants in `expect` but this way people opting in to `bson-3` get nicer error behavior as a bonus :)

src/bson_compat.rs
```diff
@@ -1,31 +1,131 @@
...
+#[cfg(not(feature = "bson-3"))]
+impl DocumentExt for crate::bson::Document {
+    fn encode_to_vec(&self) -> crate::bson::ser::Result<Vec<u8>> {
+        let mut out = vec![];
+        self.to_writer(&mut out)?;
+        Ok(out)
+    }
+}
+
+macro_rules! use_either {
```
abr-egn, line 109: I know this is kind of a frivolous use of a macro but I got _real_ tired of writing these out by hand and I find the tabular format a lot easier to read.  I can revert if you feel strongly :)

Cargo.toml
abr-egn: Note: this is also the cause of the `cargo deny` failure.

PR #1412, opened 2025-07-08
RUST-1992 Update the driver for bson cstr API changes

abr-egn:
RUST-1992

This involved unwinding all the fallibility I'd introduced in #1401, and introducing a compatibility layer for `CStr` that redirects to plain `str` for bson 2.x.

### DIFF
--- a/src/client/executor.rs
+++ b/src/client/executor.rs
@@ -879,7 +879,10 @@ impl Client {
         }
     }
 
-    async fn select_data_bearing_server(&self, operation_name: &str) -> Result<()> {
+    async fn select_data_bearing_server(
+        &self,
+        operation_name: &crate::bson_compat::CStr,
+    ) -> Result<()> {
         let topology_type = self.inner.topology.topology_type();
         let criteria = SelectionCriteria::Predicate(Arc::new(move |server_info| {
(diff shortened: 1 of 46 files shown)


src/client/executor.rs
```diff
@@ -320,10 +320,11 @@ impl Client {
                 .and_then(|s| s.transaction.pinned_mongos())
                 .or_else(|| op.selection_criteria());
 
+            let op_name: &str = op.name().into();
```
abr-egn: `op.name().as_str()` would be the easier way to spell this but unfortunately that doesn't compile for bson 2.x when `op.name()` is a `str` itself because `str::as_str` isn't stabilized yet 😞 (https://github.com/rust-lang/rust/issues/130366)

src/operation.rs
```diff
@@ -108,7 +109,7 @@ pub(crate) trait Operation {
     type O;
 
     /// The name of the server side command associated with this operation.
-    const NAME: &'static str;
+    const NAME: &'static CStr;
```
abr-egn, line 112: In the large majority of cases, `NAME` is primarily used as a key in `rawdoc!` construction.

isabelatkinson (approved):

PR #1413, opened 2025-07-08
RUST-2235 Implement GSSAPI auth support for Linux and macOS

mattChiaravalloti:
This PR implements the GSSAPI auth mechanism, as described by the [spec](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#gssapi). This took a bit longer than anticipated because manual end-to-end testing revealed that the initial implementation was not correct. Subsequent fixes made slow progress towards correctness. Ultimately, the implementation you see here has been manually tested on macOS (locally, against a local KDC and mongod) and an ubuntu2204-arm64-small evergreen host (against the LDAPTEST KDC/mongod). I still need to test this on Windows. I have a separate follow-up ticket to implement automated end-to-end testing, RUST-2236, which I will put up a PR for after confirming this works on Windows and actually writing end-to-end tests.

This implementation is guarded by a feature flag, and passes the existing spec tests when the feature flag is enabled.


src/client/auth.rs
```diff
@@ -67,8 +69,7 @@ pub enum AuthMechanism {
     /// Kerberos authentication mechanism as defined in [RFC 4752](http://tools.ietf.org/html/rfc4752).
     ///
     /// See the [MongoDB documentation](https://www.mongodb.com/docs/manual/core/kerberos/) for more information.
-    ///
-    /// Note: This mechanism is not currently supported by this driver but will be in the future.
+    #[cfg(feature = "gssapi-auth")]
```
mattChiaravalloti, line 74: I followed the same pattern as the `aws-auth` feature flag.

src/client/auth/gssapi.rs
```diff
@@ -0,0 +1,348 @@
...
+    let response_doc = conn.send_message(command).await?;
+    let sasl_response =
+        SaslResponse::parse(GSSAPI_STR, response_doc.auth_response_body(GSSAPI_STR)?)?;
+
+    let mut conversation_id = Some(sasl_response.conversation_id);
+    let mut payload = sasl_response.payload;
+
+    // Limit number of auth challenge steps (typically, only one step is needed, however
+    // different configurations may require more).
+    for _ in 0..10 {
```
mattChiaravalloti, line 80: This idea is taken directly from the python driver's implementation ([link](https://github.com/mongodb/mongo-python-driver/blob/c77c15e369ddb9d9c63abb0adb9ac7d3f8ab11ef/pymongo/asynchronous/auth.py#L278)).

It seems there may be some circumstances where multiple challenges need to be stepped through, so we need to do this multiple times. However, based on the python implementation's comments, it seems there may be protocol and/or library issues that could result in endless looping so a limit is put in place (10 tries).

src/client/auth/gssapi.rs
```diff
@@ -0,0 +1,348 @@
...
+            GSSAPI_STR,
+            "GSSAPI authentication failed after 10 attempts",
+        ))
+    }
+}
+
+impl GssapiProperties {
+    pub fn from_credential(credential: &Credential) -> Result<Self> {
+        let mut properties = GssapiProperties {
+            service_name: "mongodb".to_string(),
```
mattChiaravalloti, line 144: You may notice that in `options.rs`, we also ensure that `mongodb` is the default service name. That is required in `options.rs` for spec compliance (as in, the spec tests cannot pass _without_ doing that). At this point, although we know `mechanism_properites.get(SERVICE_NAME)` must _always_ return `Some(...)`, it felt cleanest to me to implement this code the way you see here (including `"mongodb"` as a default value).

src/client/auth/gssapi.rs
```diff
@@ -0,0 +1,347 @@
...
+        if let Some(service_realm) = properties.service_realm.as_ref() {
+            service_principal = format!("{}@{}", service_principal, service_realm);
+        } else if let Some(user_principal) = user_principal.as_ref() {
+            if let Some(idx) = user_principal.find('@') {
+                // If no SERVICE_REALM was specified, use realm specified in the
+                // username. Note that `realm` starts with '@'.
+                let (_, realm) = user_principal.split_at(idx);
+                service_principal = format!("{}{}", service_principal, realm);
+            }
+        }
```
mattChiaravalloti: This is not mentioned in the spec but it did end up being required since there is no other way to specify a realm for the service principal if not done manually. Perhaps other language libraries default to this behavior, but not `cross-krb5`/`libgssapi` in Rust. I spoke to James about this and he said he had to do something similar in the C# driver's GSSAPI implementation.

src/client/auth/gssapi.rs
```diff
@@ -0,0 +1,341 @@
...
+            ))
+        }
+    }
+
+    fn is_complete(&self) -> bool {
+        self.is_complete
+    }
+}
+
+fn canonicalize_hostname(hostname: &str, mode: &CanonicalizeHostName) -> Result<String> {
```
mattChiaravalloti: I wrote some unit tests for this but ultimately removed them since they are trivial. The more interesting testing of this will be indirect via the end-to-end GSSAPI auth tests in the next ticket.

mattChiaravalloti:
@abr-egn I am not able to add other reviewers, so if you know who else could/should be tagged on this, can you add them? Thank you!

mattChiaravalloti:
Ok I believe I've addressed the majority of failures now. The two remaining issues are:
1. Compiling with the MSRV (1.81.0) fails because `libgssapi-sys` (a deeply nested indirect dependency) uses `unsafe extern "C"` blocks, which are disallowed in that version. I'm still trying to figure out how best to address this. Any ideas from reviewers are appreciated.
2. It seems the Graviton tests are failing because `libgssapi-sys` cannot compile without setting an ENV variable. The relevant error seems to be: `Unable to find libclang: "couldn't find any valid shared libraries matching: ['libclang.so', 'libclang-*.so', 'libclang.so.*', 'libclang-*.so.*'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [])"`. So we should be able to set `LIBCLANG_PATH` for that distro and be good to go. I know nothing about this distro though. Perhaps @abr-egn can help out with this one when he's back!

Beyond addressing those two issues, the last thing I need to do is double check the spec for how to handle password being provided on Windows. It looks like there is supposed to be some special handling that happens only on Windows if the password is provided as a uri option. I'll look into that and make any relevant changes.

Cargo.toml
```diff
@@ -58,6 +58,9 @@ gcp-oidc = ["dep:reqwest"]
 # This can only be used with the tokio-runtime feature flag.
```
pmeredit: @abr-egn

Unrelated to this PR, really (though we could fix it here), should we update this comment since async-std support was dropped?

src/client/auth.rs
```diff
@@ -148,6 +149,25 @@ impl AuthMechanism {
...
+            AuthMechanism::Gssapi => {
+                if credential.username.is_none() {
+                    return Err(ErrorKind::InvalidArgument {
+                        message: "No username provided for GSSAPI authentication".to_string(),
+                    }
+                    .into());
+                }
+
+                if credential.source.as_deref().unwrap_or("$external") != "$external" {
+                    return Err(ErrorKind::InvalidArgument {
```
pmeredit, line 164: Hmm, just noticed that the "must be $external" message for OIDC is slightly different than everything else, maybe we should update it 🤔

src/client/auth/gssapi.rs
```diff
@@ -0,0 +1,343 @@
...
+impl GssapiAuthenticator {
+    // Initialize the GssapiAuthenticator by creating a PendingClientCtx and
+    // getting an initial token to send to the server.
+    async fn init(
+        user_principal: Option<String>,
+        properties: GssapiProperties,
+        hostname: &str,
+    ) -> Result<(Self, Vec<u8>)> {
+        let service_name: &str = properties.service_name.as_ref();
+        let mut service_principal = format!("{}/{}", service_name, hostname);
```
pmeredit:
```suggestion
        let mut service_principal = format!("{service_name}/{hostname}");
```
Let's just fix this before clippy starts whining about it the next time they update it :)

src/client/auth/gssapi.rs
```diff
@@ -0,0 +1,343 @@
...
+        if let Some(service_realm) = properties.service_realm.as_ref() {
+            service_principal = format!("{}@{}", service_principal, service_realm);
```
pmeredit:
```suggestion
            service_principal = format!("{service_principal}@{service_realm}");
```
ditto

src/client/auth/gssapi.rs
```diff
@@ -0,0 +1,343 @@
...
+        } else if let Some(user_principal) = user_principal.as_ref() {
+            if let Some(idx) = user_principal.find('@') {
+                // If no SERVICE_REALM was specified, use realm specified in the
+                // username. Note that `realm` starts with '@'.
+                let (_, realm) = user_principal.split_at(idx);
+                service_principal = format!("{}{}", service_principal, realm);
```
pmeredit:
```suggestion
                service_principal = format!("{service_principal}{realm}");
```

src/client/auth/gssapi.rs
pmeredit: Yea, I'm personally trying to move away from unit tests that just explicitly encode basic code.

pmeredit (approved):
LGTM once everything is working on every OS/tests are passing. Good design. Just fix the format!s because clippy will start complaining whenever they update.

Cargo.toml
abr-egn: Hah, yes, good catch.

Cargo.toml
```diff
@@ -80,8 +83,10 @@ chrono = { version = "0.4.7", default-features = false, features = [
     "clock",
     "std",
 ] }
+cross-krb5 = { version = "0.4.2", optional = true, default-features = false }
 derive_more = "0.99.17"
 derive-where = "1.2.7"
+dns-lookup = { version = "2.0", optional = true }
```
abr-egn: Why not use `hickory-resolver` which we already have as an optional dep?

Cargo.toml
mattChiaravalloti: I'm not sure what the update should be here. Do I just delete this comment on line 58? Or change it to say something else (if so, what should it say)?

src/client/auth.rs
mattChiaravalloti: Updated 👍

src/client/auth/gssapi.rs
mattChiaravalloti: Updated all 👍

Cargo.toml
mattChiaravalloti: I didn't even realize we had that imported at first to be honest. Just updated to use it. The interface is a bit clunkier than `dns-lookup` but at the end of the day I think it's preferable to use the same crate. Thanks for pointing this out.

mattChiaravalloti:
> The MSRV issue can be solved by just bumping up MSRV to 1.82 which introduced the `unsafe extern` block syntax; that's well past our six month rolling window.

Thanks! I believe I updated 1.81 -> 1.82 in all appropriate places. Time to see if it works as expected 🤞 

> I agree with Patrick that it makes sense to turn off the GSSAPI feature flag for the Graviton variants; I don't think there's enough value in the marginal test coverage increase to justify the engineer time to figure out if it can be made to work.

Amazing, love when this is a solution. Given this, and the upcoming end-to-end GSSAPI auth testing coming up, I refactored `gssapi-auth` to be its own variant with its own tasks/functions, similar to all other auth mechanisms.

Cargo.toml
abr-egn: Sorry, yup, just delete the comment.  The driver only supports tokio now so the there's nothing that needs to be documented about this feature.

src/client/auth/gssapi.rs
```diff
@@ -0,0 +1,379 @@
...
+        self.is_complete
+    }
+}
+
+async fn canonicalize_hostname(hostname: &str, mode: &CanonicalizeHostName) -> Result<String> {
+    if mode == &CanonicalizeHostName::None {
+        return Ok(hostname.to_string());
+    }
+
+    let resolver = TokioAsyncResolver::tokio_from_system_conf().map_err(|e| {
```
abr-egn: I think this should use `ClientOptions::resolver_config`, so that'll likely need to be added to `HandshakerOptions` and plumbed through in `EstablisherOptions::from_client_options` and `Credential::authenticate_stream`.

If you're feeling ambitious it's feeling like it's about time for an `AuthenticateOptions` struct that bundles together the non-`Connection` arguments to `authenticate_stream` and has a `From<ClientOptions>` impl so it can just get passed through `Handshaker` without having to update all the call sites each time new data like this is needed, but absolutely reasonable if you'd rather not include that in this PR 🙂

mattChiaravalloti: So far, I only piped through `resolver_config` so that it makes it all the way down to `canonicalize_hostname`. Let me know if you think it all looks good since it involved touching a big number of files. I decided to go with `[cfg(any(feature = "dns-lookup", feature = "gssapi-auth"))]` in the places where it's possible that only `gssapi-auth` is enabled without explicitly enabling `dns-lookup` (and vice versa). Curious to see what you think about that decision in particular.

Also, I was considering how to best introduce the `AuthenticateOptions` idea you proposed, but I'm a bit confused what you're envisioning. You suggested having it implement `From<ClientOptions>`, but where would those `ClientOptions` come from? By the time we call `authenticate_stream` we're already 2 steps removed from `ClientOptions`: it goes `ClientOptions` --into--> `HandshakerOptions` --into--> `Handshaker`, and then `Handshaker` calls `authenticate_stream`. Are you suggesting that `Handshaker` be updated to contain an `AuthenticateOptions` struct field _instead of_ all the non-connection individual fields it currently has? Would that same thing apply to `HandshakerOptions`? Maybe it would be best to leave this to someone on your team in a follow-up ticket since you guys have better expertise on the design practices you use in this repo!

abr-egn: > So far, I only piped through `resolver_config` so that it makes it all the way down to `canonicalize_hostname`. Let me know if you think it all looks good since it involved touching a big number of files. I decided to go with `[cfg(any(feature = "dns-lookup", feature = "gssapi-auth"))]` in the places where it's possible that only `gssapi-auth` is enabled without explicitly enabling `dns-lookup` (and vice versa). Curious to see what you think about that decision in particular.

I think the right thing here is for the 'gssapi-auth' feature to require 'dns-lookup', not 'dep:hickory-resolver' (after all, it doesn't really care how the dns resolution happens).  That would mean all the existing places that are conditioned on `dns-lookup` can continue to be conditioned on just that and don't need to change, and anything added for GSSAPI support specifically can be conditioned on just `#[cfg(feature = "gssapi-auth")]`.

> Also, I was considering how to best introduce the `AuthenticateOptions` idea you proposed, but I'm a bit confused what you're envisioning. You suggested having it implement `From<ClientOptions>`, but where would those `ClientOptions` come from? By the time we call `authenticate_stream` we're already 2 steps removed from `ClientOptions`: it goes `ClientOptions` --into--> `HandshakerOptions` --into--> `Handshaker`, and then `Handshaker` calls `authenticate_stream`. Are you suggesting that `Handshaker` be updated to contain an `AuthenticateOptions` struct field _instead of_ all the non-connection individual fields it currently has? Would that same thing apply to `HandshakerOptions`? Maybe it would be best to leave this to someone on your team in a follow-up ticket since you guys have better expertise on the design practices you use in this repo!

Agreed, definitely better for one of us in a follow-up ticket, there's a lot of context here 🙂

mattChiaravalloti: Smart, yes that's a better solution. I'll work on that change and get it back in shortly. Thanks!

mattChiaravalloti: @abr-egn It is all updated appropriately now 👍 Do you want me to file the follow-up ticket for `AuthenticationOptions`, or do you want to file it?

mattChiaravalloti:
MacOS hosts are pretty busy now so those might take a while. They've been sporadically failing the last few commits. Hopefully those are just transient errors that don't need to be addressed as part of this PR.

What's more concerning was `oidc-auth-test-gcp-latest`'s first run failing (I restarted it once for now). I double checked and nothing in this PR _should_ have impacted that test. Not sure if that kind of failure is transient though. Hoping the second run sheds more light on this situation.

abr-egn:
> MacOS hosts are pretty busy now so those might take a while. They've been sporadically failing the last few commits. Hopefully those are just transient errors that don't need to be addressed as part of this PR.
> 
> What's more concerning was `oidc-auth-test-gcp-latest`'s first run failing (I restarted it once for now). I double checked and nothing in this PR _should_ have impacted that test. Not sure if that kind of failure is transient though. Hoping the second run sheds more light on this situation.

There's an uncommon but persistent flaky failure in that test ([example](https://spruce.mongodb.com/task/mongo_rust_driver_oidc_linux_oidc_auth_test_gcp_latest_bc4ae960c7c9993e8103b6bc7996c0815b933f71_25_06_23_17_13_37/logs?execution=0)) that we haven't been able to nail down, I suspect that's what you ran into.

src/client/auth/gssapi.rs
abr-egn: Filed RUST-2247; it'll make a nice capper for the end of spec fest 😁

PR #1416, opened 2025-07-14
RUST-2240 Enable bson `serde_json-1` feature flag where necessary

isabelatkinson:
Enables `serde_json-1` for testing and when `tracing-unstable` is enabled.


Cargo.toml
```diff
@@ -207,6 +207,13 @@ reqwest = { version = "0.12.2", features = ["rustls-tls"] }
 serde-hex = "0.1.0"
 serde_path_to_error = "0.1"
 
+[dev-dependencies.bson3]
```
isabelatkinson, line 210: There's no equivalent `bson3?/serde_json-1`-esque syntax for a dev dependency as far as I can tell, so this has to be duplicated.

PR #1426, opened 2025-07-17
RUST-2161 Support auto encryption in unified tests

abr-egn:
RUST-2161


src/test/csfle.rs
```diff
@@ -310,3 +329,39 @@ async fn fle2v2_ok(name: &str) -> bool {
     }
     true
 }
+
+pub(crate) fn fill_kms_placeholders(
```
abr-egn, line 333: This makes more sense here than in test_runner.rs; the only logic change is the conditional for temp creds for AWS.

PR #1427, opened 2025-07-17
RUST-2074 Test that pool cleared error has "TransientTransactionError" label

isabelatkinson:
This uses the behavior introduced in RUST-1222 to trigger a pool cleared error.

### DIFF
--- a/src/cmap/test.rs
+++ b/src/cmap/test.rs
@@ -9,18 +9,20 @@ use tokio::sync::{Mutex, RwLock};
 use self::file::{Operation, TestFile, ThreadedOperation};
 
 use crate::{
+    bson::{doc, Document},
     cmap::{
         establish::{ConnectionEstablisher, EstablisherOptions},
         ConnectionPool,
         ConnectionPoolOptions,
     },
-    error::{Error, ErrorKind, Result},
+    error::{Error, ErrorKind, Result, TRANSIENT_TRANSACTION_ERROR},
     event::cmap::{
         CmapEvent,
         ConnectionCheckoutFailedReason,
         ConnectionClosedReason,
         ConnectionPoolOptions as EventOptions,
     },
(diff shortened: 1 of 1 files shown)


src/client/executor.rs
```diff
@@ -349,12 +349,19 @@ impl Client {
...
+                        Retryability::Read if err.is_pool_cleared() => true,
+                        _ => {
+                            retryability.can_retry_error(&err)
+                                && !session
+                                    .as_ref()
+                                    .is_some_and(|session| session.in_transaction())
+                        }
                     };
-                    if err.is_pool_cleared() || op_retry {
+                    if can_retry {
```
isabelatkinson, line 364: Incidentally found a bug here: we were unconditionally retrying pool cleared errors, but that doesn't align with the retryability specs. The only special behavior specified is that we can retry pool cleared errors in transactions for reads when retryReads is not false (see the [pseudocode](https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.md#pseudocode)).

The rest of the changes in this file are just cleanup.

src/test/util/fail_point.rs
```diff
@@ -162,3 +172,27 @@ impl Drop for FailPointGuard {
         }
     }
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn app_name_fail_point_is_disabled() {
```
isabelatkinson: I wrote this to make sure that the above changes actually disable the failpoint, will delete before merging because I don't think there's much value in continuing to run it.

isabelatkinson (commented):
I ran out of time trying to make this test not flaky, but I still think it's worth it to merge in the retryability fix. I'll put RUST-2074 back into the backlog and investigate next spec fest.

abr-egn (dismissed):

PR #1428, opened 2025-07-18
RUST-2247 Bundle extra arguments to auth mechanisms

abr-egn:
RUST-2247

This tidies things up a bit and means that any future data needed for auth mechanisms will be much simpler - they can be added to `AuthOptions` + the `From<ClientOptions>` impl, and `Handshaker`, `ConnectionEstablisher`, etc. won't need to be touched.


isabelatkinson (approved):
nice!

PR #1435, opened 2025-07-24
RUST-1529 Use AWS SDK to get credentials 

JamieTsai1024:
Replace existing implementation for getting AWS credentials by using the AWS SDK. 

Currently, we have [5 places where we search for AWS credentials](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#credential-fetching-order):
1. The MongoDB URI
2. Environment variables
3. ~~A custom AWS credential provider if the driver supports it~~
4. STS AssumeRole or AssumeRoleWithWebIdentity
5. The ECS or EC2 endpoint

### Testing 
For the sake of testing, the existing implementation was replaced instead of being hidden behind a feature flag. 
- Methods 1-4 were tested locally. Clone https://github.com/isabelatkinson/drivers-evergreen-tools/tree/local-aws and run the following commands:
```
./.evergreen/run-aws-auth.sh

cd .evergreen/auth_aws
# run this script to reconfigure the type of AWS authentication to use
# copy the block of unset and export commands output by the script and run them in the terminal window you're using to run the driver tests
./aws_setup.sh (regular | env-creds | assume-role | session-creds | web-identity)

cd ~/mongo-rust-driver 
# Run the unset and set export commands printed by `./aws_setup.sh` for environment variables 
cargo nextest run auth_aws --features aws-auth
```

### Work to be done for [RUST-1529](https://jira.mongodb.org/browse/RUST-1529)
- Generate signature using AWS SDK (`compute_authorization_header(...)`)
- Determine feature flags required
  - May restore deleted implementation from this PR, depending on whether we want to keep the AWS SDK as an optional dependency or replace the original non-SDK implementation entirely


Cargo.toml
```diff
@@ -121,6 +121,18 @@ webpki-roots = "0.26"
 zstd = { version = "0.11.2", optional = true }
 macro_magic = "0.5.1"
 rustversion = "1.0.20"
+aws-credential-types = "1.2.4"
+aws-types = "1.3.7"
```
isabelatkinson: These dependencies are only used for AWS authentication, so they can be put behind the `aws-auth` feature flag as well

src/client/auth/aws.rs
```diff
@@ -1,11 +1,23 @@
-use std::{fs::File, io::Read, time::Duration};
+#[cfg(feature = "aws-auth")]
+use aws_config::BehaviorVersion;
+
+#[cfg(feature = "aws-auth")]
+use aws_credential_types::provider::ProvideCredentials;
+
+#[allow(unused_imports)]
+#[cfg(feature = "aws-auth")]
+use aws_types::sdk_config::SharedCredentialsProvider;
```
isabelatkinson: does this need to be imported?

src/client/auth/aws.rs
```diff
@@ -90,24 +103,60 @@ async fn authenticate_stream_inner(
...
+                .and_then(|mp| mp.get_str("AWS_SESSION_TOKEN").ok())
+                .map(str::to_owned),
+            None,
+        )
+    } else {
+        // If credentials are not provided in the URI, use the AWS SDK to load
+        let creds = aws_config::load_defaults(BehaviorVersion::latest())
+            .await
+            .credentials_provider()
+            .expect("no credential provider configured")
```
isabelatkinson: `expect` is similar to `unwrap` in that it will panic if the optional value it's called on is `None`. I suggest using a method like [`ok_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else) to return an error in the `None` case instead

src/client/auth/aws.rs
```diff
@@ -90,24 +103,60 @@ async fn authenticate_stream_inner(
     let server_first = ServerFirst::parse(server_first_response.auth_response_body(MECH_NAME)?)?;
     server_first.validate(&nonce)?;
 
-    let aws_credential = {
-        // Limit scope of this variable to avoid holding onto the lock for the duration of
-        // authenticate_stream.
-        let cached_credential = CACHED_CREDENTIAL.lock().await;
```
isabelatkinson, line 94: Does the SDK handle credential caching? We should make sure the new implementation still caches credentials as outlined in the [spec](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#caching-credentials)

src/client/auth/aws.rs
```diff
@@ -246,6 +297,20 @@ impl AwsCredential {
         }
     }
 
+    fn from_sdk_creds(
```
isabelatkinson: Would it be feasible to replace `AwsCredential` in the code with `aws_credential_types::Credentials` (i.e. the type returned from `provide_credentials`)? That could remove some of the type translation we're doing here; it looks like they have very similar fields.

isabelatkinson (commented):
I forgot to mention this regarding the testing commit I provided, but we should leave line 226 (`patchable: false`) in [config.yml](https://github.com/mongodb/mongo-rust-driver/blob/b4308cf34843dae9b4098f6f6bf9b063e273718c/.evergreen/config.yml#L226) commented out while this PR is being worked on. `patchable: false` disables tests from running on every commit to a pull request branch - normally, we don't need to run the AWS authentication tests that frequently, but since we're making AWS-related changes we should be running the tests on this PR.

src/client/auth/aws.rs
JamieTsai1024: I think it makes sense to keep `AwsCredential` for now since we’re still using `compute_authorization_header(...)`, which relies on it. Once that method is removed in the next PR, I can look into fully switching over to `aws_credential_types::Credentials`.

To make the transition easier later, I’ll update this PR to type `creds` as `aws_credential_types::Credentials` and then map it into `AwsCredential` before using `compute_authorization_header(...)`. That should make it easier to remove `from_sdk_creds(...)` and `AwsCredential` in the follow-up.

Also, I missed that another file uses `AwsCredential::get`, so I’ve moved the credential-fetching logic into a public method to support that use case!

src/client/auth/aws.rs
JamieTsai1024: Following up on this — as we discussed, the SDK does implement credential caching. I’ve linked the relevant documentation [here](https://docs.aws.amazon.com/sdk-for-rust/latest/dg/credproviders.html#credproviders-identity-caching) in case it's helpful for future reference!

src/action/csfle/create_data_key.rs
```diff
@@ -13,6 +13,7 @@ impl ClientEncryption {
     #[deeplink]
     #[options_doc(create_data_keys)]
     pub fn create_data_key(&self, master_key: impl Into<MasterKey>) -> CreateDataKey {
+        dbg!("create_data_key inside src/action/csfle/create_data_key.rs");
```
isabelatkinson:
```suggestion

```

src/test/csfle/on_demand_aws.rs
isabelatkinson, line 1: suggest removing the debug statements in this file

src/test/auth/aws.rs
```diff
@@ -1,125 +1,130 @@
...
+    // client::auth::aws::test_utils::*,
+    // test::DEFAULT_URI,
     Client,
 };
 
 #[tokio::test]
 async fn auth_aws() {
     let client = Client::for_test().await;
     let coll = client.database("aws").collection::<Document>("somecoll");
+    let _ = tracing_subscriber::fmt::try_init();
```
isabelatkinson:
```suggestion

```

src/test/auth/aws.rs
```diff
@@ -1,125 +1,130 @@
...
+//     }
+//     assert!(cached_credential().await.is_none());
+
+//     coll.find_one(doc! {}).await.unwrap();
+//     assert!(cached_credential().await.is_some());
+// }
+
+// todo: tests for aws sdk caching
+// TODO: find documentation for rust aws sdk credential caching
+// can delete caching logic if the aws sdk handles it properly for our needs
```
isabelatkinson:
```suggestion

```

src/test/auth/aws.rs
```diff
@@ -1,125 +1,130 @@
...
-    remove_var("AWS_ACCESS_KEY_ID");
-    remove_var("AWS_SECRET_ACCESS_KEY");
-    remove_var("AWS_SESSION_TOKEN");
-}
+// async fn get_client() -> Client {
+//     Client::with_uri_str(DEFAULT_URI.clone()).await.unwrap()
+// }
+
+// #[tokio::test]
+// async fn credential_caching() {
```
isabelatkinson: Rather than commenting out these tests, we can add the [`#[ignore]`](https://doc.rust-lang.org/reference/attributes/testing.html#the-ignore-attribute) annotation to skip running them for now with a comment indicating that they'll be addressed in a future PR for RUST-1529

src/test/csfle/on_demand_aws.rs
isabelatkinson, line 1: suggest removing the debug statements in this file

Cargo.toml
```diff
@@ -122,6 +122,27 @@ zstd = { version = "0.11.2", optional = true }
...
+[dependencies.aws-config]
+version = "1"
+optional = true
+default-features = false
+features = ["default-https-client", "rt-tokio"]
+
+[dependencies.aws-sdk-sts]
+version = "=1.73"
+optional = true
+default-features = false
```
isabelatkinson: I thought of a slightly nicer workaround for this issue: rather than include this dependency directly in our `Cargo.toml` file, we can add the dependency before generating the lockfile for the MSRV check. This would entail adding the following before [this line](https://github.com/mongodb/mongo-rust-driver/blob/be56c790e24b1031f3f4547e86dc188425fb96d4/.evergreen/compile-only.sh#L12) in `compile-only.sh`:

```
# The MSRV resolver does not properly select an MSRV-compliant version
# for this transient dependency.
cargo add aws-sdk-sts@1.73
```

This will add the pinned dependency for `aws-sdk-sts` _only_ for the MSRV check, so users will be unaffected.

src/action/csfle/create_data_key.rs
isabelatkinson:
```suggestion

```

src/client/auth/aws.rs
```diff
@@ -171,168 +226,183 @@ pub(crate) struct AwsCredential {
...
-        if let (Some(access_key), Some(secret_key)) = (access_key, secret_key) {
-            return Ok(Self {
-                access_key_id: access_key,
-                secret_access_key: secret_key,
-                session_token,
-                expiration: None,
-            });
+    // /// Derives the credentials for an authentication attempt given the set of credentials the
+    // user /// passed in.
+    // pub(crate) async fn get(credential: &Credential, http_client: &HttpClient) -> Result<Self> {
```
isabelatkinson: Rather than comment out all of these methods, I suggest just adding the `#[allow(dead_code)]` annotation with a note that they'll be addressed when we decide on the feature flag situation - that should make this diff a bit more readable

src/client/auth/aws.rs
```diff
@@ -153,7 +174,41 @@ async fn authenticate_stream_inner(
     Ok(())
 }
 
+// Find credentials using MongoDB URI or AWS SDK
+pub async fn get_aws_credentials(credential: &Credential) -> Result<Credentials> {
```
isabelatkinson:
```suggestion
pub(crate) async fn get_aws_credentials(credential: &Credential) -> Result<Credentials> {
```

We generally only mark items `pub` when we intend for them to be part of the public API. `pub(crate)` will allow for internal usage in other parts of the driver.

isabelatkinson (commented):
sorry if some of my comments are duplicated! the github UI was being weird

Cargo.toml
JamieTsai1024: Thanks Isabel! This seems nicer :)

isabelatkinson (approved):
LGTM! tagging Abraham for review, and then once he approves you can remove the `patchable: false` change in `.evergreen/config.yml`

Cargo.toml
```diff
@@ -41,7 +41,7 @@ dns-resolver = ["dep:hickory-resolver", "dep:hickory-proto"]
 cert-key-password = ["dep:pem", "dep:pkcs8"]
 
 # Enable support for MONGODB-AWS authentication.
-aws-auth = ["dep:reqwest"]
+aws-auth = ["dep:reqwest", "aws-config", "aws-types", "aws-credential-types"]
```
abr-egn: These should use the `"dep:"` syntax - that avoids implicitly creating features for these optional dependencies since we don't want them to be able to be individually enabled (https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies).

src/client/auth/aws.rs
```diff
@@ -246,6 +300,21 @@ impl AwsCredential {
         }
     }
 
+    // Creates AwsCredential from keys.
+    fn from_sdk_creds(
```
abr-egn: Maybe have this take a `Credentials` instead?

JamieTsai1024: Will do! Just as a heads-up, the code for `from_sdk_creds` will be removed in my next PR when I switch to the AWS SDK for SigV4 signing, so I didn’t prioritize cleaning it up too much. That said, your suggestion is definitely cleaner! :)

isabelatkinson (approved):

PR #1437, opened 2025-07-30
RUST-2228 Sync the driver to the BSON conversion API changes

abr-egn:
RUST-2228

Fortunately, most of the conversion traits we use are present in 2.x, so not much needed in the way of compat machinery - in fact I think the removal of the `encode_`/`decode_` prefixes means this is a net reduction there.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -138,8 +138,8 @@ optional = true
 features = ["serde"]
 
 [dependencies.mongocrypt]
-git = "https://github.com/abr-egn/libmongocrypt-rust.git"
-branch = "RUST-2228/bson-conversion-cleanup"
+git = "https://github.com/mongodb/libmongocrypt-rust.git"
+branch = "main"
 version = "0.3.1"
 default-features = false
 optional = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1642,7 +1642,7 @@ dependencies = [
 [[package]]
 name = "mongocrypt"
 version = "0.3.1"
-source = "git+https://github.com/abr-egn/libmongocrypt-rust.git?branch=RUST-2228%2Fbson-conversion-cleanup#6149ed3e24e2041942a3a98148f7d52fe7d4764a"
+source = "git+https://github.com/mongodb/libmongocrypt-rust.git?branch=main#694951462f80e3157b5e5a217543e5c30088604b"
 dependencies = [
  "bson 2.15.0",
  "bson 3.0.0",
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -138,8 +138,8 @@ optional = true
 features = ["serde"]
 
 [dependencies.mongocrypt]
-git = "https://github.com/mongodb/libmongocrypt-rust.git"
-branch = "main"
+git = "https://github.com/abr-egn/libmongocrypt-rust.git"
```
abr-egn: I'll revert this once https://github.com/mongodb/libmongocrypt-rust/pull/47 goes in.

src/bson_compat.rs
```diff
@@ -29,20 +29,18 @@ pub(crate) fn cstr_to_str(cs: &CStr) -> &str {
     }
 }
 
+#[cfg(feature = "bson-3")]
 pub(crate) trait RawDocumentBufExt: Sized {
-    fn append_ref_compat<'a>(
+    fn append_ref<'a>(
```
abr-egn, line 34: This was unrelated but I realized I could drop the ugly `_compat` suffix by having this trait only be present when using `bson` 3.0.

src/operation/update.rs
```diff
@@ -144,18 +146,15 @@ impl OperationWithDefaults for Update {
             }
 
             if let Some(ref let_vars) = options.let_vars {
-                body.append(
-                    cstr!("let"),
-                    crate::bson_compat::serialize_to_raw_document_buf(&let_vars)?,
-                );
+                body.append(cstr!("let"), RawDocumentBuf::try_from(let_vars)?);
```
abr-egn, line 149: Also unrelated but a nice example of why it's valuable to have serialization be visually distinct from encoding - happened to notice that this didn't need to go via serde as I was reading through.

src/client/auth/aws.rs
```diff
@@ -75,7 +72,8 @@ async fn authenticate_stream_inner(
         // channel binding is not supported.
         "p": 110i32,
     };
-    let client_first_payload_bytes = client_first_payload.encode_to_vec()?;
+    let mut client_first_payload_bytes = vec![];
```
abr-egn, line 75: I don't think the few places `to_vec` would be used is worth keeping the extra compat overhead for it.

isabelatkinson (dismissed):

isabelatkinson (approved):

PR #1438, opened 2025-07-30
RUST-1529 Use AWS SDK for sigv4 signing 

JamieTsai1024:
Replace existing implementation for getting sigv4 signing using the AWS SDK. 

### Testing 
Clone https://github.com/isabelatkinson/drivers-evergreen-tools/tree/local-aws and run the following commands:
```
./.evergreen/run-aws-auth.sh

cd .evergreen/auth_aws
# run this script to reconfigure the type of AWS authentication to use
# copy the block of unset and export commands output by the script and run them in the terminal window you're using to run the driver tests
./aws_setup.sh (regular | env-creds | assume-role | session-creds | web-identity)

cd ~/mongo-rust-driver 
# Run the unset and set export commands printed by `./aws_setup.sh` for environment variables 
cargo nextest run auth_aws --features aws-auth
```
### Previously 
- Used AWS SDK to retrieve credentials in [PR #1435 ](https://github.com/mongodb/mongo-rust-driver/pull/1435)

### Work to be done for [RUST-1529](https://jira.mongodb.org/browse/RUST-1529)
- Decide whether to keep AWS SDK behind feature flag or replace original implementation


src/client/auth/aws.rs
```diff
@@ -117,27 +126,34 @@ async fn authenticate_stream_inner(
     let creds = get_aws_credentials(credential).await.map_err(|e| {
         Error::authentication_error(MECH_NAME, &format!("failed to get creds: {e}"))
     })?;
-    let aws_credential = AwsCredential::from_sdk_creds(creds);
```
JamieTsai1024, line 120: In a previous [PR](https://github.com/mongodb/mongo-rust-driver/pull/1435), we introduced `from_sdk_creds(...)` to convert from `aws_credential_types::Credentials` to locally defined `AwsCredential` type. Now that we've converted the rest of the authentication to use the AWS SDK, we no longer need this conversion

.evergreen/config.yml
```diff
@@ -223,7 +223,7 @@ buildvariants:
 
   - name: aws-auth
     display_name: "AWS Authentication"
-    patchable: false
+    # patchable: false
```
JamieTsai1024: will uncomment once PR is approved

src/client/auth/aws.rs
```diff
@@ -197,6 +214,111 @@ pub(crate) async fn get_aws_credentials(credential: &Credential) -> Result<Crede
...
+        })?
+        .into();
+
+    let signable_request = SignableRequest::new(
+        request.method().as_str(),
+        request.uri().to_string(),
+        request.headers().iter().map(|(k, v)| {
+            (
+                k.as_str(),
+                std::str::from_utf8(v.as_bytes()).expect("Header value should be valid UTF-8"),
```
isabelatkinson: This will panic if the conversion fails; instead, we should propagate an authentication error

src/client/auth/aws.rs
```diff
@@ -197,6 +214,111 @@ pub(crate) async fn get_aws_credentials(credential: &Credential) -> Result<Crede
...
+        .get("authorization")
+        .ok_or_else(|| Error::authentication_error(MECH_NAME, "Missing authorization header"))?
+        .to_str()
+        .map_err(|e| Error::authentication_error(MECH_NAME, &format!("Invalid header value: {e}")))?
+        .to_string();
+
+    let token_header = headers
+        .get("x-amz-security-token")
+        .map(|v| {
+            v.to_str().map(|s| s.to_string()).map_err(|e| {
```
isabelatkinson:
```suggestion
            v.to_str().map_err(|e| {
```

No need to convert to an owned `String` here - a `&str` can be inserted on line 316

src/client/auth/aws.rs
```diff
@@ -197,6 +214,111 @@ pub(crate) async fn get_aws_credentials(credential: &Credential) -> Result<Crede
...
+
+    signing_instructions.apply_to_request_http1x(&mut request);
+
+    let headers = request.headers();
+    let authorization_header = headers
+        .get("authorization")
+        .ok_or_else(|| Error::authentication_error(MECH_NAME, "Missing authorization header"))?
+        .to_str()
+        .map_err(|e| Error::authentication_error(MECH_NAME, &format!("Invalid header value: {e}")))?
+        .to_string();
```
isabelatkinson: ditto here - no need to convert to a `String`

src/client/auth/aws.rs
```diff
@@ -197,6 +214,111 @@ pub(crate) async fn get_aws_credentials(credential: &Credential) -> Result<Crede
...
+        }),
+        SignableBody::Bytes(request.body().as_bytes()),
+    )
+    .map_err(|e| {
+        Error::authentication_error(MECH_NAME, &format!("Failed to create SignableRequest: {e}"))
+    })?;
+
+    let (signing_instructions, _signature) = sign(signable_request, &signing_params)
+        .map_err(|e| Error::authentication_error(MECH_NAME, &format!("Signing failed: {e}")))?
+        .into_parts();
```
isabelatkinson, line 296: nit: since we're not using the returned signature, I think this can call [`output`](https://docs.rs/aws-sigv4/latest/aws_sigv4/struct.SigningOutput.html#method.output) rather than `into_parts`

JamieTsai1024: Thanks Isabel! I tried switching to `.output()`, but ran into a borrow checker error when calling `apply_to_request_http1x` on line 291. I double-checked the method signature and saw that it needs an owned value which we get with `.into_parts()`, whereas `.output()` returns a reference.

The `.into_parts()` also matches what I saw in the two references we looked at ([1](https://github.com/aws-samples/sigv4-signing-examples/blob/main/sdk/rust/src/main.rs), [2](https://docs.rs/aws-sigv4/1.3.3/aws_sigv4/http_request/index.html)). Let me know if you had a different approach in mind!

isabelatkinson: ah I missed the reference vs. owned value distinction, feel free to leave as-is

src/client/auth/aws.rs
```diff
@@ -197,6 +214,118 @@ pub(crate) async fn get_aws_credentials(credential: &Credential) -> Result<Crede
...
+            .map(|(k, v)| {
+                let value = std::str::from_utf8(v.as_bytes()).map_err(|_| {
+                    Error::authentication_error(
+                        MECH_NAME,
+                        "Failed to convert header value to valid UTF-8",
+                    )
+                })?;
+                Ok((k.as_str(), value))
+            })
+            .filter_map(Result::ok),
```
isabelatkinson: This will ignore any failures that occur when converting the strings and proceed with authentication. This would likely cause the signature to be computed incorrectly, and the reported error might be confusing for users to parse. Instead, I suggest doing the following to propagate an error here:

```rust
let headers: Result<Vec<_>> = request
    .headers()
    .iter()
    .map(|(k, v)| {
        let v = v.to_str().map_err(|_| {
            Error::authentication_error(
                MECH_NAME,
                "Failed to convert header value to valid UTF-8",
            )
        })?;
        Ok((k.as_str(), v))
    })
    .collect();
```
This uses a handy iterator trick: if an iterator is mapped into `Result` values, you can `collect` it into a `Result<Vec<_>>` (or a `Result` of any collection type). That result will either contain `Ok(Vec<_>)` if all of the values were `Ok`, or an `Err(_)` containing the first error that was encountered.

We then need to return an error if needed using `?` and turn the value back into an iterator, which is the type accepted by `SignableRequest::new`:

```rust
headers?.into_iter()
```

This should make what happened more clear if something goes wrong with the headers.

src/client/auth/aws.rs
JamieTsai1024: sounds good, thanks!

src/client/auth/aws.rs
```diff
@@ -197,6 +214,119 @@ pub(crate) async fn get_aws_credentials(credential: &Credential) -> Result<Crede
     }
 }
 
+pub async fn compute_aws_sigv4_payload(
```
abr-egn: Does this need to be `async`?  I didn't see anything in the body that required it.

abr-egn (approved):
LGTM!

isabelatkinson (approved):

abr-egn (approved):

PR #1444, opened 2025-08-06
RUST-2245 Implement GSSAPI auth support for Windows

mattChiaravalloti:
This PR implements GSSAPI auth support for Windows as described by the [spec](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#gssapi). Similar to the Linux and Mac implementation in #1413, this one took longer than expected. In the end, it turned out that the `sspi` crate was not as amenable to this task as initially thought at scope time. I decided to attempt to implement this using the native Windows api (via `windows-sys`) and modeling it _exactly_ off of `libmongoc` since I know the C implementation works. Of course, that results in a bit more C-style code than Rust-style code, but after several weeks of investigating and implementing, this is finally working.

I updated the GSSAPI codepath to share common code between Linux/Mac and Windows as much as possible. I only separated out the platform-specific implementations and guarded each with the relevant `cfg` attributes.

This PR also updates the e2e tests to run on Windows. I observed this implementation passing my local testing on a Windows host, and the e2e tests passing on [this patch](https://spruce.mongodb.com/version/6893a446ba5677000783105c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). I anticipate the evergreen patch associated with this PR to succeed off the bat, as well.


.evergreen/config.yml
```diff
@@ -1405,7 +1413,7 @@ functions:
       type: test
       params:
         binary: bash
-        working_dir: ${PROJECT_DIRECTORY}
+        working_dir: src
```
mattChiaravalloti, line 1417: Using `${PROJECT_DIRECTORY}` on `windows-64-vs2017-small` was causing "working_dir could not be found" errors. I checked the rest of `config.yml` and it seems like the majority of `working_dir` values are set to `src`. I only chose `${PROJECT_DIRECTORY}` initially since I based this function on the aws auth functions.

.evergreen/run-gssapi-tests.sh
```diff
@@ -16,50 +16,56 @@ FEATURE_FLAGS+=("gssapi-auth")
...
-  $SASL_DOMAIN = $SASL_REALM
-" > krb5.conf
-
-export KRB5_CONFIG=krb5.conf
-
-# Authenticate the user principal in the KDC before running the e2e test
-echo "Authenticating $PRINCIPAL"
-echo "$SASL_PASS" | kinit -p $PRINCIPAL
-klist
+if [[ "Windows_NT" != "$OSTYPE" ]]; then
```
mattChiaravalloti: I will add a comment that explains `kinit`/`kdestroy` and other krb5 setup is not required on Windows.

Cargo.toml
```diff
@@ -235,6 +234,13 @@ features = ["serde", "serde_json-1"]
 rustdoc-args = ["--cfg", "docsrs"]
 all-features = true
 
+# Target-specific dependencies for GSSAPI authentication
```
mattChiaravalloti, line 237: To be honest, I've never had to do this before in a `Cargo.toml` file so I'm not sure if this is the idiomatic way.

src/client/auth/gssapi/mod.rs
```diff
@@ -54,11 +58,20 @@ pub(crate) async fn authenticate_stream(
...
 
     let user_principal = credential.username.clone();
-    let (mut authenticator, initial_token) =
-        GssapiAuthenticator::init(user_principal, properties.clone(), &hostname).await?;
-
+    let service_principal = properties.service_principal_name(&hostname, user_principal.as_ref());
     let source = credential.source.as_deref().unwrap_or("$external");
 
+    #[cfg(target_os = "windows")]
+    let (mut authenticator, initial_token) = windows::SspiAuthenticator::init(
```
mattChiaravalloti: As you see in the `windows.rs` file, I created the `SspiAuthenticator` struct to mirror the `GssapiAuthenticator` struct. It has the same interface (although I did not go so far as to make them both implement a common trait since they'll never be compiled together). This allowed the core authentication steps in this function to remain unchanged.

src/client/auth/gssapi/mod.rs
```diff
@@ -220,98 +219,11 @@ impl GssapiAuthenticator {
...
-    }
-
-    fn is_complete(&self) -> bool {
-        self.is_complete
+        service_principal
     }
 }
 
-async fn canonicalize_hostname(
+pub(crate) async fn canonicalize_hostname(
```
mattChiaravalloti: Ah this was more relevant when I made `canonicalize_hostname` calls in the child modules; I'll revert this to private.

src/client/auth/gssapi/nix.rs
```diff
@@ -0,0 +1,108 @@
+use cross_krb5::{ClientCtx, InitiateFlags, K5Ctx, PendingClientCtx, Step};
```
mattChiaravalloti, line 1: This file is essentially just cut-pasted from the original `gssapi.rs` file. The only notable change is a minor tweak to the function signatures: they are no longer marked as `async` since that is not required. I left that in basically by accident after doing all the refactoring and trial-and-erroring the first time around.

src/client/auth/gssapi/windows.rs
```diff
@@ -0,0 +1,408 @@
...
+
+pub(super) struct SspiAuthenticator {
+    cred_handle: Option<SecHandle>,
+    ctx_handle: Option<SecHandle>,
+    auth_complete: bool,
+    service_principal: Vec<u16>,
+    user_principal: String,
+}
+
+impl SspiAuthenticator {
```
mattChiaravalloti, line 46: This is the core implementation for the SSPI auth protocol on Windows. It is based **heavily** on [the C implementation](https://github.com/mongodb/mongo-c-driver/blob/3385ba9c2251ec184e23c12559796ce49d667c88/src/libmongoc/src/mongoc/mongoc-sspi.c#L170). I truly tried (for weeks) to get a higher level, more Rust-like library to work first. At the end of the day, though, the best way to do this was to resort to the low-level Windows API. It is a bit harder to read, but fortunately the SSPI api is not something that changes often (or will/should change in the future). The C implementation has been largely untouched for years and it still works!

src/client/auth/gssapi/windows.rs
```diff
@@ -0,0 +1,408 @@
...
+            Ok(result_buf)
+        }
+    }
+
+    pub(super) fn is_complete(&self) -> bool {
+        self.auth_complete
+    }
+}
+
+impl Drop for SspiAuthenticator {
```
mattChiaravalloti, line 399: Required since these handles are set and used basically exclusively in `unsafe` blocks.

mattChiaravalloti:
> Is it possible to make exceptions to the `semgrep` policy to allow for the use of these `unsafe` blocks, or will we need to get rid of them?

I investigated further and am not confident SSPI support can be implemented the way our drivers require if we do not use these `unsafe` blocks.

There are limited crate options for kerberos auth. We've ruled out `libgssapi-sys`/`libgssapi`/`cross-krb5` since they rely on MIT KRB5. I discovered `sspi` is insufficient for our needs (it is incapable of producing the token types needed for MongoDB's SASL auth flow when using `GSSAPI`). Looking through crates.io, there are not any other viable candidates for this, so the `windows` crate is likely our best option.

Using the native Windows API inherently requires `unsafe` code. Hopefully we can make an exception for this. I know in the ODBC driver, we have some required `unsafe` code that is permitted.

abr-egn:
> > Is it possible to make exceptions to the `semgrep` policy to allow for the use of these `unsafe` blocks, or will we need to get rid of them?
> 
> I investigated further and am not confident SSPI support can be implemented the way our drivers require if we do not use these `unsafe` blocks.
> 
> There are limited crate options for kerberos auth. We've ruled out `libgssapi-sys`/`libgssapi`/`cross-krb5` since they rely on MIT KRB5. I discovered `sspi` is insufficient for our needs (it is incapable of producing the token types needed for MongoDB's SASL auth flow when using `GSSAPI`). Looking through crates.io, there are not any other viable candidates for this, so the `windows` crate is likely our best option.
> 
> Using the native Windows API inherently requires `unsafe` code. Hopefully we can make an exception for this. I know in the ODBC driver, we have some required `unsafe` code that is permitted.

I haven't had the chance to look into this PR in detail, but it should work to use a `nosemgrep` comment, we've had to do that in a couple other places in the code base ([example](https://github.com/mongodb/mongo-rust-driver/blob/main/src/client/auth/scram.rs#L320)).  My recollection of the policy (I'll verify later) is that as long as we've investigated and have manually validated that the thing semgrep would complain about is not, in fact, a problem it's fine.

src/client/auth/gssapi/nix.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use crate::{
+    client::auth::GSSAPI_STR,
+    error::{Error, Result},
+};
+
+pub(super) struct GssapiAuthenticator {
```
abr-egn, line 16: Not changed in this PR, but this being a fresh file made me realize - this is only ever either
```
GssapiAuthenticator { pending_ctx: Some(...), established_ctx: None, is_complete: false, ... }
```
or
```
GssapiAuthenticator { pending_ctx: None, established_ctx: Some(...), is_complete: true, ... }
```
right?  That seems like it would be better modeled via a state enum, e.g.
```
enum CtxState {
  Pending(PendingClientCtx),
  Established(ClientCtx),
}

pub(super) struct GssapiAuthenticator {
  ctx: CtxState,
  user_principal: String,
}
```

src/client/auth/gssapi/windows.rs
```diff
@@ -0,0 +1,408 @@
...
+            service_principal,
+            user_principal: user_principal.clone(),
+        };
+
+        let initial_token = authenticator.acquire_credentials_and_init(password)?;
+        Ok((authenticator, initial_token))
+    }
+
+    fn acquire_credentials_and_init(&mut self, password: Option<String>) -> Result<Vec<u8>> {
+        unsafe {
```
abr-egn: I believe the only unsafe portion of this is calling `AcquireCredentialsHandleW`, right?  I'd rather see the `unsafe` blocks as narrowly scoped as possible.

src/client/auth/gssapi/windows.rs
abr-egn: I'm going to have to defer to @kevinAlbs to review the actual semantics of the implementation here :)

src/client/auth/gssapi/nix.rs
mattChiaravalloti: Ah that's a nice refactor. I'll make this change.

src/client/auth/gssapi/windows.rs
mattChiaravalloti: Good point! I started broad to allow for whatever had to be done, but I'll narrow these down now like you suggested 👍

src/client/auth/gssapi/nix.rs
```diff
@@ -0,0 +1,108 @@
...
+            Err(Error::authentication_error(
+                GSSAPI_STR,
+                "Authentication context not initialized",
+            ))
+        }
+    }
+
+    // Perform the final step of Kerberos authentication by gss_unwrap-ing the
+    // final server challenge, then wrapping the protocol bytes + user principal.
+    // The resulting token must be sent to the server.
```
pmeredit, line 87: This is like a monad transformer :joy:

.evergreen/config.yml
kevinAlbs: Using `src` SGTM. I expect the error on Windows is due to `${PROJECT_DIRECTORY}` being a Cygwin path (e.g. `/cygdrive/c/data/mci/7649/src` rather than `C:\data\mci\7649\src`). I expect the other tasks using `${PROJECT_DIRECTORY}` are not running on Windows.

.evergreen/config.yml
```diff
@@ -274,6 +274,14 @@ buildvariants:
     tasks:
       - test-gssapi-auth
 
+  - name: gssapi-auth-windows
+    display_name: "GSSAPI Authentication - Windows"
+    patchable: true
+    run_on:
+      - windows-64-vs2017-small
```
kevinAlbs:
```suggestion
      - windows-vsCurrent-small
```

Suggest using `windows-vsCurrent-small`. Quoting [Guidelines around Evergreen distros](https://wiki.corp.mongodb.com/spaces/DBDEVPROD/pages/116563465/Guidelines+around+Evergreen+distros):

> All previous windows distros (*-vs2013, *-vs2015, *-vs2017, *-vs2019) are deprecated

A patch build using `windows-vsCurrent-small` appears to [pass](https://spruce.mongodb.com/version/6899e5ad7cef440007d50e60).

src/client/auth/gssapi/mod.rs
isabelatkinson, line 1: This filename should be able to stay the same - we stopped using `mod.rs` files a while ago (https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames)

src/test/auth/gssapi.rs
```diff
@@ -9,6 +9,7 @@ use crate::{
 ///   - auth_mechanism_properties is an optional set of authMechanismProperties to append to the uri
```
kevinAlbs, line 11: Minor: suggest documenting `password_var` and noting why it is Windows-specific:

```rust
///   - password_var is the name of the environment variable that stores the Windows password. Only applies to Windows since GSSAPI on non-Windows obtains credentials through `kinit`.
```

src/client/auth/gssapi/windows.rs
```diff
@@ -0,0 +1,423 @@
...
+        let package_name: Vec<u16> = "kerberos\0".encode_utf16().collect();
+
+        // nosemgrep: rust.lang.security.unsafe-usage.unsafe-usage
+        unsafe {
+            let result = AcquireCredentialsHandleW(
+                std::ptr::null(),
+                package_name.as_ptr(),
+                SECPKG_CRED_OUTBOUND,
+                ptr::null_mut(),
+                if password.is_some() {
```
kevinAlbs, line 117: Suggest documenting why passing credentials is conditioned on the password (the C driver does not have this comment, but it might help future readers).

```suggestion
                if password.is_some() {
                // Only pass credentials if a password is present. See DRIVERS-2180.
```

src/client/auth/gssapi/windows.rs
```diff
@@ -0,0 +1,423 @@
...
+                "Server message is too short",
+            ));
+        }
+
+        let data_ptr = wrap_bufs[1].pvBuffer as *const u8;
+
+        // nosemgrep: rust.lang.security.unsafe-usage.unsafe-usage
+        unsafe {
+            let first_byte = *data_ptr;
+            if (first_byte & 1) == 0 {
```
kevinAlbs: What is this first byte referring to? I did not see this same check in the C driver.

.evergreen/run-gssapi-tests.sh
```diff
@@ -16,50 +16,60 @@ FEATURE_FLAGS+=("gssapi-auth")
...
-
-# Authenticate the user principal in the KDC before running the e2e test
-echo "Authenticating $PRINCIPAL"
-echo "$SASL_PASS" | kinit -p $PRINCIPAL
-klist
+# On Windows, `kinit`/`kdestroy` and other krb5 config settings are
+# not available, nor are they required steps. Windows uses SSPI which
+# is similar to but distinct from (KRB5) GSSAPI. Therefore, we only
+# run the following steps if we are not on Windows.
+if [[ "Windows_NT" != "$OSTYPE" ]]; then
```
kevinAlbs:
```suggestion
if [[ "cygwin" != "$OSTYPE" ]]; then
```

To fix logs [in patch](https://parsley.mongodb.com/evergreen/mongo_rust_driver_gssapi_auth_windows_test_gssapi_auth_patch_3e587b937d1c316eac933be9b7b46037c43895f6_6893a446ba5677000783105c_25_08_06_18_52_04/0/task?bookmarks=0,2165&shareLine=1334):
```
.evergreen/run-gssapi-tests.sh: line 42: kinit: command not found
```

.evergreen/run-gssapi-tests.sh
```diff
@@ -16,50 +16,60 @@ FEATURE_FLAGS+=("gssapi-auth")
...
+  klist
+fi
 
 TEST_OPTIONS=()
 cargo_test test::auth::gssapi_skip_local::with_service_realm_and_host_options
 
-# Unauthenticate
-echo "Unuthenticating $PRINCIPAL_CROSS"
-kdestroy
+if [[ "Windows_NT" != "$OSTYPE" ]]; then
```
kevinAlbs:
```suggestion
if [[ "cygwin" != "$OSTYPE" ]]; then
```

.evergreen/run-gssapi-tests.sh
```diff
@@ -16,50 +16,60 @@ FEATURE_FLAGS+=("gssapi-auth")
...
+fi
 
 # Run end-to-end auth tests for "$PRINCIPAL" user
 TEST_OPTIONS+=("--skip with_service_realm_and_host_options")
 cargo_test test::auth::gssapi_skip_local
 
-# Unauthenticate
-echo "Unauthenticating $PRINCIPAL"
-kdestroy
+if [[ "Windows_NT" != "$OSTYPE" ]]; then
```
kevinAlbs:
```suggestion
if [[ "cygwin" != "$OSTYPE" ]]; then
```

kevinAlbs (commented):
The contribution is much appreciated! I have mostly minor comments and questions.

src/client/auth/gssapi/mod.rs
mattChiaravalloti: Thank you for pointing this out, I updated this to use the appropriate naming conventions.

.evergreen/config.yml
mattChiaravalloti: Thanks for running a patch to confirm `windows-vsCurrent-small` works. Updated to use that in the config file.

src/test/auth/gssapi.rs
mattChiaravalloti: done

src/client/auth/gssapi/windows.rs
mattChiaravalloti: done

.evergreen/run-gssapi-tests.sh
mattChiaravalloti: All 3 of these done -- thanks for catching it.

src/client/auth/gssapi/windows.rs
mattChiaravalloti: We don't do the check in the C driver but I thought we had to do this. Basically, I was trying to confirm that the security layer we get back from the server matches the one we send (1). But on second thought, I don't think that's necessary at all. I just re-read the RFC and took a look through the C code again, and I decided to remove this check. At best it's unnecessary, and at worst it could reject valid server responses.

mattChiaravalloti:
I suspect the "project fails validation: sync with base branch & run `evergreen validate -p <project>`" error is related to the ongoing Github problems. When I ran `evergreen validate` locally just now after merging main, it said the config was valid 👍 We should be able to run this patch whenever Github service is resolved and the evergreen team gives the green light for things to run as expected again.

kevinAlbs (approved):
LGTM!

mattChiaravalloti:
I retagged everyone, but feel free to just take a look at the parts you commented on previously. I know this is a big and confusing PR. Plus, with 4 reviewers on this it could get hectic having too many cooks. Kevin's LGTM on the core functionality will be the primary indicator this is ready to merge 👍 Thanks for all the review and guidance on this one.

abr-egn (approved):

isabelatkinson (approved):

PR #1445, opened 2025-08-07
RUST-2078 Support QE with bulk write

isabelatkinson:


src/bson_util.rs
```diff
@@ -246,6 +248,49 @@ pub(crate) fn get_or_prepend_id_field(doc: &mut RawDocumentBuf) -> Result<Bson>
...
+pub(crate) trait RawDocumentCollection: Default {
+    /// Calculates the total number of bytes that would be added to a collection of this type by the
+    /// given document.
+    fn bytes_added(index: usize, doc: &RawDocumentBuf) -> Result<usize>;
+
+    /// Adds the given document to the collection.
+    fn push(&mut self, doc: RawDocumentBuf);
+
+    /// Adds the collection of raw documents to the provided command.
+    fn add_to_command(self, identifier: &'static CStr, command: &mut Command);
```
abr-egn: Extremely minor nit: I don't think this needs the `'static`?

src/operation/bulk_write.rs
```diff
@@ -222,39 +233,104 @@ where
...
                 let namespace_doc = rawdoc! { "ns": namespace.to_string() };
-                let length_added = namespace_doc.as_bytes().len();
-                self.namespaces.push(namespace_doc);
-                let next_index = self.cache.len();
-                self.cache.insert(namespace, next_index);
-                (next_index, length_added)
+                let index = self.cache.len();
+                let bytes_added = T::bytes_added(index, &namespace_doc)?;
+                self.pending_namespace = Some(namespace_doc);
+                self.cache.insert(namespace, index);
```
abr-egn: I know in current usage it doesn't matter because a pending document not being added terminates the outer loop, but it still bugs me that the namespace is added to the cache even if `add_pending` is never called.  Not sure what a good way to fix that would be, though.

src/bson_util.rs
isabelatkinson: it does not, updated!

src/operation/bulk_write.rs
isabelatkinson: updated to defer adding to the cache til `add_pending` is called

PR #1451, opened 2025-08-15
RUST-2157 Initial framework for autogenerated Atlas Search helpers

abr-egn:
RUST-2157

This has the initial yaml file fetch and the sketch of the code generator, along with the output for three operators I've been using as a testbed to iterate on the API (`autocomplete`, `text`, and `compound`).  Codegen can quickly get impenetrable so I've tried to keep it reasonable; being able to use `parse_quote!` rather than raw strings was a real help.

Given that this is the same set of core Rust metaprogramming crates that macros use, I'm somewhat tempted to see if we can shift the action macro machinery to code generation, since that would likely knock off a chunk of compile time.

### DIFF
--- a/etc/gen_atlas_search/src/main.rs
+++ b/etc/gen_atlas_search/src/main.rs
@@ -8,7 +8,6 @@ use syn::parse_quote;
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 struct Operator {
     name: String,
-    #[expect(dead_code)]
     link: String,
     #[serde(rename = "type")]
     #[expect(dead_code)]
@@ -70,12 +69,18 @@ impl Operator {
         }
 
         let desc = &self.description;
+        let link = format!(
+            "For more details, see the [{name_text} operator reference]({}).",
+            self.link
+        );
         parse_quote! {
             #[allow(missing_docs)]
             pub struct #name_ident;
 
             impl AtlasSearch<#name_ident> {
                 #[doc = #desc]
+                #[doc = ""]
+                #[doc = #link]
                 pub fn #constr_ident(#required_args) -> Self {
                     AtlasSearch {
                         name: #name_text,
--- a/src/atlas_search.rs
+++ b/src/atlas_search.rs
@@ -50,7 +50,23 @@ impl<T> From<AtlasSearch<T>> for Document {
 
 impl<T> AtlasSearch<T> {
     /// Erase the type of this builder.  Not typically needed, but can be useful to include builders
-    /// of different types in a single `Vec`.
+    /// of different types in a single `Vec`:
+    /// ```no_run
+    /// # async fn wrapper() -> mongodb::error::Result<()> {
+    /// # use mongodb::{Collection, atlas_search::AtlasSearch, bson::{Document, doc}};
+    /// # let collection: Collection<Document> = todo!();
+    /// let cursor = collection.aggregate(vec![
+    ///     AtlasSearch::compound()
+    ///         .must(vec![
+    ///             AtlasSearch::text("description", "varieties").unit(),
+    ///             AtlasSearch::compound()
+    ///                 .should(AtlasSearch::text("description", "Fuji"))
+    ///                 .unit(),
+    ///         ])
+    ///         .into(),
+    /// ]).await?;
+    /// # }
+    /// ```
     pub fn unit(self) -> AtlasSearch<()> {
         AtlasSearch {
             name: self.name,
--- a/src/atlas_search/gen.rs
+++ b/src/atlas_search/gen.rs
@@ -8,6 +8,8 @@ impl AtlasSearch<Autocomplete> {
     fields that you intend to query with the autocomplete operator must be
     indexed with the autocomplete data type in the collection's index definition.
     */
+    ///
+    ///For more details, see the [autocomplete operator reference](https://www.mongodb.com/docs/atlas/atlas-search/autocomplete/).
     pub fn autocomplete(path: impl StringOrArray, query: impl StringOrArray) -> Self {
         AtlasSearch {
             name: "autocomplete",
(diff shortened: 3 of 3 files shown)


etc/gen_atlas_search/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+    #[expect(dead_code)]
+    link: String,
+    #[serde(rename = "type")]
+    #[expect(dead_code)]
+    type_: Vec<OperatorType>,
+    #[expect(dead_code)]
+    encode: EncodeType,
+    description: String,
+    arguments: Vec<Argument>,
+    tests: Vec<serde_yaml::Value>,
```
abr-egn, line 19: The yaml file test definitions aren't terribly useful here, so I just cheated the parsing.

etc/gen_atlas_search/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+impl Operator {
+    fn clear_tests(mut self) -> Self {
```
abr-egn: This is to make debugging output easier to read.

etc/gen_atlas_search/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+            return ArgumentRustType::MatchCriteria;
+        }
+        match self.type_.as_slice() {
+            [ArgumentType::String] => ArgumentRustType::String,
+            [ArgumentType::Object] => ArgumentRustType::Document,
+            [ArgumentType::SearchScore] => ArgumentRustType::Document,
+            [ArgumentType::SearchPath] => ArgumentRustType::StringOrArray,
+            [ArgumentType::SearchOperator, ArgumentType::Array] => ArgumentRustType::Operator,
+            [ArgumentType::Int] => ArgumentRustType::I32,
+            _ => panic!("Unexpected argument types: {:?}", self.type_),
```
abr-egn, line 141: Since this is a standalone binary, I figure `panic!` / `unwrap` is actually reasonable error handling.

src/atlas_search.rs
```diff
@@ -0,0 +1,270 @@
...
+        doc! {
+            "$search": {
+                "index": index.as_ref(),
+                self.name: self.stage,
+            }
+        }
+    }
+}
+
+impl<T> IntoIterator for AtlasSearch<T> {
```
abr-egn, line 90: This is handy for things like the methods on `compound` which expect an `IntoIterator<Item = AtlasSearch<T>>`, it means you can just pass in a single value without having to wrap it in `vec![]`.

src/atlas_search.rs
```diff
@@ -0,0 +1,270 @@
...
+    type IntoIter = std::iter::Once<AtlasSearch<T>>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        std::iter::once(self)
+    }
+}
+
+/// Order in which to search for tokens.
+#[derive(Debug, Clone, PartialEq)]
+pub enum TokenOrder {
```
abr-egn, line 103: Auxiliary types like this abstractly could fit better into the `gen` module since that's where they're used but since these don't actually depend on the input files it seemed better to hoist them up.

etc/gen_atlas_search/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+convert_case = "0.8.0"
+prettyplease = "0.2.36"
+proc-macro2 = "1.0.97"
+quote = "1.0.40"
+serde = { version = "1.0.219", features = ["derive"] }
+serde_yaml = "0.9.34"
```
abr-egn, line 12: This has been unmaintained for a while now but the alternatives aren't yet settled and it works fine for what's needed here.

src/atlas_search.rs
```diff
@@ -0,0 +1,267 @@
...
+        doc! {
+            "$search": {
+                value.name: value.stage
+            }
+        }
+    }
+}
+
+impl<T> AtlasSearch<T> {
+    /// Erase the type of this builder.  Not typically needed, but can be useful to include builders
```
abr-egn, line 52: This awkward little method is the downside of using a type parameter to differentiate the builders.  However, this is _only_ needed for situations like multiple subclauses of `compound` that are themselves different types; if we used a fully distinct type for each builder they'd _always_ need a `.build()` tacked on at the end, which this avoids in all other circumstances.

etc/gen_atlas_search/src/main.rs
```diff
@@ -0,0 +1,218 @@
...
+                        self
+                    }
+                });
+            } else {
+                required_args.push(parse_quote! { #ident : #type_, });
+                init_doc.push(parse_quote! { #arg_name : #init_expr, });
+            }
+        }
+
+        let desc = &self.description;
```
isabelatkinson, line 71: It might be nice to add the link to the documentation as well since it's already in the yaml, i.e. something like:

```rust
let desc = format!("{self.description}\n\nFor more details, see {self.link}");
```

src/atlas_search.rs
isabelatkinson: can you add a code example here of how this would be used?

isabelatkinson (commented):
just a few doc suggestions!

etc/gen_atlas_search/src/main.rs
```diff
@@ -70,12 +69,18 @@ impl Operator {
...
+            "For more details, see the [{name_text} operator reference]({}).",
+            self.link
+        );
         parse_quote! {
             #[allow(missing_docs)]
             pub struct #name_ident;
 
             impl AtlasSearch<#name_ident> {
                 #[doc = #desc]
+                #[doc = ""]
```
abr-egn, line 82: Embedding the newline in a single docstring made rustdoc try to parse it as a code snippet for some reason 🤷

isabelatkinson (approved):

PR #1453, opened 2025-08-20
RUST-2264 Fix handling of tlsInsecure in the URI

isabelatkinson:
Fixes handling of the `tlsInsecure` option in a URI. `tlsInsecure=bool` now sets `TlsOptions::allow_invalid_certificates` and `TlsOptions::allow_invalid_hostnames` (if `openssl-tls` is enabled) to the value of `bool`.

Note that `tlsInsecure` can _only_ be set via a URI; the `ConnectionString` and `ClientOptions` types both store the `TlsOptions` type, which only has fields for the more granular options.

### DIFF
--- a/src/client/options.rs
+++ b/src/client/options.rs
@@ -55,6 +55,10 @@ pub(crate) use resolver_config::ResolverConfig;
 
 pub(crate) const DEFAULT_PORT: u16 = 27017;
 
+const TLS_INSECURE: &str = "tlsinsecure";
+const TLS_ALLOW_INVALID_CERTIFICATES: &str = "tlsallowinvalidcertificates";
+#[cfg(feature = "openssl-tls")]
+const TLS_ALLOW_INVALID_HOSTNAMES: &str = "tlsallowinvalidhostnames";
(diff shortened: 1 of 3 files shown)


src/client/options.rs
```diff
@@ -1662,41 +1668,47 @@ impl ConnectionString {
...
-                &key.to_lowercase(),
-                percent_encoding::percent_decode(&value.as_bytes()[1..])
+                &key,
+                percent_encoding::percent_decode(value.as_bytes())
                     .decode_utf8_lossy()
                     .as_ref(),
             )?;
         }
 
+        if keys.contains(TLS_INSECURE) {
```
isabelatkinson, line 1698: See https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.md#conflicting-tls-options

We don't currently parse tlsAllowInvalidHostnames (RUST-1896), but I added the logic for it here so that we have the validation once we support the option in the URI.

abr-egn (approved):

kevinAlbs (approved):

PR #1461, opened 2025-08-28
RUST-2190 Remove support for server 4.0

abr-egn:
RUST-2190

Along with removing the test targets and bumping the min wire version, I did a pass to strip out now-unnecessary comments and test checks.

### DIFF
--- a/src/cmap/test/integration.rs
+++ b/src/cmap/test/integration.rs
@@ -17,8 +17,6 @@ use crate::{
     sdam::TopologyUpdater,
     selection_criteria::ReadPreference,
     test::{
-        block_connection_supported,
-        fail_command_supported,
         get_client_options,
         log_uncaptured,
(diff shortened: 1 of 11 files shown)


isabelatkinson (commented):
We should also be able to remove these test checks:

- [transactions support](https://github.com/mongodb/mongo-rust-driver/blob/330ced1002a89804a0ffdd7a484bb35c9623ae17/src/test.rs#L240)
- [block connection support](https://github.com/mongodb/mongo-rust-driver/blob/330ced1002a89804a0ffdd7a484bb35c9623ae17/src/test.rs#L243) (I'm assuming all the 4.2 servers we test against are 4.2.9+)
- [fail command support](https://github.com/mongodb/mongo-rust-driver/blob/330ced1002a89804a0ffdd7a484bb35c9623ae17/src/test.rs#L246)

PR #1462, opened 2025-08-29
RUST-2157 Complete set of Atlas Search helpers

abr-egn:
RUST-2157

This builds on #1451 and implements the expanded machinery to generate the rest of the helpers.  I'd say most of them needed some kind of addition beyond just "include the yaml file".

### DIFF
--- a/etc/gen_atlas_search/src/main.rs
+++ b/etc/gen_atlas_search/src/main.rs
@@ -39,7 +39,7 @@ impl Operator {
         self
     }
 
-    fn gen_helper(&self) -> Helper {
+    fn gen_helper(&self) -> TokenStream {
         let name_text = &self.name;
         let name_ident = format_ident!("{}", name_text.to_case(Case::Pascal));
         let constr_ident = format_ident!("{}", name_text.to_case(Case::Snake));
@@ -80,40 +80,27 @@ impl Operator {
             "facet" => parse_quote! { true },
             _ => parse_quote! { false },
         };
-        let toplevel = parse_quote! {
+        parse_quote! {
             #[allow(missing_docs)]
             pub struct #name_ident;
 
-            impl AtlasSearch<#name_ident> {
-                #[doc = #desc]
-                #[doc = ""]
-                #[doc = #link]
-                pub fn #constr_ident(#required_args) -> Self {
-                    AtlasSearch {
-                        name: #name_text,
-                        stage: doc! { #init_doc },
-                        meta: #meta,
-                        _t: PhantomData,
-                    }
-                }
-                #setters
-            }
-        };
-        let short = parse_quote! {
             #[doc = #desc]
             #[doc = ""]
             #[doc = #link]
             pub fn #constr_ident(#required_args) -> AtlasSearch<#name_ident> {
-                AtlasSearch::#constr_ident(#required_arg_names)
+                AtlasSearch {
+                    name: #name_text,
+                    stage: doc! { #init_doc },
+                    meta: #meta,
+                    _t: PhantomData,
+                }
             }
-        };
-        Helper { toplevel, short }
-    }
-}
 
-struct Helper {
-    toplevel: TokenStream,
-    short: TokenStream,
+            impl AtlasSearch<#name_ident> {
+                #setters
+            }
+        }
+    }
 }
 
 #[derive(Debug, Deserialize)]
@@ -225,7 +212,6 @@ impl TokenStreamExt for TokenStream {
 
 fn main() {
     let mut operators = TokenStream::new();
-    let mut short = TokenStream::new();
     for name in [
         "autocomplete",
         "compound",
@@ -242,9 +228,7 @@ fn main() {
         let parsed = serde_yaml::from_str::<Operator>(&contents)
             .unwrap()
             .clear_tests();
-        let helper = parsed.gen_helper();
-        operators.push(helper.toplevel);
-        short.push(helper.short);
+        operators.push(parsed.gen_helper());
     }
 
     let file = parse_quote! {
@@ -252,14 +236,6 @@ fn main() {
         use super::*;
 
         #operators
-
-        /// Atlas Search constructor functions without the `AtlasSearch::` prefix; can be useful to
-        /// improve readability when constructing deeply nested searches.
-        pub mod short {
-            use super::*;
-
-            #short
-        }
     };
     let text = prettyplease::unparse(&file);
     println!("{text}");
--- a/etc/gen_atlas_search/src/main.rs
+++ b/etc/gen_atlas_search/src/main.rs
@@ -41,8 +41,12 @@ impl Operator {
 
     fn gen_helper(&self) -> TokenStream {
         let name_text = &self.name;
-        let name_ident = format_ident!("{}", name_text.to_case(Case::Pascal));
-        let constr_ident = format_ident!("{}", name_text.to_case(Case::Snake));
+        let ident_base = match name_text.as_str() {
+            "in" => "searchIn",
+            _ => name_text,
+        };
+        let name_ident = format_ident!("{}", ident_base.to_case(Case::Pascal));
+        let constr_ident = format_ident!("{}", ident_base.to_case(Case::Snake));
 
         let mut required_args = TokenStream::new();
         let mut required_arg_names = TokenStream::new();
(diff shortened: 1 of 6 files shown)


etc/gen_atlas_search/src/main.rs
```diff
@@ -73,21 +87,27 @@ impl Operator {
...
-                pub fn #constr_ident(#required_args) -> Self {
-                    AtlasSearch {
-                        name: #name_text,
-                        stage: doc! { #init_doc },
-                        _t: PhantomData,
-                    }
+            #[doc = #desc]
+            #[doc = ""]
+            #[doc = #link]
+            pub fn #constr_ident(#required_args) -> AtlasSearch<#name_ident> {
```
abr-egn: I pulled the constructor functions out of the `AtlasSearch` struct because it got very noisy when using helpers like `compound` that can have lots of sub-clauses; this way users can either have the names prefixed or not depending on their choice of import.

(Couldn't do `use AtlasSearch::*`, sadly, the compiler complains that `AtlasSearch` isn't a namespace, despite the same syntax working for enum variants.)

src/atlas_search.rs
```diff
@@ -35,47 +36,50 @@ use crate::bson::{doc, Bson, Document};
 pub struct AtlasSearch<T> {
     name: &'static str,
     stage: Document,
+    meta: bool,
     _t: PhantomData<T>,
 }
 
-impl<T> From<AtlasSearch<T>> for Document {
```
abr-egn, line 41: I swapped this for `stage` because the document constructed is very specific to the context of an aggregation pipeline rather than being a general conversion (and it's just one more character).

src/atlas_search.rs
```diff
@@ -142,144 +146,206 @@ impl MatchCriteria {
...
-    fn to_bson(self) -> Bson;
-}
+mod private {
+    use crate::bson::{doc, Bson};
 
-impl StringOrArray for &str {
-    fn to_bson(self) -> Bson {
-        Bson::String(self.to_owned())
+    /// An Atlas Search operator parameter that can accept multiple types.
+    pub trait Parameter {
```
abr-egn, line 311: The setup with the `Parameter` trait means both that the specific child traits are sealed (it wouldn't make sense for users to add new impls for those, they're just ways to define subsets of bson values) and that the conversion to a value for inclusion in the pipeline doc doesn't have to be defined for each one.

src/atlas_search.rs
```diff
@@ -142,144 +146,206 @@ impl MatchCriteria {
...
 
-impl<const N: usize> StringOrArray for &[&str; N] {
-    fn to_bson(self) -> Bson {
-        Bson::Array(self.iter().map(|&s| Bson::String(s.to_owned())).collect())
-    }
-}
+/// An Atlas Search operator parameter that is itself a search operator.
+pub trait SearchOperator: private::Parameter {}
+impl<T> SearchOperator for AtlasSearch<T> {}
+impl SearchOperator for Document {}
```
abr-egn: This is future-proofing, so users can always fall back to directly using documents if our builders don't have the syntax they need for a subclause.

macros/src/lib.rs
abr-egn, line 1: The changes here and to aggregate.rs are from when I was considering a new action as the entry point; I ended up not doing that but these seemed reasonable microimprovements to keep.

src/atlas_search.rs
```diff
@@ -32,65 +33,180 @@ use crate::bson::{doc, Bson, Document};
...
-    /// [`Document`], but also specify the search index to use.
-    pub fn on_index(self, index: impl AsRef<str>) -> Document {
-        doc! {
-            "$search": {
-                "index": index.as_ref(),
-                self.name: self.stage,
-            }
-        }
+/// A pending `$search` aggregation stage.
+pub struct AtlasSearch {
```
abr-egn, line 124: This needs to exist as a place to hang options for the `$search` stage rather than the individual operator.  This does mean you get a little awkward when setting options (`search(operator(param)).option(value).stage()`) but that can be sidestepped with the operator's `stage()` when no options need to be set.

Alternatives considered:
* `operator(param).search().option(value).stage()` syntax instead; subjectively it stutters, and it also doesn't reflect the underlying model as well.
* New `search` action distinct from `aggregate`.  Would require duplicating all of `aggregate`'s options and would then conflate options for the stage with options for the operation, which doesn't seem great and could run into name clashes.
* Include `search` option setters in the generic `SearchOperator<T>` impl; allows nonsensical things like trying to set a different index on a nested operator than the toplevel one, and has the same name clash issue as the action does.

src/atlas_search.rs
```diff
@@ -32,65 +33,180 @@ use crate::bson::{doc, Bson, Document};
...
+    /// Document that specifies the fields to sort the Atlas Search results by in ascending or
+    /// descending order.
+    pub fn sort(mut self, value: Document) -> Self {
+        self.stage.insert("sort", value);
+        self
+    }
+
+    /// Document that specifies the tracking option to retrieve analytics information on the search
+    /// terms.
+    pub fn tracking(mut self, value: Document) -> Self {
```
isabelatkinson: This option is scheduled for removal later this year: https://www.mongodb.com/docs/atlas/atlas-search/aggregation-stages/search/#fields

I suggest we omit it entirely from our API since it'll stop working fairly soon

src/atlas_search.rs
```diff
@@ -142,144 +258,198 @@ impl MatchCriteria {
...
+                "type": "date",
+                "path": path.as_ref(),
+                "boundaries": boundaries.into_iter().collect::<Vec<_>>(),
+            },
+            _t: PhantomData,
+        }
+    }
+    impl Facet<Date> {
+        #[allow(missing_docs)]
+        pub fn default(mut self, bucket: impl AsRef<str>) -> Self {
```
isabelatkinson: I could go either way on this, but I was initially confused here by the naming collision with `Default::default` - thoughts on naming this something like `default_bucket` instead?

src/atlas_search.rs
```diff
@@ -142,144 +258,198 @@ impl MatchCriteria {
...
+/// ```
+pub mod facet {
+    use crate::bson::{doc, Bson, Document};
+    use std::marker::PhantomData;
 
-impl<const N: usize> StringOrArray for &[String; N] {
-    fn to_bson(self) -> Bson {
-        Bson::Array(self.iter().map(|s| Bson::String(s.clone())).collect())
+    /// A facet definition.
+    pub struct Facet<T> {
```
isabelatkinson, line 356: Suggest adding a link to the facet docs for more details: https://www.mongodb.com/docs/atlas/atlas-search/facet/

Also, can you add docs for the options setters? They can probably just be copied from what's in the docs page

etc/gen_atlas_search/src/main.rs
```diff
@@ -77,17 +91,17 @@ impl Operator {
             #[allow(missing_docs)]
             pub struct #name_ident;
```
isabelatkinson, line 96: Suggest adding a basic doc for these structs, something like:
```
/// {operator} Atlas Search operator. Construct with {link to constructor function}.
```

This is primarily motivated by the generated docs, which look a little bare in their current state:
Screenshot 2025-09-05 at 15 04 50

src/atlas_search.rs
```diff
@@ -32,65 +33,180 @@ use crate::bson::{doc, Bson, Document};
...
+
+    /// Document that specifies the tracking option to retrieve analytics information on the search
+    /// terms.
+    pub fn tracking(mut self, value: Document) -> Self {
+        self.stage.insert("tracking", value);
+        self
+    }
+
+    /// Convert to an aggregation stage document.
+    pub fn stage(self) -> Document {
```
isabelatkinson: I know it's more verbose, but I would prefer `into_stage` here to make it more clear that this is the final thing that needs to happen when constructing a search stage; otherwise this reads more as another constructor method

src/atlas_search.rs
isabelatkinson: nice, this is very convenient!

etc/gen_atlas_search/src/main.rs
abr-egn: Reading over the docs page for the module led me to make a couple of other updates as well:
* The example from `AtlasSearch` is now attached to the module-wide documentation as a "here's how things in this module fit together" snipper.
* I re-used the doc macros for actions to attach the setters to the documentation for the constructors so there's an easy discoverability path.

isabelatkinson (approved):
lgtm!

PR #1471, opened 2025-09-15
RUST-2203 Allow client metadata to be appended post-construction

abr-egn:
RUST-2203 / RUST-2272

This required moving metadata from something that each worker has its own copy of to a central per-`Client` (actually per-`Topology`) shared value.

I initially implemented this as an update that gets fanned out to the workers using the existing infrastructure that's there for shutdown and the like; that turned out to require a centralized copy anyway so I dropped the fan-out.  In the process, though, I did some refactoring of the related methods so that worker commands require fewer intermediate pass-through wrapper methods and division of logic among the components is clearer.

### DIFF
--- a/src/cmap/establish.rs
+++ b/src/cmap/establish.rs
@@ -65,7 +65,7 @@ impl From<&TopologySpec> for EstablisherOptions {
 #[cfg(test)]
 impl From<&ClientOptions> for EstablisherOptions {
     fn from(options: &ClientOptions) -> Self {
-        Self::from(&TopologySpec::from(options.clone()))
+        Self::from(&TopologySpec::try_from(options.clone()).unwrap())
     }
 }
 
--- a/src/cmap/establish/handshake.rs
+++ b/src/cmap/establish/handshake.rs
@@ -5,6 +5,7 @@ use crate::event::EventHandler;
 use crate::{
     bson::{rawdoc, RawBson, RawDocumentBuf},
     bson_compat::cstr,
+    error::Error,
     options::{AuthOptions, ClientOptions},
     sdam::topology::TopologySpec,
 };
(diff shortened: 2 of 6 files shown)


src/client/options.rs
```diff
@@ -656,6 +656,9 @@ pub(crate) struct TestOptions {
 
     /// Async-capable command event listener.
     pub(crate) async_event_listener: Option<TestEventSender>,
+
+    /// Callback to receive hello commands.
+    pub(crate) hello_cb: Option<EventHandler<crate::cmap::Command>>,
```
abr-egn, line 661: The prose tests required direct visibility of hello commands, so new test hook.  I used `EventHandler` for this because that's easier than typing `Arc<Box<dyn Fn(crate::cmap::Command)>>` all over the place and it already implements things like `Debug`.

src/cmap/establish.rs
```diff
@@ -48,12 +48,12 @@ pub(crate) struct EstablisherOptions {
     pub(crate) test_patch_reply: Option<fn(&mut Result<HelloReply>)>,
 }
 
-impl From<&ClientOptions> for EstablisherOptions {
-    fn from(opts: &ClientOptions) -> Self {
+impl From<&TopologySpec> for EstablisherOptions {
```
abr-egn, line 53: Because `Establisher` / `Handshaker` now require a value outside of `ClientOptions` to construct, I introduced `TopologySpec` to capture that (and anything similar that happens to be needed in the future) rather than thread through another function parameter.

src/cmap/establish/handshake/test.rs
abr-egn, line 1: These were just construction tests of metadata values and IMO didn't have enough utility to be worth updating.

src/cmap/establish/handshake.rs
```diff
@@ -38,6 +38,48 @@ pub(crate) struct ClientMetadata {
     pub(crate) os: OsMetadata,
     pub(crate) platform: String,
     pub(crate) env: Option<RuntimeEnvironment>,
+    pub(crate) appended: HashSet<DriverInfo>,
+}
+
+impl ClientMetadata {
+    pub(crate) fn append(&mut self, driver_info: DriverInfo) {
```
isabelatkinson: The spec has the following requirement:

> All strings provided as part of the driver info MUST NOT contain the delimiter used for metadata concatention. Drivers MUST throw an error if any of these strings contains that character.

So we should be checking for "|" in these strings - I think we're also missing that for metadata included in `ClientOptions`.

src/sdam/topology.rs
```diff
@@ -1020,16 +940,60 @@ impl TopologyWatcher {
...
+    pub(crate) fn logical_session_timeout(&self) -> Option<Duration> {
+        self.peek_latest().description.logical_session_timeout
+    }
+
+    /// Get the topology's currently highest seen cluster time.
+    pub(crate) fn cluster_time(&self) -> Option<ClusterTime> {
+        self.peek_latest().description.cluster_time().cloned()
+    }
+
+    /// Gets the latest information on whether transactions are support or not.
```
isabelatkinson:
```suggestion
    /// Gets the latest information on whether transactions are supported or not.
```

isabelatkinson (commented):
just FYI, I was confused about the ambiguity that comes with omitting unspecified fields, but looks like there's an open DRIVERS ticket to address this blocked on changes from analytics: https://jira.mongodb.org/browse/DRIVERS-3251

src/cmap/establish/handshake.rs
abr-egn: Good catch, added.

PR #1474, opened 2025-09-22
RUST-1405 Expose full server response in `Error`

isabelatkinson:
RUST-1405

### DIFF
--- a/src/trace.rs
+++ b/src/trace.rs
@@ -1,5 +1,7 @@
+#[cfg(feature = "bson-3")]
+use crate::bson_compat::RawDocumentBufExt;
 use crate::{
-    bson::{Bson, Document},
+    bson::Bson,
     client::options::{ServerAddress, DEFAULT_PORT},
 };
 
@@ -38,7 +40,7 @@ impl crate::error::Error {
             self.source
         );
         if let Some(server_response) = self.server_response() {
-            let server_response_string = match Document::try_from(server_response) {
+            let server_response_string = match server_response.to_document() {
                 Ok(document) => serialize_command_or_reply(document, max_document_length),
                 Err(_) => {
                     let mut hex_string = hex::encode(server_response.as_bytes());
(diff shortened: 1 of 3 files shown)


src/error.rs
```diff
@@ -52,20 +53,32 @@ pub type Result<T> = std::result::Result<T, Error>;
...
     pub kind: Box<ErrorKind>,
+
     labels: HashSet<String>,
+
     pub(crate) wire_version: Option<i32>,
+
     #[source]
     pub(crate) source: Option<Box<Error>>,
+
+    pub(crate) server_response: Option<Box<Document>>,
```
abr-egn: I'd lean towards storing and exposing this as a `RawDocumentBuf` - this feels like a "lower level" API where the wire representation is more appropriate, and also it allows inspecting the bytes in case the server sends a response that doesn't fully parse.

src/trace.rs
```diff
@@ -29,11 +29,21 @@ impl TracingRepresentation for crate::bson::oid::ObjectId {
     }
 }
 
-impl TracingRepresentation for crate::error::Error {
-    type Representation = String;
-
-    fn tracing_representation(&self) -> String {
-        self.to_string()
+impl crate::error::Error {
+    fn tracing_representation(&self, max_document_length: usize) -> String {
```
abr-egn, line 35: Having to feature-conditionally thread this through everywhere is a bit unpleasant; what would you think about attaching `max_document_length` as a field of `Error` instead?

src/error.rs
isabelatkinson: Agreed - I updated this to be a raw document. I did leave the tracing representation to be a parsed document for readability in logs (https://github.com/mongodb/specifications/blob/master/source/logging/logging.md#representing-documents-in-log-messages) with a hex string of the byte vec as a fallback.

src/trace.rs
isabelatkinson: I prefer this approach as it'll accommodate any future additions of fields needing truncation for the conection and SDAM emitters, and it feels consistent with other fields for the command emitter - happy to make the change if you feel strongly in the other direction, though

PR #1478, opened 2025-09-24
Remove unused dependencies

FalkWoldmann:


isabelatkinson (commented):
Sorry for the back-and-forth here: we pull in the `tokio-util` dependency to generate the docs as well, so `tokio-util` needs to be a non-dev dependency. Can you move it back to the `dependencies` list? I think we can omit the `io` feature while we're at it; we're only using `compat` in the documentation.

isabelatkinson (approved):
lgtm!

PR #1479, opened 2025-09-26
RUST-2237 Support for text indexes in explicit encryption

isabelatkinson:
RUST-2237

I added an unstable feature flag for this in case the API changes when texts indexes are GA.

### DIFF
--- a/.evergreen/run-csfle-tests.sh
+++ b/.evergreen/run-csfle-tests.sh
@@ -10,7 +10,7 @@ set -o xtrace
 
 export CSFLE_TLS_CERT_DIR="${DRIVERS_TOOLS}/.evergreen/x509gen"
 
-FEATURE_FLAGS+=("in-use-encryption" "azure-kms")
+FEATURE_FLAGS+=("in-use-encryption" "azure-kms", "text-indexes-unstable")
 CARGO_OPTIONS+=("--ignore-default-filter")
 
 if [[ "$OPENSSL" = true ]]; then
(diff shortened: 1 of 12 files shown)


src/action/csfle/encrypt.rs
```diff
@@ -150,6 +158,94 @@ pub struct RangeOptions {
...
+/// used in experimental workloads.
+#[skip_serializing_none]
+#[derive(Clone, Default, Debug, Serialize, TypedBuilder)]
+#[serde(rename_all = "camelCase")]
+#[builder(field_defaults(default, setter(into)))]
+#[non_exhaustive]
+#[cfg(feature = "text-indexes-unstable")]
+pub struct TextOptions {
+    /// Options for substring queries.
+    substring: Option<SubstringOptions>,
```
abr-egn: Any particular reason fields in the new structs aren't `pub`?

.evergreen/run-csfle-tests.sh
```diff
@@ -10,7 +10,7 @@ set -o xtrace
 
 export CSFLE_TLS_CERT_DIR="${DRIVERS_TOOLS}/.evergreen/x509gen"
 
-FEATURE_FLAGS+=("in-use-encryption" "azure-kms")
+FEATURE_FLAGS+=("in-use-encryption" "azure-kms", "text-indexes-unstable")
```
abr-egn:
```suggestion
FEATURE_FLAGS+=("in-use-encryption" "azure-kms" "text-indexes-unstable")
```

PR #1481, opened 2025-09-29
RUST-2127 lint for sign loss when casting

abr-egn:
RUST-2127

This enables the `clippy::cast_sign_loss` lint and fixes all the locations it flagged.  These fell into four fuzzy categories:
* surrounding checks ensure it's infallible, but the Rust compiler can't prove it: these are now `try_into().unwrap()` so if the checks miss something it'll cause a panic rather than silent misbehavior;
* conversion between float and integer: there's now an explicit function for that (that just calls lint-disabled `as`) to make it clear that's what's intended;
* assuming a technically-signed value from the server is always actually unsigned: this is now checked at parse time and handled as an unsigned value everywhere beyond that;
* test assertions, which now also use `try_into().unwrap()`

### DIFF
--- a/src/bson_util.rs
+++ b/src/bson_util.rs
@@ -62,6 +62,29 @@ pub(crate) fn get_int_raw(val: RawBsonRef<'_>) -> Option<i64> {
     }
 }
 
+#[allow(private_bounds)]
+pub(crate) fn round_clamp<T: RoundClampTarget>(input: f64) -> T {
+    T::round_clamp(input)
+}
+
+trait RoundClampTarget {
+    fn round_clamp(input: f64) -> Self;
+}
+
+impl RoundClampTarget for u64 {
+    #[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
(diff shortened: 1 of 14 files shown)


src/bson_util.rs
```diff
@@ -62,14 +62,39 @@ pub(crate) fn get_int_raw(val: RawBsonRef<'_>) -> Option<i64> {
     }
 }
 
+#[allow(private_bounds)]
+pub(crate) fn round_clamp<T: RoundClampTarget>(input: f64) -> T {
```
abr-egn, line 66: Doing it this way allows the type to be specified inline with the call rather than the awkwardness that `Into` and friends have.

src/bson_util.rs
```diff
@@ -291,6 +316,31 @@ impl RawDocumentCollection for RawArrayBuf {
     }
 }
 
+pub(crate) mod option_u64_as_i64 {
```
abr-egn, line 319: This could be just `bson::serde_helpers::u64::AsI64` but we can't rely on bson 3.0 stuff :(

PR #1484, opened 2025-09-29
RUST-2226 Add a prose test for OIDC reauthentication when a session is involved

Julia-Garland:
[link](https://github.com/mongodb/specifications/blob/master/source/auth/tests/mongodb-oidc.md#45-reauthentication-succeeds-when-a-session-is-involved)


abr-egn (approved):
Thank you!

PR #1485, opened 2025-09-30
RUST-2113 Add has_next to cursor API

### LINKED ISSUE #1258: cursor.hasNext() is not pub
I would like to use the cursor.hasNext() method documented here: 

https://www.mongodb.com/docs/upcoming/reference/method/cursor.hasNext/

It seems like this is not a public method on Cursor objects. Can this be made pub?

https://github.com/mongodb/mongo-rust-driver/blob/0ac9fdfca571322dd49bc0b70e4f2637951e5a10/src/cursor.rs#L152

Julia-Garland:
Resolves #1258

- Made `has_next()` pub + added doc comments
- Added `has_next()` to sync cursor API
- Added simple test

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -142,24 +142,6 @@ impl<T> Cursor<T> {
             .and_then(|c| c.post_batch_resume_token())
     }
 
-    /// Whether this cursor has exhausted all of its getMore calls. The cursor may have more
-    /// items remaining in the buffer.
-    pub(crate) fn is_exhausted(&self) -> bool {
-        self.wrapped_cursor.as_ref().unwrap().is_exhausted()
-    }
-
-    /// Whether this cursor has any additional items to return.
-    pub(crate) fn has_next(&self) -> bool {
-        !self.is_exhausted()
-            || !self
-                .wrapped_cursor
-                .as_ref()
-                .unwrap()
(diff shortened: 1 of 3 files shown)


src/cursor.rs
```diff
@@ -240,6 +222,26 @@ impl<T> Cursor<T> {
...
+    /// Whether this cursor has exhausted all of its getMore calls. The cursor may have more
+    /// items remaining in the buffer.
+    pub(crate) fn is_exhausted(&self) -> bool {
+        self.wrapped_cursor.as_ref().unwrap().is_exhausted()
+    }
+
+    /// Returns true if the cursor has any additional buffered items to return and false otherwise.
+    ///
+    /// This method only checks the local buffer and exhaustion state; it does not contact the
+    /// server.
```
isabelatkinson:
```suggestion
    /// Returns true if the cursor has any additional items to return and false otherwise.
```

The call to `is_exhausted` checks whether the server has more batches based on its last `getMore` response, so the "buffered" distinction isn't necessary here.

(ditto for the doc on the sync method)

isabelatkinson (commented):
just a doc suggestion! otherwise looks good

isabelatkinson (approved):
lgtm!

PR #1486, opened 2025-10-01
RUST-1996 SCRAM-SHA-256 FIPS Compliance

Julia-Garland:
### Motivation

The prior implementation of PBKDF2 key derivation uses the `pbkdf2` library which is just a pure Rust implementation and not FIPS-compliant.

### Changes

Switch to the `aws-lc-rs` version of `pbkdf2` which can be configured to be FIPS-compliant.


Julia-Garland:
Reason for closing: using a SHA256 HMAC with PBKDF2 is enough to ensure FIPS _compliance_, and the driver already does this. Figured its not worth adding a whole new dependency (which requires cmake and go installed) just to ensure FIPS _certification_.

PR #1494, opened 2025-10-09
RUST-1308 Implement Display and Serialize for ConnectionString

Julia-Garland:
### Summary

- Makes `ConnectionString` serializable.
- Supports `ConnectionString::to_string()` by implementing Display (added an internal helper `to_uri_str`)

### DIFF
--- a/src/client/options.rs
+++ b/src/client/options.rs
@@ -1713,6 +1713,45 @@ impl ConnectionString {
 
     /// Un-parses a [`ConnectionString`] struct back into a MongoDB connection string.
     fn to_uri_str(&self) -> String {
+        let ConnectionString {
+            host_info,
+            app_name,
+            tls,
+            heartbeat_frequency,
+            local_threshold,
+            read_concern,
+            replica_set,
+            write_concern,
+            server_selection_timeout,
+            max_pool_size,
+            min_pool_size,
+            max_connecting,
+            max_idle_time,
+            #[cfg(any(
+                feature = "zstd-compression",
+                feature = "zlib-compression",
+                feature = "snappy-compression"
+            ))]
+            compressors,
+            connect_timeout,
+            retry_reads,
+            retry_writes,
+            server_monitoring_mode: _,
+            direct_connection,
+            credential,
+            default_database,
+            load_balanced,
+            socket_timeout,
+            read_preference,
+            uuid_representation,
+            srv_max_hosts,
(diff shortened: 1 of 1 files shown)


src/client/options.rs
```diff
@@ -784,10 +784,36 @@ impl Serialize for ClientOptions {
     }
 }
 
+// Copy of UuidRepresentation for serialization
+#[non_exhaustive]
+#[derive(Serialize)]
+#[serde(remote = "UuidRepresentation")]
+enum UuidRepresentationForSerialize {
```
abr-egn: Since this is only used for `serialize_uuid_rep_option`, it can go inside that fn definition (common pattern for seder helper types).

src/client/options.rs
```diff
@@ -806,11 +832,13 @@ pub struct ConnectionString {
     /// The TLS configuration for the Client to use in its connections with the server.
     ///
     /// By default, TLS is disabled.
+    #[serde(flatten, serialize_with = "Tls::serialize")]
```
abr-egn: Is there a reason to flatten fields here?

src/client/options.rs
```diff
@@ -1021,25 +1060,22 @@ impl From<TlsOptions> for Option<Tls> {
...
-            }
+            Some(Tls::Enabled(tls_options)) => TlsOptions::serialize(tls_options, serializer),
             _ => serializer.serialize_none(),
         }
     }
 }
 
 /// Specifies the TLS configuration that the [`Client`](../struct.Client.html) should use.
-#[derive(Clone, Debug, Default, Deserialize, PartialEq, TypedBuilder)]
+#[derive(Clone, Debug, Default, Deserialize, Serialize, PartialEq, TypedBuilder)]
```
abr-egn: Having both a `Serialize` impl and a `serialize` fn is a bit confusing here.

src/client/options.rs
```diff
@@ -1673,6 +1712,316 @@ impl ConnectionString {
         Ok(conn_str)
     }
 
+    /// Un-parses a [`ConnectionString`] struct back into a MongoDB connection string.
+    fn to_uri_str(&self) -> String {
```
abr-egn, line 1715: I'd recommend doing a destructuring assignment here listing all the fields to ensure that they're all used (or intentionally omitted) and the compiler will complain if fields are added without updating this method.

PR #1495, opened 2025-10-10
RUST-2138 Add optional support for OpenTelemetry

abr-egn:
RUST-2138

The bulk of the added code is in the new `crate::otel` submodule; beyond that it's mostly some additional tracking in the executor and new fields for operations.

### DIFF
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -6,7 +6,7 @@ set -o pipefail
 source .evergreen/env.sh
 source .evergreen/cargo-test.sh
 
-FEATURE_FLAGS+=("tracing-unstable" "cert-key-password" "opentelemetry")
+FEATURE_FLAGS+=("tracing-unstable" "cert-key-password" "opentelemetry" "error-backtrace")
 
 if [ "$OPENSSL" = true ]; then
   FEATURE_FLAGS+=("openssl-tls")


src/bson_util.rs
abr-egn, line 1: The utility functions here just got moved out of `tracing` since they're needed for otel as well.

src/client/executor.rs
```diff
@@ -496,9 +518,21 @@ impl Client {
             let should_redact = cmd.should_redact();
             let cmd_name = cmd.name.clone();
             let target_db = cmd.target_db.clone();
+            #[cfg(feature = "opentelemetry")]
+            let cmd_attrs = crate::otel::CommandAttributes::new(&cmd);
```
abr-egn, line 529: `otel::CommandAttributes` exists because the opentelemetry trace needs values from both the `Command` struct and the `Message` struct, and constructing the latter consumes the former so they can't just both be passed in directly.

src/client/session.rs
```diff
@@ -142,6 +152,14 @@ impl Transaction {
         self.options = None;
         self.pinned = None;
         self.recovery_token = None;
+        self.drop_span();
+    }
+
+    pub(crate) fn drop_span(&mut self) {
```
abr-egn, line 158: I'm annoyed that this is needed but `commit` / `abort` are called _before_ the relevant operation is executed and that op needs to be included in the span.

src/client/session/action.rs
```diff
@@ -79,7 +79,11 @@ impl ClientSession {
                 }
 
                 self.increment_txn_number();
-                self.transaction.start(options);
+                self.transaction.start(
+                    options,
+                    #[cfg(feature = "opentelemetry")]
+                    self.client.start_transaction_span(),
```
abr-egn, line 85: I was hoping to keep span construction internal to `Transaction` but it doesn't have access to the `Client` (specifically, the client's `tracer`) so it has to happen here and get passed in.

src/lib.rs
```diff
@@ -44,6 +44,10 @@ mod hello;
 pub(crate) mod id_set;
 mod index;
 mod operation;
+#[cfg(feature = "opentelemetry")]
+pub mod otel;
+#[cfg(not(feature = "opentelemetry"))]
+mod otel_stub;
```
abr-egn, line 51: Some of the executor opentelemetry stuff (future method chaining mostly) would be really awkward and noisy to keep behind `#[cfg(feature = "opentelemetry")]`; the `otel_stub` module exists to sidestep that.

src/otel.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// Maximum length of the `db.query.text` attribute of command spans.  If unset, will use the
+    /// value of the `OTEL_RUST_INSTRUMENTATION_MONGODB_QUERY_TEXT_MAX_LENGTH` environment
+    /// variable.
+    pub query_text_max_length: Option<usize>,
+    /// Tracer provider to use.  If unset, will use the global instance.
+    #[serde(skip)]
+    #[derive_where(skip)]
+    #[builder(
+        setter(
+            fn transform<S, T, P>(provider: P) -> Option<Arc<dyn ObjectSafeTracerProvider + Send + Sync>>
```
abr-egn, line 47: This is what necessitated the bump on `typed_builder` - older versions don't support setters with constraints.

src/test/spec/open_telemetry.rs
```diff
@@ -0,0 +1,21 @@
+use crate::test::spec::unified_runner::run_unified_tests;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn run_unified_operation() {
+    // TODO:
```
abr-egn, line 5: These and the skip reasons are going to turn into a follow-up PR for the driver otel spec&tests.

src/operation.rs
```diff
@@ -176,6 +176,14 @@ pub(crate) trait Operation {
 
     /// The name of the server side command associated with this operation.
     fn name(&self) -> &CStr;
+
+    #[cfg(feature = "opentelemetry")]
+    type Otel: crate::otel::OtelWitness<Op = Self>;
```
abr-egn, line 181: Motivating problem: I didn't want to just splat the new data extraction methods needed for otel into `Operation` - that's already got a ton of methods, it would be very noisy with the compiler conditional on each one, and it would mean otel-specific updates would always be touching core operation code.

_Ideally_, those methods would (a) be in their own trait and (b) that trait would be a supertrait of `Operation` when otel is enabled.  (a) is easy enough but (b) is ugly - it would mean either duplicating the definition of `Operation` based on whether the feature is enabled or not, or making a no-op `OperationSuper` trait that itself gets duplicated.  Either way it's a lot of noise and wouldn't scale well if we end up needing other optional traits on `Operation`.

So, I took a trick from my old Haskell days and used a type witness.  That lets us prove to the compiler that any _particular_ `&impl Operation` value can be converted into a `&impl OtelInfo` value, and the casts themselves are no-ops so it all compiles away.

src/otel.rs
```diff
@@ -0,0 +1,410 @@
...
+        }
+    }
+}
+
+pub(crate) trait OtelWitness {
+    type Op: OtelInfo;
+    fn otel(op: &Self::Op) -> &impl OtelInfo {
+        op
+    }
+    fn output_cursor_id(output: &<Self::Op as Operation>::O) -> Option<i64> {
```
abr-egn, line 351: In theory this shouldn't be needed but the compiler would lose the associated type constraints somewhere, so this propagates them.

src/client/executor.rs
```diff
@@ -106,7 +110,25 @@ impl Client {
...
     ) -> Result<ExecutionDetails<T>> {
-        // Validate inputs that can be checked before server selection and connection checkout.
+        let session = session.into();
+        let ctx = self.start_operation_span(op, session.as_deref());
+        let result = self
+            .execute_operation_with_details_inner(op, session)
+            .with_context(ctx.clone())
+            .await;
+        #[cfg(feature = "opentelemetry")]
+        self.record_error(&ctx, &result);
```
isabelatkinson: Will this record a duplicate error if an error is also recorded from the call to `record_command_result` on line 673?

src/otel.rs
```diff
@@ -0,0 +1,410 @@
...
+        if !self.options().otel_enabled() {
+            return;
+        }
+        if let Err(error) = result {
+            let span = context.span();
+            span.set_attributes([
+                KeyValue::new("exception.message", error.to_string()),
+                KeyValue::new("exception.type", error.kind.name()),
+                #[cfg(test)]
+                KeyValue::new("exception.stacktrace", error.bt.to_string()),
```
isabelatkinson: WDYT about adding a feature flag for this rather than keeping it behind `cfg(test)`?

src/otel.rs
```diff
@@ -0,0 +1,410 @@
...
+#[cfg(test)]
+pub(crate) mod testing;
+
+/// Configuration for OpenTelemetry.
+#[derive(Clone, serde::Deserialize, typed_builder::TypedBuilder)]
+#[derive_where(Debug, PartialEq)]
+#[builder(field_defaults(default, setter(into)))]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct Options {
```
isabelatkinson: nit: I'd lean towards `OpentelemetryOptions` or `TracingOptions` here for consistency with other options structs; the generated docs also look a little odd with this name:

Image

abr-egn: Done.

src/client/executor.rs
abr-egn: Nope - this will record an error on the operation span, line 673 is recording for the command span.  I had to go back and look to confirm this, so I refactored the code to make the different types of spans in use have different concrete types to hopefully make this easier for future reading.

src/otel.rs
abr-egn: Seems reasonable to me; done.

isabelatkinson (dismissed):

isabelatkinson (commented):
whoops, hit approve before looking at CI

isabelatkinson (approved):

isabelatkinson (approved):

PR #1502, opened 2025-10-22
minor: fix gssapi tests

abr-egn:
The root cause here is that the test domain was switched from an actual DNS entry to a line in `/etc/hosts`; the latter means it has no CNAME record, so the lookup for forward-only canonicalization would error.  _However_, the [spec says](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#gssapi):

> The driver MUST fallback to the provided host if any lookup errors or returns no results.

so this fixes the behavior to do that rather than propagate up the error.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -261,7 +261,7 @@ buildvariants:
     display_name: "GSSAPI Authentication - Linux"
     patchable: true
     run_on:
-      - ubuntu2204-small
+      - rhel87-small
     tasks:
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
abr-egn, line 1: These changes didn't actually fix it; I'm leaving them in because it means these tests run on the same hosts as the rest of our standard tests, which seems like a vaguely good thing.

PR #1517, opened 2025-11-01
Make chrono optional dependency

tottoto:
Makes `chrono` optional dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ dns-resolver = ["dep:hickory-resolver", "dep:hickory-proto"]
 cert-key-password = ["dep:pem", "dep:pkcs8"]
 
 # Enable support for MONGODB-AWS authentication.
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -217,6 +217,7 @@ features = ["v4"]
 anyhow = { version = "1.0", features = ["backtrace"] }
 approx = "0.5.1"
 backtrace = { version = "0.3.68" }
+chrono = { version = "0.4.32", default-features = false }
```
isabelatkinson:
```suggestion
chrono = { version = "0.4.32", default-features = false, features = ["std"] }
```

One of our tests needs this feature enabled

isabelatkinson (changes requested):

Cargo.toml
tottoto: Fixed.

PR #1520, opened 2025-11-06
Update to pbkdf2 0.12

tottoto:
Updates to `pbkdf2` 0.12.


isabelatkinson (approved):
LGTM! (For future reference, [internal docs](https://docs.rs/pbkdf2/latest/src/pbkdf2/lib.rs.html#174) indicate that the result being unwrapped should never be an error.)

PR #1549, opened 2025-12-02
RUST-2281 Cache test binary for cross-topololgy tests

abr-egn:
RUST-2281

This converts all of these to using cached binaries shared across the server topology tasks:
* the main test suite
* openssl
* stable API
* in use encryption
* load balancer
* graviton

A lot of yaks to shave here:
* Archiving test runner binaries turns out to be nontrivial (rustc hash, dependent libraries, etc.).  Fortunately, `cargo nextest` has support for this.
* Tests needed a little bit of an update to pick up paths at runtime rather than compile-time since even on the same host tag different tasks will have different paths.
* Speaking of host tags, we were using an outdated one for rhel8.
* Some tests (e.g. compression) use the same tasks as the main suites, so I had to use some trickery to make the archived runner portion optional.
* `cargo nextest` doesn't support doctests (they're apparently built in a different and hard to manage way); we were running those alongside the main tests but that doesn't work if the run is using an archive, so I've split that out into its own lint task.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -54,7 +54,7 @@ buildvariants:
   - name: compile
     display_name: "Compile Only"
     run_on:
-      - rhel87-small
+      - rhel8-latest-small
     tasks:
       - name: compile-on-latest
       - name: compile-on-msrv
@@ -62,25 +62,28 @@ buildvariants:
   - name: lint
     display_name: "Lint"
     run_on:
-      - rhel87-small
+      - rhel8-latest-small
+    expansions:
+      LIBMONGOCRYPT_OS: rhel-80-64-bit
     tasks:
       - name: .lint
 
   - name: cargo-deny
     display_name: "Cargo Deny"
     run_on:
-      - rhel87-small
+      - rhel8-latest-small
     tasks:
       - name: check-cargo-deny
 
   - name: rhel-8
     display_name: "RHEL 8"
     run_on:
-      - rhel87-small
+      - rhel8-latest-small
     expansions:
       AUTH: auth
       SSL: ssl
     tasks:
+      - name: build-nextest-archive
       - name: .standalone
       - name: .replicaset
       - name: .sharded
@@ -93,6 +96,7 @@ buildvariants:
       AUTH: auth
       SSL: ssl
     tasks:
+      - name: build-nextest-archive
       # Ubuntu 22.04 does not support MongoDB versions below 6.0.
       - name: .standalone !.4.2 !.4.4 !.5.0
       - name: .replicaset !.4.2 !.4.4 !.5.0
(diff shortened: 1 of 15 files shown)


.evergreen/config.yml
```diff
@@ -802,27 +822,82 @@ tasks:
 
   - name: check-cargo-deny
     commands:
-      - func: "check cargo deny"
+      - command: shell.exec
```
abr-egn: These were tasks calling a function that itself was a single command and not used anywhere else; I got annoyed tracking down the indirections and collapsed them.  I can revert this portion if you don't think it's an improvement :)

.evergreen/config.yml
```diff
@@ -1529,6 +1662,12 @@ functions:
     - command: expansions.update
       params:
         file: src/uri-expansions.yml
+    # The server needs some time after startup otherwise it won't include things like
```
abr-egn, line 1709: Turns out the time of the compilation step between mongo orchestration and running the tests was masking this behavior; took me a while to figure out why the relevant tests were failing.

.evergreen/generate-tasks/src/main.rs
```diff
@@ -19,8 +19,13 @@ tasks:"
                 "
   - name: test-{version}-{top_name}
     tags: [{version}, {top_name}]
+    depends_on:
+      - name: build-nextest-archive
+        patch_optional: true
```
abr-egn, line 24: Despite the name, what this _actually_ means is "if the named task doesn't exist in this buildvariant, don't actually depend on it".

driver/src/change_stream/session.rs
```diff
@@ -83,7 +83,7 @@ where
     /// # let mut session = client.start_session().await?;
     /// let mut cs = coll.watch().session(&mut session).await?;
     /// while let Some(event) = cs.next(&mut session).await? {
-    ///     let id = bson::to_bson(&event.id)?;
+    ///     let id = bson::serialize_to_bson(&event.id)?;
```
abr-egn, line 86: These failed because the new doctest target is running against bson3.  I figure it's better to have the newer version be the one documented.

.evergreen/install-dependencies.sh
```diff
@@ -32,8 +32,7 @@ for arg; do
 
     # This file is not created by default on Windows
     echo 'export PATH="$PATH:${CARGO_HOME}/bin"' >>${CARGO_HOME}/env
-    echo "export CARGO_NET_GIT_FETCH_WITH_CLI=true" >>${CARGO_HOME}/env
```
abr-egn, line 35: This was a fix for a years-old issue with git authentication on macos that's no longer a problem.  Removed it because I was hitting very hard to diagnose issues with cargo not being able to locate the git binary on windows, likely something to do with cygwin path manipulation.

.evergreen/config.yml
isabelatkinson: Yeah I like this change - can these be subprocess.exec, though?

abr-egn: Done :)

isabelatkinson (approved):

PR #1550, opened 2025-12-02
RUST-1054 SOCKS5 Proxy Support

isabelatkinson:
RUST-1054

### DIFF
--- a/driver/src/client/options.rs
+++ b/driver/src/client/options.rs
@@ -422,6 +422,7 @@ pub struct ServerApi {
 #[non_exhaustive]
 pub struct Socks5Proxy {
     /// The hostname or IP address on which the proxy is listening.
+    #[builder(!default)]
     pub host: String,
 
     /// The port on which the proxy is listening. Defaults to 1080 if unset.


driver/src/runtime/stream.rs
```diff
@@ -33,40 +36,129 @@ pub(crate) enum AsyncStream {
     Tcp(TcpStream),
 
     /// A TLS connection over TCP.
-    Tls(TlsStream),
+    Tls(TlsStream<TcpStream>),
```
isabelatkinson, line 39: I used concrete types in the TLS variants rather than a unified one with a generic bound because doing so would propagate generics to every type that uses a connection.

driver/src/client/options.rs
```diff
@@ -407,6 +415,59 @@ pub struct ServerApi {
...
+    pub port: Option<u16>,
+
+    /// A username/password pair to authenticate to the proxy.
+    pub authentication: Option<(String, String)>,
+}
+
+/// Dummy struct for internal use.
+#[cfg(not(feature = "socks5-proxy"))]
+#[derive(Clone, Debug)]
+pub(crate) struct Socks5Proxy;
```
abr-egn, line 438: I'm guessing this is to avoid having to annotate every single place that just passes one of these around with `#[cfg(feature = "socks5-proxy")]` - handy.  Going to remember this technique.

abr-egn (approved):

PR #1558, opened 2025-12-15
Add 'indexmap_2' feature to serde_with dependency

richarddd:
Many projects have moved to `indexmap2` causing `serde_with` to pull in duplicate deps


richarddd:
> The typical approach for this sort of situation is for crates that use `mongodb` and `serde_with` to include their own `[dependencies]` entry for `serde_with` including that feature, and the dependency resolution [feature unification](https://doc.rust-lang.org/cargo/reference/resolver.html#features) will propagate it. Does that not work in this case?

Thats a valid workaround, but shouldn't the the default behavior be what benefits most users? I.E use `indexmap2` so explicit feature flags per transitive dependency can be avoided for most cases?

PR #1560, opened 2025-12-17
RUST-2295 Bevy plugin to read assets from mongodb

abr-egn:
RUST-2295

This new crate provides a data backend (in bevy parlance, an [`AssetSource`](https://docs.rs/bevy/latest/bevy/asset/io/struct.AssetSource.html)) for loading assets from either BSON documents or GridFS.

Potential areas for future improvement:
* Provide a Bevy [`Asset`](https://docs.rs/bevy/latest/bevy/asset/trait.Asset.html) data format plugin for BSON documents themselves.  This could hypothetically replace the current hardcoded expected document structure for asset loading with one based on the Bevy [`AssetTransformer`](https://docs.rs/bevy/latest/bevy/asset/transformer/trait.AssetTransformer.html) setup that would provide substantially more flexibility.
* Bridge the GridFS streaming and Bevy asset streaming.  In the current implementation, the asset is loaded entirely by the plugin worker thread and then passed as a buffer back to Bevy; this means that things like loading bars would only see "0%" and "100%" states.
* More ergonomic construction; right now there's a little bit of a dance required to construct the plugin in the same Tokio runtime as the mongodb `Client`.  It seems possible that this could be handled automatically in the common case and leave the manual construction for when something unusual is required.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -467,6 +467,7 @@ buildvariants:
 
   - name: bevy
     display_name: "Bevy"
+    patchable: false
     run_on: [rhel8-latest-small]
     tasks:
       - name: test-bevy-asset


bevy/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name = "bevy_mongodb_asset"
+version = "0.1.0"
+authors = ["Abraham Egnor <abraham.egnor@mongodb.com>"]
+license = "Apache-2.0"
+repository = "https://github.com/mongodb/mongo-rust-driver"
+edition = "2024"
+
+[workspace]
```
abr-egn, line 9: This declares that the crate is _not_ part of the shared workspace defined in the parent directory, mostly because bevy pulls in a ton of dependencies and it didn't seem worth it to dump them all in to that `Cargo.lock`.

bevy/README.md
```diff
@@ -0,0 +1,43 @@
...
+    .add_plugins(plugin)
+    ...
+    .run();
+```
+
+## Usage
+
+Assets can be loaded either from BSON documents or from GridFS.
+
+BSON document assets use the path structure `mongodb://document/<database>/<collection>/<name>`, and fetch data from a BSON object with the shape:
```
abr-egn, line 28: I originally intended for this to be `mongodb://document/<database>/<collection>/<objectid>`, but it turns out Bevy uses the extensions on asset paths to autodetect the asset format.  It's possible to sidestep that but it's not the typical path and requires a lot of ceremony.  This scheme makes the name arbitrary so it can have whatever extension is appropriate to the data, at the cost of allowing duplicates and the resulting odd behavior.

bevy/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+Bevy metadata is loaded from an object (if found) with the same shape and an additional `meta: true` entry.
+
+GridFS assets use the path structure `mongodb://gridfs/<database>/<bucket>/<name>`.  Bevy metadata is loaded from a GridFS file (if found) of the same name with an additional `.meta` extension.
+
+## Bevy Compatibility
+
+|Bevy|bevy_mongodb_asset|
+|---|---|
+|0.17|0.1|
```
abr-egn, line 43: I was also planning to keep the version in sync with the compatible Bevy version, but that turns out to be less common than I thought, so I'm following convention here with providing a table.

bevy/src/reader.rs
```diff
@@ -0,0 +1,244 @@
...
+
+use bevy::asset::io::{AssetReaderError, PathStream, Reader, VecReader};
+use mongodb::{
+    bson::doc,
+    error::{ErrorKind as MdbErrorKind, GridFsErrorKind},
+    options::GridFsBucketOptions,
+};
+use tokio::sync::{mpsc, oneshot};
+
+pub(crate) struct Worker(mpsc::Sender<AssetMessage>);
```
abr-egn, line 14: The underlying implementation spins up a worker task and communicates with the Bevy side via message-passing.  While the Bevy `AssetReader` trait is async, it uses `smol` as the runtime so we can't just plunk the code in directly.  `smol` does provide compatibility wrappers but those just spin up a new Tokio runtime under the hood, so that wouldn't work either (the driver needs all operations to be kept in the same runtime).

bevy/src/reader.rs
```diff
@@ -0,0 +1,244 @@
...
+
+impl Worker {
+    pub(crate) async fn new(client: &mongodb::Client) -> Self {
+        let (tx, rx) = mpsc::channel::<AssetMessage>(1);
+        let client = client.clone();
+        tokio::spawn(async move {
+            let mut rx = rx;
+            while let Some(message) = rx.recv().await {
+                let client = client.clone();
+                tokio::spawn(async move {
```
abr-egn, line 24: Having all asset load requests be serialized via the channel is a potential bottleneck, but since the worker on the other end is just taking the requests and tossing the handling into its own task for each request I suspect it'll be completely irrelevant next to actual asset load time.

bevy/src/reader.rs
```diff
@@ -0,0 +1,244 @@
...
+    async fn read_meta<'a>(&'a self, path: &'a Path) -> Result<impl Reader + 'a, AssetReaderError> {
+        self.send_request(AssetRequest {
+            path: path.to_owned(),
+            is_meta: true,
+        })
+        .await
+    }
+
+    async fn is_directory<'a>(&'a self, _path: &'a Path) -> Result<bool, AssetReaderError> {
+        Ok(false)
```
abr-egn, line 68: We could hypothetically support "directories" by doing a find, but the utility of that is unclear to me.

bevy/src/reader.rs
```diff
@@ -0,0 +1,244 @@
...
+    is_meta: bool,
+}
+
+impl AssetRequest {
+    async fn process(self, client: mongodb::Client) -> AssetResponse {
+        let asset_path =
+            AssetPath::parse(&self.path).map_err(|e| AssetReaderError::Io(Arc::new(e)))?;
+        match asset_path {
+            AssetPath::Document { namespace, name } => {
+                self.process_document(client, namespace, name).await
```
abr-egn, line 91: My kingdom for enum subtypes.  It looks like the current active proposal for this kind of thing is [pattern types](https://gist.github.com/joboet/0cecbce925ee2ad1ee3e5520cec81e30), which has (IMO) fallen into the trap of trying to be a very general solution where enum subtypes would have been a much narrower and easier sell.

bevy/README.md
```diff
@@ -0,0 +1,43 @@
...
+BSON document assets use the path structure `mongodb://document/<database>/<collection>/<name>`, and fetch data from a BSON object with the shape:
+```
+{
+    name: <String>,
+    data: <Binary>
+}
+```
+Bevy metadata is loaded from an object (if found) with the same shape and an additional `meta: true` entry.
+
+GridFS assets use the path structure `mongodb://gridfs/<database>/<bucket>/<name>`.  Bevy metadata is loaded from a GridFS file (if found) of the same name with an additional `.meta` extension.
```
kevinAlbs: Curious: would returning the GridFS [metadata](https://github.com/mongodb/specifications/blob/192976b194afdb1f458cbba2530c73de6b2c700f/source/gridfs/gridfs-spec.md#definitions) field be appropriate? As I understand it, we can choose the convention.

bevy/README.md
```diff
@@ -0,0 +1,43 @@
...
+Assets can be loaded either from BSON documents or from GridFS.
+
+BSON document assets use the path structure `mongodb://document/<database>/<collection>/<name>`, and fetch data from a BSON object with the shape:
+```
+{
+    name: <String>,
+    data: <Binary>
+}
+```
+Bevy metadata is loaded from an object (if found) with the same shape and an additional `meta: true` entry.
```
kevinAlbs: I expect querying with the same name and `meta: true` means the `name` field could not have a unique index.

Could this instead query a different collection `<collection>.meta` for the same name?

bevy/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+                    bytes: PNM_IMAGE_DATA.to_owned(),
+                },
+            };
+
+            let coll = client
+                .database("bevy_test")
+                .collection::<mongodb::bson::RawDocumentBuf>("doc_images");
+            coll.drop().await.unwrap();
+
+            coll.insert_one(doc).await.unwrap();
```
kevinAlbs: Could this test, and the gridfs test, also test inserting and loading metadata?

.evergreen/config.yml
```diff
@@ -465,6 +465,12 @@ buildvariants:
       #- name: .8.0
       #- name: .7.0
 
+  - name: bevy
```
isabelatkinson, line 468: Suggest making this `patchable: false` before merging

bevy/README.md
abr-egn: Oh hey, there's a gridfs metadata field.  Done!

bevy/README.md
abr-egn: Indexing hadn't occurred to me, that's a good point.  Done.

bevy/src/lib.rs
abr-egn: Done!  As a side effect, I moved the tests out to their own file since there's a now a substantial amount there (and it makes it easier to find if people are looking for examples).

bevy/src/tests.rs
```diff
@@ -0,0 +1,240 @@
...
+                subtype: BinarySubtype::Generic,
+                bytes: PNM_IMAGE_DATA.to_owned(),
+            },
+        };
+        let coll = db.collection::<RawDocumentBuf>("doc_images2");
+        coll.drop().await.unwrap();
+        coll.insert_one(doc).await.unwrap();
+
+        let meta =
+            bevy::asset::meta::AssetMeta::<(), ()>::new(bevy::asset::meta::AssetAction::Ignore);
```
abr-egn, line 83: The metadata is used for the asset processing stuff; the easiest way that I could find to test that it's actually being loaded properly is to use it to flag a file as ignored and then check for the corresponding error.

bevy/src/tests.rs
```diff
@@ -0,0 +1,240 @@
...
+        };
+        let meta_coll = db.collection::<RawDocumentBuf>("doc_images2.meta");
+        meta_coll.drop().await.unwrap();
+        meta_coll.insert_one(meta_doc).await.unwrap();
+
+        MongodbAssetPlugin::new(&client).await
+    });
+
+    let result = test_load_image(plugin, "mongodb://document/bevy_test/doc_images2/pixel.pbm");
+    let Err(ImageLoadFailure::LoadFailed(inner)) = result else {
```
abr-egn, line 99: `inner` here is an `Arc<AssetLoadError>` so we can't directly pattern match on the whole thing :(

bevy/src/tests.rs
```diff
@@ -0,0 +1,240 @@
...
+    static MAX_FRAMES: u32 = 60 * 60 * 5;
+    #[derive(Resource)]
+    struct TestImage(Handle<Image>);
+
+    let load_images = move |mut commands: Commands, asset_server: Res<AssetServer>| {
+        let handle = asset_server.load::<Image>(path);
+        commands.insert_resource(TestImage(handle));
+    };
+
+    let failure = Arc::new(std::sync::OnceLock::<ImageLoadFailure>::new());
```
abr-egn, line 194: The Bevy `App` interface is pretty deeply assumed to be the outer loop of a process so it only allows numeric failure codes.  This little dance is smuggling out a structured error.

bevy/src/tests.rs
```diff
@@ -0,0 +1,240 @@
...
+        system::{Commands, Res},
+    },
+    image::Image,
+};
+use mongodb::{
+    bson::{RawDocumentBuf, doc, rawdoc, spec::BinarySubtype},
+    options::GridFsBucketOptions,
+};
+
+const MONGODB_URI: LazyCell<String> = LazyCell::new(|| std::env::var("MONGODB_URI").unwrap());
```
kevinAlbs: Minor:

```suggestion
const MONGODB_URI: LazyCell<String> = LazyCell::new(|| std::env::var("MONGODB_URI").expect("MONGODB_URI environment variable"));
```

To indicate error cause. Before:
```
called `Result::unwrap()` on an `Err` value: NotPresent
```

After:
```
MONGODB_URI environment variable: NotPresent
```

kevinAlbs (dismissed):

isabelatkinson (dismissed):
lgtm!

PR #1573, opened 2026-01-06
RUST-2300 Provide a raw batched cursor mode for all cursor-returning actions

abr-egn:
RUST-2300

This required changing the internals of cursor to defer some response parsing to allow the zero-copy path to be possible; the operation, executor, and action paths have been updated so there doesn't need to be (as much) code duplication across the various axes of cursor type.

There's some obvious followup work to be done - I've filed RUST-2324 and RUST-2325 for that.

### DIFF
--- a/driver/src/action/list_collections.rs
+++ b/driver/src/action/list_collections.rs
@@ -48,7 +48,7 @@ impl Database {
     ///
     /// `await` will return d[`Result<Vec<String>>`].
     #[deeplink]
-    #[options_doc(list_collections)]
+    #[options_doc(list_collection_names)]
     pub fn list_collection_names(&self) -> ListCollections<'_, ListNames> {
         ListCollections {
             db: self,
@@ -75,7 +75,7 @@ impl crate::sync::Database {
     ///
     /// [`run`](ListCollections::run) will return d[`Result<Vec<String>>`].
     #[deeplink]
-    #[options_doc(list_collections, "run")]
+    #[options_doc(list_collection_names, "run")]
     pub fn list_collection_names(&self) -> ListCollections<'_, ListNames> {
         self.async_database.list_collection_names()
     }
@@ -92,8 +92,8 @@ pub struct ListCollections<'a, M = ListSpecifications, S = ImplicitSession> {
 }
 
 #[option_setters(crate::db::options::ListCollectionsOptions)]
-#[export_doc(list_collections, extra = [session, batch])]
-impl<M, S> ListCollections<'_, M, S> {}
+#[export_doc(list_collection_names, extra = [session])]
+impl<S> ListCollections<'_, ListNames, S> {}
 
 impl<'a, M> ListCollections<'a, M, ImplicitSession> {
     /// Use the provided session when running the operation.
@@ -110,6 +110,8 @@ impl<'a, M> ListCollections<'a, M, ImplicitSession> {
     }
 }
 
+#[option_setters(crate::db::options::ListCollectionsOptions)]
+#[export_doc(list_collections, extra = [session, batch])]
 impl<'a, S: ActionSession<'a>> ListCollections<'a, ListSpecifications, S> {
     async fn exec_generic<C: crate::cursor::NewCursor>(self) -> Result<C> {
         let mut list_collections =
--- a/driver/src/action/list_indexes.rs
+++ b/driver/src/action/list_indexes.rs
@@ -1,6 +1,6 @@
 use std::{marker::PhantomData, time::Duration};
 
-use crate::bson::Bson;
+use crate::{action::ActionSession, bson::Bson};
 use futures_util::stream::TryStreamExt;
 
 use crate::{
@@ -50,7 +50,7 @@ where
     ///
     /// `await` will return d[`Result<Vec<String>>`].
     #[deeplink]
-    #[options_doc(list_indexes)]
+    #[options_doc(list_index_names)]
     pub fn list_index_names(&self) -> ListIndexes<'_, ListNames> {
         ListIndexes {
             coll: CollRef::new(self),
@@ -80,7 +80,7 @@ where
     ///
     /// [`run`](ListIndexes::run) will return d[`Result<Vec<String>>`].
     #[deeplink]
-    #[options_doc(list_indexes, "run")]
+    #[options_doc(list_index_names, "run")]
     pub fn list_index_names(&self) -> ListIndexes<'_, ListNames> {
         self.async_collection.list_index_names()
     }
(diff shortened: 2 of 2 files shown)


driver/src/action/aggregate.rs
```diff
@@ -167,6 +167,24 @@ impl<'a, Session, T> Aggregate<'a, Session, T> {
     }
 }
 
+macro_rules! agg_exec_generic {
```
abr-egn, line 170: I couldn't get the lifetimes to compile with this as a proper method, and I'm frustrated to say that I have no idea why _this_ particular impl had issues when the others don't.  Putting the body in a macro sidesteps that while still meaning the code is technically not duplicated (for the human reader, at least).

driver/src/client/executor.rs
```diff
@@ -657,21 +645,28 @@ impl Client {
                         effective_criteria: effective_criteria.clone(),
                     };
 
-                    match op.handle_response(&response, context).await {
-                        Ok(response) => Ok(response),
+                    let handle_result = match op.wants_owned_response() {
+                        true => op.handle_response(Cow::Owned(response), context).await,
```
abr-egn, line 649: Cost of the zero-copy path for those operations that flag it: errors won't have the full server response (because that would require copying it).  We could add a "zero-copy _except_ for errors" mode if this turns out to actually be an issue, but I suspect it won't come up.

driver/src/lib.rs
```diff
@@ -77,6 +77,7 @@ pub use crate::{
     client::{session::ClientSession, Client},
     coll::Collection,
     cursor::{
+        raw_batch as raw_batch_cursor,
```
abr-egn, line 80: This exposes the new cursor types as `crate::raw_batch::[Session]RawBatchCursor`.  I declare shed-painting season open 🙂

driver/src/options.rs
```diff
@@ -112,8 +112,6 @@ macro_rules! resolve_rw_concern_with_session {
         } else {
             resolve_options!($obj, $opts, [$concern]);
         }
-        let result: Result<()> = Ok(());
```
abr-egn, line 115: This is unrelated but I noticed these macros were set up to return a result for no reason that I could tell, so I removed that.

driver/src/cursor/raw_batch.rs
```diff
@@ -0,0 +1,509 @@
...
+            .and_then(RawBsonRef::as_document)
+            .ok_or_else(|| Error::invalid_response("missing cursor subdocument"))?;
+
+        let docs = cursor
+            .get("firstBatch")?
+            .or_else(|| cursor.get("nextBatch").ok().flatten())
+            .ok_or_else(|| Error::invalid_response("missing firstBatch/nextBatch"))?;
+
+        docs.as_array()
+            .ok_or_else(|| Error::invalid_response("missing firstBatch/nextBatch"))
```
isabelatkinson:
```suggestion
            .ok_or_else(|| Error::invalid_response("invalid firstBatch/nextBatch"))
```

driver/src/action/list_collections.rs
isabelatkinson:
```suggestion
#[export_doc(list_collections, extra = [session, batch])]
```

driver/src/action/list_indexes.rs
isabelatkinson, line 1: Any reason this didn't get a `batch` method?

driver/src/action/aggregate.rs
isabelatkinson:
```suggestion
#[export_doc(aggregate, extra = [session, batch])]
```

driver/src/cursor/raw_batch.rs
```diff
@@ -0,0 +1,509 @@
...
+    ClientSession,
+};
+
+use super::common::CursorInformation;
+
+/// A raw batch response returned by the server for a cursor getMore/find.
+///
+/// This provides zero-copy access to the server's batch array via
+/// [`doc_slices`](RawBatch::doc_slices).
+#[derive(Debug)]
```
isabelatkinson:
```suggestion
#[derive(Clone, Debug)]
```

driver/src/client/executor.rs
isabelatkinson: IIUC this will omit the raw reply when response handling fails for all ops that specify `ZERO_COPY=true`, regardless of whether we're working with raw batches. Was this intended as an overall perf improvement, or for implementation reasons? I'm guessing this is a less common path for error cases than an ok:0 response so it's probably not a huge deal, but I was curious about the motivation.

driver/src/lib.rs
isabelatkinson: SGTM! I would _maybe_ put the types directly into `crate` to align with the other cursor types, but a separate module also makes sense given the more niche use case.

driver/src/operation.rs
```diff
@@ -215,19 +222,31 @@ pub(crate) trait OperationWithDefaults: Send + Sync {
...
+            message: format!("response handling not implemented for {}", Self::NAME),
         }
         .into())
     }
 
+    /// Interprets the server response taking ownership of the body to enable zero-copy handling.
+    ///
+    /// Default behavior delegates to the borrowed [`handle_response`]; operations that set
+    /// [`ZERO_COPY`] to `true` should override this to consume the response.
+    fn handle_response_cow<'a>(
```
isabelatkinson, line 234: Is this additional method necessary? This has the same signature as `handle_response` and the executor passes either owned or borrowed data based on the val for `ZERO_COPY`, so I'm not sure why `handle_response` can't be used for all cases.

abr-egn: `Operation` just has the single `handle_response` method; `OperationWithDefaults` has three, forming a chain: `handle_response` (sync, ref), `handle_response_cow` (sync, `Cow`), `handle_response_async` (async, `Cow`).  The main utility of the extra method is the same as for the other two: only those few implementors that need the extra complication have to deal with it, all the rest can remain unchanged.

driver/src/client/executor.rs
abr-egn: I considered having this be a runtime switch rather than a constant and only set it true for raw batch mode, but didn't for a few reasons:
* It's an extra bit to carry around on those operations and make sure is set to the right thing at the right time
* a constant means the compiler can optimize this branch away

I'm not thrilled that this has user-visible implications but the other alternative was preemptively copying the response in case it's needed by the error, which had a measurable (~10%) performance impact.

This is also a decision that can be changed in the future if it turns out that the lack of full error response is a problem.

driver/src/action/list_indexes.rs
abr-egn: Whoops, nope, just an oversight.  Fixed.

PR #1576, opened 2026-01-08
RUST-2310 Deprioritize servers in server selection for all topologies

isabelatkinson:
RUST-2310

### DIFF
--- a/driver/src/client.rs
+++ b/driver/src/client.rs
@@ -471,7 +471,7 @@ impl Client {
         criteria: Option<&SelectionCriteria>,
     ) -> Result<ServerAddress> {
         let (server, _) = self
-            .select_server(criteria, "Test select server", Vec::new(), |_, _| None)
+            .select_server(criteria, "Test select server", &[], |_, _| None)
             .await?;
         Ok(server.address.clone())
     }
@@ -483,7 +483,7 @@ impl Client {
         criteria: Option<&SelectionCriteria>,
         #[allow(unused_variables)] // we only use the operation_name for tracing.
         operation_name: &str,
-        deprioritized: Vec<&ServerAddress>,
+        deprioritized: &[&ServerAddress],
         override_criteria: OverrideCriteriaFn,
     ) -> Result<(SelectedServer, SelectionCriteria)> {
         let criteria =
@@ -526,7 +526,7 @@ impl Client {
                 effective_criteria,
                 &state.description,
                 &state.servers(),
-                &deprioritized,
+                deprioritized,
             );
             match result {
                 Err(error) => {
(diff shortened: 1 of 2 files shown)


driver/src/sdam/description/topology/server_selection.rs
isabelatkinson, line 1: A lot of the changes in this file are just shuffling things around; the functional changes are based on [this diff](https://github.com/mongodb/specifications/pull/1865/changes#diff-fe3d136605671ac7f5a201d1359e5a5406043b88df4e8c837134b8de51a283d6) from the spec PR.

driver/src/sdam/description/topology/server_selection.rs
```diff
@@ -59,34 +59,36 @@ pub(crate) fn attempt_to_select_server<'a>(
     criteria: &'a SelectionCriteria,
     topology_description: &'a TopologyDescription,
     servers: &'a HashMap<ServerAddress, Arc<Server>>,
-    deprioritized: Option<&ServerAddress>,
+    deprioritized: &[&ServerAddress],
```
isabelatkinson, line 62: Made this a list in anticipation of CSOT and because the tests use lists of multiple addresses.

driver/src/sdam/description/topology/server_selection.rs
```diff
@@ -104,57 +106,115 @@ impl TopologyDescription {
...
-
-    pub(crate) fn has_available_servers(&self) -> bool {
-        self.servers.values().any(|server| server.is_available())
+        }
     }
 
-    fn suitable_servers(
+    fn filter_servers_in_replica_set<'a>(
         &self,
+        servers: impl Iterator<Item = &'a ServerDescription> + Clone,
```
abr-egn, line 164: Why not just `&[&'a ServerDescription]`?

driver/src/client.rs
```diff
@@ -483,7 +483,7 @@ impl Client {
         criteria: Option<&SelectionCriteria>,
         #[allow(unused_variables)] // we only use the operation_name for tracing.
         operation_name: &str,
-        deprioritized: Option<&ServerAddress>,
+        deprioritized: Vec<&ServerAddress>,
```
abr-egn: Likewise, it doesn't seem like this needs to be owned?

abr-egn (approved):
LGTM!  Some very small questions, no need for a re-review.

driver/src/sdam/description/topology/server_selection.rs
isabelatkinson: `filter_servers_by_selection_criteria` (which calls this) already has an iterator with the deprioritized servers filtered out, so accepting that directly allows us to avoid an intermediary structure

PR #1582, opened 2026-01-12
RUST-2321 Deprecate the socketTimeoutMS option

abr-egn:


driver/src/client/options/parse.rs
```diff
@@ -12,6 +12,12 @@ impl Action for ParseConnectionString {
 
     async fn execute(self) -> Result<ClientOptions> {
         let mut conn_str = self.conn_str?;
+        #[allow(deprecated)]
```
abr-egn, line 15: I didn't want to error in the `ConnectionString` parsing because that could hypothetically be used independent of the driver, but I also think the driver shouldn't accept a URI option it will silently ignore.

PR #1584, opened 2026-01-13
RUST-2267 Avoid clearing connection pool when server connection rate limiter triggers

isabelatkinson:
RUST-2267

### DIFF
--- a/driver/src/test/spec/sdam.rs
+++ b/driver/src/test/spec/sdam.rs
@@ -370,7 +370,6 @@ async fn connection_pool_backpressure() {
             _ => None,
         });
         if !pool_cleared_events.is_empty() {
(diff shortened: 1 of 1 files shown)


driver/src/sdam/description/topology/test/sdam.rs
```diff
@@ -708,75 +710,3 @@ async fn direct_connection() {
         .await
         .expect("write should succeed with directConnection unspecified");
 }
-
-#[tokio::test(flavor = "multi_thread")]
-async fn pool_cleared_error_does_not_mark_unknown() {
```
isabelatkinson, line 713: The updated error handling logic broke this test. it's not a prose test defined in any specs, so I just removed it.

PR #1588, opened 2026-01-14
RUST-2268 Add jitter to convenient transaction retries

isabelatkinson:
Unfortunately, we can't wrap the entire sync convenient transaction in `block_on`. If the closure provided to `and_run` contains calls that use `block_on` (i.e. any driver operation), then tokio returns an error because `block_on` is called from within `block_on`.

Fortunately, I was able to figure out a way to pass `await` to the macro - it's a little weird but minimally invasive.

### DIFF
--- a/driver/src/test/spec/transactions.rs
+++ b/driver/src/test/spec/transactions.rs
@@ -291,9 +291,12 @@ async fn convenient_api_retry_backoff_is_enforced() {
             options.hosts.drain(1..);
         }
         let client = Client::for_test().options(options).await;
-        let coll = client
-            .database("db")
-            .collection("convenient_retry_backoff_is_enforced");
+
+        let coll_name = "convenient_retry_backoff_is_enforced";
+        let db = client.database("db");
+        let coll = db.collection(coll_name);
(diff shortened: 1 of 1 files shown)


driver/src/client/session/action.rs
```diff
@@ -164,6 +175,54 @@ macro_rules! convenient_run {
...
+    ) => {{
+        const BACKOFF_INITIAL_MS: f64 = 5f64;
+        const BACKOFF_MAX_MS: f64 = 500f64;
+        const RETRY_BASE: f64 = 1.5f64;
+
+        let jitter = rand::random_range(0f64..1f64);
+        #[cfg(test)]
+        let jitter = $session.convenient_transaction_jitter.unwrap_or(jitter);
+
+        let computed_backoff = jitter * BACKOFF_INITIAL_MS * RETRY_BASE.powf(f64::from($retries));
```
isabelatkinson: Our `Checked` type doesn't work for these operations, but if something goes awry mathematically we'll still end up with a reasonable value on lines 200-201 so I didn't think it was worth trying to catch underflow/overflow.

PR #1594, opened 2026-01-23
RUST-2341 Consolidate operation feature inheritance

abr-egn:
RUST-2341

This PR takes the notion of "operation target" that already existed in a few place, merges them, and then uses _that_ to shift the implementation of inheriting selection criteria/read concern/write concern from macros called in action implementations to the executor path itself.  The result is cleaner code on both sides of the path, less mutability, and plenty of room for future inheritable features if the need comes up :)

This does come with the cost of operations carrying around a little bit more data and needing to clone it occasionally, but since the targets are all `Arc`-wrapped both storage and cloning are cheap.

### DIFF
--- a/driver/src/otel.rs
+++ b/driver/src/otel.rs
@@ -22,7 +22,6 @@ use crate::{
     options::{ClientOptions, ServerAddress, DEFAULT_PORT},
     Client,
     ClientSession,
-    Namespace,
 };
 
 #[cfg(test)]
@@ -293,25 +292,6 @@ fn record_error<T>(context: &Context, result: &Result<T>) {
     });
 }
 
-impl OperationTarget {
-    fn name(&self) -> TargetName<'_> {
-        match self {
-            OperationTarget::Database(db) => TargetName {
-                database: db.name(),
-                collection: None,
-            },
-            OperationTarget::Collection(coll) => TargetName {
-                database: coll.db().name(),
-                collection: Some(coll.name()),
-            },
-            OperationTarget::Namespace(ns) => TargetName {
-                database: &ns.db,
-                collection: Some(&ns.coll),
-            },
-        }
-    }
-}
-
 fn op_target(op: &impl OtelInfo) -> String {
     let target = op.target();
     let name = target.name();
@@ -423,20 +403,21 @@ pub(crate) struct TargetName<'a> {
     pub(crate) collection: Option<&'a str>,
 }
 
-impl<'a> From<&'a str> for TargetName<'a> {
-    fn from(value: &'a str) -> Self {
-        TargetName {
-            database: value,
-            collection: None,
-        }
-    }
-}
-
-impl<'a> From<&'a Namespace> for TargetName<'a> {
-    fn from(value: &'a Namespace) -> Self {
-        TargetName {
-            database: &value.db,
-            collection: Some(&value.coll),
+impl OperationTarget {
+    fn name(&self) -> TargetName<'_> {
+        match self {
+            OperationTarget::Database(db) => TargetName {
+                database: db.name(),
+                collection: None,
+            },
+            OperationTarget::Collection(coll) => TargetName {
+                database: coll.db().name(),
+                collection: Some(coll.name()),
+            },
+            OperationTarget::Namespace(ns) => TargetName {
+                database: &ns.db,
+                collection: Some(&ns.coll),
+            },
         }
     }
 }


driver/src/action/aggregate.rs
```diff
@@ -262,38 +248,19 @@ enum AggregateTargetRef<'a> {
 }
 
 impl AggregateTargetRef<'_> {
```
abr-egn, line 250: Hypothetically, `AggregateTargetRef` could also be replaced with `OperationTarget`, but the benefits are unclear and it would introduce some corner cases since not all operation targets are valid aggregate targets.

driver/src/client/options/bulk_write.rs
```diff
@@ -54,7 +54,7 @@ pub struct BulkWriteOptions {
     pub let_vars: Option<Document>,
 
     /// The write concern to use for this operation.
-    #[serde(skip_serializing_if = "write_concern_is_empty")]
+    #[serde(skip_serializing)]
```
abr-egn, line 57: This is the one aspect I'm not entirely happy with; operation options need to not serialize the write concern so it can be handled by `Command`.  This was already true for read concerns, so I don't think this is making the situation _worse_, but it still feels like an easy footgun to fire.

driver/src/client/executor.rs
```diff
@@ -697,14 +738,20 @@ impl Client {
         let stream_description = connection.stream_description()?;
         let is_sharded = stream_description.initial_server_type == ServerType::Mongos;
         let mut cmd = op.build(stream_description)?;
+        // Clear inherited read/writeconcern when in a transaction
```
abr-egn, line 741: Technically the more correct thing to do would be not set them in the first place, but that would require plumbing the session down into `Command` construction and that didn't seem reasonable.

driver/src/cmap/conn/command.rs
```diff
@@ -52,11 +51,24 @@ pub(crate) struct Command {
...
 
+fn write_concern_is_empty(write_concern: &Option<WriteConcern>) -> bool {
+    write_concern
+        .as_ref()
+        .is_none_or(|write_concern| write_concern.is_empty())
+}
+
 impl Command {
-    pub(crate) fn new(name: impl ToString, target_db: impl ToString, body: RawDocumentBuf) -> Self {
+    pub(crate) fn new_raw(
```
abr-egn, line 67: I renamed this to make the distinction between this and `from_operation` more apparent, and to discourage this from being picked out of routine in the future if the other one should be used instead.

driver/src/operation/aggregate/change_stream.rs
```diff
@@ -39,7 +39,7 @@ impl ChangeStreamAggregate {
     }
 }
 
-impl OperationWithDefaults for ChangeStreamAggregate {
+impl Operation for ChangeStreamAggregate {
```
abr-egn, line 42: Since this is wrapping another operation, it needs to impl `Operation` to make sure the passthrough works properly; I don't know why I didn't update this back when we did the split to create `OperationWithDefaults`.

Debugging this made me genuinely suspect a compiler bug for a bit ("why it it calling the _wrong_ write_concern?!?").

driver/src/operation/aggregate/change_stream.rs
```diff
@@ -83,68 +83,99 @@ impl OperationWithDefaults for ChangeStreamAggregate {
...
-
-        let description = context.connection.stream_description()?;
-        if self.args.options.as_ref().is_none_or(has_no_time)
-            && description.max_wire_version.is_some_and(|v| v >= 7)
-            && spec.is_empty
-            && spec.post_batch_resume_token.is_none()
-        {
-            data.initial_operation_time = op_time;
-        }
+    ) -> crate::BoxFuture<'a, Result<Self::O>> {
```
abr-egn, line 90: There's no change here, just wrapping the body with `async move { }`

driver/src/operation/count.rs
```diff
@@ -1,26 +1,28 @@
...
     error::{Error, Result},
     operation::{OperationWithDefaults, Retryability},
-    selection_criteria::SelectionCriteria,
 };
 
 use super::{append_options_to_raw_document, ExecutionContext};
 
 pub(crate) struct Count {
-    ns: Namespace,
+    target: Collection<Document>,
```
abr-egn, line 16: For those operations that target a collection, I just sidestepped the type issue by using `Document` since the value type doesn't matter here.

driver/src/client.rs
```diff
@@ -741,3 +739,28 @@ impl Drop for Client {
         }
     }
 }
+
+/// Operation-specific parameters to server selection
+pub(crate) struct OpSelectionInfo<'a> {
```
abr-egn, line 744: This originated in an earlier approach to the larger problem; it seems neutral at worst on its own so I left it in, but I'd be happy to revert it if you think that'd be better.

driver/src/operation.rs
```diff
@@ -151,16 +154,14 @@ pub(crate) trait Operation {
     fn handle_error(&self, error: Error) -> Result<Self::O>;
 
     /// Criteria to use for selecting the server that this operation will be executed on.
-    fn selection_criteria(&self) -> Option<&SelectionCriteria>;
+    fn selection_criteria(&self) -> Feature<&SelectionCriteria>;
 
-    /// Whether or not this operation will request acknowledgment from the server.
-    fn is_acknowledged(&self) -> bool;
```
abr-egn, line 157: This was literally never implemented for any operation; everything used the default logic.  I dropped that in favor of moving the method to the write concern.

driver/src/operation.rs
```diff
@@ -189,9 +193,95 @@ pub(crate) trait Operation {
     }
 }
 
+#[derive(Debug, Copy, Clone)]
+pub(crate) enum Feature<T> {
```
abr-egn, line 197: `Option<T>` didn't quite work for these, since there needs to be a way to distinguish between the `None` meaning "unsupported for this operation" and the `None` meaning "no value set".

driver/src/operation.rs
```diff
@@ -180,6 +181,9 @@ pub(crate) trait Operation {
     /// The name of the server side command associated with this operation.
     fn name(&self) -> &CStr;
 
+    /// The noun to this operation's verb.
```
isabelatkinson, line 184: 😁

PR #1604, opened 2026-02-03
RUST-2351 skip BSON size limits

kevinAlbs:
Skip encryption BSON size limits when testing mongocryptd >= 7.0 due to SERVER-118428.

Patch build: https://spruce.mongodb.com/version/6982607367710400072179d6

Removing the skip is tracked in RUST-2351.

### DIFF
--- a/driver/src/test/csfle/prose.rs
+++ b/driver/src/test/csfle/prose.rs
@@ -122,7 +122,9 @@ async fn custom_key_material() -> Result<()> {
 #[tokio::test]
 async fn bson_size_limits() -> Result<()> {
     if *DISABLE_CRYPT_SHARED && server_version_gte(7, 0).await {
-        log_uncaptured("skipping BSON size limits test on mongocryptd due to SERVER-118428: server >= 7.0");
+        log_uncaptured(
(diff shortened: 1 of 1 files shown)


driver/src/test/csfle/prose.rs
```diff
@@ -121,6 +121,11 @@ async fn custom_key_material() -> Result<()> {
 // Prose test 4. BSON Size Limits and Batch Splitting
 #[tokio::test]
 async fn bson_size_limits() -> Result<()> {
+    if *DISABLE_CRYPT_SHARED && server_version_gte(7, 0).await {
+        log_uncaptured("skipping BSON size limits test on mongocryptd due to SERVER-118428: server >= 7.0");
```
abr-egn:
```suggestion
        log_uncaptured(
            "skipping BSON size limits test on mongocryptd due to SERVER-118428: server >= 7.0",
        );
```

abr-egn (approved):

driver/src/test/csfle/prose.rs
kevinAlbs: Oop. Thank you for the lint fix.

PR #1620, opened 2026-02-24
RUST-2324 Implement `Cursor` on top of `RawBatchCursor`

abr-egn:
RUST-2324

This uses `RawBatchCursor` to handle the underlying commands and batch fetches, `Cursor` to break up those batches into buffers and iterate over individual items, and `ChangeStream` to handle resume tokens and automatic resume.  The code in each layer is (subjectively) simpler and easier to understand in isolation, and (objectively) has much less duplication of functionality.

### DIFF
--- a/driver/src/change_stream.rs
+++ b/driver/src/change_stream.rs
@@ -1,4 +1,5 @@
 //! Contains the functionality for change streams.
+pub(crate) mod common;
 pub mod event;
 pub(crate) mod options;
 pub mod session;
@@ -6,32 +7,20 @@ pub mod session;
 #[cfg(test)]
 use std::collections::VecDeque;
 use std::{
-    future::Future,
     pin::Pin,
     task::{Context, Poll},
 };
 
-#[cfg(test)]
-use crate::bson::RawDocumentBuf;
-use crate::{
-    bson::{Document, Timestamp},
-    operation::OperationTarget,
-};
+use crate::error::Error;
 use derive_where::derive_where;
 use futures_core::{future::BoxFuture, Stream};
+use futures_util::FutureExt;
 use serde::de::DeserializeOwned;
 #[cfg(test)]
 use tokio::sync::oneshot;
 
-#[cfg(feature = "bson-3")]
-use crate::bson_compat::RawBsonRefExt as _;
-use crate::{
-    change_stream::event::{ChangeStreamEvent, ResumeToken},
-    cursor::{stream_poll_next, BatchValue, CursorStream, NextInBatchFuture},
-    error::{ErrorKind, Result},
-    ClientSession,
-    Cursor,
-};
+use crate::{change_stream::event::ResumeToken, error::Result, Cursor};
+use common::{ChangeStreamData, WatchArgs};
 
 /// A `ChangeStream` streams the ongoing changes of its associated collection, database or
 /// deployment. `ChangeStream` instances should be created with method `watch` against the relevant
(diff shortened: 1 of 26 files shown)


driver/src/change_stream/common.rs
```diff
@@ -0,0 +1,142 @@
...
+            initial_operation_time: self.initial_operation_time,
+            resume_token: self.resume_token.clone(),
+            document_returned: self.document_returned,
+            implicit_session: self.implicit_session.take(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(super) struct CursorWrapper<Inner> {
```
abr-egn, line 61: Everything before this is just moved from the parent module.

driver/src/cursor/stream.rs
```diff
@@ -0,0 +1,221 @@
...
+                                Ok(true) => Some(ar.buffer.deserialize_current()),
+                            };
+                            self.state = StreamState::Idle(ar.buffer);
+                            return Poll::Ready(out);
+                        }
+                    }
+                }
+                StreamState::Polling => {
+                    return Poll::Ready(Some(Err(Error::internal(
+                        "attempt to poll cursor already in polling state",
```
abr-egn, line 220: This setup is used in a couple of different places here (a state enum with accessors that can error).  This means the code can be substantially simpler, at the cost of not being able to prove a valid state by construction.  In practice, the invalid states are only possible if a future is dropped without being polled to completion, and we already document that as not something we support.

driver/src/test/client.rs
```diff
@@ -861,6 +861,8 @@ async fn find_one_and_delete_serde_consistency() {
         .database("find_one_and_delete_serde_consistency")
         .collection("test");
 
+    coll.drop().await.unwrap();
```
abr-egn, line 864: Leftover state from previous runs caused some extremely baffling failures without this.

driver/src/change_stream.rs
```diff
@@ -83,31 +72,16 @@ pub struct ChangeStream<T>
...
-    /// Arguments to `watch` that created this change stream.
-    args: WatchArgs,
-
-    /// Dynamic information associated with this change stream.
-    data: ChangeStreamData,
-
-    /// A pending future for a resume.
-    #[derive_where(skip)]
-    pending_resume: Option<BoxFuture<'static, Result<ChangeStream<T>>>>,
+    inner: StreamState<T>,
```
abr-egn, line 75: This wrapping layer isn't strictly needed, but it'll be useful to have the distinct type in case we want to support `SessionChangeStream`, er, `Streams` in the future.

driver/src/change_stream.rs
```diff
@@ -161,164 +132,149 @@ where
...
 
-/// Dynamic change stream data needed for resume.
-#[derive(Debug, Default)]
-pub(crate) struct ChangeStreamData {
-    /// The `operationTime` returned by the initial `aggregate` command.
-    pub(crate) initial_operation_time: Option<Timestamp>,
-
-    /// The cached resume token.
-    pub(crate) resume_token: Option<ResumeToken>,
+type CursorWrapper = common::CursorWrapper<Cursor<()>>;
```
abr-egn, line 165: Because `ChangeStream`s only access the values via `current()`, there's no need for the inner `Cursor` to carry a type.

driver/src/event/cmap.rs
```diff
@@ -176,6 +176,8 @@ pub struct ConnectionClosedEvent {
     #[serde(skip)]
     #[derive_where(skip)]
     pub(crate) error: Option<crate::error::Error>,
+
+    pub(crate) service_id: Option<ObjectId>,
```
abr-egn, line 180: This was debugging for an issue that turned out to be local test environment weirdness, but it seems useful to preserve.

driver/src/cursor/stream.rs
isabelatkinson: makes sense, can you add comment explaining that here for future reference?

driver/src/cursor/raw_batch.rs
isabelatkinson, line 136: Very small nit: I was briefly confused reading references to this field because `initial_reply` implied to me that it stored the server's initial reply to the cursor-returning command. Can we call this something like `buffered_reply` instead to make clear that it can also store a `getMore` reply?

PR #1626, opened 2026-03-03
RUST-2273 Exponential backoff and jitter in retry loops

isabelatkinson:
RUST-2273

### DIFF
--- a/driver/src/client.rs
+++ b/driver/src/client.rs
@@ -57,7 +57,7 @@ use crate::{
     TopologyType,
 };
 
-pub(crate) use executor::{HELLO_COMMAND_NAMES, REDACTED_COMMANDS};
+pub(crate) use executor::{retry::Retry, HELLO_COMMAND_NAMES, REDACTED_COMMANDS};
 pub(crate) use session::{ClusterTime, SESSIONS_UNSUPPORTED_COMMANDS};
 
 use session::{ServerSession, ServerSessionPool};
--- a/driver/src/client/executor.rs
+++ b/driver/src/client/executor.rs
@@ -1,31 +1,20 @@
-mod retry;
+pub(super) mod retry;
+
+use std::{
+    borrow::{BorrowMut, Cow},
+    collections::HashSet,
+    sync::{atomic::Ordering, Arc, LazyLock},
+    time::Instant,
+};
+
+#[cfg(feature = "in-use-encryption")]
+use futures_core::future::BoxFuture;
+use serde::de::DeserializeOwned;
 
 #[cfg(feature = "in-use-encryption")]
 use crate::bson::RawDocumentBuf;
 use crate::{
-    bson::{doc, RawBsonRef, RawDocument, Timestamp},
-    cursor::{
-        common::{CursorInformation, CursorSpecification},
-        NewCursor,
-    },
-    error::SYSTEM_OVERLOADED_ERROR,
-    operation::{is_commit_or_abort, Feature, OperationTarget},
-};
-#[cfg(feature = "in-use-encryption")]
-use futures_core::future::BoxFuture;
-use serde::de::DeserializeOwned;
-use std::{borrow::Cow, sync::LazyLock};
-
-use std::{
-    borrow::BorrowMut,
-    collections::HashSet,
-    sync::{atomic::Ordering, Arc},
-    time::Instant,
-};
-
-use super::{session::TransactionState, Client, ClientSession};
-use crate::{
-    bson::Document,
+    bson::{doc, Document, RawBsonRef, RawDocument, Timestamp},
     change_stream::{
         common::{ChangeStreamData, WatchArgs},
         event::ChangeStreamEvent,
@@ -42,12 +31,18 @@ use crate::{
         RawCommandResponse,
         StreamDescription,
     },
-    cursor::{session::SessionCursor, Cursor},
+    cursor::{
+        common::{CursorInformation, CursorSpecification},
+        session::SessionCursor,
+        Cursor,
+        NewCursor,
+    },
     error::{
         Error,
         ErrorKind,
         Result,
         RETRYABLE_WRITE_ERROR,
+        SYSTEM_OVERLOADED_ERROR,
         TRANSIENT_TRANSACTION_ERROR,
         UNKNOWN_TRANSACTION_COMMIT_RESULT,
     },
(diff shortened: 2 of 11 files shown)


driver/src/client/executor/retry.rs
```diff
@@ -0,0 +1,166 @@
...
+        is_transaction_op,
+        txn_number,
+        false,
+    )
+}
+
+/// Handles a failure by either creating a new `Retry` or updating the existing `Retry` with the new
+/// error. If the error cannot be retried, the appropriate error to return from the retry loop is
+/// returned and no further retries should be performed.
+#[allow(clippy::too_many_arguments)]
```
isabelatkinson: This is indeed too many arguments. It could be nice in the future to have some broader context tracker for an operation, but it didn't seem worth adding that to an already-large PR.

driver/src/client/executor.rs
```diff
@@ -102,31 +181,137 @@ impl Client {
...
+        let mut session = ExecutionSession::explicit(session.into());
+        self.execute_operation_with_details(op.borrow_mut(), &mut session)
             .await
             .map(|details| details.output)
     }
 
+    /// Execute the given operation, returning the cursor created by the operation.
+    ///
+    /// Server selection be will performed using the criteria specified on the operation, if any.
+    pub(crate) async fn execute_cursor_operation<Op, C>(
```
isabelatkinson, line 196: I did some reordering of these methods to make the file more readable; the only functional change to the top-level `execute` methods is adding the new session wrapper.

driver/src/client/executor.rs
```diff
@@ -980,6 +1029,27 @@ impl Client {
         }
     }
 
+    pub(crate) fn is_load_balanced(&self) -> bool {
+        self.inner.options.load_balanced.unwrap_or(false)
+    }
+
+    pub(crate) fn pin_connection_for_cursor(
```
isabelatkinson, line 1040: no changes to these, they were just moved

driver/src/test/client.rs
```diff
@@ -619,132 +618,69 @@ async fn x509_auth_skip_ci() {
         .unwrap();
 }
 
-/// Test verifies that retrying a commitTransaction operation after a checkOut
-/// failure works.
+/// Test that commitTransaction succeeds on retry after a checkout failure.
```
isabelatkinson, line 621: This test was relying on incorrect behavior: we were adding RetryableWriteError labels to ok:0 hello responses from the server, which the spec does not say to do. I changed the overall logic of this test while I was here because it's been flaky in the past, so hopefully it's more consistent and easier to understand now.

driver/src/client.rs
```diff
@@ -760,3 +778,52 @@ impl<'a> OpSelectionInfo<'a> {
         }
     }
 }
+
+/// Retry token bucket functionality. Note that the values used are scaled by a factor of 10 from
+/// those defined in the spec to allow for the use of integers.
```
isabelatkinson, line 779: Integers aren't strictly necessary with the design I landed on here, but I think they're more pleasant to work with so I left this as-is.

driver/src/client/executor/retry.rs
```diff
@@ -0,0 +1,166 @@
...
+            let better_error = if (error.is_server_error()
+                || error.is_read_retryable()
+                || error.is_write_retryable())
+                && !error_is_connection_establishment
+                && !error.contains_label(NO_WRITES_PERFORMED)
+            {
+                error
+            } else {
+                retry.last_error
+            };
```
isabelatkinson: Note: I realized that this logic is slightly incorrect compared to the spec while impementing RUST-2335. Will fix and test in the PR for that ticket.

driver/src/client/executor/retry.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        let computed_backoff =
+            jitter * BACKOFF_INITIAL_MS * RETRY_BASE.powf(f64::from(self.attempt - 1));
+        let max_backoff = jitter * BACKOFF_MAX_MS;
+        let backoff: u64 = std::cmp::min(round_clamp(computed_backoff), round_clamp(max_backoff));
+        Duration::from_millis(backoff)
+    }
+}
+
+pub(super) fn handle_connection_establishment_failure<T: Operation>(
```
abr-egn: The `handle_*` functions seem like they'd be a little more readable as constructor functions on `Retry`, i.e.
```
impl Retry {
  pub(super) fn for_connection_establishment_failure<T: Operation>(...) -> Result<Self> { ... }
  .. etc ...
}
```

driver/src/client/executor/retry.rs
```diff
@@ -0,0 +1,166 @@
...
+#[allow(clippy::too_many_arguments)]
+fn handle_failure<T: Operation>(
+    retry: Option<Retry>,
+    error: Error,
+    op: &T,
+    client: &Client,
+    server: ServerAddress,
+    is_transaction_op: bool,
+    txn_number: Option<i64>,
+    error_is_connection_establishment: bool,
```
abr-egn: This might be better as a `FailureKind { ConnectionEstablishment, Execution }` enum.

driver/src/client/executor/retry.rs
```diff
@@ -0,0 +1,166 @@
...
+                    .unwrap_or_default(),
+                last_error: error,
+                txn_number,
+                overloaded,
+            })
+        }
+    }
+}
+
+pub(super) fn return_last_error(retry: &mut Option<Retry>) -> Result<()> {
```
abr-egn, line 166: Likewise, I think these would be a little more readable as methods on `Option<Retry>`.  That would require an extension trait, though, which might not be worth it for only two methods.

driver/src/operation.rs
```diff
@@ -167,10 +174,12 @@ pub(crate) trait Operation {
...
 
     /// The level of retryability the operation supports.
-    fn retryability(&self) -> Retryability;
+    fn retryability(&self, options: &ClientOptions) -> Retryability;
+
+    fn is_backpressure_retryable(&self, options: &ClientOptions) -> bool;
 
     /// Updates this operation as needed for a retry.
-    fn update_for_retry(&mut self);
+    fn update_for_retry(&mut self, overloaded: bool);
```
abr-egn: I wonder if it would be better from a future-proofing / consistency perspective to pass in a `&Retry` instead of just a bool?

PR #1657, opened 2026-04-08
RUST-2275 Update change streams spec test

abr-egn:
RUST-2275

This required adding `#[skip_serializing_none]` to the relevant struct, which we probably should have had there in the first place; otherwise the test fails because `None` gets interpreted as bson `null`, which is different from "field not present".


BorisDog (approved):
LGTM

PR #1658, opened 2026-04-08
RUST-2394 Fix a race condition deadlock in Client::shutdown

abr-egn:
RUST-2394

Credit for both the find and the fix to @jyemin 🙂

### DIFF
--- a/driver/src/sdam/test.rs
+++ b/driver/src/sdam/test.rs
@@ -349,3 +349,45 @@ fn ipv6_invalid_me() {
     };
     assert!(!desc.invalid_me().unwrap());
 }
+
+/// Regression test: Client::shutdown() must not deadlock.
+///
+/// The topology worker's shutdown sequence closes monitors and waits for them to exit.
+/// Meanwhile, a monitor in streaming mode may complete a hello and call
+/// topology_updater.update(), which waits for acknowledgment from the topology worker.
+/// Since the worker has already exited its processing loop, neither side makes progress.
+///
+/// The fix is to drop the topology worker's update_receiver before closing monitors,
+/// so that the monitor's update() call fails immediately.
+#[tokio::test(flavor = "multi_thread")]
+async fn shutdown_does_not_deadlock() {
+    let mut options = get_client_options().await.clone();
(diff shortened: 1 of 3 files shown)


abr-egn:
cargo-deny failure is unrelated and fixed in #1659.

driver/src/sdam/test.rs
```diff
@@ -349,3 +349,45 @@ fn ipv6_invalid_me() {
...
+///
+/// The topology worker's shutdown sequence closes monitors and waits for them to exit.
+/// Meanwhile, a monitor in streaming mode may complete a hello and call
+/// topology_updater.update(), which waits for acknowledgment from the topology worker.
+/// Since the worker has already exited its processing loop, neither side makes progress.
+///
+/// The fix is to drop the topology worker's update_receiver before closing monitors,
+/// so that the monitor's update() call fails immediately.
+#[tokio::test(flavor = "multi_thread")]
+async fn shutdown_does_not_deadlock() {
```
BorisDog, line 363: Is there a way to reproduce this situation deterministically?
Something like:
1. Wait for `update` to be called, and block it
2. Call `shutdown` during a blocked `update`
3. Unblock the `update`

BorisDog (commented):
Looks good, question about the new test.

driver/src/sdam/test.rs
abr-egn: Not simply or cleanly, I think.  The best way I can think of to implement a version that doesn't rely on sleep timing would be to thread a testing [`Barrier`](https://docs.rs/tokio/1.51.0/tokio/sync/struct.Barrier.html) into both the topology monitor and worker, but it would need to be conditionally enabled by the test right before the `shutdown` call to avoid blocking on pre-test hearbeats, and that sounds iffy to me.

In practice, the timing test here is effectively deterministic; it fails on the first iteration of the loop.

BorisDog: Thanks for explaining that `update` can't be blocked by the event listener (driver user).

BorisDog (approved):
LGTM

PR #1662, opened 2026-04-10
Add feature flag to use `aws-lc-rs` instead of `ring` as the `rustls` crypto provider

TimTheBig:
This patch adds a TLS provider option to use `aws-lc-rs` instead of `ring` as the `rustls` crypto provider.

This is not a braking change as it only adds another option.

It saves dependencies on builds with the `aws-auth` feature flag.

Since [`aws-lc-rs`](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/) is designed to be a drop-in replacement for [`ring`](https://docs.rs/ring/latest/ring/) it requires little the way of testing.


driver/src/runtime.rs
```diff
@@ -43,10 +43,12 @@ use crate::{error::Result, options::ServerAddress};
...
 use tls_openssl as tls;
-#[cfg(all(feature = "rustls-tls", not(feature = "openssl-tls")))]
+#[cfg(all(any(feature = "rustls-tls", feature = "rustls-tls-aws-lc"), not(feature = "openssl-tls")))]
 use tls_rustls as tls;
-#[cfg(not(any(feature = "rustls-tls", feature = "openssl-tls")))]
-compile_error!("At least one of the features 'rustls-tls' or 'openssl-tls' must be enabled.");
+#[cfg(not(any(feature = "rustls-tls", feature = "rustls-tls-aws-lc", feature = "openssl-tls")))]
+compile_error!("At least one of the features 'rustls-tls', 'rustls-tls-aws-lc' or 'openssl-tls' must be enabled.");
+#[cfg(all(feature = "rustls-tls", feature = "rustls-tls-aws-lc"))]
+compile_error!("Only one of the features 'rustls-tls' or 'rustls-tls-aws-lc' can be enabled at once.");
```
abr-egn: Mutually exclusive features [should be avoided](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features) and we test that the crate will compile with `--all-features`.  This should work the same way having `rustls-tls` and `openssl-tls` both enabled does: the non-default option (in this case, `rustls-tls-aws-lc`) should take precedence.

TimTheBig: Good catch
in terms of priority `openssl-tls`, `rustls-tls-aws-lc`, then `rustls-tls` seems like the most reasonable order

TimTheBig: Fix in latest commit

abr-egn:
Generally looks good - can you rebase to pick up some fixes that would block merging?  It also looks like it needs a rustfmt pass.

TimTheBig:
> Generally looks good - can you rebase to pick up some fixes that would block merging? It also looks like it needs a rustfmt pass.

Done

abr-egn (approved):

PR #1664, opened 2026-04-13
RUST-2397 fix semgrep check

abr-egn:
RUST-2397

It turns out that [semgrep 1.158.0](https://github.com/semgrep/semgrep/releases/tag/v1.158.0), released on Friday, introduced a new requirement on glibc >=2.35; RHEL8 doesn't satisfy this requirement.  RHEL10 does, but we don't have evergreen test hosts for that yet AFAICT.


BorisDog (approved):
LGTM

PR #1668, opened 2026-04-15
fix(client/executor): correct 'occured' -> 'occurred' in retry-decision doc

SAY-5:
A doc comment in `driver/src/client/executor.rs:1170` explaining when an operation is eligible for write-retry used `the error occured is write-retryable`. Doc-only Rust `///` comment change.


abr-egn:
The originating account appears to be spamming LLM-generated PRs at open-source projects.

PR #1674, opened 2026-04-20
release v3.6.0

abr-egn:


abr-egn:
The failures are unrelated; it looks like server `latest` has a new field it's emitting that breaks a test.  I'm looking to see if this is something other drivers are running into or if there's something about the Rust unified runner that's at fault here.

BorisDog (approved):
LGTM

PR #1676, opened 2026-04-21
RUST-2405 listIndexes returns root documents

abr-egn:
RUST-2405

The rules for when extra fields are allowed in the spec are complicated and require flagging some specific operations as returning an array of root-level documents; we'd forgotten to tag `listIndexes` with this flag and that was causing the current batch of test failures.

I also removed the unrelated warning about having both `rustls-tls` and `rustls-tls-aws-lc` enabled because (a) it's annoying when developing with `--all-features` to have an unfixable warning and (b) we don't have a warning like that for the similar `rustls` vs `openssl` overlap.


BorisDog (approved):
LGTM

PR #1692, opened 2026-05-06
RUST-2014 Refactor `Client::shutdown` to use tokio utilities

abr-egn:
RUST-2014

Tokio provides [`CancellationToken`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) to signal for an explicit stop and [`TaskTracker`](https://docs.rs/tokio-util/latest/tokio_util/task/struct.TaskTracker.html) to wait for a set of spawned tasks to complete; this refactors the machinery for `Client::shutdown` to use those, making things substantially simpler.

This also demonstrates what we'd need to support something CSOT-like: threading a `CancellationToken` through the executor layer to select on at the appropriate points.


driver/src/cmap/test/integration.rs
```diff
@@ -50,6 +50,7 @@ async fn acquire_connection_and_send_command() {
         ConnectionEstablisher::new(EstablisherOptions::from(&client_options)).unwrap(),
         TopologyUpdater::channel().0,
         Some(pool_options),
+        TopologyShutdown::new(),
```
abr-egn, line 53: This does require plumbing through the shutdown struct to various places, including making placeholders in tests.  I didn't see any clear case for bundling it with any of the existing related types :(

driver/src/cmap/worker.rs
```diff
@@ -256,11 +260,12 @@ impl ConnectionPoolWorker {
             server_updater,
             max_connecting,
             cancellation_sender,
+            shutdown,
         };
 
-        runtime::spawn(async move {
+        runtime::spawn(worker.shutdown.tasks.clone().track_future(async move {
```
abr-egn, line 266: The idiom of cloning the task pool (cheap, it shares internal state) is necessary to avoid the borrow checker thinking that there's a simultaneous borrow of (in this case) `worker` both for the `track_future` call and in the async closure.  Unfortunately, this can't be turned into a utility function because that hits the same issue.

driver/src/cmap/worker.rs
```diff
@@ -337,10 +341,6 @@ impl ConnectionPoolWorker {
                     PoolManagementRequest::Broadcast(msg) => {
                         let (msg, ack) = msg.into_parts();
                         match msg {
-                            BroadcastMessage::Shutdown => {
```
abr-egn, line 340: Workers no longer need to explicitly handle a shutdown message because the toplevel `Topology` causes them to terminate via the normal drop path, and the tracker means we're still waiting for the tasks to complete without needing an explicit ack.

driver/src/sdam/monitor.rs
```diff
@@ -119,7 +123,12 @@ impl Monitor {
 
             if self.topology_version.is_some() && self.allow_streaming {
                 if let Some(rtt_monitor) = self.pending_rtt_monitor.take() {
-                    runtime::spawn(rtt_monitor.execute());
+                    runtime::spawn(
```
abr-egn, line 126: We weren't actually waiting for these for `shutdown` before.  Oops!

driver/src/client.rs
```diff
@@ -170,8 +167,8 @@ struct ClientInner {
 }
 
 #[derive(Debug)]
-struct Shutdown {
-    pending_drops: SyncMutex<IdSet<crate::runtime::AsyncJoinHandle<()>>>,
+struct ClientShutdown {
```
abr-egn, line 170: Because there are now two `Shutdown` variants with slightly different semantics, I renamed them to make the distinction clear.

driver/src/client.rs
```diff
@@ -170,8 +167,8 @@ struct ClientInner {
...
+    pending_drops: TaskTracker,
     executed: AtomicBool,
```
abr-egn, line 172: `TaskTracker` has its own `closed` bool, but that has to be set _before_ waiting on the tasks, and we still need this as a flag to be set _after_ the tasks have completed to prevent any further commands from being executed.

driver/src/sdam/monitor.rs
```diff
@@ -106,9 +109,10 @@ impl Monitor {
             connection: None,
             allow_streaming,
             topology_version: None,
+            shutdown: shutdown.clone(),
         };
 
-        runtime::spawn(monitor.execute());
+        runtime::spawn(shutdown.tasks.track_future(monitor.execute()));
```
isabelatkinson: These tasks don't need to care about the cancellation token because they automatically terminate when the topology drops, right?

isabelatkinson (approved):
nice!

rigens:
> nice!

Very nice, now (v3.7) it takes ~10 seconds to shutdown a client.

```rust
#[tokio::main]
async fn main() {
    let client = Client::with_uri_str("mongodb://localhost:27017/db33") // empty database
        .await
        .unwrap();

    let db = client.default_database().unwrap();

    let collection = db.collection::<Document>("coll33"); //empty collection
    let cursor = collection.find(doc! {}).await.unwrap();
    let docs: Vec<_> = cursor.collect().await;
    println!("{:?}", docs);

    let start = Instant::now();
    client.shutdown().await;
    let duration = start.elapsed();
    println!("Time elapsed: {:?}", duration);
}
```

Output (v3.7.0):
`Time elapsed: 10.001053699s`

Output (v3.6.0):
`Time elapsed: 1.083255ms`

PR #1682, opened 2026-05-01
RUST-2418 bump to use hickory-* 0.26

aviramha:
Hey folks!
There's an [advisory](https://github.com/hickory-dns/hickory-dns/security/advisories/GHSA-3v94-mw7p-v465) that we want to fix in our software and need mongodb crate to bump hickory to solve that.
I used Claude to do the update, then manually reviewed it.

PR #1693, opened 2026-05-07
RUST-2342 bump MSRV to 1.88

abr-egn:
RUST-2342

The newer versions of the `hickory` crates in the *actual* fix (https://github.com/mongodb/mongo-rust-driver/pull/1682) require 1.88, which fortunately is past our six-month threshold.


BorisDog (approved):
LGTM

PR #1694, opened 2026-05-08
RUST-2342 Skip flaky timing-based tests on macos

abr-egn:
RUST-2342

I went back through a week of waterfall runs; these tests were the cause of all of the failures on macos runs in that span ("Topology lifecycle" spec test by far the most).

### DIFF
--- a/driver/src/sdam/test.rs
+++ b/driver/src/sdam/test.rs
@@ -30,6 +30,10 @@ use crate::{
 
 #[tokio::test(flavor = "multi_thread")]
 async fn min_heartbeat_frequency() {
+    if cfg!(target_os = "macos") {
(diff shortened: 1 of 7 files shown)


driver/src/sdam/test.rs
```diff
@@ -30,6 +30,10 @@ use crate::{
 
 #[tokio::test(flavor = "multi_thread")]
 async fn min_heartbeat_frequency() {
+    if cfg!(target_os = "macos") {
```
abr-egn, line 33: I did this rather than putting the whole function behind `#[cfg(not(target_os = "macos"))]` so the skip isn't silent; also, this means that imports don't need to be conditional on target_os.

isabelatkinson (approved):

PR #1700, opened 2026-05-15
RUST-2425 Document error code buckets

abr-egn:
RUST-2425

This adds documentation for the error code category constants, turns a few that were inline into named, and adds tests for the spec-defined relationships to make sure we don't accidentally break that later.

### DIFF
--- a/driver/src/error.rs
+++ b/driver/src/error.rs
@@ -25,7 +25,7 @@ pub use bulk_write::{BulkWriteError, PartialBulkWriteResult};
 /// SHUTTING_DOWN_CODES is a strict subset of these.
 pub(crate) const RECOVERING_CODES: [i32; 5] = [11600, 11602, 13436, 189, 91];
 /// Codes indicating the node is not a writable primary (per the SDAM spec).
-const NOTWRITABLEPRIMARY_CODES: [i32; 3] = [10107, 13435, 10058];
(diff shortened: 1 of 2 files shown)


driver/src/error.rs
```diff
@@ -21,17 +21,34 @@ use crate::{
 
 pub use bulk_write::{BulkWriteError, PartialBulkWriteResult};
 
-const RECOVERING_CODES: [i32; 5] = [11600, 11602, 13436, 189, 91];
+/// Codes indicating the node is in a recovering state (per the SDAM spec).
+/// SHUTTING_DOWN_CODES is a strict subset of these.
+pub(crate) const RECOVERING_CODES: [i32; 5] = [11600, 11602, 13436, 189, 91];
+/// Codes indicating the node is not a writable primary (per the SDAM spec).
 const NOTWRITABLEPRIMARY_CODES: [i32; 3] = [10107, 13435, 10058];
```
BorisDog: Should we also test that NOTWRITABLEPRIMARY_CODES don't overlap with RECOVERING_CODES ?

BorisDog (approved):
Nice, LGTM.

driver/src/error.rs
abr-egn: Reasonable; added.

PR #1708, opened 2026-05-25
RUST-2376 Optionally redact errors containing URI elements

abr-egn:
RUST-2376

This adds a new `redact-errors` feature flag that will cause URI components contained in error messages to be replaced with `"<redacted>"`.

There's also a minor unrelated cleanup - some code didn't have the proper `#[cfg(feature = "in-use-encryption")]` gating, causing warnings when compiling with default features.

### DIFF
--- a/driver/src/sdam/topology.rs
+++ b/driver/src/sdam/topology.rs
@@ -5,11 +5,6 @@ use std::{
     time::Duration,
 };
 
-use crate::{
-    bson::oid::ObjectId,
-    cmap::establish::handshake::ClientMetadata,
-    error::SYSTEM_OVERLOADED_ERROR,
-};
 use futures_util::{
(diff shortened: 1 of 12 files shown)


driver/src/sdam/topology.rs
```diff
@@ -5,11 +5,6 @@ use std::{
     time::Duration,
 };
 
-use crate::{
```
abr-egn, line 8: We had two different `use crate::{stuff}` blocks here for no particular reason (probably auto-imports), I just merged them.

BorisDog (changes requested):
Do we need testing for that? Some basic sanity tests maybe?

abr-egn:
> Do we need testing for that? Some basic sanity tests maybe?

Added.

BorisDog (approved):
LGTM

PR #1712, opened 2026-05-27
RUST-2299 Remove warnings about hickory performance

abr-egn:
RUST-2299

It turns out the reason I was having trouble making a repro case is because ... it doesn't repro any more.  Happily, it _does_ repro with the version of hickory (then trust-dns) that was current at the time the bug originally was filed, so I was able to validate the repro and behavior change.


BorisDog (approved):
LGTM

PR #1714, opened 2026-05-28
RUST-2434 Run SBOM update inline with version bump

abr-egn:
RUST-2434

While the workflow could be updated to make the improvements suggested in the ticket, by far the simplest and most robust solution turned out to be to bundle the SBOM update with our existing version-bump script.

Benefits:
* Removes the need for a distinct SBOM update PR requiring an evergreen run and human approval.
* Entirely eliminates the PR spam and branch garbage collection problems.
* Doesn't add any additional human steps, just an update in the release process doc to use the shell script instead of the binary directly.
* Naturally ties SBOM generation to the release process.

Drawbacks:
* Requires a couple new binaries installed locally for devs.  IMO, this is pretty minor: they're lightweight installs and the script includes pointers.


BorisDog (approved):
LGTM

PR #1716, opened 2026-05-29
RUST-2248 Add deprecation warnings for explicit MONGODB-AWS auth properties

abr-egn:
RUST-2248

Since this is a breaking change, we can't actually enable it until 4.0, but we can emit warnings.

### DIFF
--- a/driver/src/client/auth.rs
+++ b/driver/src/client/auth.rs
@@ -202,25 +202,30 @@ impl AuthMechanism {
             }
             #[cfg(feature = "aws-auth")]
             AuthMechanism::MongoDbAws => {
-                if credential.username.is_some() {
-                    return Err(Error::invalid_argument(
-                        "Username cannot be provided for MONGODB-AWS authentication",
-                    ));
-                }
-                if credential.password.is_some() {
-                    return Err(Error::invalid_argument(
-                        "Password cannot be provided for MONGODB-AWS authentication",
-                    ));
+                if credential.username.is_some() || credential.password.is_some() {
+                    crate::log_warning!(
+                        "URI credentials for MONGODB-AWS are deprecated and will be removed in \
+                         4.0."
+                    );
                 }
                 if credential
                     .mechanism_properties
                     .as_ref()
-                    .map_or(false, |mp| mp.contains_key("AWS_SESSION_TOKEN"))
+                    .map_or(false, |mp| mp.contains_key(aws::AWS_SESSION_TOKEN))
                 {
-                    return Err(Error::invalid_argument(
-                        "AWS_SESSION_TOKEN mechanism property cannot be provided for MONGODB-AWS \
-                         authentication",
-                    ));
+                    crate::log_warning!(
+                        "Auth mechanism {:?} for MONGODB-AWS is deprecated and will be removed in \
+                         4.0",
+                        aws::AWS_SESSION_TOKEN
+                    );
+                }
+                if credential.username.is_some() && credential.password.is_none() {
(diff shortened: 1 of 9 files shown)


driver/src/lib.rs
```diff
@@ -97,3 +97,14 @@ compile_error!(
     "The feature 'compat-3-3-0' must be enabled to ensure forward compatibility with future \
      versions of this crate."
 );
+
+macro_rules! log_warning {
+    ($($arg:tt)+) => {
+        #[cfg(debug_assertions)]
```
BorisDog, line 105: Does this mean that the warning will be visible in debug build only?

driver/src/client/auth/aws.rs
```diff
@@ -104,17 +104,22 @@ pub(super) async fn authenticate_stream(
...
 }
 
+pub(super) const AWS_SESSION_TOKEN: &'static str = "AWS_SESSION_TOKEN";
+
 // Find credentials using MongoDB URI or AWS SDK
 pub(crate) async fn get_aws_credentials(credential: &Credential) -> Result<Credentials> {
     if let (Some(access_key), Some(secret_key)) = (&credential.username, &credential.password) {
+        crate::log_warning!(
+            "URI credentials for MONGODB-AWS are deprecated and will be removed in 4.0."
+        );
```
BorisDog: Are is `validate_credential` always invoked prior to `get_aws_credentials`?
If so, does it make sense to emit the warning in a single place?

driver/src/client/auth.rs
```diff
@@ -202,6 +202,23 @@ impl AuthMechanism {
...
+                    );
+                }
+                if credential
+                    .mechanism_properties
+                    .as_ref()
+                    .map_or(false, |mp| mp.contains_key(aws::AWS_SESSION_TOKEN))
+                {
+                    crate::log_warning!(
+                        "Auth mechanism {:?} for MONGODB-AWS is deprecated and will be removed in \
+                         4.0",
```
BorisDog: nit: period in the end. "4.0."

BorisDog (changes requested):

driver/src/client/auth/aws.rs
abr-egn: Indeed it does, fixed.

driver/src/client/auth.rs
abr-egn: Fixed.

driver/src/lib.rs
abr-egn: Yep - it seems rude to add new non-error stderr output to prod builds.

BorisDog: Agreed about stderr, but I am concerted that there is a good chance that this warning might go unnoticed.

Should we use the logging infra instead?

BorisDog (changes requested):
Looks good, left a question about logging.

driver/src/lib.rs
abr-egn: We don't really have a great option here - logging is an optional feature in the crate that as far as I am aware isn't commonly used, so that seems like it would have a _higher_ risk of not being seen.  Printing to stderr but only doing so in debug builds was the best I could come up with for a reasonable middle path.

driver/src/client/auth/aws.rs
abr-egn: Updated - now if logging is enabled it'll use the standard `log::warn!`, only falling back to a stderr-print-on-debug if it's not enabled.

BorisDog (approved):
LGTM

PR #1720, opened 2026-06-09
RUST-2421 Add regression tests for QE case/diacritic sensitivity

abr-egn:
RUST-2421

The spec ticket adds cases 8-11; I ended up doing a lot of refactoring of the existing tests as part of this:
* changed the container for the overall test from an increasingly giant function to a submodule; beyond readability, this also means the case groups get distinct test results and can be run individually
* replaced the setup function with a `Setup` struct and associated constructor, mostly to give the return values names and make it easy to see which case is using which values
* folded together those cases that were running the same routines with different inputs/expectations


isabelatkinson (approved):
lgtm!

PR #1721, opened 2026-06-09
RUST-2366 Switch to run-mongodb.sh

abr-egn:


.evergreen/benchmarks.yml
```diff
@@ -133,7 +133,7 @@ functions:
             SSL=${SSL} \
             REQUIRE_API_VERSION=${REQUIRE_API_VERSION} \
             SKIP_LEGACY_SHELL=1 \
-            sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
+            sh ${DRIVERS_TOOLS}/.evergreen/run-mongodb.sh start
     # run-orchestration generates expansion file with the MONGODB_URI for the cluster
```
BorisDog: minor: update the stale comment.

.evergreen/benchmarks.yml
```diff
@@ -133,7 +133,7 @@ functions:
             SSL=${SSL} \
             REQUIRE_API_VERSION=${REQUIRE_API_VERSION} \
             SKIP_LEGACY_SHELL=1 \
-            sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
+            sh ${DRIVERS_TOOLS}/.evergreen/run-mongodb.sh start
```
BorisDog, line 136: I think `run-mongodb.sh stop` needs to be run at some point? Does `mongo-orchestration stop` need to be removed?

BorisDog (changes requested):

.evergreen/benchmarks.yml
abr-egn: Good catch.  Fun facts:
* The evergreen config directly called `mongo-orchestration stop` rather than using `run-orchestration.sh` (this is why my initial search didn't turn it up).
* This had an implicit assumption that the `mongo-orchestration` binary was installed by the initial server startup script
* That hasn't been true since [January 2025](https://github.com/mongodb-labs/drivers-evergreen-tools/pull/578), and our evergreen runs have just been silently failing to properly tear down the server since then.

BorisDog (approved):
LGTM

PR #1725, opened 2026-06-22
RUST-2447 Update crypto dependencies

abr-egn:
RUST-2447

Some minor API differences; mostly just bumping to pick up `block-buffer` 0.12.1.  Verified that it's updated and there are no lingering older versions via `cargo tree`.


BorisDog (approved):
LGTM

PR #1726, opened 2026-06-24
RUST-2445 Fix explicit encryption prose test case 2

abr-egn:
RUST-2445

Also did a bit of readability tidying of the tests.

### DIFF
--- a/driver/src/test/csfle/prose.rs
+++ b/driver/src/test/csfle/prose.rs
@@ -932,82 +932,92 @@ mod deadlock {
 mod explicit_encryption {
     use super::*;
 
-    struct ExplicitEncryptionTestData {
+    struct TestData {
         key1_id: Binary,
         client_encryption: ClientEncryption,
         encrypted_client: Client,
     }
 
-    async fn explicit_encryption_setup() -> Result<Option<ExplicitEncryptionTestData>> {
-        if server_version_lt(6, 0).await {
-            log_uncaptured("skipping explicit encryption test: server below 6.0");
-            return Ok(None);
+    impl TestData {
+        async fn new() -> Result<Self> {
+            let key_vault_client = Client::for_test().await;
+
+            let encrypted_fields = load_testdata("data/encryptedFields.json")?;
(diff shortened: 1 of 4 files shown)


driver/src/test/csfle/prose.rs
```diff
@@ -932,82 +932,102 @@ mod deadlock {
...
-        if server_version_lt(6, 0).await {
-            log_uncaptured("skipping explicit encryption test: server below 6.0");
-            return Ok(None);
-        }
-        if topology_is_standalone().await {
-            log_uncaptured("skipping explicit encryption test: cannot run on standalone");
-            return Ok(None);
-        }
+    impl TestData {
+        async fn new() -> Result<Self> {
```
abr-egn, line 942: Aside from omitting the server checks, this is unchanged from `explicit_encryption_setup`.

driver/src/test/csfle/prose.rs
```diff
@@ -932,82 +932,102 @@ mod deadlock {
...
+                return Ok(());
+            }
+            if server_version_lt(6, 0).await {
+                log_uncaptured(format!(
+                    "skipping explicit encryption test [{}]: server below 6.0",
+                    $name
+                ));
+                return Ok(());
+            }
+            if topology_is_standalone().await {
```
BorisDog: is fle2v2_ok sufficient here?
Looks like` fle2v2_ok == true` implies `server_version_lt(6, 0) == false` and `topology_is_standalone == false`?

BorisDog (changes requested):
minor comment

driver/src/test/csfle/prose.rs
abr-egn: Derp, I didn't spot that merging those two checks made this part redundant.  With just a single function I've dropped the macro entirely.

abr-egn:
Side note: the cargo-deny failure was already fixed by the weekly dependabot update.

BorisDog (approved):
LGTM

PR #1727, opened 2026-06-24
RUST-2402 QE string API GA

abr-egn:
RUST-2402, RUST-2449

This marks the `string` (nee `textPreview`) algorithm and related methods GA, removing the need for the `text-indexes-unstable` feature.  A bunch of the tests had their predicate swapped from <9.0 to >=9.0, and a few need special handling to make sure things still work with the `preview` names.


driver/src/action/csfle/encrypt.rs
```diff
@@ -125,12 +125,8 @@ pub struct EncryptOptions {
     pub range_options: Option<RangeOptions>,
 
     /// Set the text options. This should only be set when the algorithm is
-    /// [`Algorithm::TextPreview`].
-    ///
-    /// NOTE: This option is unstable and subject to backwards-breaking changes. It should only be
-    /// used in experimental workloads.
-    #[cfg(feature = "text-indexes-unstable")]
-    pub text_options: Option<TextOptions>,
```
BorisDog, line 133: Is this (and the rest of Text removal) a breaking change? Is this expected in minor release?

BorisDog (changes requested):

driver/src/action/csfle/encrypt.rs
abr-egn: This would be a breaking change, but these fields were gated behind the `text-indexes-unstable` feature (off by default) that was itself flagged as subject to breaking changes.  From the perspective of the stable driver API, this is an addition 🙂

BorisDog (approved):
LGTM

PR #1729, opened 2026-06-29
RUST-2417 build libmongocrypt from source (driver)

abr-egn:
RUST-2417

(driver edition)

The old tarball location was deprecated and stale; this changes the libmongocrypt install to fetch source and build that.  It's pinned to a specific release to avoid random changes at head from breaking us, but this means we'll need to remember to update that version as necessary.


abr-egn:
The failing CSFLE tests are fixed by https://github.com/mongodb/mongo-rust-driver/pull/1726 🙂

isabelatkinson (approved):

PR #1735, opened 2026-07-07
RUST-2381 test KMS provider `accessToken`

abr-egn:
RUST-2381

Unfortunately, instead of just being another variable to fit into the existing placeholder scheme, this one is a distinct form that requires special handling.

### DIFF
--- a/etc/update-spec-tests.sh
+++ b/etc/update-spec-tests.sh
@@ -28,7 +28,11 @@ tmpdir=`perl -MFile::Temp=tempdir -wle 'print tempdir(TMPDIR => 1, CLEANUP => 0)
 curl -sL "https://github.com/mongodb/specifications/archive/$REF.zip" -o "$tmpdir/specs.zip"
 unzip -q -d "$tmpdir" "$tmpdir/specs.zip"
 mkdir -p "${DEST}/$1"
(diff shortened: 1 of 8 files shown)


etc/update-spec-tests.sh
```diff
@@ -28,7 +28,11 @@ tmpdir=`perl -MFile::Temp=tempdir -wle 'print tempdir(TMPDIR => 1, CLEANUP => 0)
 curl -sL "https://github.com/mongodb/specifications/archive/$REF.zip" -o "$tmpdir/specs.zip"
 unzip -q -d "$tmpdir" "$tmpdir/specs.zip"
 mkdir -p "${DEST}/$1"
-rsync -ah "$tmpdir/specifications-$REF"*"/source/$1/tests/" "${DEST}/$1" --delete
+EXCLUDE=""
+if [ "$1" = "client-side-encryption" ]; then
+  EXCLUDE="--exclude=legacy/"
```
abr-egn, line 33: The `client-side-encryption/legacy` subdir has a bunch of obsolete spec tests and having them get pulled in each time this script is run just to manually `rm` them was getting annoying.

BorisDog (approved):
LGTM

PR #1738, opened 2026-07-13
RUST-2412 Provide an operation wrapper abstraction

abr-egn:
RUST-2412

Not directly related to the main work for this ticket, but that involves adding a field to `Operation` and I got annoyed at having to always update the various wrappers.

This adds an `OperationWrapper` trait that knows how to dispatch to a generic inner `Operation`; this means that the various places we have such (four as of this writing) are both easier to read, since they only need to list out the methods where the behavior is changed by the wrapper, and don't have to be updated when we add new methods to the core trait.

The downside is that, because we can't have two blanket traits that impl `Operation` for `OperationWithDefaults` and `OperationWrapper`, those now work via a helper trait `OperationDispatch` and tag trait `OperationImpl`.  IMO, this is worth it because the added boilerplate per-operation is minimal and the helpers are largely ignorable (nothing outside the one blanket impl in `operation.rs` needs to care about `OperationDispatch`).


driver/src/operation.rs
```diff
@@ -411,10 +411,100 @@ pub(crate) trait OperationWithDefaults: Send + Sync {
     type Otel: crate::otel::OtelWitness<Op = Self>;
 }
 
-impl<T: OperationWithDefaults> Operation for T
+pub(crate) trait OperationDispatch<Kind> {
```
abr-egn, line 422: This intermediate trait is needed because the rule is that there can be only one blanket impl per trait, so the blanket impl is `Operation for OperationDispatch<K>`.  The blanket impls for the helper traits don't run into this issue because `OperationDispatch<WithDefaults>` and `OperationDispatch<Wrapper>` are considered to be distinct traits for the purposes of overlap checks.

Ideally we could just include `OperationImpl<Kind=WithDefaults>` / `<Kind=Wrapper>` in a blanket `Operation` impl but it turns out there's a [longstanding compiler bug](https://github.com/rust-lang/rust/issues/20400) that prevents that 🙁

abr-egn:
The failures here are fixed by https://github.com/mongodb/mongo-rust-driver/pull/1739.

driver/src/operation/create.rs
```diff
@@ -65,6 +71,10 @@ impl OperationWithDefaults for Create {
     type Otel = crate::otel::Witness<Self>;
 }
 
+impl OperationImpl for Create {
+    type Kind = WithDefaults;
```
isabelatkinson: I think the `WithDefaults` naming is more descriptive of the implementation than the concept. Can we change this to be something like `Base` or `Root`? And ditto for the `OperationWithDefaults` trait - something like `BaseOperation` or `RootOperation` would make it more clear that the types that define that trait are ones that don't wrap another operation.

(Not necessarily a new concern on this PR, but something I got to thinking about more while reviewing.)

driver/src/operation.rs
isabelatkinson: Makes sense - can you add a comment on this file explaining the basic trait hierarchy? I don't think I'll remember this rationale next time I look at this 🙂

abr-egn: Good thought, done - also tidied up some other related docs here.

driver/src/operation/create.rs
abr-egn: Done, also renamed `OperationWrapper` to `WrappedOperation` to follow the pattern.

isabelatkinson (approved):

PR #1742, opened 2026-07-15
RUST-2412 Send afterClusterTime on writes in causally-consistent sessions

abr-egn:
RUST-2412

This one's a little fiddly because I didn't want to hardcode a list but the set of commands that need this isn't _quite_ "commands that support a write concern", there are a few in the latter that are excluded in the former.  Adding a new method with a default impl and specifically tagging those few operations seemed the least-bad option here.

### DIFF
--- a/driver/src/test/spec/unified_runner/operation.rs
+++ b/driver/src/test/spec/unified_runner/operation.rs
@@ -6,6 +6,7 @@ mod connection;
 mod count;
 #[cfg(feature = "in-use-encryption")]
 mod csfle;
+mod database;
 mod delete;
 mod failpoint;
 mod find;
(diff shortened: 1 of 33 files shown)


driver/src/test/spec/unified_runner/operation.rs
```diff
@@ -120,6 +122,7 @@ pub(crate) trait TestOperation: Debug + Send + Sync {
 macro_rules! with_mut_session {
```
abr-egn, line 122: When I was implementing the `dropDatabase` test operation I changed these to not leak required imports into callers.

isabelatkinson (approved):

PR #1749, opened 2026-07-22
RUST-2446 Use `baseBackoffMS` in retry backoff calculations

isabelatkinson:


driver/src/client/session/action.rs
BorisDog: Consistency:
Should both `compute_backoff` and `calculate_backoff` use or not use `RETRY_BASE`?
Also should both methods have same name?

BorisDog (approved):
LGTM + small suggestion

driver/src/client/session/action.rs
isabelatkinson: 👍 updated the names and removed RETRY_BASE

PR #1756, opened 2026-08-06
RUST-2409 CSFLE/QE Support for HTTP Proxies

isabelatkinson:


driver/src/client/csfle/client_encryption.rs
```diff
@@ -458,3 +470,72 @@ impl MasterKey {
...
+    #[allow(clippy::type_complexity)]
+    Arc<dyn Fn(ServerAddress) -> BoxFuture<'static, Result<Box<dyn KmsStream>>> + Send + Sync>,
+);
+
+impl KmsConnectCallback {
+    /// Create a new KMS connect callback.
+    ///
+    /// Due to limitations in the Rust compiler, this type cannot be constructed directly from an
+    /// async closure. Instead, this function accepts a sync closure that returns a [`BoxFuture`]
+    /// containing the logic to establish the KMS stream.
```
isabelatkinson, line 500: If/when this issue goes away, the opaque design of the `KmsConnectCallback` type will make it easy to add a new constructor that can take an async callback

driver/src/client/csfle/state_machine.rs
```diff
@@ -187,30 +200,59 @@ impl CryptExecutor {
...
+                                    Error::internal(format!("buffer size overflow: {e}"))
+                                })?;
+                                buf.resize(buf_size, 0);
+                                let count = stream.read(&mut buf).await?;
+                                kms_ctx.feed(&buf[0..count])?;
+                            }
+                            Ok(())
+                        }
+
+                        match crypt_executor.kms_connect_callback {
```
abr-egn, line 234: It might be a little more readable to add an `AsyncStream` variant for the return value of `kms_connect_callback`?

abr-egn (approved):
LGTM!  One optional readability suggestion, no need for re-review either way.

driver/src/client/csfle/state_machine.rs
isabelatkinson: Going to leave this as-is, adding `KmsStream` into `AsyncStream` requires it to be `Sync`, which is an unnecessary bound for where the variant is actually being used

PR #1757, opened 2026-08-07
RUST-1820 Use raw documents and smart truncation for tracing and otel

abr-egn:
RUST-1820

This fixes the problem of large command documents causing extreme performance degradation when logging is enabled by skipping the `RawDocument` -> `Document` conversion step and truncating within the string formatting step rather than after.

Large doc insertMany benchmarks before this change:

without tracing: 217.224 MB/s
with tracing: 61.941 MB/s (~3.5x slower)

After this change:

without tracing: 217.057 MB/s
with tracing: 199.361 MB/s (~1.1x slower)

### DIFF
--- a/driver/src/bson_util.rs
+++ b/driver/src/bson_util.rs
@@ -410,7 +410,7 @@ fn rawdoc_to_json_str_inner(
     let mut current = doc.iter_elements();
     let mut is_array = false;
     let mut is_first = true;
-    push_trunc!(out, max_length_bytes, "{ ");
+    push_trunc!(out, max_length_bytes, "{");
     'outer: loop {
         while let Some(elt) = current.next() {
             let elt = elt?;
@@ -431,7 +431,7 @@ fn rawdoc_to_json_str_inner(
             }
 
             if !is_first {
-                push_trunc!(out, max_length_bytes, ", ");
+                push_trunc!(out, max_length_bytes, ",");
             }
             is_first = false;
 
@@ -441,11 +441,11 @@ fn rawdoc_to_json_str_inner(
                 )
                 .to_string();
                 push_trunc!(out, max_length_bytes, &key);
-                push_trunc!(out, max_length_bytes, ": ");
+                push_trunc!(out, max_length_bytes, ":");
             }
             match value {
                 RawBsonRef::Document(d) => {
-                    push_trunc!(out, max_length_bytes, "{ ");
+                    push_trunc!(out, max_length_bytes, "{");
                     let mut tmp = d.iter_elements();
                     std::mem::swap(&mut current, &mut tmp);
                     stack.push((tmp, is_array));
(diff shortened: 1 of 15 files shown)


driver/src/trace/topology.rs
```diff
@@ -167,6 +167,13 @@ impl TopologyTracingEventEmitter {
             target: TOPOLOGY_TRACING_EVENT_TARGET,
             TracingOrLogLevel::Debug
         ) {
+            let reply = (&event.reply)
```
abr-egn, line 170: Since the heartbeat response is entirely server-controlled and very unlikely to be Large, it didn't seem worth defining a whole shadow `RawSdamEvent` structure for this one case.  It _also_ didn't seem worth keeping around the old `doc_to_json_str` for this one case either, so this cheats a little and converts the parsed doc back into a rawdoc to use the new function.

driver/src/bson_util.rs
```diff
@@ -325,33 +325,173 @@ pub(crate) mod option_u64_as_i64 {
 /// Truncates the given string at the closest UTF-8 character boundary >= the provided length.
 /// If the new length is >= the current length, does nothing.
 #[cfg(any(feature = "tracing-unstable", feature = "opentelemetry"))]
-pub(crate) fn truncate_on_char_boundary(s: &mut String, new_len: usize) {
+#[must_use]
+pub(crate) fn truncate_on_char_boundary(s: &mut String, new_len: usize) -> bool {
```
abr-egn, line 329: This needed to change from appending `"..."` itself on truncation to returning a bool for truncation because the raw doc formatter always needs the ellipsis even if _this_ value wasn't truncated because there's more of the doc that's being omitted.

driver/src/bson_util.rs
```diff
@@ -325,33 +325,173 @@ pub(crate) mod option_u64_as_i64 {
...
+        }
+    }};
 }
 
 #[cfg(any(feature = "tracing-unstable", feature = "opentelemetry"))]
-pub(crate) fn doc_to_json_str(doc: crate::bson::Document, max_length_bytes: usize) -> String {
-    let mut ext_json = Bson::Document(doc).into_relaxed_extjson().to_string();
-    truncate_on_char_boundary(&mut ext_json, max_length_bytes);
-    ext_json
+fn rawdoc_to_json_str_inner(
```
abr-egn, line 401: IMO, this function is a very nice illustration of the strengths of Rust: it does a pointer walk through the nested documents in the doc byte buffer to append to a string buffer, keeping a minimal stack of resume states so it's a flat loop rather than recursion.  In C/C++ this would be _terrifying_ to ensure none of the pointer logic does anything squiffy, but here it's entirely safe by construction and fairly ergonomic at that.

driver/src/event/command/raw.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub(crate) service_id: Option<ObjectId>,
+}
+
+impl TryInto<CommandEvent> for RawCommandEvent {
+    type Error = crate::error::Error;
+
+    fn try_into(self) -> Result<CommandEvent, Self::Error> {
+        Ok(match self {
+            RawCommandEvent::Started(ev) => CommandEvent::Started(CommandStartedEvent {
+                command: ev.command.try_into()?,
```
isabelatkinson: Can we make this method infallible and sub this with a "serialization error" doc if the conversion fails? That way users will still observe an event in the failure case

isabelatkinson (approved):
lgtm, just a small suggestion!

PR #1760, opened 2026-08-12
RUST-2465 Fix an infinite loop in ChangeStream iteration

abr-egn:
RUST-2465

The immediate cause was that we forgot to check for exhaustion in the non-session path - `StreamState::poll_next` would `continue` for any `Ok(None)` case, which could be either "waiting for next result" or "exhausted".  In the latter, the next layers down would early-exit and `poll_next` would keep looping.

The root cause (IMO) is that flattening the trinary `AdvanceResult` into a binary `bool`/`Option` was done way too early and made this sort of mistake very easy to make - it was just a forgotten `if` guard in a match.  This PR extends `AdvanceResult` with an optional payload and makes it the data carrier all the way through the layers up to the public API so hopefully it's less of a footgun.

### DIFF
--- a/driver/src/change_stream.rs
+++ b/driver/src/change_stream.rs
@@ -227,6 +227,9 @@ impl<T: DeserializeOwned> Stream for StreamState<T> {
                         *self = StreamState::Idle(Box::new(state));
                         match out {
                             Ok(Some(v)) => return Poll::Ready(Some(Ok(v))),
+                            Ok(None) if self.state().cursor.raw().is_exhausted() => {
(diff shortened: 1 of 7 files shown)


abr-egn:
Oh, also, as always when I work on the change stream code, I'm Not A Fan of the structure, so I'm going to take another stab at making things less of an eldritch nest with a follow-up PR.

isabelatkinson (commented):
Changes look good - can you add a test for the infinite loop this fixes?

isabelatkinson (approved):

PR #1761, opened 2026-08-13
RUST-2465 Simplify cursor/changestream internals

abr-egn:
RUST-2465

This adds a `PollState` helper that handles the "stream but also expose inner state" logic, used for both cursors and change streams; unlike my previous attempts, this one dodges the variance issues primarily by not attempting to be a stream itself and being parameterized on the stream _output_. I also took this opportunity to make the state access (which is technically fallible) propagate the failure and only panic on public APIs that can't be changed before 4.0.  If this seems reasonable, I'll file a ticket to make those APIs return `Result` in 4.0.

This also takes the logic for raw batch cursors that was duplicated across the session/no boundary and pulls it into a `CursorState` struct, making `RawBatchCursor` and `SessionRawBatchCursor` thin wrappers around common logic.

### DIFF
--- a/driver/src/cursor.rs
+++ b/driver/src/cursor.rs
@@ -1,4 +1,5 @@
 pub(crate) mod common;
+pub(crate) mod poll_state;
 pub mod raw_batch;
 pub(crate) mod session;
 pub(crate) mod stream;
@@ -15,6 +16,7 @@ use serde::{de::DeserializeOwned, Deserialize};
 use crate::{
     bson::RawDocument,
     cmap::conn::PinnedConnectionHandle,
+    cursor::poll_state::poll_panic,
     error::Result,
     Client,
     ClientSession,
@@ -86,6 +88,16 @@ pub struct Cursor<T> {
     stream: stream::Stream<'static, raw_batch::RawBatchCursor, T>,
 }
 
+#[allow(dead_code)]
+const _: fn() = || {
+    /// `Cursor<T>` must stay covariant in `T` for zero-copy deserialization to work: a cursor
+    /// yielding a type that borrows for a long lifetime has to be usable where one borrowing for a
+    /// shorter lifetime is expected.
+    fn assert_covariant<'long: 'short, 'short>(cursor: Cursor<&'long str>) -> Cursor<&'short str> {
+        cursor
+    }
+};
+
 impl<T> Cursor<T> {
     /// Move the cursor forward, potentially triggering requests to the database for more results
     /// if the local buffer has been exhausted.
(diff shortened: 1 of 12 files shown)


driver/src/cursor/raw_batch.rs
```diff
@@ -133,12 +133,124 @@ const _: fn() = || {
...
+                continue;
+            }
+
+            // Otherwise, we're done.
+            return Poll::Ready(None);
+        }
+    }
+}
+
+/// A server reply along with cached "has documents" status.
```
abr-egn, line 233: Tiny unrelated improvement: `has_next` was re-parsing the raw doc every time to get the answer, this saves that.

driver/src/cursor.rs
```diff
@@ -86,6 +88,16 @@ pub struct Cursor<T> {
     stream: stream::Stream<'static, raw_batch::RawBatchCursor, T>,
 }
 
+#[allow(dead_code)]
+const _: fn() = || {
```
abr-egn, line 92: This makes the invariance requirement give an understandable error if it's violated.

driver/src/cursor/poll_state.rs
```diff
@@ -0,0 +1,98 @@
...
+    Running(#[derive_where(skip)] BoxFuture<'a, (S, O)>),
+    Polling,
+}
+
+impl<'a, S, O> PollState<'a, S, O> {
+    pub(crate) fn new(state: S) -> Self {
+        Self::Idle(state)
+    }
+
+    pub(crate) fn state(&self) -> Result<&S> {
```
abr-egn, line 27: Despite the bodies of these methods being textually identical, they're distinct across the owned/&/&mut axis typewise so the only way to reduce the repetition would be a macro, which doesn't seem worth it here.

driver/src/cursor/session.rs
```diff
@@ -256,7 +262,7 @@ impl<T> SessionCursor<T> {
     where
         T: Deserialize<'a>,
     {
-        self.buffer().deserialize_current()
+        poll_panic!(self.buffer()).deserialize_current()
```
isabelatkinson:
```suggestion
        self.buffer()?.deserialize_current()
```
should be able to propagate here instead of unwrap

isabelatkinson (approved):
LGTM, just a tiny suggestion!

> If this seems reasonable, I'll file a ticket to make those APIs return `Result` in 4.0.

My understanding is that these methods can only fail when the futures aren't polled to completion, which we have generally treated as something we don't support. I'd lean towards leaving the unwraps as-is for high-traffic methods like `current`/`has_next` as I expect it's not ideal to add error handling everywhere users are calling them. Not a blocker here though; we can discuss this more in the 4.0 design 🙂

PR #1762, opened 2026-08-13
RUST-2470 Test against 9.0 builds

isabelatkinson:


.evergreen/config.yml
```diff
@@ -1198,6 +1198,34 @@ tasks:
       - func: "start csfle servers"
       - func: "run csfle tests"
 
+  - name: test-in-use-encryption-8.0
```
isabelatkinson, line 1201: this variant was missing (ditto below for load balancer)

BorisDog (approved):
LGTM

PR #1772, opened 2026-08-20
RUST-2483 Deserialize FindOptions maxTimeMS from milliseconds

Pongchi:
[RUST-2483](https://jira.mongodb.org/browse/RUST-2483)

`FindOptions::max_time` serializes to an integer number of milliseconds via `serialize_duration_option_as_int_millis`, but it has no matching `deserialize_with`. Deserialization falls back to the default serde representation of `std::time::Duration` and rejects the very value the driver itself emits:

```
invalid type: integer `5000`, expected struct Duration
```

It only accepts `{ "secs": 5, "nanos": 0 }`, an internal representation that no MongoDB document would ever carry.

Every other option type in `driver/src/coll/options.rs` that carries `max_time` already applies `deserialize_duration_option_from_u64_millis`: `AggregateOptions`, `CountOptions`, `DistinctOptions`, `EstimatedDocumentCountOptions`, `CreateIndexOptions`, `DropIndexOptions` and `ListIndexesOptions`. `FindOptions` is the only one missing it.

The test suite shows the same expectation: `driver/src/test/spec/unified_runner/operation/find.rs` replicates `FindOptions`' fields because serde cannot combine `flatten` with `deny_unknown_fields`, and it applies that helper to `max_time` by hand.

This reaches end users through Prisma, whose MongoDB connector deserializes the `options` argument of `findRaw` straight into `FindOptions`. Passing `{ maxTimeMS: 25000 }` to `findRaw` fails, while the identical value passed to `aggregateRaw` succeeds because `AggregateOptions` has the helper.

## Changes

Apply the helper (with `default`, matching `AggregateOptions`) and add three tests: deserialization from milliseconds, a serialize/deserialize round trip, and the field remaining optional.

## Testing

The new tests pass with the fix; two of the three fail without it.

`cargo nextest run --no-fail-fast` against a local three-node replica set gives 318 passed / 47 failed on this branch, and 315 passed / 47 failed on `main` under the same setup. The failing sets are identical — the difference of three is the tests added here, all passing. Those 47 failures are an artifact of my local server not being started with `enableTestCommands=1`, so `configureFailPoint` is unavailable.

`cargo +nightly fmt --check` passes and `cargo clippy --lib --all-features --tests` reports nothing for the changed file.

### DIFF
--- a/driver/src/coll/options.rs
+++ b/driver/src/coll/options.rs
@@ -842,7 +842,9 @@ pub struct FindOptions {
     /// across the wire as an integer number of milliseconds.
     #[serde(
         rename = "maxTimeMS",
-        serialize_with = "serde_util::serialize_duration_option_as_int_millis"
+        serialize_with = "serde_util::serialize_duration_option_as_int_millis",
+        deserialize_with = "serde_util::deserialize_duration_option_from_u64_millis",
+        default
     )]
     pub max_time: Option<Duration>,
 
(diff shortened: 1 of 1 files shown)


isabelatkinson:
See comment on RUST-2483.

PR #1781, opened 2026-08-31
RUST-2133 Allow custom aws credential provider

abr-egn:
RUST-2133 / RUST-2249

This adds `aws_credential_provider` to `Credential`; while the spec says it should be in `mechanism_properties` that's just a `Document`, and there's precedent for callbacks in `Credential` in the form of `oidc_callback`.

The new prose test validates both that authentication via the custom provider succeeds and that it's used preferentially over any other mechanism; [passing run](https://spruce.corp.mongodb.com/version/6a955dfc00536f000714381f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

### DIFF
--- a/driver/src/client/csfle/options.rs
+++ b/driver/src/client/csfle/options.rs
@@ -72,6 +72,8 @@ pub(crate) struct AutoEncryptionOptions {
     pub(crate) key_cache_expiration: Option<Duration>,
     #[serde(skip)]
     pub(crate) kms_connect_callback: Option<KmsConnectCallback>,
+    #[serde(skip)]
+    pub(crate) credential_providers: Option<CredentialProviders>,
 }
 
 fn default_key_vault_namespace() -> Namespace {
@@ -96,6 +98,7 @@ impl AutoEncryptionOptions {
             disable_crypt_shared: None,
             key_cache_expiration: None,
             kms_connect_callback: None,
+            credential_providers: None,
         }
     }
 }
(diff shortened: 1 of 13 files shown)


isabelatkinson (approved):
LGTM. The spec changes also include new client encryption option - is the plan to add that in a separate PR?

abr-egn:
> LGTM. The spec changes also include new client encryption option - is the plan to add that in a separate PR?

Derp, missed that.  Added, PTAL!

driver/src/client/csfle/options.rs
```diff
@@ -216,3 +245,12 @@ pub(crate) const EO_MONGOCRYPTD_SPAWN_ARGS: ExtraOptionArray =
     ExtraOptionArray("mongocryptdSpawnArgs");
 pub(crate) const EO_CRYPT_SHARED_LIB_PATH: ExtraOptionStr = ExtraOptionStr("cryptSharedLibPath");
 pub(crate) const EO_CRYPT_SHARED_REQUIRED: ExtraOptionBool = ExtraOptionBool("cryptSharedRequired");
+
+/// Custom credential providers.
+#[derive(Debug, Clone, Default)]
```
isabelatkinson: suggest adding `TypedBuilder` here as well

isabelatkinson (approved):
one small suggestion, no need for re-review

PR #1782, opened 2026-08-31
minor: sanity checks

abr-egn:
This splits out a subset of our current `lint` checks into a new `sanity` tag; the idea is to have the sanity checks be the ones required for a submit rather than the whole lint set.

Motivating incident: a recent update changed comment line flowing in rustfmt, causing a blocking failure.  With that kind of external change possible, we shouldn't gate our processes on it.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -64,6 +64,15 @@ buildvariants:
       - name: compile-on-latest
       - name: compile-on-msrv
 
+  - name: sanity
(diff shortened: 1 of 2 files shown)


isabelatkinson (commented):
Another thought: how about pinning to a Rust version for the rustfmt check like we do for clippy? That should prevent new rules from making our build red; as far as I can recall, we haven't run into this with clippy since we started pinning it.

PR #1786, opened 2026-09-02
RUST-1740 Break up rest of evergreen config

isabelatkinson:
Everything here is copy-paste with a few exceptions:
- Inlined some functions that were only being called in one place
- Moved the calls to start and stop the CSFLE servers directly into the `run-csfle-tests.sh` script to remove "stop csfle servers" from the general teardown tasks
- Removed some Windows-specific logic in `run-csfle-tests.sh` because we haven't run those tests on Windows in years

I will revert 9f30d4c before merging.


abr-egn (approved):
Very nice 😁

PR #1787, opened 2026-09-02
RUST-2492 Check for inherited unacknowledged write concerns

isabelatkinson:
Checks for an inherited unacknowledged write concern at the beginning of command execution. I also removed a few spots that check for an unacknowledged write concern that would need a similar fix. They're unreachable code with the existing validation, but in the _very_ unlikely case that we decide to support unacknowledged writes in the future, we can add them back in.

### DIFF
--- a/driver/src/operation.rs
+++ b/driver/src/operation.rs
@@ -43,7 +43,7 @@ use crate::{
         StreamDescription,
     },
     error::{CommandError, Error, ErrorKind, Result},
-    options::{ClientOptions, ReadConcern, ReadConcernInternal, WriteConcern},
+    options::{ClientOptions, ReadConcern, WriteConcern},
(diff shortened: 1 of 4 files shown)


driver/src/operation.rs
```diff
@@ -154,8 +154,7 @@ pub(crate) trait Operation {
     /// The read concern to use for this operation, if any.
     fn read_concern(&self) -> Feature<&ReadConcern>;
 
-    /// The write concern to use for this operation, if any.  If this is implemented,
-    /// `set_write_concern` MUST also be.
```
isabelatkinson, line 158: Unrelated fix, but I think this is a leftover comment from a previous design. There's no `set_write_concern` method.

abr-egn (approved):

PR #1795, opened 2026-09-11
RUST-2489 Fix GSSAPI tests

isabelatkinson:
patch: https://spruce.corp.mongodb.com/version/6aa46cdba0d58c000747ed11/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


driver/src/test/auth/gssapi.rs
```diff
@@ -1,135 +1,82 @@
...
-        "GSSAPI_DB",
-        Some("CANONICALIZE_HOST_NAME:forwardAndReverse"),
-    )
-    .await
+    run_test(Some("forwardAndReverse")).await;
 }
 
 #[tokio::test]
-async fn with_service_realm_and_host_options() {
-    // This test uses a "cross-realm" user principal, however the service principal is not
```
isabelatkinson, line 121: We were the only driver running a test like this, and the necessary secrets/setup are no longer available after the host change, so I removed it.

abr-egn (approved):

PR #1797, opened 2026-09-16
RUST-2503 Bump rustls to 0.23.45

abr-egn:
RUST-2503

Dependabot hadn't picked this up yet because the advisory was reported after this week's run.  Technically we don't _need_ to bump this in `Cargo.toml` because the fixed version is compatible with the version we specify, but we may as well to ensure the fix is used.


abr-egn:
Failures are unrelated - a python dependency used by evergreen-tools `setup-secrets.sh` is broken.

isabelatkinson (approved):

